PR #1, opened 2026-07-21
Add saved-login Codex provider

cneuralnetwork:
## What

Adds a local Codex CLI provider backed by the official Codex SDK and a saved ChatGPT login, so users can run `copperhead do ... --model codex` without an OpenAI or Anthropic API key. Direct providers and the deterministic `check` command are unchanged.

## Why

Copperhead previously required direct OpenAI or Anthropic API credentials even when a user already had authenticated Codex access through ChatGPT. This closes that gap by reusing the existing saved login.

## Design

The provider lives in [codex.ts](src/agent/providers/codex.ts) and drives the Codex SDK. Copperhead's structural tool gating is preserved: Codex threads run read-only with web access disabled, and the tool schema is built dynamically from the same gated tool list the other providers see, so `edit_file`/`write_file` remain absent until an OpenSpec proposal validates. Model selection is wired through [loop.ts](src/agent/loop.ts), [cli.ts](src/cli.ts), [config.ts](src/config.ts), and [types.ts](src/agent/types.ts).

## Spec / OpenSpec

- OpenSpec change: `add-codex-cli-provider` (`openspec validate add-codex-cli-provider` passes).
- Delta spec: [agent-core/spec.md](openspec/changes/add-codex-cli-provider/specs/agent-core/spec.md).
- [SPEC.md](openspec/specs/SPEC.md) updated to document the Codex provider alongside the direct providers.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | pass (147 passed, 10 skipped) |
| Live-LLM (opt-in) | keyed on `COPPERHEAD_TEST_CODEX` | pass (4 passed, 6 skipped; run by author) |

New tests: [codex-provider.test.ts](test/codex-provider.test.ts) covers the provider; [agent-integration.test.ts](test/agent-integration.test.ts) gains Codex live cases gated on `COPPERHEAD_TEST_CODEX`. No known pre-existing failures.

### Manual test log

- Sandbox variant used (`create` / `edit`): direct fixture (author validation)
- Commands run and outcome:

```text
$ npm run typecheck        # pass (re-run after merging main)
$ npm run build            # pass (re-run after merging main)
$ npm test                 # 147 passed, 10 skipped (re-run after merging main)
$ COPPERHEAD_TEST_CODEX=1 npm exec -- vitest run test/agent-integration.test.ts
  4 passed, 6 skipped; every applicable Codex live acceptance and safety case passed
$ kicad-cli version        # 9.0.9
  direct fixture ERC: 0 violations
  direct fixture DRC: 0 violations, 0 unconnected items
$ ASTRO_TELEMETRY_DISABLED=1 npm run docs:build   # pass
```

## Docs

Updates to [README.md](README.md), [.env.example](.env.example), [quickstart.mdx](docs/src/content/docs/getting-started/quickstart.mdx), [cli.md](docs/src/content/docs/reference/cli.md), [configuration.md](docs/src/content/docs/reference/configuration.md), and a new [AGENTS.md](AGENTS.md).

---

## Invariant checklist

- [x] **Spec-gated in**: `edit_file`/`write_file` stay structurally absent from the agent tool list until an OpenSpec proposal validates (Codex builds its tool schema from the same gated list).
- [x] **Verification-gated out**: mutation verification (ERC/DRC, repair up to `maxRepairCycles`, then rollback) is unchanged by this provider.
- [x] **`check`/`verify` stays LLM-free and network-free**: the Codex provider is only reachable from the agent loop, never from `src/commands/check`.
- [x] **No sexp serialization**: `src/kicad/` is untouched; editing stays anchored exact-match text replace.
- [x] **Sync-obligations ledger**: unchanged; the post-tool-call hooks and commit refusal behave as before.
- [x] **Secrets**: Codex uses the saved ChatGPT login, no API key is introduced; transcript redaction and `.gitignore` entries are unchanged.
- [x] **Spec coherence**: SPEC.md and the `add-codex-cli-provider` artifacts (proposal, design, delta spec, tasks) move together and `openspec validate` passes.


animesh-chouhan:
Thanks for this! The architecture is exactly right: Codex as a reasoning-only backend behind the normal `Provider` interface, read-only sandbox, structured output with a per-turn tool-name enum, and no silent fallback to a paid API. The OpenSpec change and doc updates are appreciated too. A few things before I can merge:

## Blocking

1. **Unavailable tool names shouldn't kill the run.** With the OpenAI/Anthropic providers, a hallucinated tool name reaches `dispatchTool`, which returns a corrective tool result and lets the model self-correct. `CodexProvider` throws from `parseStructuredTurn` instead; that error isn't retryable, so the loop fails the run and rolls back the snapshot. The pre-validation window is exactly when the model is most tempted to call `edit_file`, so this will be the most common way a Codex run dies, and it breaks the AC-3.10 parity you added. Suggestion: on parse/validation failure, re-run the thread once with the validation error as feedback, or pass unknown names through and let `dispatchTool` reject them.

2. **`messageCursor` advances before parsing.** In `chat()`, the cursor is set before `parseStructuredTurn` runs, so any retry after a parse failure sends a contentless prompt with no error context. Advance it only after a successful parse (a test pinning the retry prompt content would be great).

3. **Rebase needed.** The PR conflicts with main (the CLAUDE.md prose changed in `ff43b8b`). Please rebase and re-check the CLAUDE.md/README hunks against the new wording.

4. **Dependency weight.** `@openai/codex-sdk` pulls in `@openai/codex`, which installs a platform binary for every user, even those on API keys, and the provider never uses the bundled binary since it always sets `codexPathOverride`. Could this be an optional peerDependency with a lazy `import()` in `makeProvider`?

## Editorial

5. Please keep the docs neutral on provider choice: revert the SPEC.md example config back to `"model": "gpt-5"` and the demo examples back to `--model claude`. The dedicated "Use your existing Codex login" section is great and can stay.

## Smaller

6. design.md overstates read isolation: a read-only sandbox with `approvalPolicy: 'never'` can still read arbitrary paths (including the repo's `.env`) via Codex's own tools, so that part is prompt-enforced only. The mutation invariant holds, which is what matters; just soften the claim or confirm reads are restricted to the workdir.

7. Prefer `fs.mkdtemp` per provider instance over the fixed shared `tmpdir()` path.

8. `networkAccessEnabled` / `webSearchMode` are only asserted against a mock; I'll run the live suite with `COPPERHEAD_TEST_CODEX=1` on my side before merging to confirm they're honored by 0.144.6.

On AGENTS.md: happy to take it, with the understanding that it needs to stay in sync with CLAUDE.md going forward.

animesh-chouhan:
Related follow-up idea, for this PR or a separate one: the same saved-login pattern should work for Claude Code users via the **Claude Agent SDK** (`@anthropic-ai/claude-agent-sdk`, docs at code.claude.com/docs/en/agent-sdk). It drives the locally installed Claude Code and reuses its existing login, so a `--model claude-code` provider would let subscription users run copperhead without an `ANTHROPIC_API_KEY`, exactly like this PR does for ChatGPT users.

A few notes if anyone picks this up:

- The provider seam this PR establishes (`makeProvider` routing plus the injectable client used in `CodexProvider`) should carry over almost unchanged, so most of the work is mapping copperhead's `Provider.chat(messages, tools) -> Turn` contract onto the SDK's query surface.
- Tool gating may come out *cleaner* than the Codex version: instead of mirroring the gate in a structured-output enum, the Agent SDK lets you register custom in-process tools and restrict the built-in ones, so copperhead could expose exactly the tools `availableTools(ctx)` returns each turn and disable the SDK's own file-editing and bash tools. The spec-gated-in invariant stays structural rather than schema-mirrored.
- The same invariants from this PR apply: reasoning-only backend, isolated working directory, no built-in edit capability, all mutations through copperhead's dispatcher, no credential handling in copperhead, and no silent fallback to a paid API.
- Same dependency question too: it should be an optional dependency with a lazy import, not a hard install for every user.

I'll open a tracking issue so this PR doesn't grow scope: flagging it here since the design decisions being settled now (provider seam, gating approach, optional-dep packaging) determine how easy that follow-up is.

cneuralnetwork:

cneuralnetwork:
recheck

animesh-chouhan (commented):
Thanks for the contribution — this is a well-specified change that takes the repo's invariants seriously. Review below.

## Overview

The PR adds a third provider: the locally installed Codex CLI via `@openai/codex-sdk` and its saved ChatGPT login (`--model codex` / `codex:<model-id>`). Since Codex is itself an agent with filesystem access, it's deliberately demoted to a reasoning-only backend: read-only sandbox, `approvalPolicy: never`, network/web search disabled, throwaway temp working directory, and all actions returned as structured output whose tool-name enum is rebuilt each turn from `availableTools(ctx)` — so the "spec-gated in" invariant survives structurally.

## What's done well

- **Invariant preservation is genuinely structural, not cosmetic.** The per-turn schema enum mirrors the tool gating, and returned names/arguments are re-validated in `src/agent/providers/codex.ts` before entering the normalized `Turn`, so the enum isn't the only line of defense.
- **The SDK options are real.** I unpacked `@openai/codex-sdk@0.144.6` and confirmed every `ThreadOptions` key the provider sets (`sandboxMode`, `approvalPolicy`, `networkAccessEnabled`, `webSearchMode`, `skipGitRepoCheck`, `workingDirectory`) plus `codexPathOverride`, `outputSchema`, and the `Usage` field names exist in the published types — nothing is silently ignored.
- **No silent fallback to a paid API.** `otherProvider()` matches only the `openai`/`anthropic` provider names, so a rate-limited Codex run fails through rollback rather than switching to a metered key — consistent with the design doc's claim. Preserving `status`/`statusCode` for the shared `withRetry` 429 logic (with a test) is a nice touch.
- **Cursor semantics are correct.** `messageCursor` advances only after a turn is accepted, so rejected structured turns and thrown errors leave the unprocessed input to be re-sent; the one-corrective-retry behavior is tested for both unavailable tools and malformed args.
- **`makeProvider` going async is safe** — it had exactly one call site, which this PR updates.
- Offline test coverage is thoughtful: sandbox settings asserted verbatim, unique temp dirs, thread continuation without assistant replay, retry paths, `resolveModel` precedence.

## Issues and risks

**1. Read exposure is broader than the docs imply (security, medium).** D2 honestly admits the read-only sandbox doesn't confine *reads* and that avoiding native reads is prompt-enforced. Two concrete consequences deserve a doc note in `design.md`'s security properties: (a) Codex can natively read the target repo's `.env` or any file on disk despite the temp working directory; (b) the Codex CLI keeps its own session logs under `~/.codex`, entirely outside Copperhead's transcript redaction guarantee (AC-4.1). The redaction story users rely on no longer fully holds on this path and the docs should say so.

**2. Unescaped pseudo-XML framing (robustness, low-medium).** `renderTurnPrompt` embeds tool results and user content inside `<tool_result>`/`<user_request>` tags with no escaping. An ERC report or KiCad file containing a literal `</tool_result>` breaks framing, and hostile file content could inject instructions. The schema enum caps the blast radius to misuse of currently-allowed tools, but escaping (or fencing) would be cheap insurance.

**3. Arguments are validated as "a JSON object", not against the tool's parameter schema.** Native tool-calling providers get schema-constrained arguments; here the per-tool schemas appear only as prose in the prompt. If the dispatcher in `tools.ts` validates args, fine — but if it assumes provider-side validation, Codex is the one provider that can hand it structurally wrong arguments. Worth confirming, or folding per-tool `parameters` into the output schema via a `oneOf` discriminated on `name`.

**4. Duplicate context on retries (minor).** The correction prompt re-sends the original input into a thread that already received it, and a 429 during the correction call makes `withRetry` send it a third time. By design given the acceptance-gated cursor — just noting the token cost and potential model confusion.

**5. Misleading error suffix (minor).** `runThread` appends "Ensure codex is on PATH and authenticated (run: codex login status)" to *every* failure, including 429s and schema errors. Consider only attaching it to spawn/auth-shaped failures.

**6. Likely-wrong docs URL.** The README links Codex CLI docs at `https://learn.chatgpt.com/docs/codex/cli`; the official location is `https://developers.openai.com/codex/cli`. Please verify.

**7. Suggestion: derive `CodexClientLike` from the SDK's own types.** The hand-rolled structural interface duplicates a subset of `ThreadOptions`, and TypeScript's method-parameter bivariance means a renamed SDK option in a future version would *not* be a type error — the option would be silently dropped and the sandbox claim silently weakened. Since the SDK is already a devDependency, a types-only `import type { ThreadOptions } from '@openai/codex-sdk'` (no runtime cost) would make option drift a compile error. Highest-value non-blocking suggestion given how load-bearing those options are.

## Smaller notes

- `usage` totals ignore `cached_input_tokens` — consistent with the existing `Turn` type, fine.
- The `codex:` colon namespace diverges from the `claude*` prefix convention, but deliberately (D5), and it avoids the documented typo-routing hazard. Good trade.
- The `key: 'saved-codex-login'` sentinel in the test matrix is a bit hacky but reuses the existing skip logic cleanly.
- `tasks.md` marks live verification (3.4, 3.5) complete; the PR body reports the live-run evidence, which I can't reproduce without a Codex login — a maintainer spot-check of `COPPERHEAD_TEST_CODEX=1` before merge would be good.
- `AGENTS.md` duplicates much of `CLAUDE.md` and both now carry a "keep aligned" obligation — acceptable, just a standing maintenance cost to be aware of.

## Verdict

Approve with minor changes. Before merge I'd ask for: the docs-URL fix (#6), the security-properties doc note about native reads and Codex's own logs (#1), and ideally the types-only SDK import (#7). The rest can land as follow-ups.

animesh-chouhan:
the docs url is correct

animesh-chouhan:
## Independent live test run of this PR (`--model codex`)

Follow-up to my review above — I checked out the PR head (35e6072) and ran it live against a real saved ChatGPT login, KiCad 10.0.4 (kicad-cli), Node 24, Linux x64.

### Setup notes

- The `codex` CLI was **not** installed on PATH on this machine. Pointing `COPPERHEAD_CODEX_PATH` at the SDK's own bundled launcher (`node_modules/@openai/codex/bin/codex.js`, codex-cli 0.144.6) worked fine with the existing `~/.codex/auth.json` login. Worth documenting as a fallback — it means the npm install alone is sufficient, no separate CLI install needed.

### Direct CLI run — passed

`copperhead do "rename net KEY_DAH to KEY_DASH" --model codex` on the open-key fixture succeeded first try (~2 min):

- Spec gate observed working end-to-end: turn 1's schema enum had no edit tools; `propose_change` → `validate_change` unlocked them for turn 2.
- ERC clean, drift clean, surgical diff (4 changed lines in the `.kicad_sch`), single commit, `provider: codex` and token usage (37,274 in / 1,750 out) recorded in the transcript. No correction retries were needed.
- Codex made **zero** native shell/file calls (its rollout log shows no command executions) and its temp working directory stayed empty — the prompt-enforced reasoning-only contract held in practice.

### Live vitest suite — 3 passed, 1 failed

`COPPERHEAD_TEST_CODEX=1 npm exec vitest run test/agent-integration.test.ts`:

| Test | Result |
| --- | --- |
| AC-3.1 net rename, ERC clean, surgical diff | ✅ 35 s |
| AC-3.4 budget-violating pullup refused | ✅ 30 s |
| AC-3.6 persistent violations roll back | ❌ `expected 'success' to be 'failure'` |
| AC-4.1 no key material in tree | ✅ |

**Why AC-3.6 fails (and why it's flaky, not random):** the test forces failure with `maxTurns: 2`, assuming no provider can complete the whole flow in two turns. But this provider batches arbitrarily many tool calls into one structured turn — in my run Codex did `search, read_file, propose_change, validate_change` in turn 1 and `edit_file×3, record_constraint, resolve_affected×2, record_decision, search, list_nets, run_erc, check_drift, finish` in turn 2, legitimately finishing inside the budget. The test only passes when Codex happens to spread work across 3+ turns, which presumably is what happened in the run reported in the PR description.

Suggested fix: use `maxTurns: 1` in AC-3.6 — that forces failure *structurally* for every provider, since edit tools cannot appear in turn 1's tool list before validation. (Alternatively, cap `toolCalls` per structured turn via `maxItems` for turn-count parity with the native tool-calling providers — turn budgets are a cost/safety rail, and unbounded batching dilutes them.)

### Other observations

- The per-instance temp working directories (`/tmp/copperhead-codex-*`) are never removed; three accumulated over the suite run. Consider cleanup on run end.
- Confirming review point 1 empirically: each run left a ~90 KB rollout log under `~/.codex/sessions/` containing the full Copperhead system prompt and design content. AC-4.1 passes because it only scans the repo tree — Codex's own logs are outside Copperhead's redaction boundary and the docs should say so.

Overall: the provider works well live — the gating design held up under a real run — but AC-3.6 needs the test fix above before the parity claim (AC-3.10) is trustworthy for codex.

animesh-chouhan (commented):
# Review: Add saved-login Codex provider (PR #1)

## Verdict

**Approve with changes requested on three points.** High-quality PR: sound architecture, an unusually honest threat model, good offline tests, thorough docs. Before merge: resolve the AC-3.6 test/spec mismatch (finding 1), add a JWT redaction pattern (finding 3), and decide the landing order with sibling PR #37 explicitly (finding 2).

Verification performed on the PR branch (`agent/add-codex-cli-provider`, head `ce87ca4`), actual results, not assumed: `npm run typecheck` clean; `npm test` 119 passed, 10 skipped (live-provider tests, no key configured); `openspec validate add-codex-cli-provider` reports the change valid.

## Overview

Adds a third provider, `--model codex` / `codex:<model-id>`, backed by the official `@openai/codex-sdk` driving the locally installed, ChatGPT-authenticated Codex CLI. Codex is used strictly as a reasoning backend: read-only sandbox, approvals/network/web search disabled, and every action returns as structured output validated against the current Copperhead tool catalog before entering the normalized `Turn` type. The PR also adds an OpenSpec change (`openspec/changes/add-codex-cli-provider/`), offline provider tests, docs across the site and README, and an `AGENTS.md`.

## Repo invariant checks (all performed)

- **Spec-gated in: PASS.** The structural gate is mirrored, not weakened: `turnSchema()` in `src/agent/providers/codex.ts` builds the tool-name enum per turn from the tools the loop passes in, so `edit_file`/`write_file` cannot be expressed in the output schema before a proposal validates, and `parseStructuredTurn()` rejects any name outside the catalog with one corrective retry.
- **Verification-gated out: PASS.** No verification, repair, or rollback path is touched; provider failures flow into the existing `fail()` path that restores the snapshot. But see finding 1: the live test that proved rollback-after-persistent-violations was repurposed.
- **`check` stays LLM-free and network-free: PASS.** `src/commands/check` and its import graph are untouched; the Codex SDK is loaded lazily inside `makeProvider` only for `codex*` models.
- **No sexp serialization: PASS.** `src/kicad/` is untouched.
- **Sync-obligations ledger: PASS.** Ledger and commit gates are untouched.
- **Secrets: PASS with a gap.** No key is written anywhere; the saved ChatGPT credential is never read by Copperhead; `.gitignore` is unchanged. See finding 3 for a redaction gap in error text.

## Findings, ranked by severity

### 1. Major: shared live test AC-3.6 no longer tests what SPEC.md AC-3.6 specifies

`test/agent-integration.test.ts`: the test was renamed to "a structurally forced failure rolls back to a byte-identical tree" and its budget changed from `maxTurns: 2` to `maxTurns: 1`. With one turn, the run fails on budget exhaustion before edit tools can ever unlock, for every provider including the existing keyed ones. SPEC.md AC-3.6 still reads "If violations persist after `maxRepairCycles`, working tree equals the pre-run state", and repair cycles are now exercised by no live test. Failure scenario: a regression in the ERC-repair-then-rollback path would pass this suite. Fix: update SPEC.md AC-3.6 alongside (repo convention is that scenarios map 1:1 onto ACs), or keep the original violation-driven test for keyed providers and add the structural one-turn variant as a separate case.

### 2. Major (coordination, not a defect in this PR): guaranteed conflicts and divergent conventions with PR #37

Both this PR and #37 (saved-login claude-code provider) edit `src/agent/loop.ts` (`makeProvider`), `src/config.ts`, `test/agent-integration.test.ts` (the providers array), `.env.example`, README, `docs/reference/configuration.md`, and SPEC.md. Two incompatibilities go beyond textual conflicts: this PR makes `makeProvider` async (justified by lazy optional-SDK loading) while #37 extends the synchronous signature, and this PR adds `Provider.close?()` plus loop-level provider cleanup that #37 does not know about. The two saved-login providers share no abstraction: each reimplements namespace parsing (`codex:` vs `claude-code:`), no-fallback semantics, setup-error hinting, and live-test opt-in wiring. Defensible for two genuinely different transports, so no blocking extraction needed now, but land one first, require the other to rebase onto the async `makeProvider` and `close?()` conventions, and write the saved-login provider conventions down once so a third provider does not invent a fourth pattern.

### 3. Minor: redaction misses JWT-shaped tokens that this PR newly routes into transcripts

`src/util/redact.ts` covers `sk-*`, `Bearer ...`, npm, and GitHub tokens, but not bare JWTs (`eyJ...`). The saved Codex login token is JWT-shaped, and `runThread()` in `src/agent/providers/codex.ts` copies raw SDK/CLI error text into `Codex CLI provider failed: ${original.message}`, which reaches transcripts and summaries via the loop's `provider error:` fail path. Failure scenario: a Codex CLI error that echoes token material lands unredacted in `.copperhead/runs/`. Fix: add a JWT pattern (three dot-separated base64url segments above a minimum length) to `PATTERNS`; it also hardens the existing providers.

### 4. Minor: hand-rolled `validateJsonSchema()` fails closed on unsupported keywords, breaking only Codex, only at runtime

The validator rejects any schema keyword outside `type`, `description`, `properties`, `required`, `enum`, `items`, `additionalProperties`. Every schema in `src/agent/tools.ts` fits today. Failure scenario: someone adds `minimum` or `oneOf` to a tool schema; OpenAI and Anthropic keep working, and the Codex provider starts failing every call to that tool with an "unsupported schema keyword" error that looks like a model mistake. Fix: add an offline test iterating the real `TOOLS` catalog asserting every `parameters` schema stays within the supported subset, so CI catches it.

### 5. Minor: inaccurate comment on the retry cursor invites a wrong mental model

In `chat()`: "The input remains unseen until Copperhead accepts a structured turn" is not accurate; the thread has already seen the prompt, only the cursor is held back. Consequence: when `withRetry` re-invokes `chat()` after a rate limit, or a caller retries after a double parse failure, the same messages are resent into a thread that already contains them (the malformed-JSON test's third call demonstrates the duplication). Acceptable tradeoff; the comment should say what actually happens.

### 6. Minor: the spawned `codex` process inherits the full environment, including `.env`-loaded secrets

By spawn time, `process.env` contains `OPENAI_API_KEY`, `SYNAP_API_KEY`, etc. from the startup `.env` load, and the Codex CLI keeps its own local logs. If the SDK exposes an env option, pass a scrubbed environment; otherwise note it in the same doc paragraph that already covers native reads.

### 7. Nits

- `otherProvider()` correctly returns null for a rate-limited Codex run (no silent fallback to a paid API, matching design.md), but only via an implicit name mismatch; a one-line comment stating the guarantee would protect it (PR #37 adds exactly such a comment).
- `chat()` silently drops `ChatOpts.maxTokens` (`_opts`). No caller passes it today; a comment ("Codex CLI manages its own output budget") would prevent a parity assumption later.
- The new change directory lacks `.openspec.yaml`, unlike `build-copperhead-phase-1` and PR #37's change; validation passes without it, so cosmetic.
- Scope creep: `AGENTS.md` (80 lines of general repo guidance) and the CLAUDE.md current-state edit are not part of adding a provider, and AGENTS.md creates a second guidance file to keep in sync by hand. Consider a separate PR.
- `@openai/codex-sdk` in devDependencies makes every contributor's `npm install` pull platform-specific Codex binaries. The peer-optional arrangement for consumers is right; the dev cost deserves a CONTRIBUTING note.

## Done well

- Credential handling: Copperhead never reads, copies, or logs the saved ChatGPT credential; the CLI owns `~/.codex/auth.json` entirely, stated as a requirement in the proposal, design.md, and delta spec. Subprocess invocation is safe: `codexPathOverride` is passed as a path to the SDK, no shell interpolation anywhere.
- Honest threat model: the read-only sandbox's unconfined native reads (Codex can technically read `.env`) and the out-of-boundary `~/.codex/sessions/` logs are documented in README, configuration.md, SPEC.md section 7, and the delta spec rather than papered over.
- Prompt-injection hardening: all Copperhead messages and tool results are JSON-framed with an explicit policy line, and a dedicated test feeds hostile `</tool_result>`-style content through the framing.
- Provider abstraction conformance: normalized `Turn`, usage summed across the corrective retry (correct accounting), `provider: codex` recorded via the existing run metadata, rate-limit `status` preserved for the shared `withRetry`/failover machinery, and an injected `CodexClientLike` seam plus type-only SDK imports keep offline tests and typecheck SDK-independent.
- Cleanup: the optional `Provider.close?()` plus the providers set in `runAgentLoop` guarantees temp-dir removal even across failover, with cleanup errors logged rather than thrown.
- Docs and spec coherence: .env.example, README, quickstart, CLI reference, configuration, demo page, the `init` scaffold text, SPEC.md, and a complete OpenSpec change all move together.
- Test coverage: ten focused offline provider tests covering routing, sandbox options, temp-dir lifecycle, continuation without assistant replay, JSON framing, unavailable-tool and schema-mismatch retries, malformed-argument failure, rate-limit passthrough, and setup-hint gating.

## Merge and staleness

Against main right now: none. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`; the branch is freshly rebased despite being the oldest open PR. The real staleness risk is finding 2: whichever of #1 and #37 lands second takes the conflicts.

animesh-chouhan:
🤖 **Automated pr-review pass** (not a human review; a maintainer should still sign off).

## Verdict: Approve with nits

PR #1 adds a local Codex CLI provider (saved ChatGPT login) as a third reasoning backend. The change is well-scoped, preserves every repo invariant, and is unusually well-tested. Merge-blocking issues: none.

## Verification I ran on the PR head (merge commit `dee9088`)

| Check | Result |
| --- | --- |
| `npm ci` | OK |
| `npm run typecheck` | pass |
| `npm test` | **147 passed, 10 skipped** (live-provider tests skip offline) |
| `openspec validate add-codex-cli-provider` | valid |

- **Spec-gated in** ✅ — [codex.ts](src/agent/providers/codex.ts) builds its tool schema (`turnSchema`, line 203) purely from the `tools` list `chat()` receives, which is still `availableTools(ctx)`. The provider never injects `edit_file`/`write_file`; the Codex thread runs `sandboxMode: 'read-only'`, `networkAccessEnabled: false`, `webSearchMode: 'disabled'` (lines 76-79). Gating stays structural.
- **Verification-gated out** ✅ — the provider only returns tool calls; ERC/DRC, repair, and rollback in [loop.ts](src/agent/loop.ts) are untouched.
- **`check`/`verify` LLM-free & network-free** ✅ — `src/commands/check.ts` has no provider references, and the Codex SDK is loaded via `await import('@openai/codex-sdk')` inside `makeProvider` ([loop.ts:70](src/agent/loop.ts#L70)), so nothing on the check path can even pull it in.
- **No sexp serialization** ✅ — `src/kicad/` is untouched.
- **Sync-obligations ledger** ✅ — unchanged.
- **Secrets** ✅ — no API key introduced; Codex uses its saved login. `.env.example` adds only an empty `COPPERHEAD_CODEX_PATH`. The one honest caveat (Codex's own `~/.codex/sessions/` logs sit outside Copperhead's redaction boundary) is disclosed explicitly in [SPEC.md](openspec/specs/SPEC.md) §7.
- **Spec coherence** ✅ — SPEC.md, the `add-codex-cli-provider` proposal/design/delta-spec/tasks all move together; `openspec validate` passes.

## Correctness / coverage notes

- `makeProvider` becomes `async`; the sole production caller ([loop.ts:188](src/agent/loop.ts#L188)) awaits it, and provider cleanup is handled by a `providers` Set drained in `finally` with `close?.()`, so the Codex temp dir is always removed.
- [codex-provider.test.ts](test/codex-provider.test.ts) (10 cases) covers the read-only thread, structured tool-call mapping, thread continuation without replaying assistant output, prompt-injection framing of untrusted content, the single validation-correction retry, malformed-JSON and schema-mismatch rejection, rate-limit status passthrough, and setup-hint gating. [agent-integration.test.ts](test/agent-integration.test.ts) adds Codex to the AC-3.10 parity matrix, gated on `COPPERHEAD_TEST_CODEX`.

## Nits (non-blocking)

1. **Codex is excluded from rate-limit failover.** `otherProvider` ([loop.ts:91](src/agent/loop.ts#L91)) only swaps openai↔anthropic. A Codex-primary run that hits a rate limit fails rather than failing over, even when `OPENAI_API_KEY`/`ANTHROPIC_API_KEY` are set. This looks intentional (Codex is a distinct auth path), but it is asymmetric with the other providers and undocumented; worth a one-line comment or a follow-up.
2. **The `opts.provider` test seam is now `close()`d.** Any injected provider is added to the `providers` Set and closed in `finally`. Harmless today (openai/anthropic have no `close`), but it means the loop now takes ownership of an injected provider's lifecycle. Fine to leave; just noting the behavior change.

Neither nit blocks merge. The PR is mergeable and CI is green.

PR #17, opened 2026-07-21
Preflight hardening: unborn-HEAD gate, deferred affects-revisit, and why/remedy failure messages

animesh-chouhan:
Three pieces of preflight/loop hardening, developed and merged as a stack on the fork (animesh-chouhan#1, animesh-chouhan#2):

## 1. Fail preflight cleanly on an unborn HEAD

`copperhead create` in a fresh `git init` repo (zero commits) crashed in the `spec-seed` stage with raw git output (`exit code 128: fatal: ambiguous argument 'HEAD'`). The preflight only checked `isGitRepo`, which passes on an unborn HEAD, and then `snapshot()` ran `git rev-parse HEAD` unguarded. Adds `hasCommits()` and a preflight gate that fails up front with a friendly message, plus offline scenario tests covering non-git dirs, unborn HEAD, staged-but-uncommitted state, branch-name independence, and no `.copperhead` footprint on a failed preflight (AC-3.8).

## 2. Defer affects-revisit obligations until the target artifact exists

Affects-revisit obligations no longer open against artifacts that don't exist yet; they're deferred and re-opened at the start of the first run where the target artifact does exist (landing in that run's fresh ledger, so `finish` gates on them).

## 3. Explain why preflight failed and how to fix it

Every run-blocking failure — non-git directory, unborn HEAD, dirty tree, missing `kicad-cli` — now throws a structured `PreflightError` carrying the reason, a "why it failed" line, and numbered remedy steps:

```
repository has no commits; copperhead requires at least one commit for snapshots and rollback

why it failed: the pre-run snapshot is the current HEAD commit; with an unborn HEAD there is nothing to roll back to if verification fails
to fix:
  1. git add -A && git commit -m "initial commit"
  2. rerun the same copperhead command
```

The explanation is embedded in `err.message`, so every CLI catch path surfaces it with no rendering changes; `reason`/`why`/`remedy` remain available as structured fields. The three inline git gates are consolidated into `gitPreflight()` in `src/util/git.ts`, and `KicadCliMissingError` joins the same format.

## Verification

Full offline suite green: 92 passed, 7 skipped (LLM integration tests skip without an API key). The branch is behind `main` by the three docs-site commits but touches none of the same files — merges clean.

### DIFF
--- a/openspec/changes/build-copperhead-phase-1/specs/agent-core/spec.md
+++ b/openspec/changes/build-copperhead-phase-1/specs/agent-core/spec.md
@@ -47,7 +47,7 @@ The system prompt SHALL include the verbatim rules of SPEC.md §4.3, the `budget
 - **THEN** the agent refuses or proposes an alternative, citing the budget from SPEC.md, and does not silently comply
 
 ### Requirement: Sync obligations block commit
-The loop SHALL maintain an obligations ledger fed by deterministic post-tool-call hooks: editing a KiCad file records ERC/DRC, drift-check, and changelog obligations; recording a constraint records dual-write verification and a revisit obligation for every item in its `affects[]`; a non-trivial decision records a `docs/DECISIONS.md` append obligation. The commit step SHALL refuse to run while any obligation is open, and the ledger's final state SHALL be written into the run's `summary.md`.
+The loop SHALL maintain an obligations ledger fed by deterministic post-tool-call hooks: editing a KiCad file records ERC/DRC, drift-check, and changelog obligations; recording a constraint records dual-write verification and a revisit obligation for every item in its `affects[]` whose target artifact exists — an item targeting a not-yet-built artifact (schematic, board, BOM before their pipeline stage) is marked `deferred` in the registry instead and re-opens at the start of the first run where the artifact exists; a non-trivial decision records a `docs/DECISIONS.md` append obligation. The commit step SHALL refuse to run while any obligation is open, and the ledger's final state SHALL be written into the run's `summary.md`.
 
 #### Scenario: Stale doc blocks commit
 - **WHEN** a run edits a schematic value referenced by BOM.md but has not yet updated the doc
@@ -56,6 +56,10 @@ The loop SHALL maintain an obligations ledger fed by deterministic post-tool-cal
 #### Scenario: Constraint change forces affects revisit
 - **WHEN** a constraint with `affects: ["U2", "R7-absent"]` is modified during a run
 - **THEN** the run cannot commit until each affected item is explicitly resolved as changed or "no change needed" with a reason
+
+#### Scenario: Deferred revisit for unbuilt artifacts
+- **WHEN** a docs-only stage records a constraint with `affects: ["layout", "R1"]` before any board exists
+- **THEN** only the R1 revisit obligation opens now; the layout item is marked deferred and its obligation re-opens automatically at the start of the first run where a board is configured
 
 #### Scenario: Aborted run shows open obligations
 - **WHEN** a run fails or is aborted with obligations open
--- a/openspec/specs/SPEC.md
+++ b/openspec/specs/SPEC.md
@@ -208,6 +208,8 @@ The agent maintains `.copperhead/constraints.json` — a live, machine-readable 
 
 Loaded into every run's system prompt; `check` validates the design against it mechanically where possible (leakage sums, keepout geometry, forbidden pins). The `affects` field is what makes propagation reliable — change a constraint and the agent knows exactly which parts to revisit.
 
+An `affects` item that targets an artifact that does not exist yet (schematic, board, or BOM recorded before their pipeline stage) opens no revisit obligation at record time — it is marked `deferred` in the registry instead, and the obligation re-opens automatically at the start of the first run where the artifact exists. This keeps early docs-only stages from burning turns on ceremonial "not yet created" resolutions without losing the reconciliation guarantee: the revisit still happens, at the moment it can actually change the design.
+
 ### Change workflow (OpenSpec propose → apply → archive)
 
 - `copperhead do "<request>"` first generates `openspec/changes/<id>/` (proposal.md, spec deltas, tasks.md), then implements against it; the ERC/DRC-clean commit archives the change. Every hardware change gets a paper trail: *why → what spec changed → what files changed → verification result*
--- a/src/agent/loop.ts
+++ b/src/agent/loop.ts
@@ -8,7 +8,7 @@ import { loadConstraints, reopenDeferredAffects } from '../memory/constraints.js
 import { loadConfig, type CopperheadConfig } from '../config.js';
 import { Transcript } from './transcript.js';
 import { ObligationsLedger } from './ledger.js';
-import { isDirty, isGitRepo, hasCommits, snapshot, restore, commitAll, changedFiles } from '../util/git.js';
+import { gitPreflight, isDirty, snapshot, restore, commitAll, changedFiles } from '../util/git.js';
 import { withRetry, isRateLimit } from '../util/retry.js';
 import { openspecArchive } from '../openspec/cli.js';
 import { existsSync } from 'node:fs';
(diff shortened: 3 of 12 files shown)


animesh-chouhan:
Closing: this stack already landed on main via the feat/run-observability merge chain (9f5661e, 8e09717, 0a9e293, 9796b65). Verified by trial merge — merging this branch into current main produces a tree byte-identical to main, so there is nothing left to merge. All three pieces (unborn-HEAD preflight gate, deferred affects-revisit with reopen-on-next-run, PreflightError why/remedy messages) are live on main.

🤖 Verified with [Claude Code](https://claude.com/claude-code)

animesh-chouhan:
# Review: PR #17, Preflight hardening

**Verdict: approve with suggestions.** The three pieces are well designed, well tested, and coherent with the spec artifacts. Verified locally: `npm test` green (109 passed, 7 skipped) and `openspec validate build-copperhead-phase-1` passes.

**Process note first:** all four content commits of this PR (`9796b65`, `0a9e293`, `8e09717`, `9f5661e`) are already reachable from `main`, pulled in through the `feat/run-observability` merge chain (PR #28). The only commit main lacks is the fork-side merge commit `f5de683`, which carries no content. Merging this PR is a no-op; it can be closed as already-merged. The findings below therefore apply to code that is live on main today.

## Overview

1. Unborn-HEAD gate: `gitPreflight()` in `src/util/git.ts` consolidates the three git gates (repo, commits, dirty) in a guaranteed order, closing the `exit code 128` crash in `snapshot()`.
2. Deferred affects-revisit: `record_constraint` (src/agent/tools.ts) classifies each `affects[]` item via `classifyAffectsTarget()` and defers items whose artifact does not exist; `reopenDeferredAffects()` (src/memory/constraints.ts) re-opens them at run start in src/agent/loop.ts, before `loadConstraints()` so the prompt sees the updated registry.
3. `PreflightError` (src/util/preflight.ts, new) carries reason/why/remedy, formats them into `err.message` so no CLI catch path needs changes; `KicadCliMissingError` joins the format.

Answering the obvious question about the new gate: `copperhead init` still works on an unborn HEAD, since `runInit` (src/memory/scaffold.ts:166) never touches git. Only the agent loop and `create` (via `runAgentLoop`) hit the gate. Good.

## Correctness

**Major: the deferred marker can be silently dropped when rollback fails or the process dies mid-run.** `reopenDeferredAffects()` consumes the `deferred` marker and rewrites `constraints.json` at run start (src/agent/loop.ts:163, before any turn runs), while the re-opened obligations live only in the in-memory ledger. In the normal failure path this is safe: `snapshot()` runs first (loop.ts:119) and `restore()` (`reset --hard` plus stash apply) puts the marker back. But in two paths the repo explicitly supports, the guarantee breaks:

- a failed rollback (the `cee78f3` "survive a failed rollback" path) leaves the marker consumed on disk with the obligations gone,
- a hard crash or Ctrl-C mid-run never restores at all.

In both cases the next run finds no marker, re-opens nothing, and the revisit is lost, which is exactly what the PR body says cannot happen ("without losing the reconciliation guarantee"). Suggestion: keep the marker in the registry when re-opening, and remove it only when `resolve_affected` closes the obligation (mirroring the constraint dual-write pattern), or persist the registry rewrite at successful finish rather than at run start.

**Minor: same-run artifact creation defers one commit too late.** `affectsTargetExists()` reads `ctx.config`, which is loaded once at run start (loop.ts:115) and never refreshed. If a single `do` run creates the board (or schematic) and updates `.copperhead/config.json`, a constraint recorded later in that same run still sees `board: null`, gets deferred, and the run commits without the revisit; it re-opens only at the next run's start, after the commit it should have gated. The create pipeline is fine because stages are separate runs, but the spec text promises re-opening "at the start of the first run where the artifact exists", and here the artifact exists in the current run. Suggestion: re-check deferred items at `finish` time (pre-commit) against the on-disk config and open any whose artifact now exists.

**Minor: keyword classification can wrongly defer refdes-specific items.** `classifyAffectsTarget()` (src/memory/constraints.ts) checks board keywords without first checking for a concrete refdes, so an item like `"U5 thermal pad"` or `"J2 mounting"` classifies as `board` and gets deferred during schematic work, even though the item names a part whose schematic treatment is decidable now. The code comment argues misclassification is cheap, but in this direction the revisit is pushed past the moment it could change the design. Suggestion: test for a refdes pattern (for example `/\b[A-Z]{1,4}\d+\b/`) first and return `null`, which the existing test `classifyAffectsTarget('U2 EN pullup') === null` already gestures at.

**Correct and verified:** the reopen-before-`loadConstraints` ordering in loop.ts is right and commented; re-opened obligations land in the fresh ledger so `finish` gates on them; `record_constraint` preserves the full `affects[]` list with `deferred` as a subset (registry propagation is unaffected); the `run_erc`/`run_drc` early returns do not call `ledger.clear`, so a run that writes a `.kicad_sch` without configuring it still cannot commit (verification-gated out holds); `hasCommits` uses `rev-parse --quiet --verify HEAD`, which is branch-name independent and treats staged-but-uncommitted as no commits, both covered by tests.

## Repo invariant checks (all performed, all pass)

- Spec-gated in: `requiresUnlock` gating untouched; no mutation tool exposed unconditionally.
- Verification-gated out: ERC/DRC obligations still block commit; the new "does not apply yet" returns do not clear them.
- `check` stays LLM-free and network-free: `src/commands/check.ts` and its imports untouched.
- No sexp serialization: `src/kicad/sexp.ts` untouched.
- Ledger: `record_constraint` still feeds it; only the intended subset change, and the delta spec moved with it.
- Secrets: no key material in any new message; redaction and `.gitignore` untouched.

## Project conventions

- **Minor: spec coherence is asymmetric within the PR.** The deferral behavior correctly updates both `openspec/specs/SPEC.md` and the agent-core delta spec. But the unborn-HEAD gate is also new spec-level behavior (a new refusal condition) and appears nowhere in SPEC.md's safety rails or the safety-rails delta spec, which still only covers the dirty-tree refusal. The tests and `gitPreflight` docstring label it AC-3.8, but SPEC.md defines AC-3.8 as the dirty-tree criterion, so the label now covers behavior the spec does not state. Suggestion: add the no-commits refusal to SPEC.md's safety rails (and a scenario to the safety-rails delta), or give it its own AC number.
- **Nit: em dashes in new user-facing prose.** The new SPEC.md paragraph, the dirty-tree remedy strings ("git stash — to set them aside"), the ERC/DRC "does not apply yet — skip it" messages, and the prompts.ts turn-budget paragraph all use em dashes; house style for user-facing prose prefers colons or commas.
- **Nit: mild scope creep.** The empty-pattern `search` guard, the ERC/DRC message rewording, and the prompts.ts turn-batching paragraph are not among the PR's three stated pieces. All are small, tested, and sensible, but a line in the PR body would have made the diff self-explaining.

## Performance

No concerns. `gitPreflight` adds at most three short-lived git subprocesses per run start; `reopenDeferredAffects` is one registry read and at most one write. Both are dwarfed by provider latency.

## Test coverage

Strong overall: test/preflight.test.ts covers non-git, unborn HEAD, staged-only, branch-name independence, no `.copperhead` footprint on failed preflight, gate ordering, the `create` bug-report path, structured fields, and the offered `--allow-dirty` flag actually working. test/gating-sync.test.ts covers defer-at-record, reopen-exactly-once, and classification. Gaps worth closing:

- No test that rollback restores the `deferred` marker, which is the core guarantee under failure (and the major finding above).
- The partial-reopen branch in `reopenDeferredAffects` (some items re-open, `stillDeferred` keeps the rest) is untested; existing tests only exercise all-or-nothing.
- No classification test for a refdes combined with a board keyword ("U5 thermal pad"), the misclassification case above.

## Security

No issues. All git invocations go through `execa` with argument arrays (no shell interpolation); error messages contain no secret material; the preflight throws before any provider construction or network access, and the tests prove no run directory is written on a failed preflight.

## Done well

- Consolidating three inline gates into `gitPreflight` with an explicitly tested ordering is the right structural move; future gates have one home.
- Embedding why/remedy into `err.message` so every existing catch path benefits with zero rendering changes is an elegant constraint to design under, and the structured `reason`/`why`/`remedy` fields keep programmatic callers first-class.
- The tests are scenario-named after real failure modes (the bug-report path, exit-128 noise) rather than implementation details, and run the real loop offline by relying on proven throw ordering.
- Code comments consistently explain why, not what: the reopen-before-loadConstraints ordering, the misclassification cost argument, and the 13-obligations/25-turns regression note in the test.

PR #18, opened 2026-07-21
Prompt to continue on turn exhaustion + agent-loop efficiency fixes

### LINKED ISSUE #15: On turn-budget exhaustion, prompt to continue (with token usage) + agent-loop efficiency fixes
## Context

Two consecutive `copperhead create` (spec-seed) runs on a USB-C breakout brief failed with `turn budget exhausted (40 turns, N files touched but unverified)` and were hard-rolled-back, discarding ~90%-complete spec docs. Combined cost: ~500k input tokens with nothing kept.

Trace analysis (`.copperhead/runs/2026-07-21T10-44-49-392Z/`): 12 `record_constraint` calls opened 33 `affects-revisit` obligations plus a drift obligation. `finish` was refused at turn 25 with 15 turns left; the model then resolved obligations one per turn (every resolution was "no change needed: schematic/layout/BOM not yet created") and ran dry with 20 still open. A zero-waste run of this stage needs ~60 turns — the failure is structural, not model slowness.

## Feature: on turn exhaustion, ask the user instead of rolling back

When `maxTurns` is reached (`src/agent/loop.ts:385`), instead of immediately failing and restoring the snapshot:

1. Print run stats so far: turns used, files touched, open obligations count, and **token usage** (`tokensIn` / `tokensOut` are already tracked in the loop — currently they only surface in summary.md).
2. Prompt: `Turn budget exhausted (40 turns, 247k in / 9.5k out). Continue with N more turns? [y/N]` — on yes, extend the budget and keep the conversation going; on no (or non-TTY), fall back to the current fail-and-restore behavior.
3. Precedent exists: `--interactive` already pauses for approval after proposal validation (`src/cli.ts:103`), so there's a prompt mechanism to reuse. For non-interactive/CI runs the current behavior stays the default.

Related quality-of-life: even when the user declines, consider preserving the failed work via `git stash create` + `git stash store` (or a `copperhead/failed/<run-id>` branch) before `restore()` in `fail()` (`src/agent/loop.ts:182`), so budget-exhaustion failures are recoverable rather than destroyed.

## Optimizations (from the same trace analysis)

1. **Prompt the model to batch tool calls.** The loop executes every tool call in one assistant message within a single turn (`src/agent/loop.ts:251`), and both providers support parallel calls — but nothing in `src/agent/prompts.ts` says so. gpt-5 emitted exactly one call per message for all 40 turns. Add to `WORKFLOW`: "Emit multiple independent tool calls in a single response: all `record_constraint` calls in one message, all `resolve_affected` calls in one message, independent `write_file` calls together." Also add it to the "Only 5 turns remain, converge now" nudge (`src/agent/loop.ts:267`). This alone compresses the failed run's 40 turns to ~10.

2. **Don't open `affects-revisit` obligations for artifacts that don't exist.** 27 of the 34 obligations targeted schematic/layout/BOM when none existed yet (spec-seed stage). In the `record_constraint` handler (`src/agent/tools.ts:377`), skip or auto-defer obligations whose target artifact isn't configured (`ctx.config.schematic` / `ctx.config.board`) or whose doc doesn't exist; re-open deferred obligations when the artifact is later configured, so the reconciliation guarantee is kept.

3. **Array form for `resolve_affected`.** Accept `resolutions: [{constraint_key, item, resolution}]` alongside the single form, so one call can clear the backlog even with models that won't parallelize tool calls.

4. **Prompt caching in the Anthropic provider.** The loop resends the full conversation every turn (quadratic input cost — this is how 9.5k output tokens cost 247k input). OpenAI gets automatic prefix caching; `src/agent/providers/anthropic.ts` sends no `cache_control` at all, so Claude runs pay full price every turn. Add `cache_control: {type: 'ephemeral'}` breakpoints on the system prompt, the last tool definition, and the final message block per turn — roughly an order of magnitude off input cost on Claude runs.

5. **Smaller dials:**
   - `record_constraint`'s return value should include the running open-obligation count, so the model sees debt accumulating instead of discovering it at `finish`.
   - Make `maxTurns` configurable per stage — spec-seed (docs + bookkeeping) has a very different turn profile from a layout stage.
   - `search` should reject empty patterns with a corrective hint (turn 1 of the failed run was wasted on `pattern: ""`).
   - `run_erc`'s "no schematic configured" response should say ERC isn't applicable yet, so the model doesn't retry it.

## Expected impact

Items 1+2 take the spec-seed stage from ~60 required turns to ~10, comfortably inside the default 40. Item 4 cuts Claude input cost ~10×. The continue-prompt turns the remaining edge cases from "destroy the work" into a user decision, with token usage visible at the decision point.

## Acceptance criteria

Binary criteria, mirrored 1:1 by the delta specs in `openspec/changes/turn-budget-continue-and-loop-efficiency/` and by SPEC.md's new AC-15 block. Each is covered by an offline test in `test/budget-efficiency.test.ts` / `test/create-stage-turns.test.ts`.

**Continue prompt**

- [ ] **AC-15.1** When `maxTurns` is reached in an attended run and the user grants extra turns, the loop continues from the same conversation state, records a `budget-extended` transcript event, and the run can still finish successfully. The prompt can fire again at the next exhaustion.
- [ ] **AC-15.2** Declining at the prompt fails and restores the pre-run snapshot exactly as before.
- [ ] **AC-15.3** Non-interactive/CI runs (no TTY) fail and restore exactly as before this change.
- [ ] **AC-15.4** The decision point shows turns used, files touched, open obligation count, and cumulative token usage in/out, matching what summary.md reports.

**Work preservation**

- [ ] **AC-15.16** Any run failure with touched files leaves a `copperhead failed run <run-id>` git stash entry holding the work (tracked and untracked), while the working tree is still restored byte-identical; the CLI prints the recovery hint.
- [ ] **AC-15.17** A failure that touched nothing creates no stash entry.

**Efficiency**

- [ ] **AC-15.5 / AC-15.6** The system prompt workflow and the 5-turns-remaining nudge both instruct emitting multiple independent tool calls in a single response.
- [ ] **AC-15.7** `record_constraint` with `affects` targeting a schematic/board/doc that does not exist yet opens zero `affects-revisit` obligations; the items are reported as deferred and never block `finish`.
- [ ] **AC-15.8** `affects` items whose artifact exists still open obligations exactly as before.
- [ ] **AC-15.9 / AC-15.10** `resolve_affected` accepts `resolutions: [{constraint_key, item, resolution}, ...]`; entries resolve independently, so one bad entry does not invalidate the rest.
- [ ] **AC-15.11** `record_constraint`'s return value includes the running open-obligation count.
- [ ] **AC-15.12** `run_erc`/`run_drc` without a configured schematic/board say the check is not applicable yet, so the model does not retry.
- [ ] **AC-15.13** `search` with an empty pattern returns a corrective hint with an example instead of a generic missing-argument error.
- [ ] **AC-15.14 / AC-15.15** The Anthropic provider sends exactly three `cache_control: ephemeral` breakpoints (system prompt, last tool definition, last block of the final message) and reports `inputTokens` as uncached + cache-read + cache-creation.
- [ ] **AC-15.18 / AC-15.19** `stageMaxTurns` in `.copperhead/config.json` overrides `maxTurns` for the named create-pipeline stages; stages without an entry, or an absent map, behave exactly as before.

### LINKED ISSUE #19: copperhead create can't bootstrap a KiCad project — schematic stage always refuses on a fresh repo
## Summary

On a fresh repo, `copperhead create` runs the docs-only stages (spec-seed, architecture, part-selection) fine, but the **schematic** stage always ends in a refusal: the agent has no way to bring a KiCad project into existence, and nothing earlier in the pipeline scaffolds one.

Observed in a real run (copperhead-test, run `2026-07-21T11-08-30-594Z`): the agent proposed and validated the `schematic-stage` change, then called `finish(refuse)` with:

> Blocked: the repo has no KiCad project/schematic files and the tools cannot create .kicad_* files. ... Please add a blank KiCad project and top-level schematic.

## Why it's boxed in

- `write_file` hard-refuses KiCad files by design (`src/agent/filetools.ts` — `write_file refuses KiCad files (...); use edit_file with anchors instead`), so the agent can never create a `.kicad_sch`/`.kicad_pro` from scratch.
- `edit_file` needs an existing file to anchor into — nothing to edit.
- `list_symbols` / `list_nets` / `run_erc` all return `no schematic configured` because `config.schematic` is unset.
- `copperhead init` only *discovers* an existing schematic (`src/memory/scaffold.ts`: "no .kicad_sch found ... point copperhead at a KiCad project") — it doesn't scaffold one either.

So the create pipeline assumes a KiCad project pre-exists, but `create` is supposed to bootstrap the whole project. The guardrail (never let the LLM generate a whole `.kicad_sch`) is sound; the gap is that no *deterministic* code path creates the blank project.

## Proposed fix

Have the schematic stage (or `create` itself, before that stage) scaffold a minimal valid blank project deterministically — template-based, not LLM-generated:

- write a blank `<project>.kicad_pro` + `<project>.kicad_sch` (empty top sheet with title block) from templates or via `kicad-cli`
- set `config.schematic` so `list_symbols`/`run_erc`/`edit_file` work from there
- keep the `write_file` KiCad refusal exactly as is — the agent then builds the schematic through anchored `edit_file` calls as designed

Alternative considered: a narrowly-scoped `create_schematic` tool that only writes a fixed blank template (no LLM-provided content), which keeps bootstrapping inside the agent loop.

animesh-chouhan:
Closes #15. Also fixes #19, #21, #23, #25 (create-pipeline hardening from live runs).

Two consecutive spec-seed runs burned ~500k input tokens and were hard-rolled-back at the 40-turn budget. This PR turns exhaustion into a user decision, makes failures recoverable, removes the structural turn waste the trace analysis found, and hardens the create pipeline against the blank-sheet and corruption failure modes that live verification exposed.

## What changed

**Continue prompt on turn exhaustion (attended runs)**
- At `maxTurns`, TTY runs are asked: `Turn budget exhausted (40 turns, 247.3k in / 9.5k out, 3 file(s) touched, 20 open obligation(s)). Continue with 20 more turns? [y/N]`. Yes extends the budget in place (repeatable; the offer is a constant `ceil(original maxTurns / 2)` with fresh numbers each time); No / non-TTY keeps today's fail-and-restore. A throwing prompt reads as declined, so the rollback path always runs. Wired for both `do` and every `create` stage via an `onBudgetExhausted` callback in `RunOptions`.

**Failed work is preserved, not destroyed**
- Before any failure rollback, touched work (tracked + untracked) is stored as a `copperhead failed run <run-id>` stash entry; the CLI prints the recovery hint. Clean failures create no stash. The in-flight transcript dir is explicitly unstaged first so preservation can never delete the audit trail.

**Loop efficiency (from the issue's trace analysis)**
- System prompt workflow step 8 + the 5-turns-remaining nudge now instruct batching independent tool calls in one response.
- `record_constraint` defers `affects-revisit` obligations whose target artifact does not exist yet (schematic/board not configured, doc absent) instead of opening dead obligations, and returns the running open-obligation count. Doc matching is case-insensitive and explicit `.md` names classify as docs before board keywords. Deferred items are surfaced in the tool result and `summary.md`.
- `resolve_affected` accepts `resolutions: [...]` to clear a backlog in one call, with per-entry results.
- Anthropic provider sends three `cache_control: ephemeral` breakpoints (system prompt, last tool, last message block) and counts cache read/creation tokens in reported usage.
- `stageMaxTurns` config for per-stage create budgets (invalid entries dropped on load); `search` rejects empty patterns with a corrective hint; `run_erc`/`run_drc` without a configured artifact read as not-applicable-yet; only *consecutive* tool-less turns count as a stall, so sporadic empty completions no longer roll back a converging run.

**Create-pipeline hardening (live-run follow-ups)**
- Stage completion is content-aware: the schematic stage requires symbols plus drift-clean BOM/PINOUT, layout-draft requires a board with a placed footprint plus the LAYOUT.md marker. `runCreate` re-checks the contract after each successful run and halts for resume instead of advancing over planning-only output.
- Every `edit_file` on a `.kicad_sch`/`.kicad_pcb` is validated with a kicad-cli load probe: a newly unloadable file is reverted with kicad-cli's own error; an already-unloadable file keeps repair edits (no incremental-repair deadlock). `.kicad_pro`/`.kicad_sym`/`.kicad_mod` are never probed (no standalone load command exists for them).
- Zero-symbol schematics are drift-exempt bootstrap state (docs legitimately lead the schematic); `check` surfaces a non-failing warning (log + optional `drift.warning` JSON field) when an empty schematic coexists with a populated BOM.md, so an accidentally emptied schematic can't hide behind the exemption.
- Missing ERC/DRC reports raise kicad-cli's own output instead of an opaque report-read ENOENT; the transcript dir is re-created before every write so the audit trail survives rollbacks.

## Spec artifacts

- OpenSpec change `openspec/changes/turn-budget-continue-and-loop-efficiency/`: proposal, design D1–D10, delta specs for `agent-core`/`safety-rails`/`create-pipeline`/`docs-memory`/`kicad-tooling` with scenarios AC-15.1 through AC-15.28, tasks all checked.
- `openspec/specs/SPEC.md` updated in step: sections 4.5, 5, and the AC-15 block.

## Testing

- 26 new offline tests across `test/budget-efficiency.test.ts`, `test/create-stage-turns.test.ts`, and `test/create-hardening.test.ts`: scripted-provider loop runs for continue/decline/absent/throwing callback, stash preservation and recovery, deferral and classifier edge cases, batch resolution, prompt guidance, mocked-SDK cache breakpoints and token accounting, per-stage budgets and config sanitization, probe scoping, corrupting-edit revert, already-corrupt repair-keep, and the drift bootstrap exemption + check warning.
- Full suite: 101 passed, 7 skipped (live-LLM tests, key-gated as designed). `tsc` build clean. `openspec validate` passes.

## Review trail

Three review rounds in the PR comments; all findings resolved in-branch (`0c3fbda`, `c1a6d04`). Remaining documented trade-offs (run-start config for deferral decisions, possible duplicate convergence nudge after an extension) are follow-up material, not blockers.


animesh-chouhan:
# Review

## Overview

The PR converts turn-budget exhaustion from an unconditional rollback into a TTY decision point (`src/agent/loop.ts`, `src/cli.ts`), preserves failed work in a named git stash before restore (`src/util/git.ts`), and attacks the issue #15 turn-waste causes: batching guidance in the prompt, deferral of `affects-revisit` obligations for not-yet-existing artifacts, a batch form for `resolve_affected`, Anthropic prompt caching, and per-stage `stageMaxTurns`. Spec artifacts (proposal, design D1–D6, delta specs, SPEC.md AC-15) move in step, which matches this repo's OpenSpec discipline. Overall this is a well-scoped, well-tested change; the design doc honestly records alternatives and risks. One significant bug in the pushed code, plus a handful of smaller issues.

## Correctness issues

**1. (Important) `preserveFailedRun` can delete the in-flight transcript directory.**
It does `git add -A` then stash. When `.copperhead/runs/` is *not* gitignored in the target repo, `add -A` stages it; `restore()`'s `reset --hard` then deletes staged-but-not-in-HEAD paths, and its `clean -fd -e .copperhead/runs` guard only spares *untracked* files (`src/util/git.ts`). The fail path afterwards writes `summary.md` into a directory that no longer exists. Copperhead-initialized repos gitignore `runs/` (AC-4.3), but `restore()` explicitly defends the non-ignored case, so the code elsewhere assumes it happens. Fix: unstage the runs dir after staging (`git reset -q -- .copperhead/runs`) before `stash create`.

**2. Doc-name matching in `missingArtifactReason` is case-sensitive.**
An affects item like `bom.md` produces candidates `docs/bom.md` / `bom.md`; on a case-sensitive filesystem `existsSync` misses the real `BOM.md`, so the item is wrongly deferred and never opens an obligation (`src/agent/tools.ts`). The bare-word path (`BOM`, `bom`) handles case via `toUpperCase()`, but the `.md` regex path preserves the original case. Check the uppercased basename as an additional candidate.

**3. `LAYOUT.md` is classified board-ish, not doc-ish.**
`/\b(layout|board|pcb)\b/` matches the "layout" inside `layout.md` before the doc check runs, so a reference to the LAYOUT doc defers with "no board configured yet" even when `docs/LAYOUT.md` exists. Run the `.md`-filename check before the board-keyword check.

**4. A throwing `onBudgetExhausted` callback bypasses rollback entirely.**
`await opts.onBudgetExhausted(stats)` in `src/agent/loop.ts` is not wrapped; if the readline prompt throws (stdin closed mid-prompt, EIO on a dying terminal), the exception propagates out of the loop with no `restore()`, no stash, and no summary — the exact silent-destroy failure mode this PR exists to remove, minus even the restore. Wrap it and treat a throw as "declined" (0).

**5. Extension size diverges from design D1 and escalates.**
The design says the prompt offers `ceil(maxTurns / 2)`; the CLI computes `Math.ceil(stats.turnsUsed / 2)` (`src/cli.ts`). At the second exhaustion `turnsUsed` equals the already-extended budget, so offers grow 20 → 30 → 45 → … Each is still a fresh consented decision, so this is arguably fine behavior, but code and design.md should agree — pick one and update the other.

## Minor observations

- **Stale config within a run:** `record_constraint` checks `ctx.config.schematic` loaded at run start. A single `do` run that creates and configures a schematic and then records constraints will still defer them, and deferred items never re-materialize in-run. The cross-run reconciliation story (registry `affects[]` + drift) covers it eventually, and design.md's non-goals acknowledge this, so a note is all it needs.
- **`stageMaxTurns` values are unvalidated** — `0` or a negative value makes the stage exhaust on turn 0 (attended runs immediately prompt). A one-line clamp or validation error in `src/config.ts` would help.
- The 5-turns-remaining nudge can be injected a second time after an extension (>5 extra turns). Harmless, possibly even useful; just noting it's not deduped.
- `provider?: Provider` on `RunOptions` as a test seam is pragmatic and clearly labeled; fine.

## What's good

- **The Anthropic caching implementation is correct**: three breakpoints (system block, last tool, last block of the final message — the moving conversation breakpoint still hits via longest-prefix matching, and `cache_control` on `tool_result` blocks is supported). Counting cache read/creation tokens into `inputTokens` is the right call for a volume-honest summary, and the rationale comment says exactly that.
- **Deferral is visible everywhere it needs to be** (tool result, `summary.md`, ledger accessor) — the "defer, don't skip" audit story from D3 is actually implemented, not just written down.
- **Per-entry results in batch `resolve_affected`** avoid the one-bad-key-wastes-the-call trap, and the corrective errors (empty search pattern, not-applicable ERC/DRC, neither-form-given) consistently tell the model what to do next rather than just what failed.
- **Test coverage is strong**: the scripted-provider loop tests exercise continue/decline/absent-callback end to end including transcript events and byte-identical restore; the stash test verifies actual recoverability via `git stash apply`, not just that a stash exists; the mocked-SDK test pins the exact breakpoint count. Findings 2 and 3 above are the only untested seams found.

## Security

No concerns. `git add -A` respects `.gitignore`, so `.env` stays out of the preservation stash (guaranteed by AC-4.3 in initialized repos); transcript redaction is unchanged; stash entries contain only working-tree design files.

## Verdict

Approve with changes: fix the transcript-dir deletion (finding 1 — blocking), wrap the exhaustion callback (finding 4), and fix the two classifier misses (findings 2–3, small). Findings 5 onward are polish.

---

animesh-chouhan:
# Re-review at `0c3fbda`

**Disclosure:** of the two commits added since the first review, `0c3fbda` was authored by Claude as the review-fix commit (findings 2–5: classifier ordering/case, callback throw safety, constant extension offer, `stageMaxTurns` sanitization — all verified present, suite green, `openspec validate` passing). The genuinely unreviewed material is **`3930740` "Harden the create pipeline against blank-sheet and corruption failure modes"**, and that's where the new findings are.

## Overview of the new commit

`3930740` reacts to live-run failures (#19, #21, #23, #25): KiCad-file edits are probe-validated with `kicad-cli` and auto-reverted if the file would become unloadable (`src/agent/tools.ts`, `src/kicad/cli.ts`); `create` stage completion is content-aware (schematic needs symbols + drift-clean docs, layout needs a placed footprint) and re-checked after each "successful" run (`src/commands/create.ts`); zero-symbol schematics are treated as bootstrap state by drift (`src/memory/drift.ts`); ERC/DRC report-parse failures now surface kicad-cli's own error; the transcript dir is re-created defensively.

## Correctness issues (new)

**1. (Important) Edits to `.kicad_pro`, `.kicad_sym`, and `.kicad_mod` are *always* auto-reverted.**
The validation triggers on `isKicadFile()`, which matches five extensions (`src/util/paths.ts`), but `kicadLoadError()` only knows two probes: `.kicad_sch` → netlist export, **everything else** → `pcb export pos`. Feed a project file (JSON) or a symbol/footprint library to `pcb export pos` and kicad-cli fails to load it as a board → non-zero exit → the handler reverts a perfectly good edit and tells the agent it "would make the file unloadable." The agent can never edit a `.kicad_pro`, symbol lib, or footprint again. Fix: probe only `.kicad_sch`/`.kicad_pcb` and skip validation for the other three (or add real probes for them).

**2. An already-corrupt file can never be incrementally repaired.**
The handler captures `before`, applies the edit, probes, and reverts on any load error — but never probes `before` itself. If the file was already unloadable, every partial repair edit is reverted unless one single edit makes the whole file load. That's a deadlock for exactly the corruption-recovery scenario this commit is about. Suggestion: when the post-edit probe fails, re-probe the pre-edit content; if it was already broken, keep the edit and say so instead of reverting.

**3. The drift zero-symbol carve-out weakens `check` for established repos.**
`checkDrift` returning no mismatches for an empty schematic is right for bootstrap, but `copperhead check` is the LLM-free CI gate — a repo whose schematic gets truncated/emptied by accident now passes drift while `BOM.md` still claims twenty parts. A false negative in the one command that's contractually trustworthy. Consider emitting a single informational finding ("schematic has zero symbols but BOM lists N refdes") from `check`'s path, while keeping the create-pipeline exemption.

## Process / conventions

**4. `3930740`'s spec-level changes have no OpenSpec artifacts.**
CLAUDE.md is explicit: spec-level behavior changes update SPEC.md and change artifacts together. This commit changes the create-pipeline stage-completion *contract*, drift semantics, and edit-tool behavior — all capability-level — yet touches nothing under `openspec/`, and the PR body doesn't mention any of it. Either fold these into the existing change (new requirements/scenarios + design decisions) or split them into their own change. As merged history, `check`-visible behavior now diverges from SPEC.md.

## Test coverage (new commit)

- No tests for: `kicadLoadError` itself, the edit-revert path (finding 1 would have been caught by a `.kicad_pro` edit test), `layout-draft.isComplete`, the drift zero-symbol carve-out, or the transcript-dir re-creation.
- `test/create-stage-turns.test.ts` was repurposed: a test named "per-stage turn budgets (AC-15.18)" now also asserts the pipeline-halt behavior. It passes, but the halt contract deserves its own named test — right now a regression in stage gating would fail a turn-budget test.

## Minor

- Every KiCad edit now spawns a kicad-cli export (~0.5–2 s). Acceptable for correctness, but on a schematic stage with dozens of edits it's real wall-clock; memoize per content-hash if it ever hurts.
- The transcript `mkdir` hardening and the runCheck "no report → here's kicad-cli's actual error" change are both good, low-risk fixes; `toolEditFile` throwing before the probe means failed edits don't pay the probe cost.

## Verdict

The `53c7609` + `0c3fbda` material is done: reviewed, fixed, tested, validated. The new `3930740` commit fixes real live-run failures but ships one significant regression (finding 1 — library/project-file edits are bricked), one repair deadlock (finding 2), and skips the repo's own spec discipline (finding 4). Findings 1–2 should be fixed with tests, the missing OpenSpec coverage added, and finding 3 considered, before merge.

---

animesh-chouhan:
# Review round 3 at `c1a6d04` — approve

**Provenance disclosure:** this PR is now a mix of reviewed and reviewer-authored work. `53c7609` was reviewed in round 1; `3930740` in round 2; `0c3fbda` and `c1a6d04` are the review-fix commits authored by Claude; `b354882` is the stall-counter fix. The two newest commits were re-checked fresh rather than assumed correct.

## State of the PR

- Head `c1a6d04` is byte-identical to the validated tree: `openspec validate` passes, `tsc --noEmit` clean, **101 passed / 7 key-gated skips**. CI on the PR runs only the CLA check (passing) — the repo has no test workflow, so the green suite is a local claim, not a CI-enforced one.
- File list matches the five commits exactly — no stray files, nothing sensitive.
- All findings from rounds 1 and 2 are fixed in-branch: transcript-dir preservation, callback-throw safety, classifier ordering/case, constant extension offers, `stageMaxTurns` sanitization, probe scoping to `.kicad_sch`/`.kicad_pcb`, already-corrupt repair-keep, the `check` empty-schematic warning, and full OpenSpec coverage (D1–D10, AC-15.1–15.28, five delta specs).

## Fresh pass over the unreviewed delta (`b354882`, `c1a6d04`)

- **`b354882` (stall counter): correct.** The reset sits after the tool-less early-continue, so it fires exactly on productive turns; only three *consecutive* empty completions fail the run. Comment and spec scenario (AC-15.27) match the code.
- **`c1a6d04` self-review — holds up, with three honest nits:**
  - If `kicad-cli` vanishes mid-run, the first probe throws `KicadCliMissingError`, so the edit stays applied unvalidated and the error surfaces through the dispatcher. Acceptable degradation; the revert guarantee is only as available as kicad-cli.
  - `emptySchematicWarning` re-parses the schematic that `checkDrift` just parsed (duplicate `listSymbols` in `check`). Negligible; pass the symbol list in if `check` ever gets hot.
  - The agent-core delta spec says "the *loop* SHALL probe" — it's the tool handler that probes. Cosmetic wording.

## Remaining known-open items (documented, none blocking)

- Deferral decisions use run-start config, so a single `do` run that creates + configures a schematic mid-run still defers constraints recorded afterwards (design non-goal; covered cross-run by the registry + drift).
- The 5-turns-remaining nudge can be injected twice after a budget extension.
- No unit tests for `layout-draft.isComplete` or the transcript-dir re-creation; probe cost is ~1 s per schematic/board edit.

## Verdict

**Approve.** Every finding across three rounds is resolved on the branch, spec artifacts are coherent with the code, and the suite is green. The PR description has been updated to cover the full five-commit scope; the open items above are follow-up-issue material, not blockers.

---

animesh-chouhan:
# Review: PR #18, Prompt to continue on turn exhaustion + agent-loop efficiency fixes

**Verdict: request changes.** The work itself is high quality and the offline suite passes on the PR head (verified locally: 101 passed, 7 skipped, `tsc` build clean). The blocker is not the code in isolation: the branch is stale against main, GitHub reports it as CONFLICTING, and main has since merged a different implementation of several of the same features, including a deferral design that carries a spec guarantee this branch's design explicitly declines. Merging by resolving conflicts in this branch's favor would regress merged spec behavior.

## Overview

The PR turns turn-budget exhaustion into a TTY-gated user decision (`onBudgetExhausted` callback in `RunOptions`, wired in `src/cli.ts` for `do` and every `create` stage), preserves failed work in a named git stash before rollback (`preserveFailedRun` in `src/util/git.ts`), and adds a set of loop-efficiency and create-pipeline hardening fixes: obligation deferral for nonexistent artifacts, batch `resolve_affected`, Anthropic prompt caching, per-stage `stageMaxTurns`, kicad-cli load-probe validation of schematic/board edits, content-aware stage completion, the zero-symbol drift exemption, and consecutive-only stall counting. Spec artifacts (proposal, design D1-D10, five delta specs, tasks, SPEC.md 4.5/config/AC-15) move together with the code, which is exactly what this repo requires.

## Blocker: stale branch, conflicting with main

Merge base is `603a2c5`; since then main merged four PRs that overlap this branch heavily. `gh pr view` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`.

- **`0a9e293` (deferral, already on main)**: main's `record_constraint` already defers affects items for missing artifacts, but persists the `deferred` marker in `constraints.json` (`src/memory/constraints.ts`) and re-opens the obligations via `reopenDeferredAffects` at the start of the first run where the artifact exists. Main's `openspec/specs/SPEC.md` (the paragraph at line 211) states this re-open guarantee explicitly. This PR's `design.md` D3 declares cross-run persistence a non-goal and implements deferral as in-memory ledger state only (`deferredList` in `src/agent/ledger.ts`), relying on the system-prompt registry dump for eventual reconciliation. That is a strictly weaker guarantee than what main's SPEC.md now promises. On rebase, keep main's registry-persisted deferral and layer this PR's genuinely new additions on top (deferred items in `summary.md`, the running open-obligation count in the tool result, the case-insensitive doc matching and `.md`-beats-board-keyword classifier refinements, batch `resolve_affected`).
- **`0a9e293` also already shipped** the batching instruction in `src/agent/prompts.ts`, the empty-pattern `search` rejection, and the not-applicable `run_erc`/`run_drc` messages, with slightly different wording. Conflict resolution has to pick one wording per site and keep the corresponding tests in sync.
- **`cee78f3` / `ff3f792` (PR #27)**: main already re-creates the transcript dir before writes in `src/agent/transcript.ts`, duplicating this PR's task 7.6.
- **`3ee3eed` / `309f29b` (PR #28, run observability)**: main's `run-start` event now carries the full AC-8.1 metadata (resolved config snapshot including effective `maxTurns`, model selection source, git state) and `fail()` on main takes a second argument. The loop rewrite in this PR is built on the old `loop.ts` and will conflict throughout; the `budget-extended` event should also be reconciled with the AC-8 transcript surface (it currently duplicates all of `BudgetExhaustedStats` into the event, which is fine, but the effective budget after extension should agree with what `run-start` reported).

Concrete ask: rebase onto main, reconcile the deferral model in favor of main's persisted design (and update this change's D3 and the agent-core delta spec accordingly), re-run `openspec validate turn-budget-continue-and-loop-efficiency`, and re-run the suite.

## Invariant checks performed (all on the PR head)

- **Spec-gated in: pass.** `edit_file`/`write_file`/`resolve_affected` keep `requiresUnlock: true`; `availableTools` filtering is untouched. The probe logic runs inside the already-gated handler.
- **Verification-gated out: pass.** A budget extension only enlarges the turn loop; `finish` gating, repair cycles, and rollback are untouched. A declined, absent, or throwing callback all take the same `fail()` path, and the throwing-callback-reads-as-declined handling in `src/agent/loop.ts` is exactly right: a broken prompt must not skip the rollback.
- **`check` stays LLM-free and network-free: pass.** The additions to `src/commands/check.ts` call only `emptySchematicWarning` (fs reads + the read-only sexp parser). The optional `drift.warning` key is omitted when clean, preserving the stable-key contract.
- **No sexp serialization: pass.** `kicadLoadError` probes via throwaway kicad-cli exports; the revert path restores the exact pre-edit bytes with `writeFile`, no round-tripping.
- **Sync-obligations ledger: pass with the deferral caveat above.** Deferral is a deliberate, spec-documented change, not a silent bypass; the concern is solely that this branch's version drops main's re-open guarantee.
- **Redaction (AC-4.1): pass.** The new `budget-extended` and `work-preserved` events go through `Transcript.event`, and the deferred-obligations section of `summary.md` goes through `writeSummary`; both call `redactSecrets` at write time. `preserveFailedRun` uses `git add -A`, which honors `.gitignore`, so `.env` and `.copperhead/runs/` stay out of the stash in any repo scaffolded per AC-4.3. Note for the design doc: in a repo that does not gitignore `.env`, the stash would capture it; the stash is local-only, so this is a documentation note, not a leak.
- **Non-interactive/CI safety: pass.** `budgetContinuePrompt` in `src/cli.ts` returns `undefined` unless both stdin and stdout are TTYs, so CI and piped runs never receive a callback and keep fail-and-restore by construction. The prompt cannot hang a headless run. Repeated extension cannot loop unattended: each round requires an explicit `y` with fresh numbers, default No.

## Correctness

- The extendable-budget loop is sound: `Math.floor` plus the `Number.isFinite` guard rejects NaN/Infinity callbacks, `turnsUsed` and the final `turn budget exhausted (${budget} turns...)` message use the right values, and `maxTurns` in the stats is deliberately the original budget so the CLI offers a constant increment (D1). Good.
- `nudges = 0` reset after a tool-calling turn correctly converts the stall counter to consecutive-only, and both directions are tested (sporadic empties succeed, three consecutive fail).
- `preserveFailedRun` is carefully reasoned: stage everything, unstage `.copperhead/runs` so `restore()`'s `reset --hard` cannot delete the in-flight audit trail, `stash create` + `stash store`, never throw so preservation can never block the rollback. The comment explaining why unstaging beats an exclude pathspec is exactly the kind of load-bearing comment this codebase does well.
- Minor: in the `edit_file` handler (`src/agent/tools.ts`), `await readFile(abs, 'utf8')` runs before `toolEditFile` for probeable files. If the target `.kicad_sch` does not exist, this throws a raw ENOENT that `dispatchTool` surfaces as `error: ENOENT: no such file...`, replacing whatever friendlier message `toolEditFile` would have produced. Guard with `existsSync` or catch and fall through to `toolEditFile`.
- Minor: deferral classification reads `ctx.config`, which is loaded at run start. A schematic configured mid-run (the create schematic stage does exactly this) means later `record_constraint` calls in the same run still defer. Acknowledged in the PR body as a follow-up; note that main's run-start `reopenDeferredAffects` bounds the damage to one run, another reason to keep main's model.
- Minor: the 5-turns-remaining convergence nudge (`remaining === 5` in `src/agent/loop.ts`) can fire again after each extension, so a twice-extended run sees the message three times. Harmless, acknowledged in the PR body, a `nudgedAtRemaining5` flag would clean it up.
- The Anthropic changes are correct: three breakpoints is within the API's limit of four, string user content is now uniformly block arrays, and the tool-result merge no longer needs the `Array.isArray` check. Counting `cache_read_input_tokens + cache_creation_input_tokens` into `inputTokens` is a real semantic change to the summary and to the continue prompt's `247.3k in` figure: it now reports volume, not a cost proxy. That is the honest choice and D4 documents it, but consider surfacing the cached/uncached split in `summary.md` so a user does not read 247k as billable.
- `stageMaxTurns` sanitization in `src/config.ts` (drop zero/negative/non-integer entries) prevents a config typo from creating an instantly-exhausted stage. Good defensive detail, and tested.
- The probe revert logic (revert, re-probe the reverted content, restore the edit if the file was already broken) correctly avoids the incremental-repair deadlock, and skipping `markTouched` on a reverted edit keeps `filesTouched` truthful.

## Performance

- Prompt caching is the single biggest cost fix here and the breakpoint placement matches Anthropic's recommended pattern for growing conversations.
- The load probe adds roughly one kicad-cli invocation per schematic/board edit (two more on the failure path). Paid only on the mutation path, acceptable.
- `layout-draft.isComplete` reads the whole board file to look for `(footprint`; fine at these sizes.

## Test coverage

Strong overall: 26 new offline tests with a scripted provider, covering grant/decline/absent/throwing callbacks, stash recovery end to end (apply after restore), classifier edge cases (`bom.md` casing, `LAYOUT.md` beating the board keyword), per-entry batch resolution, the mocked-SDK cache assertions (exactly three breakpoints, summed usage), config sanitization, probe scoping, revert, already-corrupt keep, and both drift-exemption directions. The `provider` test seam in `RunOptions` is a reasonable price for making the loop testable.

Gaps worth closing:

- AC-15.28 (missing ERC/DRC report raises kicad-cli's own output) has no test, though `src/kicad/cli.ts` implements it.
- No test exercises a second extension (callback firing twice), which is the D1 repeat-decision path.
- The blank-sheet schematic halt (AC-15.23) is only covered indirectly via the mocked pipeline in `test/create-stage-turns.test.ts`; a direct `isComplete` test against a zero-symbol schematic would pin it.

## Done well

- Failure-path discipline throughout: throwing prompt reads as declined, preservation can never block rollback, audit trail explicitly protected from `reset --hard`, clean trees produce no stash.
- Spec workflow followed properly: proposal, design with real rationale and rejected alternatives, delta specs with WHEN/THEN scenarios mapping 1:1 onto the AC-15 block, SPEC.md updated in step. This is the repo's contract and the PR honors it.
- The already-unloadable-keeps-repair-edits carve-out shows real understanding of the failure mode it prevents.
- Comments explain why, not what, at every non-obvious decision point.

## Summary of asks

1. Rebase onto main and resolve the conflicts; this is mandatory, the PR cannot merge as is.
2. Reconcile deferral with main's registry-persisted `reopenDeferredAffects` design; update D3 and the agent-core delta spec so the branch does not weaken SPEC.md's re-open guarantee.
3. De-duplicate the batching prompt, empty-pattern search, not-applicable ERC/DRC messages, and transcript-dir re-creation against what main already ships.
4. Fix the `edit_file` ENOENT pre-read edge; consider the cached-token split in summaries and the small test gaps above.

animesh-chouhan:
## Reconciled with main (`0d71e15`)

Main advanced under this PR — it absorbed the run-observability stack (#28: runmeta, progress renderer, `exitPath`/`run-end`, restore-failure handling) and, importantly, a **persisted affects-revisit deferral** (`classifyAffectsTarget` + `reopenDeferredAffects` writing `deferred[]` into `constraints.json`) that independently solves what this PR's in-run deferral did. Merged main in and resolved the 8-file conflict as follows:

- **Base on main** everywhere the two overlapped; grafted this PR's unique features on top: budget-continue prompt + extendable turn loop, `preserveFailedRun` stash before rollback, Anthropic prompt caching, `resolve_affected` batch form, `stageMaxTurns`, the `edit_file` loadability probe, the drift bootstrap carve-out + `check` warning, and the consecutive-only stall counter.
- **Dropped this PR's in-run deferral** (`ObligationsLedger.defer`/`describeDeferred`, `missingArtifactReason`, the summary section) in favor of main's stronger persisted+reopen mechanism — which does the cross-run reconciliation this PR's own design had listed as a non-goal. The `resolve_affected` batch form now clears the obligations main opens.
- Deduped a duplicate `RunOptions.provider`; renamed a shadowing `stats` local; took main's search/ERC/DRC message wording (one copy of the same feature).
- **SPEC.md keeps both** the AC-8 (observability) and AC-15 blocks; AC-15.7/15.8/15.11 dropped as superseded. Delta specs, design D3, proposal, and tasks updated to record the reconciliation; tests realigned to main's contract.

Post-merge: **137 passed / 7 skipped**, `tsc` + `npm run build` clean, `openspec validate` passes. PR now shows MERGEABLE.

🤖 Reconciled with [Claude Code](https://claude.com/claude-code)

animesh-chouhan:
## pr-review skill — PR #18 at `0d71e15`

**Verdict: approve** (two nits, both non-blocking). Reviewed against the repo's SPEC.md invariants and spec workflow.

*Disclosure: this reviewer authored the reconciliation merge and two of the five commits, so every check was re-derived from the code and each candidate finding was actively refuted rather than assumed.*

### Verification (actual, run on this exact commit)

- `tsc --noEmit`: clean
- `vitest run`: **137 passed / 7 skipped** (the 7 are key-gated live-LLM tests, by design)
- `openspec validate turn-budget-continue-and-loop-efficiency`: valid

### Invariant checks — all performed, all pass

- **Spec-gated in** ✓ — `edit_file`/`write_file` stay `requiresUnlock: true`; `availableTools` gating untouched. Mutation tools remain structurally absent until an OpenSpec proposal validates.
- **Verification-gated out** ✓ — the budget-continue path extends turns but still must reach `finish` with ERC/DRC clean; decline/exhaust fails and rolls back. The edit-probe "keep already-corrupt file" branch calls `markTouched`, re-opening the ERC obligation, so a corrupt file cannot ride through to a clean finish.
- **`check`/`verify` LLM-free & network-free** ✓ — the new `emptySchematicWarning` in `check.ts`'s chain resolves to `drift.ts` → `kicad/sexp` (read-only) + `fs` only. No provider, key, or network.
- **No sexp serialization** ✓ — `kicadLoadError` probes via a `kicad-cli` export subprocess; `edit_file` reverts with a byte-identical `writeFile(before)`. The parser stays read-only.
- **Sync-obligations ledger** ✓ — `resolve_affected`'s batch form routes every entry through `ledger.clear`; commit still refuses on open obligations.
- **Secrets** ✓ — `redactSecrets` wraps every transcript event (including the new `budget-extended`/`work-preserved` events); `.gitignore` keeps `.env` and `.copperhead/runs/`.
- **Spec coherence** ✓ — SPEC.md carries both AC-8 (from main) and AC-15; superseded AC-15.7/8/11 dropped; delta specs, design D3, proposal, and tasks moved in step; `openspec validate` passes.

### Refuted (checked, not a finding)

- **Board-probe false-revert** — tested empirically: a valid, footprint-less `.kicad_pcb` exports `pos` with exit 0, so early layout edits are not wrongly reverted by the probe.

### Findings (nits only)

1. **Progress counter shows a stale denominator after a budget extension** — `src/agent/loop.ts`, the `r.turnStart(turn + 1, maxTurns, …)` call passes `maxTurns` while the loop bound uses the grown `budget`, so the status line reads "turn 41/40" after an extension. Cosmetic; pass `budget` to fix.
2. **Two probe paths lack direct unit tests** — `test/create-hardening.test.ts` covers the `.kicad_sch` probe (revert + already-corrupt keep) but not the `.kicad_pcb` path or `layout-draft.isComplete`. The `runCreate` contract-halt itself is covered by `test/create-stage-turns.test.ts`. Low priority (shared code path with the tested schematic probe).

Neither blocks merge. The PR does what its updated description claims, the reconciliation preserved all of main's run-observability and persisted-deferral work, and there is no scope creep.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

Issue #21, opened 2026-07-21
init-installed pre-commit hook deadlocks the create pipeline: repo-wide drift check blocks stage commits while docs legitimately lead the schematic


animesh-chouhan:
## Summary

`copperhead init` installs a pre-commit hook that runs `copperhead check`. During `copperhead create`, this hook blocks the pipeline's **own** stage commits: the docs-first stages (part-selection writes BOM.md) legitimately run before the schematic stage populates the sheet, so the repo-wide drift check fails and every stage commit dies with exit 1 until the schematic exists.

Related to #19 (create can't bootstrap a KiCad project) — this was hit immediately after working around #19 by hand-scaffolding a blank schematic and running `copperhead init`.

## Reproduction (observed in copperhead-test, 2026-07-21)

1. `copperhead create` runs spec-seed → architecture → part-selection on a fresh repo. These stages write SPEC.md, SUBSYSTEMS.md, BOM.md. No hook exists yet (no `init` was run), and no schematic is configured, so `check_drift` reports "vacuously clean". All commits succeed.
2. Schematic stage refuses (#19). To unblock: scaffold a blank `usb-c-power-breakout.kicad_pro`/`.kicad_sch` by hand and run `copperhead init` — which sets `config.schematic` **and installs the pre-commit hook**.
3. Re-run `copperhead create`. The spec-seed stage finishes (its run-level ledger is clean — it touched no docs) and tries to commit:

```
Command failed with exit code 1: git commit -m 'copperhead: create pipeline stage: spec-seed...'

ERC ✓
DRC skipped (no board configured)
drift: BOM.md claims "J1 exists" but actual is "J1 not in schematic"
drift: BOM.md claims "F1 exists" but actual is "F1 not in schematic"
drift: BOM.md claims "D1 exists" but actual is "D1 not in schematic"
drift: BOM.md claims "R2, R3 exists" but actual is "R2, R3 not in schematic"
drift: BOM.md claims "D2 exists" but actual is "D2 not in schematic"
drift: BOM.md claims "R1 exists" but actual is "R1 not in schematic"
drift: BOM.md claims "J2 exists" but actual is "J2 not in schematic"
drift: BOM.md claims "J3 exists" but actual is "J3 not in schematic"
constraints ✓
```

The pipeline is now deadlocked: no stage can commit until the schematic contains every BOM refdes, but the schematic stage can only get there through commits.

## Why it happens

Two consistency gates disagree about scope and timing:

[...]

PR #26, opened 2026-07-21
Record full run metadata and add interactive progress output

### LINKED ISSUE #22: Feature: record full run metadata (model, config, versions, git state) at run start — in the transcript, summary, and CLI output
## Problem

A run's context is mostly invisible. Today `run-start` records only three fields (`src/agent/loop.ts:168`):

```ts
await transcript.event('run-start', { request: opts.request, model: opts.model, provider: provider.name });
```

`summary.md` shows request/outcome/change/tokens, and the live CLI output shows nothing about the environment at all — just `stage spec-seed: running`. When triaging a failed or surprising run after the fact (e.g. the schematic-stage refusal in #19, or the drift gate-gaming in #21), you can't answer basic questions from the artifacts alone: which copperhead version ran (repo build vs stale global install)? what turn budget applied? what did the repo look like when the run started? was a schematic even configured?

## Request

Emit a structured metadata block **before the first agent turn**, written to all three surfaces:

1. `run-start` event in `transcript.jsonl` (machine-readable, complete)
2. a `## Environment` section in `summary.md`
3. a one/two-line header in live CLI output (at minimum: version, model/provider, stage, turn budget)

## Suggested metadata

**Tooling & versions**
- copperhead version + install path (catches the stale-global-install problem: running `dist/` from the repo vs `~/.nvm/.../node_modules/copperhead`)
- `kicad-cli` version (ERC/DRC behavior differs across KiCad 8/9/10), openspec CLI version if present
- Node version, OS/platform

**Model & provider**
- model id, provider, and how the model was selected (flag > `COPPERHEAD_MODEL` > config > available-key fallback — the winning source matters when debugging "why did this run on gpt-5?")
- sampling/reasoning params if set; API base URL if overridden

**Run identity & mode**
- run id, ISO timestamp, command (`do`/`create`/`sync`), stage name + position in the create pipeline
- interactive vs autonomous, and whether human gates were auto-approved
- for `create`: brief path + content hash (detects a brief edited mid-pipeline)

**Config snapshot** (resolved, not just the file)
- `schematic`/`board` paths (or null — #19 would have been diagnosable from `schematic: null` in the header alone), `docs` dir, `maxTurns`/stage override, `maxRepairCycles`, budgets verbatim

**Repo state**
- git commit, branch, dirty/clean + count of uncommitted files at run start
- whether the copperhead pre-commit hook is installed (relevant to #21)
- counts of open constraints and prior runs

**Post-run addenda** (same event stream, at finish)
- tokens in/out (already in summary) plus per-stage/per-turn breakdown and estimated cost
- turns used vs budget, repair cycles used vs budget, wall-clock duration
- exit path: done / refuse / turn-budget-exhausted / commit-failed (the `git add -A` failure mode from the .history/ embedded repo surfaced only as a raw exit-128 stack — it should land in summary.md as an outcome)

## Why it pays off

Every debugging session in copperhead-test so far (#19, #21) started by reverse-engineering exactly this information from transcripts, `which copperhead`, and `git status`. Runs are long, non-deterministic, and cost real tokens — a cheap deterministic header makes each one self-describing and comparable across model/config changes.

animesh-chouhan:
Closes #22.

> [!NOTE]
> Stacked on #17 (`Preflight hardening`) — this branch builds on those changes (`src/util/git.ts`, `src/agent/loop.ts`), so its 5 commits appear here until #17 merges. Only the top commit (`3ee3eed`) is new.

## What

Every agent-loop run is now self-describing from its artifacts alone, and the live CLI output got a Claude Code-style interactive upgrade. Planned and implemented via the `record-run-metadata` OpenSpec change included in this PR (proposal / design / delta spec / tasks), with binary acceptance criteria added to SPEC.md as **AC-8.1 … AC-8.10**.

### Run metadata (issue #22)
- A deterministic, LLM-free metadata block is collected once before turn 1 — copperhead version + install path (catches the stale-global-install problem), `kicad-cli`/Node/platform versions, model + provider + **selection source** (which level of flag > `COPPERHEAD_MODEL` > config > key-fallback won), run identity (run id, command, stage `k/N` + brief sha256 for `create`), the **resolved** config snapshot (`schematic: null` is a present key — #19 would have been diagnosable from the header alone), and git state (commit, branch, dirty count, pre-commit hook presence, constraint/prior-run counts).
- Rendered from one source object onto three surfaces: the `run-start` transcript event (complete, machine-readable; old field names preserved), a `## Environment` section in `summary.md`, and a two-line CLI header.
- Every probe is individually caught: a failing probe yields `null` for that field, never a failed run.

### Post-run addenda
- Every terminal branch emits a `run-end` event and a `## Run stats` summary section: exit-path enum (`done`/`refused`/`turn-budget-exhausted`/`repair-cycles-exhausted`/`commit-failed`/`provider-error`/`stalled`), turns and repair cycles vs budgets, per-turn token breakdown, wall-clock duration, plus one final CLI outcome line.
- A failing end-of-run `git commit` (the embedded-repo exit-128 case from #21's triage) now rolls back per the snapshot contract and reports `commit-failed` in `summary.md` instead of escaping as a raw stack trace. An archive-commit failure after the verified commit exists is caught and reported as a warning without discarding the commit.

### Live output
- New `ProgressRenderer` with two modes chosen once at startup: **interactive** on a TTY (bottom-pinned status line redrawn in place — spinner while the provider call is in flight, elapsed time, `turn 7/40`, cumulative tokens — with assistant text and tool results scrolling above it; cursor restored on exit/Ctrl-C) and **plain** (`[turn 7/40 · 12.3k in / 4.1k out]` markers, zero ANSI escapes) when piped, with `--json`, or with the new global `--plain` flag.
- Hand-rolled ANSI (~60 lines); no new dependencies.

## Test plan

- 11 new offline tests (no API key): metadata collection + probe degradation, both renderer modes (including a fake-TTY test for redraw/cursor hygiene), exit paths (budget exhaustion, refusal, deterministic commit failure via a failing pre-commit hook), and redaction of the new surfaces (extends AC-4.1).
- Full suite: 102 passed, 7 pre-existing LLM-gated skips. `npm run build` clean, `openspec validate record-run-metadata` passes.
- Manually verified plain mode end-to-end via a scripted-provider demo; interactive mode exercised through the fake-TTY unit test (worth one live-terminal sanity run on review).


animesh-chouhan:
## Review (head `309f29b2`)

**Scope note:** the five commits under `3ee3eede` belong to stacked PR #17 and are reviewed there. This review covers `3ee3eede` (the run-metadata + interactive-output work) and the follow-up fix `309f29b2`, which addresses the two blocking findings from the first review pass. Disclosure: `309f29b2` was authored during that same review session, so treat its verification below as author-adjacent. CI is green (`cla`, `test`: 104 passed / 7 LLM-gated skips).

### First-pass findings, now resolved

- **Interactive `create` went silent after stage 1.** `InteractiveRenderer.finish()` used to tear the renderer down permanently, and `cli.ts` shares one renderer across all pipeline stages. Fixed in `309f29b2`: `finish()` now suspends (stops the spinner, clears the status line, restores the cursor) and the next `turnStart()` re-arms it with a per-run elapsed clock. Edge cases traced: stage 2's header logs while idle with no stale status line, SIGINT between stages is a clean no-op suspend before `exit(130)`, and the spinner timer cannot leak. A new loop-level test runs two `runAgentLoop` calls through one renderer and asserts the second run's header, status line, and outcome line all render.
- **`sync --json` stdout purity regression.** The always-passed renderer sent progress to stdout, undoing the previously silent resolve phase. Fixed: `makeRenderer` routes progress to stderr under `--json`, so stdout carries only the JSON (report for `sync`, `RunResult` for `do`, which is an improvement over its old interleaved output). Covered by a spy-based test.
- Side effect of the suspend design: `create`'s stage messages now only print while the renderer is idle, so they no longer risk colliding with the pinned status line.
- Spec artifacts moved with the code per CLAUDE.md: delta spec requirement + two new scenarios, SPEC.md AC-8.8/8.9 extended, `openspec validate record-run-metadata` passes.

### Correctness spot-checks that passed

- `uncommittedCount` split-on-newline is safe: the `git()` helper trims stdout.
- All three `resolveModel` call sites updated for the new `{model, source}` return.
- Commit-failure path runs `restore()` after a failed `git commit` (test verifies a clean tree); the archive-failure path deliberately preserves the verified commit, a good trade and well-commented.
- Redaction: metadata flows through `Transcript.event`/`writeSummary`, both redacting at write time; the new test proves it on both surfaces. Old `run-start` field names survive at the top level as claimed.

### Remaining minor items (fine to defer)

- Dead `'unknown'` branches: the `preCommitHookInstalled` and `priorRuns` probes in `runmeta.ts` coerce `null` to `false`/`0`, so their `| null` types and the "unknown" rendering branches are unreachable. Either drop the coercion (a permission error would then read "unknown" rather than "absent") or tighten the types.
- The hook probe reads `.git/hooks/pre-commit` directly, so it reports "absent" in git worktrees where `.git` is a file.
- `statusText()` truncates by JS string length, not display width; cosmetic.
- `- **Per turn:**` in summary.md is a single line, unwieldy at 40 turns.
- Under `--json`, `create`'s stage-level messages (`log: console.log` in `cli.ts`) still go to stdout while loop progress goes to stderr. Harmless today since `create` emits no JSON, but worth closing if `create --json` ever grows a JSON result.

### Verdict

The metadata/exit-path half was ready as-is; both merge-blocking renderer issues are fixed, tested at the right level, and reflected in the spec artifacts. Ready to merge once #17 lands and the stack rebases. One outstanding manual check: a live-terminal `create` run to eyeball the interactive stage handoff (the fake-TTY tests cover the mechanics, not real terminal behavior).

PR #27, opened 2026-07-21
fix: preserve failed-run audit trails during rollback

### LINKED ISSUE #25: Rollback deletes the in-flight run's audit trail (preserveFailedRun stages it, reset --hard removes it), then the summary writer crashes the process
## Summary

When a run fails and rolls back (e.g. maxRepairCycles exhausted), the run's own `.copperhead/runs/<id>/` directory is deleted mid-run, and the process then dies with an unhandled `ENOENT` trying to write `summary.md` into it. The audit trail of the failed run — exactly the evidence `restore()` promises to keep — is destroyed, and the create pipeline exits with a crash instead of a clean stage failure.

Observed in copperhead-test: schematic stage hit 48 ERC violations wiring the board, exhausted 5 repair cycles, rolled back, crashed:

```
  [run_erc] ERC: 48 violation(s)
  [read_file] 1: # Pinout
ENOENT: no such file or directory, open '.../.copperhead/runs/2026-07-21T13-18-28-851Z/summary.md'
```

## Mechanism

Two functions in `src/util/git.ts` interact badly:

1. `preserveFailedRun()` runs first and does `git add -A` — staging the (untracked) in-flight run directory.
2. `restore()` then does `git reset --hard <snap>` followed by `git clean -fd -e .copperhead/runs`. The `-e` exclusion only protects **untracked** files — but step 1 just made the run dir *staged*, and a staged-but-not-in-HEAD path is deleted by `reset --hard` before `clean` ever runs.

So the protection exists, and is defeated by the sibling function one line earlier. Everything else survives (the failed work does land in the named stash), only the audit trail dies — taking the process with it.

## Fix applied locally

- `preserveFailedRun()` now unstages the audit trail before creating the stash: `git add -A` then `git reset -q -- .copperhead/runs`. (An `:(exclude)` pathspec doesn't work: `git add` errors outright when a pathspec touches gitignored paths, and `.copperhead/runs/` is gitignored in some target repos but tracked in others.)
- `Transcript.event()`/`writeSummary()` now `mkdir -p` their directory before writing — losing an event is acceptable; crashing the run to report one is not.
- Suite passes 89/89 (including the AC-15.16 stash-preservation test, which caught a first attempt that used the exclude pathspec).

## Related hardening worth considering

- `restore()` could snapshot-protect `.copperhead/runs` structurally (e.g. move it aside and back) rather than relying on every caller to keep it unstaged.
- The crash surfaced because nothing between the agent loop and the CLI catches transcript-write failures; a failed run should still end as a *reported* failure.

fyzanshaik:
## What changes

This fixes #25: a failed agent run could delete its own audit directory, then crash while attempting to write its failure summary.

## How rollback preserves the audit trail

`restore()` now treats `.copperhead/runs/` as protected evidence rather than ordinary working-tree content:

1. Before calling Git, it copies the complete runs directory to a temporary directory outside the repository.
2. It runs the normal rollback sequence: `git reset --hard <snapshot>`, `git clean`, then applies the saved dirty-tree stash when present.
3. In a nested `finally` block, it copies the saved runs directory back into `.copperhead/runs/`. This executes even if reset, clean, or stash application throws, so Git can never remove the evidence before it is restored.
4. A final cleanup removes the temporary backup.

This covers the staged-path failure mode: `git clean -e .copperhead/runs` only protects untracked paths, while `git reset --hard` deletes a staged, not-in-HEAD run directory before `clean` gets a chance.

`Transcript.event()` and `Transcript.writeSummary()` also create their parent run directory immediately before writing. If an earlier failure or an external cleanup removed it, reporting the failure no longer raises `ENOENT`.

## Tests

- rollback retains a staged in-flight `transcript.jsonl`
- rollback still restores the audit trail when stash application fails
- transcript event and summary writes recover after the run directory is removed

Closes #25.

## Verification

- `npx vitest run test/safety.test.ts` — 16 passed
- `npm run typecheck` — passed
- `PATH="/Applications/KiCad/KiCad.app/Contents/MacOS:$PATH" npm test` — 73 passed, 7 skipped, 2 existing failures outside this change:
  - `test/gating-sync.test.ts`: fixture ERC does not reach the expected clean state
  - `test/init-check.test.ts`: clean fixture check returns `ok: false` under KiCad 10.0.4

### DIFF
--- a/src/agent/transcript.ts
+++ b/src/agent/transcript.ts
@@ -38,6 +38,7 @@ export class Transcript {
 
   async event(type: string, data: unknown): Promise<void> {
     const line = redactSecrets(JSON.stringify({ ts: new Date().toISOString(), type, data }));
+    await mkdir(this.dir, { recursive: true });
     await appendFile(this.jsonlPath, line + '\n', 'utf8');
   }
 
@@ -72,6 +73,7 @@ export class Transcript {
     }
     if (s.detail) lines.push('', '## Detail', '', s.detail);
     const out = path.join(this.dir, 'summary.md');
+    await mkdir(this.dir, { recursive: true });
     await writeFile(out, redactSecrets(lines.join('\n') + '\n'), 'utf8');
     return out;
   }
--- a/src/util/git.ts
+++ b/src/util/git.ts
@@ -1,4 +1,8 @@
 import { execa } from 'execa';
+import { cp, mkdir, mkdtemp, rm } from 'node:fs/promises';
+import { existsSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
 
 export interface GitSnapshot {
   head: string;
@@ -43,10 +47,49 @@ export async function snapshot(repo: string): Promise<GitSnapshot> {
  * (.copperhead/runs/) survives rollback: it is the evidence of what failed.
  */
 export async function restore(repo: string, snap: GitSnapshot): Promise<void> {
-  await git(repo, ['reset', '--hard', snap.head]);
-  await git(repo, ['clean', '-fd', '-e', '.copperhead/runs']);
-  if (snap.stash) {
-    await git(repo, ['stash', 'apply', snap.stash]);
+  // `git clean -e` only protects untracked paths. A run directory can become
+  // staged (for example while preserving failed work), and `reset --hard`
+  // deletes such paths before clean runs. Copy it outside the repository so
+  // the audit trail survives regardless of its index state.
+  const runs = path.join(repo, '.copperhead', 'runs');
+  let backupRoot: string | null = null;
+  let backup: string | null = null;
+  try {
+    try {
+      backupRoot = await mkdtemp(path.join(tmpdir(), 'copperhead-runs-'));
+      backup = path.join(backupRoot, 'runs');
+      if (existsSync(runs)) await cp(runs, backup, { recursive: true });
+    } catch (err) {
+      backup = null;
+      console.warn(`warning: could not preserve failed-run audit trail before rollback: ${(err as Error).message}`);
+    }
+
+    try {
+      await git(repo, ['reset', '--hard', snap.head]);
+      await git(repo, ['clean', '-fd', '-e', '.copperhead/runs']);
+      if (snap.stash) {
+        await git(repo, ['stash', 'apply', snap.stash]);
+      }
+    } finally {
+      if (backup && existsSync(backup)) {
+        try {
+          await mkdir(path.dirname(runs), { recursive: true });
+          // Restored runs are intentionally untracked; their audit contents
+          // are ignored by the target-repository convention.
+          await cp(backup, runs, { recursive: true, force: true });
+        } catch (err) {
+          console.warn(`warning: could not restore failed-run audit trail: ${(err as Error).message}`);
+        }
+      }
+    }
+  } finally {
+    if (backupRoot) {
+      try {
+        await rm(backupRoot, { recursive: true, force: true });
+      } catch (err) {
+        console.warn(`warning: could not clean failed-run audit backup: ${(err as Error).message}`);
+      }
+    }
   }
 }
 
--- a/test/safety.test.ts
+++ b/test/safety.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { mkdtemp, readFile, writeFile } from 'node:fs/promises';
+import { mkdir, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
 import { resolveInRepo, SandboxError, isKicadFile } from '../src/util/paths.js';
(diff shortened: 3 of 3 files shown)


fyzanshaik:

fyzanshaik:
recheck

animesh-chouhan (commented):
## Review

### Overview

This fixes #25 on two fronts: `restore()` now copies `.copperhead/runs/` to a temp directory before `git reset --hard`/`git clean` and copies it back in a nested `finally`, so a staged run directory survives rollback even when the rollback itself throws. `Transcript.event()` and `writeSummary()` recreate their run directory before writing, so failure reporting no longer dies with `ENOENT` if the directory vanished. Both fixes restore behavior the code already promised (the `restore()` docstring and AC-3.6 say the audit trail survives rollback), so no spec change is needed. The diagnosis in the PR body is correct: `git clean -e` only protects untracked paths, and `reset --hard` deletes a staged, not-in-HEAD path before `clean` runs.

### What's good

- The nested `finally` structure is right: the copy-back runs even if `reset`, `clean`, or `stash apply` throws, and the failed-rollback test (`stash: 'not-a-stash'`) locks that in.
- The comment in `restore()` explains the non-obvious constraint (why `clean -e` is insufficient) rather than narrating the code.
- The `mkdir({ recursive: true })` guards in `Transcript` are idempotent and simpler than catch-and-retry.
- Tests cover the three failure modes named in the PR body, and the fixture-repo helper keeps them hermetic.

### Issues and risks

1. **A backup failure now blocks rollback entirely.** If the initial `cp(runs, backup)` throws (permissions, full tmpfs, exotic file in a run dir), `restore()` throws before `git reset --hard` ever runs, leaving the user's tree in the failed, mutated state. That inverts the priority: rollback is the primary safety rail (AC-3.6), evidence preservation is secondary. Consider wrapping the backup `cp` in a try/catch that logs and proceeds with rollback anyway; the `-e .copperhead/runs` exclusion still protects the common untracked case.
2. **Error masking in `finally`.** If a git command throws and the copy-back (or the outer `rm`) also throws, the second error replaces the first, hiding the real rollback failure. Wrapping the `finally` bodies in their own try/catch (log and swallow) would preserve the original error.
3. **Whole-directory copy on every rollback.** The backup copies all historical runs, not just the in-flight one, so cost grows with `.copperhead/runs/` over the project's life, and each rollback copies it twice. Two lighter options if this ever matters: back up only the staged paths (`git diff --cached --name-only -- .copperhead/runs`), or skip the copy entirely by unstaging first (`git reset -q <head> -- .copperhead/runs` before `reset --hard`, which makes the files untracked and lets the existing `-e` exclusion protect them). The copy approach is more robust against weird index states, so this is a judgment call, not a blocker.
4. **Temp residue on hard kill.** If the process dies between backup and the outer `rm`, a `copperhead-runs-*` directory lingers in the OS temp dir. Contents are redacted at write time (AC-4.1) and `mkdtemp` creates it mode 0700, so exposure is low, just worth knowing.

### Minor

- After copy-back, previously staged run files become untracked. That's fine because `.copperhead/runs/` is gitignored (AC-4.3), but a one-line comment noting the intentional stage-state loss would help the next reader.
- Optional test additions: assert the temp backup root is removed after `restore()`, and add a plain untracked-run-file case as a regression guard for the `-e` path (the existing byte-identical test doesn't touch `.copperhead/runs/`).

### Verdict

Correct fix, well tested, matches project conventions. Item 1 (backup failure blocking rollback) is the one thing I'd resolve before merge, even if the resolution is a deliberate "evidence beats rollback" decision recorded in the code comment.

animesh-chouhan (commented):
## Re-review (after update `ff3f792c`)

The update addresses the substantive feedback from my previous review:

- **Backup failure no longer blocks rollback** (previous item 1): the `mkdtemp`/`cp` backup is now wrapped in try/catch, logs a warning, and rollback proceeds. The priority is now correct: AC-3.6 rollback runs even when the temp copy cannot be made, degrading to the existing `-e .copperhead/runs` protection for untracked paths. The new test ("still rolls back when temporary audit backup storage is unavailable") locks this in, and setting `TMPDIR` to a missing path does reach the degraded branch since `os.tmpdir()` reads the env var at call time.
- **Error masking** (previous item 2): the copy-back and the temp cleanup now catch and warn instead of letting a secondary error replace the original git failure. The failed-rollback test still asserts the git error propagates, which confirms the swallowing is scoped to the recovery steps only.
- **Stage-state comment** (previous minor): the copy-back now notes that restored runs are intentionally untracked.

Remaining nits, none blocking:

- The `TMPDIR` override in the new test exercises the degraded path on Linux/macOS but not Windows (where `os.tmpdir()` uses `TEMP`/`TMP`). The test still passes there, it just silently tests the happy path instead. Fine given the project's platforms, just worth knowing if CI ever adds Windows.
- The three `console.warn` calls are reasonable for a CLI, but if the agent loop has a render/output layer these could route through it later so warnings land in the transcript too.
- The whole-runs-directory copy per rollback (previous item 3) is unchanged and remains an acceptable judgment call.

Looks good to merge.

animesh-chouhan:
Congrats on being the first outside contributor to copperhead, @fyzanshaik! 🎉 Great first PR to kick things off with: real bug, careful fix and tests to match. Lmk how can I reach out to you with your Zomato Voucher.

Issue #24, opened 2026-07-21
Feature: re-run individual pipeline stages on demand, with a stage dependency graph that invalidates and updates downstream work


animesh-chouhan:
## Problem

The create pipeline is strictly forward-only: `copperhead create` resumes at the first incomplete stage and there is no way to deliberately re-run one stage. But design iteration is exactly that: the BOM changes after the schematic exists, the spec's budgets tighten after layout, a footprint swap invalidates the outline. Today the options are bad — hand-delete artifacts until the completion probes read "incomplete" (fragile, see #23), or drive everything through free-form `copperhead do` requests that don't know they're redoing a stage.

### 1. Targeted stage re-run

```
copperhead create --stage part-selection          # re-run one stage
copperhead create --from schematic               # re-run a stage and everything after it
```

The stage runs with its normal stage prompt and gates, but against the *existing* artifacts (revise, don't recreate).

### 2. An explicit stage dependency graph

The pipeline order already implies it; make it data:

```
spec-seed → architecture → part-selection → schematic → layout-draft → outputs
                                              └→ firmware (via PINOUT.md) → devplan
```

Each edge should name *what* flows across it (architecture reads SPEC budgets; schematic reads BOM refdes/values/footprints; firmware reads PINOUT.md; outputs reads the board), so invalidation can be precise rather than "everything after".

### 3. Downstream invalidation + propagation

When a stage is re-run and its outputs actually change, downstream stages that consume the changed artifacts are marked stale and either:

- **auto-update** (default in autonomous mode): each stale stage re-runs in dependency order, scoped to reconciling with the upstream change ("BOM changed R1 10k→4.7k: update schematic value, re-run ERC, refresh PINOUT/outputs"), or
- **report-only** (`--dry-run` / interactive): print the stale set and what changed, let the human pick.

No change in a stage's outputs → downstream stays valid (content-hash the crossing artifacts, in the spirit of init's `generatedHashes`).

## Building blocks already in the codebase

- `record_constraint`/`resolve_affected` + the obligations ledger already implement fine-grained affects-propagation *within* a run — this is the same idea lifted to stage granularity.
- `checkDrift` already detects BOM/PINOUT-vs-schematic disagreement, i.e. a stale schematic edge.
[...]

PR #29, opened 2026-07-21
fix: enforce create stage completion contracts

### LINKED ISSUE #23: create's stage-completion inference uses shallow existence checks — blank bootstrap files and init's own doc templates mark stages complete
## Summary

`copperhead create` infers stage completion from repo state (design D10, `src/commands/create.ts`), but several `isComplete` probes are shallow existence checks that false-positive on states where the stage's actual work never happened. Observed consequence in copperhead-test: **the pipeline skipped the schematic and layout-draft stages entirely** — over a schematic with zero symbols — and went straight to outputs, which then refused for a missing `.kicad_pcb`.

Third bug in the bootstrap series: #19 (can't create KiCad files), #21 (drift gates deadlock while docs lead the schematic), and now this — the resume heuristics can't tell "bootstrapped" from "done".

## Observed sequence

1. Worked around #19 by hand-scaffolding a blank `usb-c-power-breakout.kicad_sch` + running `copperhead init`.
2. Re-ran `copperhead create`. Output:
   - `stage schematic: already complete (resuming past it)` — `isComplete` is just `config.schematic` set + file exists; a blank sheet passes.
   - `stage layout-draft: already complete` — `isComplete` checks LAYOUT.md for the literal `## Draft quality` heading, **and init's generated LAYOUT.md template contains exactly that heading** as a placeholder. Running `init` retroactively marks layout-draft complete.
   - `stage outputs: running` → refused: `usb-c-power-breakout.kicad_pcb is missing, so Gerbers/Drill, DXF, STEP, and PCB SVG cannot be exported`.

## The general problem

Two independent flaws compound:

1. **Existence ≠ completion.** File-on-disk (`schematic`) or dir-on-disk (`outputs/` — a partially-failed export run also satisfies it) says nothing about whether the stage's contract was met.
2. **init's templates contain create's completion markers.** `generateDocs` in `src/memory/scaffold.ts` emits SPEC.md with `## Budgets` (spec-seed's marker) and LAYOUT.md with `## Draft quality` (layout-draft's marker). On any init-ed repo, create silently skips those stages. The probes and the templates were written against each other with no shared contract.

## Applied locally (stopgap, in this repo's working tree)

- `schematic.isComplete`: file must exist **and `listSymbols(...) > 0`** — a blank bootstrap sheet no longer counts.
- `layout-draft.isComplete`: additionally requires `config.board` set, the board file present, and at least one `(footprint` on it.
- Full test suite passes (89/89) with both changes.

## Suggested durable fixes

- Make each stage's `isComplete` check the stage's *contract*, not an artifact's existence: spec-seed → budgets recorded in constraints.json; schematic → symbols present + last ERC clean; layout-draft → footprints placed + DRC clean; outputs → expected artifact set present and non-empty; firmware → build marker.
- Alternatively (or additionally): record explicit stage completion in `.copperhead/` state at commit time — the pipeline already commits per stage, so the commit is the natural completion record — and use repo-state probes only as a fallback for repos not built by `create`.
- De-collide init: templates must not contain the literal completion markers, or markers must be something init never emits (e.g. a filled section vs an empty placeholder).

fyzanshaik:
## What

Strengthens `copperhead create` completion contracts for the spec-seed, schematic, and layout stages. A generated template, empty board, or malformed KiCad file cannot be treated as a usable completed artifact.

## Why

This addresses the stage-contract portion of #23. The previous checks accepted `SPEC.md` with only a `## Budgets` heading and an initialized layout template. It does not close #23: outputs and firmware require their own contract design.

## Design

- `spec-seed` requires non-template content under the exact `## Budgets` section.
- Schematic completion remains symbols plus drift-clean docs plus passing ERC. Read or parse failures safely mean incomplete.
- Layout completion requires a parseable board containing a real top-level footprint and non-template content under `## Draft quality`.
- A stage that reports success but misses its contract now logs the specific required artifact before recovery or stop.
- The boolean `isComplete` signature is unchanged for compatibility with the related PRs.

## Spec / OpenSpec

Updated [SPEC.md](openspec/specs/SPEC.md) and the active [create-pipeline delta](openspec/changes/build-copperhead-phase-1/specs/create-pipeline/spec.md) to define repo-state completion contracts and post-run rechecks. OpenSpec validation could not run because the `openspec` executable is not installed by this repository after `npm ci`.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Focused contracts | `npx vitest run test/create-stage-contracts.test.ts test/create-stage-turns.test.ts test/create-resilience.test.ts` | pass: 8 tests |
| Unit + offline | `PATH="/Applications/KiCad/KiCad.app/Contents/MacOS:$PATH" npm test` | known baseline: 329 passed, 14 skipped, 2 unrelated KiCad 10.0.4 fixture failures in `gating-sync` and `init-check` |
| Live-LLM (opt-in) | keyed on `OPENAI_API_KEY` | n/a: no live provider run for this deterministic contract change |

New coverage rejects initialized placeholders and malformed schematic/board files, accepts filled positive artifacts, and verifies actionable unmet-contract recovery output.

- Sandbox variant used (`create` / `edit`): create
- Commands run and outcome: n/a. The post-agent acceptance path requires a live keyed provider; it is exercised deterministically through the focused `runCreate` and contract tests above.

```text
$ npm run build
passed
$ npm run lint:md
passed: 109 Markdown files
```

## Docs

Updated the normative SPEC and active OpenSpec delta.

---

## Invariant checklist

- [ ] **Spec-gated in**: N/A, no tool-list or proposal-gate behavior changed.
- [x] **Verification-gated out**: schematic completion still requires ERC, and an unmet contract stops advancement.
- [ ] **`check`/`verify` stays LLM-free and network-free**: N/A, their module graph is unchanged.
- [x] **No sexp serialization**: the KiCad parser is used read-only to inspect board footprints.
- [ ] **Sync-obligations ledger**: N/A, ledger behavior is unchanged.
- [ ] **Secrets**: N/A, transcript and environment handling are unchanged.
- [ ] **Spec coherence**: SPEC and the active delta changed together; `openspec validate` is unavailable locally because the CLI is not installed after `npm ci`.

### DIFF
--- a/openspec/changes/build-copperhead-phase-1/specs/create-pipeline/spec.md
+++ b/openspec/changes/build-copperhead-phase-1/specs/create-pipeline/spec.md
@@ -27,6 +27,13 @@ Pipeline state SHALL live in the repo (docs + files + gate results), so a killed
 - **WHEN** `create` is killed after the BOM stage and re-run
 - **THEN** it skips spec/architecture/BOM and resumes at the schematic stage
 
+### Requirement: Stage contract acceptance
+Each stage SHALL define a repo-state completion contract. `create` SHALL check that contract before skipping a stage during resume and again after a successful stage run. If the contract remains unmet, `create` SHALL preserve the partial committed work, report the unmet clause, and stop without advancing to a downstream stage.
+
+#### Scenario: Agent success without complete work
+- **WHEN** an agent reports success for a stage but the stage's repo-state contract remains unmet
+- **THEN** `create` does not start the next stage, reports the failed contract clause, and a later `create` invocation resumes the same stage
+
 ### Requirement: First-draft layout with honesty gate
 The layout stage SHALL produce rule-driven placement (real coordinates in the `.kicad_pcb`) and rule-based routing of power/critical nets, with every routed net passing DRC, and SHALL auto-write a `## Draft quality` section in LAYOUT.md listing what is done and what a human or specialist tool should redo.
 
--- a/openspec/specs/SPEC.md
+++ b/openspec/specs/SPEC.md
@@ -155,7 +155,7 @@ Firmware scope: scaffold + pin definitions + driver stubs + one working happy-pa
 
 ### The pipeline (Mode A internally)
 
-**Run-to-completion guarantee:** once `create` starts, it always finishes with the complete output package. Gates are *quality checks the agent must satisfy*, not stops that wait for a human. By default the pipeline is fully autonomous: unstated decisions get `ASSUMED` flags, imperfect layout gets the `Draft quality` label, and the run ends with gerbers, firmware, renders, and DEVPLAN.md on disk — an end product, reviewable as a whole. `--interactive` turns the two human gates (spec approval, pre-export review) back on for users who want them.
+**Run-to-completion guarantee:** once `create` starts, it autonomously works toward the complete output package. Gates are *quality checks the agent must satisfy*, not stops that wait for a human. Each stage has a repo-state completion contract; Copperhead skips a stage only when its contract is met and re-checks it after an agent reports success. If a contract remains unmet, Copperhead preserves the partial committed work, reports the unmet clause, and exits so the next `create` run resumes that stage rather than silently advancing. By default unstated decisions get `ASSUMED` flags, imperfect layout gets the `Draft quality` label, and a successful run ends with gerbers, firmware, renders, and DEVPLAN.md on disk — an end product, reviewable as a whole. `--interactive` turns the two human gates (spec approval, pre-export review) back on for users who want them.
 
 ```
 brief.md
@@ -169,7 +169,7 @@ brief.md
   → DEVPLAN.md
 ```
 
-Each stage is a `do`-loop run with a stage-specific prompt. State lives in the repo (docs + files), so `create` is resumable: kill it at any stage, re-run, it continues from the docs.
+Each stage is a `do`-loop run with a stage-specific prompt and completion contract. State lives in the repo (docs + files), so `create` is resumable: kill it at any stage, or leave a stage contract unmet, re-run, and it continues from that stage.
 
 ### First-draft layout (explicitly non-optimal, explicitly useful)
 
--- a/src/commands/create.ts
+++ b/src/commands/create.ts
@@ -5,7 +5,7 @@ import { createHash } from 'node:crypto';
 import { loadConfig } from '../config.js';
 import { bootstrapKicadProject } from '../kicad/bootstrap.js';
 import { exportSvg, runErc } from '../kicad/cli.js';
-import { listSymbols } from '../kicad/sexp.js';
+import { children, isList, listSymbols, parseSexp } from '../kicad/sexp.js';
 import { isDirty, commitAll, changedFiles } from '../util/git.js';
 import type { CopperheadConfig } from '../config.js';
 import { checkDrift } from '../memory/drift.js';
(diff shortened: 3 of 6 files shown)


animesh-chouhan:
## Review

### Overview

This PR turns the `create` pipeline's shallow stage-existence probes into real completion contracts, checked in both places that matter: before a resume skip and after a stage's agent run reports success. The schematic contract requires a configured schematic with at least one non-power symbol and drift-clean BOM/PINOUT docs; the layout contract requires a board with an actual top-level `footprint` expression plus non-template content under `## Draft quality`. This directly closes the #23 failure mode where a blank bootstrap sheet or `init`'s placeholder heading advanced the pipeline.

The design is sound and fits the repo's invariants well: contracts are derived from repo state (keeping D10's free resumability), the sexp parser is used read-only, and `checkDrift` reuse means the schematic contract matches the same truth-precedence semantics `sync` uses. The tests cover the right cases, especially the quoted-`footprint` string test, which proves the check operates on parsed structure rather than text.

### Issue worth fixing

**A malformed schematic crashes `runCreate` instead of failing the contract.** `schematicIsComplete` calls `listSymbols`, and `loadSheets` throws (`not a KiCad s-expression file`) when the file is empty, truncated, or not a parseable s-expression, and `readFile` throws ENOENT when a hierarchical sheet references a missing sub-sheet file. Before this PR, `isComplete` was existence-only and could not throw; now the post-run acceptance check runs on agent-written files, exactly the ones most likely to be malformed when the agent misbehaves. An exception here escapes `runCreate` uncaught, skipping the graceful "contract not met, re-run to continue" path and the partial-progress return value. Suggest wrapping the contract body in try/catch and treating a parse failure as contract-not-met (a file the parser rejects is by definition not a complete schematic). Same reasoning applies to `checkDrift`, which parses the schematic twice more via `listSymbols`/`pinNets`.

### Minor points

- `schematicIsComplete` and `layoutIsComplete` each call `loadConfig` even though `runCreate` already holds the config. Since `isComplete` only receives `(repoRoot, docs)`, this is understandable, but passing the config through (or widening the signature) would avoid re-reading it up to twice per stage evaluation.
- The `Stage.isComplete` doc comment still reads "true when repo state shows the stage is already done (resume support)". It now also serves as the post-run acceptance gate, so it is worth updating alongside the module comment you already revised.
- `docHasFilledSection` matches the heading with `indexOf`, so `### Draft quality` (a substring match) would also satisfy it. Anchoring on line start (`\n## Draft quality` or a regex with `^`/`m`) would make it exact. Low risk in practice since the template owns this heading.
- In the first test, the `expect(init.created).toContain('docs/LAYOUT.md')` assertion is unrelated to the schematic contract being tested; it would sit more naturally in the layout-template test or a scaffold test.
- The branch appears to be based on a pre-#28 main (the import hunk lacks the `RunMetaInput`/`ProgressRenderer` imports). GitHub reports it mergeable, so no action needed, just worth a rebase-and-retest before merge since #28 touched `runCreate`'s call into `runAgentLoop`.

### Design note (no change requested)

On post-run contract failure the pipeline stops and asks the user to re-run rather than retrying the stage in-process. That is a defensible choice (it avoids burning tokens on a stage that just demonstrably failed its contract), and the PR body states it explicitly. If repeated re-runs against a stubborn stage become a real workflow, a bounded in-run retry could be layered on later without changing these contracts.

### Verdict

Approve once the parse-failure crash path is handled; everything else is polish. Nice, tightly scoped fix with tests that pin each contract clause.

fyzanshaik:
Addressed the review's required malformed-schematic path and rebased onto current `main` (including #28).

Changes:

- `schematicIsComplete()` now treats parser, drift, and missing-sheet read failures as an unmet contract rather than allowing them to escape `runCreate()`.
- `## Draft quality` matching is now exact at the Markdown-heading level; `### Draft quality` no longer matches.
- Removed the unrelated scaffold assertion and added a malformed-schematic regression.

Verification after the rebase:

- `npx vitest run test/create-stage-turns.test.ts` — 5 passed
- `npm run typecheck` — passed
- full suite with KiCad 10.0.4 — 112 passed, 7 skipped, with the same two pre-existing fixture failures described in the PR body.

animesh-chouhan (commented):
# Review: enforce create stage completion contracts

## Overview

This PR turns the `schematic` and `layout-draft` `isComplete` probes from existence checks into real contracts, and adds a post-run acceptance check in `runCreate()` so an agent that reports success without meeting the active stage's contract halts the pipeline instead of advancing. The direction is exactly right for issue #23, the new helpers are well built, and the tests are sharp on the negative cases. The main gaps: the PR fixes only two of the four false positives the issue documents, ships no SPEC/OpenSpec update for a behavior change, and gives the user (and a future rerun) no detail about which contract clause failed.

## What is done well

- The post-run acceptance check (`src/commands/create.ts`, new lines ~167-172) closes the second half of the hole: not just "do not skip past partial work" but "do not advance on an agent's word". Checking the same contract in both places is the correct symmetry.
- `layoutIsComplete` uses the read-only sexp parser (`parseSexp` + `children(root, 'footprint')`) rather than a regex, so quoted text like `(gr_text "(footprint fake)")` cannot satisfy it. The dedicated test for that case (test/create-stage-turns.test.ts, "requires a real footprint expression...") is a genuinely good adversarial test. Footprints are top-level children of `kicad_pcb`, so the top-level-only `children()` scan is correct.
- `docHasFilledSection` strips HTML comments before judging emptiness, which is precisely the right de-collision against init's `<!-- copperhead writes an honest assessment here -->` placeholder in LAYOUT.md. The `/^##\s/` next-heading regex correctly keeps `###` subsections inside the section.
- The contracts stay cheap and offline: no `kicad-cli` and no network in any probe, consistent with the repo's "deterministic checks are CI-safe" stance.
- The end-to-end halt test mocks the agent loop, openspec, and check, so it is hermetic and fast, and it asserts both the `completed` list and the log message.

## Correctness of the contract checks

### Major: spec-seed, outputs, and firmware still false-positive, but the PR says "Closes #23"

Issue #23 explicitly names two init-template collisions: LAYOUT.md's `## Draft quality` (fixed here) and SPEC.md's `## Budgets` (not fixed). `src/memory/scaffold.ts` line 58 still emits `## Budgets` with only comment placeholders, and `spec-seed.isComplete` is still `docHasContent(..., '## Budgets')`, so on any init-ed repo the spec-seed stage is silently skipped and no budgets are ever seeded. The issue also calls out `outputs.isComplete` (bare `existsSync('outputs')`, satisfied by a partially failed export run) and `firmware` has the same shape.

The fix for spec-seed is nearly free with the machinery this PR already adds: `docHasFilledSection(root, path.join(docs, 'SPEC.md'), '## Budgets')` gives the intended semantics (comments stripped, template fails, agent-written budgets pass). I would either add that one-liner (plus a test) or change "Closes #23" to "Partially addresses #23" and file the remainder; as written, merging closes the issue with half the reported false positives still live.

Notably, the halt test itself encodes the residue: it expects `completed: ['spec-seed', 'architecture', 'part-selection']` on a repo where only `init` ran, meaning three stages were marked complete purely by init templates and doc scaffolding.

### Minor: layoutIsComplete lacks the error handling schematicIsComplete has

`schematicIsComplete` wraps everything in try/catch and treats any throw as "contract unmet" (and there is a test for a malformed schematic). `layoutIsComplete` has no such guard: a throwing `loadConfig`, or `readFile` on an unreadable/directory board path, propagates out of `stage.isComplete` and crashes `runCreate` with an unhandled exception instead of the graceful halt message. `parseSexp` itself is total in practice, so the exposure is config/fs errors, but the asymmetry is unnecessary. Suggest the same try/catch, or better, a shared `asContract(fn)` wrapper so future probes cannot forget it.

(Flip side of the schematic catch: it also swallows a corrupted `.copperhead/config`, which will surface later as a confusing stage re-run rather than a config error. Acceptable trade, worth a comment.)

### Minor: the halt says nothing about which clause failed

The post-run message ("stage contract is not met yet") and the resume skip both discard everything the probes learned: which drift mismatches exist, whether the board has zero footprints, whether `## Draft quality` is empty. Two consequences:

- The user gets no actionable detail, and per SPEC §2.5 this pipeline is sold as run-to-completion; when it now stops early it should at least say why.
- A rerun launches the same stage with the same prompt, so an agent that missed the contract once has no new signal and may miss it the same way again.

Suggestion: have the probe (or a sibling `explain()`) return an unmet-reason string and log it; longer term, feed it into the stage prompt on rerun, or retry the stage once in-run with the reason appended before halting. If you change the probe signature, coordinate with #33 first (see below).

### Contract semantics look right otherwise

- Requiring drift-clean BOM/PINOUT at schematic completion matches the stage prompt ("Update PINOUT.md as you assign pins... check_drift") and correctly forces every BOM part to be placed. It does mean a BOM edit after the fact retroactively reopens the schematic stage, which is desirable and is exactly the semantics #33's staleness model wants.
- `listSymbols` excludes power symbols, so a sheet containing only `power:GND` will not count as a schematic. Good.
- `docHasFilledSection` requires `line.trim() === heading` exactly; `## Draft Quality` or extra internal spacing false-negatives. Nit, and consistent with the repo's anchored-exact-match philosophy, but worth knowing.

## Project conventions

- Major: CLAUDE.md says spec-level behavior changes must update SPEC.md and the change artifacts together. This PR changes what "stage complete" means and adds a new early-exit path from `create` (a qualification of the §2.5 run-to-completion guarantee), with no SPEC.md or delta-spec touch. Compare #33, which ships an OpenSpec change alongside its create.ts edits. At minimum, the §2.5 resumability paragraph ("State lives in the repo... it continues from the docs") should describe contracts, and the new halt behavior deserves a scenario.
- The doc comment updates on `Stage`/`isComplete` are good and accurate.
- No em dashes, no emoji, style matches the file. Good.

## Coordination with #31 and #33

- #33 (issue #24, stage graph + rerun) was explicitly designed to compose with this PR: it promises not to touch any `isComplete` body and uses these probes as both the unrecorded-stage fallback and the completion-record gate, with probe-gated records so this PR's post-run check cannot deadlock against a pre-committed record. That composition holds only if the probe signature `(repoRoot, docs) => boolean` stays stable, so if you adopt the "return a reason" suggestion above, agree on the shape with #33 first (a separate `explain()` alongside the boolean is the conflict-free option).
- Textual conflicts in `src/commands/create.ts` are guaranteed with both #31 and #33 (#33 replaces the sequential runner with a plan/queue; #31 adds `--stage`/`--from` handling in the same loop). #33's PR body already volunteers to rebase if this lands first; landing #29 first is the sensible order since both rerun PRs want trustworthy probes underneath them.
- Separately for the maintainer: #31 and #33 both implement issue #24 and cannot both merge; that choice is outside this PR but blocks sequencing.

## Test coverage

- Negative cases are well covered: blank schematic, malformed schematic, init's LAYOUT template, quoted footprint text, and the post-run halt.
- Gap (worth fixing before merge): there is no positive-case test. Nothing asserts that a genuinely complete stage satisfies its contract, so the false-negative direction, an over-strict contract that re-runs finished stages and burns LLM turns redoing committed work, has no regression guard. Both positives are cheap with the existing fixture: after `runInit` on the untouched fixture repo (real schematic with symbols, BOM/PINOUT generated from it so drift-clean), `stage('schematic').isComplete` should resolve true; a board with one real `(footprint ...)` plus a filled `## Draft quality` should make `layoutIsComplete` true.
- If spec-seed gets the `docHasFilledSection` treatment, add the template-vs-filled pair of tests for it too.

## Performance

- `schematicIsComplete` parses the schematic tree three times per call (`listSymbols`, then `checkDrift` calls `listSymbols` and `pinNets`, each doing its own `loadSheets`), and each probe reloads config; probes run up to twice per stage per pipeline pass. Negligible at fixture scale and fine to ship, but if #33's classifier starts calling probes more often, a parse-once refactor of `checkDrift` to accept preloaded sheets would be the lever. Nit, no action needed now.

## Verdict

Right mechanism, good tests for what it covers. Before merge I would: (1) fix or explicitly descope the spec-seed `## Budgets` collision and soften "Closes #23" accordingly, (2) add the try/catch to `layoutIsComplete`, (3) add the two positive-case tests, (4) log which contract clause failed, and (5) add the SPEC.md/OpenSpec touch per repo convention.

fyzanshaik:
Addressed the review and rebased onto current `main` (branch head `f0361f2`).

- `spec-seed` no longer accepts the generated `SPEC.md` template: `## Budgets` must contain real content.
- Layout and schematic contract checks now treat unreadable/malformed KiCad files as incomplete instead of throwing.
- Contract failures now log the concrete missing requirement before recovery/stop.
- Added positive and negative regressions for initialized placeholders, filled spec/layout artifacts, malformed KiCad files, and the failure message.
- Updated the normative SPEC/OpenSpec wording to define stage contracts and post-run rechecks.
- Kept the boolean `isComplete` signature unchanged for the related PRs.

I agree that outputs and firmware are still shallow and are a separate design slice. I removed `Closes #23` from the PR description; this PR now explicitly addresses only the stage-contract portion.

Verification:

- `npx vitest run test/create-stage-contracts.test.ts test/create-stage-turns.test.ts test/create-resilience.test.ts` — 8 passed
- `npm run typecheck` — passed
- full KiCad-backed suite — 329 passed, 14 skipped; the two known pre-existing KiCad 10.0.4 fixture failures remain in `gating-sync` and `init-check`.

PR #33, opened 2026-07-21
Re-run create stages on demand, with a stage dependency graph and downstream invalidation

### LINKED ISSUE #24: Feature: re-run individual pipeline stages on demand, with a stage dependency graph that invalidates and updates downstream work

pavvann:
## What

`copperhead create` learns targeted stage re-runs: `--stage <name>` re-runs one stage against the existing artifacts and reconciles every stage that consumes an output the re-run actually changed; `--from <name>` force-re-runs a stage and its graph descendants; `--dry-run` prints each stage's classification and what would run, writing nothing. A plain `create` run becomes self-healing: stages whose recorded inputs changed re-run automatically.

## Why

The create pipeline was strictly forward-only, but design iteration is backward as often as forward: the BOM changes after the schematic exists, budgets tighten after layout, a footprint swap invalidates the outline. The only workarounds were hand-deleting artifacts until the resume probes read "incomplete", or free-form `do` requests that don't know they are redoing a stage. Closes #24.

## Design

- **Stage graph as data**: each stage in [create.ts](src/commands/create.ts) declares `consumes`/`produces` over a closed artifact vocabulary (brief, spec, subsystems, bom, pinout, schematic, board, layout-intent, outputs, firmware, devplan); edges derive from it, so invalidation is per artifact, never "everything after".
- **Completion records**: each stage's completing commit records content hashes of its consumed and produced artifacts in `.copperhead/create-state.json` ([stagestate.ts](src/memory/stagestate.ts)), written via a new `beforeCommit` seam on the loop ([loop.ts](src/agent/loop.ts)) that fires only after every gate passes; a throw routes to the existing commit-failed path with snapshot rollback.
- **Probe-gated records, halt on unmet contract**: `beforeCommit` withholds the record when the stage's completion probe fails post-run, and the runner halts with `ok:false` (main's hardening semantics), so "re-run to continue" is always true and absent work can never classify fresh.
- **Classification** (fresh / stale with changed-artifact names / incomplete / assumed-complete): changed inputs outrank a failing probe, because the drift-aware schematic probe fails precisely when an upstream artifact changed; a recorded stage whose inputs match but whose probe fails (deleted work, stricter probes) is incomplete (AC-9.9). Stale stages re-run with a reconciliation preamble naming exactly what changed; specifics come from `check_drift`, deterministically.
- **Post-merge of main (v0.5.0)**: the create-pipeline hardening (drift-aware/footprint probes, halt semantics) is adopted as-is, probe bodies untouched; `stageMaxTurns` and `onBudgetExhausted` are threaded through the queue runner and upstream's `create-stage-turns` test passes unchanged.
- **Sibling PRs**: #29 remains an overlapping, narrower hardening of the same probes; #31 implements the same flags with config-stored completion and list-order invalidation, and the two state formats cannot coexist (recommended close-as-superseded in review discussion; maintainer's call).

Implementation was adversarially reviewed before submission; confirmed findings (probe-gated records, path-sensitive hashing for walked artifacts, autosave-sheet exclusion, model-free dry-run, corrupt-state preservation as `create-state.json.corrupt`) are fixed in-branch.

## Spec / OpenSpec

OpenSpec change [`rerun-create-stages`](openspec/changes/rerun-create-stages/) (proposal, design D1-D8, tasks): validates. Delta specs: new capability [`pipeline-invalidation`](openspec/changes/rerun-create-stages/specs/pipeline-invalidation/spec.md); modified [`create-pipeline`](openspec/changes/rerun-create-stages/specs/create-pipeline/spec.md), [`cli-surface`](openspec/changes/rerun-create-stages/specs/cli-surface/spec.md), [`run-observability`](openspec/changes/rerun-create-stages/specs/run-observability/spec.md). SPEC.md moved together: §2.5 (stage graph, records, invalidation), §3 (flags), new AC-9.1 through AC-9.10.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | pass: 211 passed, 10 skipped (key-gated), against kicad-cli 10.0.4 |
| Live-LLM (opt-in) | keyed on `OPENAI_API_KEY`/`ANTHROPIC_API_KEY` | skip (no key in this environment) |

New tests: [stagestate.test.ts](test/stagestate.test.ts) (20: hashing, records, classification including stale-outranks-probe and AC-9.9) and [create-stages.test.ts](test/create-stages.test.ts) (30: the issue #24 flow end to end, transitive cascades, targeted-mode non-widening, halt semantics, interactive gates, dry-run purity, trigger metadata on all three AC-8 surfaces). Hermetic: scripted providers, no LLM, kicad-cli shimmed per file.

- Sandbox variant used (`create` / `edit`): `edit` (init scaffold committed with `--no-verify`, since the sandbox has no global `copperhead` for the installed pre-commit hook), plus `examples/simple/usb-c-breakout.md` as the brief. LLM-free surfaces exercised by hand; record-based staleness and reconciliation are LLM-driven and covered by the offline suite above.
- Commands run and outcome:

```text
$ copperhead --repo manual-tests/runs/edit create --brief brief.md --dry-run
stage classification:
  spec-seed: assumed-complete
  architecture: assumed-complete
  part-selection: assumed-complete
  schematic: assumed-complete
  layout-draft: incomplete
  outputs: incomplete
  firmware: incomplete
  devplan: incomplete
would run (default): layout-draft -> outputs -> firmware -> devplan
(exit 0; git status untouched afterwards; runs without any API key configured)

$ copperhead --repo manual-tests/runs/edit create --brief brief.md --stage devplan --dry-run
warning: upstream stage(s) layout-draft, firmware are incomplete; devplan may run against missing inputs
would run (--stage devplan): devplan

$ copperhead --repo manual-tests/runs/edit create --brief brief.md --stage part-slection
unknown stage "part-slection" for --stage; valid stages: spec-seed, architecture, part-selection,
schematic, layout-draft, outputs, firmware, devplan
(exit 1, before any model or kicad-cli resolution)

$ copperhead --repo manual-tests/runs/edit create --brief brief.md --stage outputs --from schematic
--stage and --from are mutually exclusive: --stage re-runs one stage and propagates real changes;
--from force-re-runs a stage and its descendants
(exit 1)

$ copperhead --repo manual-tests/runs/edit check
ERC ✓ / DRC ✓ / drift ✓ (exit 0, real kicad-cli 10.0.4)
```

## Docs

[reference/cli.md](docs/src/content/docs/reference/cli.md) (flags, stage table with consumes/produces, classification semantics) and [workflows/create-from-brief.md](docs/src/content/docs/workflows/create-from-brief.md) (re-run walkthrough). `.copperhead/create-state.json` is design state and committed, matching the `generatedHashes` convention; `.copperhead/runs/` stays ignored.

---

## Invariant checklist

- [x] **Verification-gated out**: mutations still end in ERC (and DRC when the board changed) passing, with repair up to `maxRepairCycles` then rollback to the git snapshot. The new `beforeCommit` seam fires only after every gate and never on refuse, dry-run, or rollback paths (each covered by a test).
- [x] **`check`/`verify` stays LLM-free and network-free**: untouched; `create --dry-run` is additionally LLM-free and kicad-free by construction.
- [x] **No sexp serialization**: the parser stays read-only; artifact hashing reads bytes only.
- [x] **Sync-obligations ledger**: untouched; `beforeCommit` fires after the ledger gates, so it cannot commit past open obligations.
- [x] **Secrets**: `create-state.json` holds only stage names, run ids, timestamps, and sha256 hashes; redaction paths untouched.
- [x] **Spec coherence**: SPEC.md and the `rerun-create-stages` artifacts moved together; `openspec validate rerun-create-stages` passes.

## Summary by CodeRabbit

* **New Features**
  * Added targeted `copperhead create` reruns via `--stage` and `--from`, plus `--dry-run` that deterministically classifies stages (fresh/stale/incomplete/assumed-complete) and previews exactly what will run.
  * `copperhead create` is now resumable: it records per-stage completion and resumes at the first stale/incomplete stage.
  * Run reporting/metadata now includes stage trigger details and changed upstream artifacts.
* **Documentation**
  * Updated CLI/workflow docs to explain reruns, dependency propagation, and dry-run semantics.
* **Bug Fixes**
  * If `create-state` is missing/corrupt, `create` now warns and safely falls back to probe-based completion without aborting.
* **Tests**
  * Added extensive automated coverage for reruns, staleness propagation, dry-run/write behavior, interactive reconciliation, hooks, and corrupted-state handling.


pavvann:

animesh-chouhan (commented):
# Review: PR #33, re-run create stages with a dependency graph and downstream invalidation

**Verdict: approve with changes requested.** One robustness bug (malformed completion record aborts the run, reproduced locally against this branch) plus two sibling-PR coordination decisions that should be written down before merge. Everything else is nits or forward-looking hardening.

**Verified locally on the PR branch** (fetched `pull/33/head`): `tsc --noEmit` clean; the two new test files pass 42/42; `openspec validate rerun-create-stages` and `openspec validate build-copperhead-phase-1` both valid. The TypeError finding below was reproduced with a five-line script, not inferred from reading.

## Overview

This is a strong, carefully designed PR. It turns the create pipeline's implicit stage ordering into data (`consumes`/`produces` over a closed artifact vocabulary), adds probe-gated completion records with content hashes in `.copperhead/create-state.json`, and builds `--stage`, `--from`, and `--dry-run` on top. The OpenSpec artifacts (proposal, design D1-D8, four delta specs, AC-9 block in SPEC.md) are coherent with the code, and the 42 new offline tests are hermetic and meaningful. I found one real robustness bug, a couple of semantic edge cases worth documenting, and two significant coordination issues with the sibling PRs.

## What is done well

- **Probe-gated records** (`src/commands/create.ts`, the `beforeCommit` callback): the record asserts "this stage's work exists as committed", not "a run committed". This is the right trust model, and it is what makes composition with #29 possible at all.
- **A record never outranks a failing probe** (`src/memory/stagestate.ts`, `classifyStages`): the AC-9.9 demotion closes the "skipped forever while consumers go stale against a missing input" trap.
- **Rollback safety of the new seam**: `beforeCommit` fires only after every gate, and a throw routes to `fail(..., 'commit-failed')` in `src/agent/loop.ts`, which restores the pre-run snapshot. I verified `restore()` in `src/util/git.ts` hard-resets and cleans, so a half-written `create-state.json` record is wiped along with the failed work. The invariant "record and work are atomic" genuinely holds on every path.
- **Hashing details**: the `ABSENT` sentinel round-trips (recorded-absent plus still-absent is fresh), walked artifacts stay path-sensitive even with one file, `_autosave-*.kicad_sch` exclusion, and timestamps never participate. All tested in `test/stagestate.test.ts`.
- **Pop-time freshness re-check** and the "newly stale only" filter prevent both wasted LLM runs and infinite reconcile loops (the `ran.has` exclusion makes ping-pong structurally impossible).
- **`create-state.json.corrupt` preservation** in `saveStageRecord`: degrading a corrupt file to empty and then writing one record back would silently erase history; renaming it aside first is the right call.
- **AC-9.5 discipline**: `--dry-run` and stage-name validation run in `src/cli.ts` before model resolution and the kicad preflight, so the classification report needs neither an API key nor KiCad.

## Correctness: dependency graph and invalidation

- **Cycles**: the graph is static and closed, `descendantsOf` does a single forward pass, and `test/create-stages.test.ts` asserts the STAGES array is a valid topological order of the produces/consumes edges. Solid. Consider making the topo assertion a startup invariant (or at least a comment on STAGES saying the test enforces it), since a future stage-table edit that breaks topo order would make `descendantsOf` silently wrong while only one test catches it.
- **Major: a parseable but shape-invalid record crashes classification, violating AC-9.8.** `loadCreateState` validates only that `stages` is an object; `classifyStages` then does `record.inputs[artifact] !== current`. A state file like `{"version":1,"stages":{"schematic":{}}}` (hand edit, partial write, future schema change) makes `record.inputs` undefined and throws a TypeError that escapes the load-time try/catch, aborting the run. Reproduced on this branch: `TypeError: Cannot read properties of undefined (reading 'bom')`. AC-9.8 and the state-resilience delta spec promise "never aborts". Fix: validate each record's shape in `loadCreateState` (demote bad records to unrecorded with the warning), or defensively use `record.inputs?.[artifact]`. A test for this shape belongs next to the existing `'{"stages": 42}'` cases in `test/stagestate.test.ts`.
- **Minor: undeclared writes do not cascade within an invocation.** Post-stage rehashing covers only `stage.produces` (`src/commands/create.ts`, the `changed` loop), and `newlyStale` requires the changed input to be in `changedThisRun`. But any stage's agent can write outside its declared outputs, most notably `record_constraint`, which dual-writes `.copperhead/constraints.json`, part of the `spec` artifact. A constraint recorded by `layout-draft` therefore stales `architecture` and `part-selection` invisibly until the next default `create`. Design D8 acknowledges the spec-bundling noise but not this specific within-run blind spot. The next-run self-heal makes it acceptable; I would add one log line when the end-of-run `runCheck`/classification could detect it, or at least document it in D6/D8 so a rebase does not "fix" it into an infinite loop.
- **Minor: self-edits of consumed artifacts cause one spurious stale cycle.** Inputs are hashed at stage start (before `runAgentLoop`); if a stage's run edits an artifact it also consumes (the schematic stage touching BOM.md footprints, for example), the recorded input hash is pre-edit and the stage classifies stale on the next run. It heals after one no-op reconcile run (the record rewrites with fresh input hashes), but it costs an LLM run. Consider re-hashing inputs inside `beforeCommit` for artifacts the stage itself modified, or documenting the one-cycle cost.
- **Minor: targeted modes do not warn about incomplete upstream stages.** `create --stage layout-draft` in a repo with no schematic runs the full LLM stage against nothing; the probe gate correctly withholds the record afterward, but an upfront warning (upstream stage X is incomplete) would save a whole model run. PR #31 hard-errors here; a warning is the better middle ground.
- **Partial failures mid-rerun** are handled correctly: a failed queued stage returns `ok:false`, and because staleness is recomputed from persisted records plus the working tree, the next default `create` re-derives the remaining stale set. One wording nit: the failure message "re-run copperhead create to resume here" is slightly misleading after `--stage`/`--from`, since a plain re-run resumes but repeating the same targeted command re-runs the requested stage too.
- **Minor: the `version` field is ignored on load.** `loadCreateState` stamps `version: 1` over whatever it read. A future `version: 2` file would be silently misread instead of degrading with a warning. One `raw.version !== 1` check keeps the escape hatch open.

## Repo invariant checks (all examined, all pass)

- **Spec-gated in**: untouched and preserved. Stage runs still go through the normal proposal gate (the test scripts all call `propose_change`/`validate_change` before writing), and `beforeCommit` runs in the runner's context, not the agent's; it adds no tool surface.
- **Verification-gated out**: preserved. The hook fires after all gates and before `commitAll` only (never on refuse, dry-run, or rollback, each covered by a test), the tail `runCheck` still runs in every non-dry mode including targeted ones, and a hook throw lands on the existing commit-failed path with snapshot restore. The loop diff (`src/agent/loop.ts`, the eight-line insertion after `ledger.clear('changelog')`) is admirably minimal.
- **`check`/`verify` LLM-free and network-free**: untouched; `create --dry-run` is additionally LLM-free by construction (returns before any provider or kicad-cli resolution).
- **No sexp serialization**: the parser is untouched; hashing reads bytes only.
- **Sync-obligations ledger**: untouched; `beforeCommit` fires after the ledger gates, so it cannot be used to commit past open obligations.
- **Secrets**: no new key surface; see the security section.

## Coordination with sibling PRs

- **Major: PR #31 is a competing implementation of the same flags and cannot merge alongside this.** It stores `stageCompletion` inside `.copperhead/config.json`, invalidates by list order rather than graph edges, records only the brief hash as input, and skips the final `runCheck` in `--stage` mode (which weakens verification-gated out). PR #33 is a strict superset with better semantics on every axis. Recommendation: decide explicitly, and close #31 in favor of #33 rather than attempting a merge of the two state formats; two different completion-record stores in one repo would be actively harmful.
- **Major: composition with PR #29 is safe structurally but has two semantic wrinkles the PR body does not mention.**
  1. **Drift-aware probes shadow the stale path.** #29's `schematicIsComplete` requires BOM/PINOUT drift-cleanliness. Under #33's classification, the probe runs first and a failing probe demotes to `incomplete` before hashes are ever compared. So the flagship scenario (edit BOM.md, expect `schematic: stale (changed: bom)` plus the reconciliation preamble) instead classifies `incomplete`: the changed-input names, the revise-not-recreate preamble, and the `stale` trigger metadata are all lost precisely in the case this PR was built for, and the stage re-runs with the bare first-time prompt. Whoever rebases second should special-case this: when a record exists and inputs mismatch, prefer `stale` over probe-driven `incomplete` for probes that fail only on drift (or have #29's probes report existence and drift separately).
  2. **Halt versus continue.** #29 stops the whole pipeline when a post-run contract is unmet (`ok:false`, "re-run to continue"); #33 logs the withheld record and continues to the next queued stage. Both are defensible, but they are different behaviors for the same event and the merged code must pick one deliberately.
- The PR body's claim that no probe body is touched holds, and the expected textual conflicts in `src/commands/create.ts` are called out honestly. Good coordination hygiene overall; the two points above just need to be on the rebase checklist.

## Project conventions

- OpenSpec artifacts, delta specs mapping to AC-9, and SPEC.md updated together, exactly per CLAUDE.md. `openspec validate` is claimed to pass.
- `create-state.json` is committed (design state, like `generatedHashes`) while `.copperhead/runs/` stays ignored; consistent with AC-4.3.
- Observability lands on all three AC-8 surfaces with LLM-free collection; `stageLabel` in `src/agent/runmeta.ts` is a clean extension.
- Nit: the docs-site additions (`docs/src/content/docs/reference/cli.md`, `workflows/create-from-brief.md`) lean heavily on em dashes; the project owner prefers colons/commas in user-facing prose.

## Performance

- Fine at today's scale: classification is 8 probes plus hashing of consumed artifacts, and full reclassification happens only after a stage whose outputs changed, bounding it at 8x8 per invocation.
- Two forward-looking notes: (1) once #29 lands, each probe call may parse the schematic and run `checkDrift`, so reclassify-after-every-stage becomes meaningfully expensive; caching probe results keyed by the artifact hashes already being computed would keep it flat. (2) `hashArtifact` reads whole files into memory; `outputs/` STEP files can be tens of MB. Streaming hashes (`createHash` with a read stream) is a cheap future-proofing change.
- `walkFiles` recursion has no symlink-cycle guard, but `Dirent.isDirectory()` is false for symlinks, so a symlinked directory degrades to an `ABSENT` file entry rather than infinite recursion. Acceptable; a one-line comment would prevent someone "fixing" it with `stat` and introducing the cycle.

## Test coverage

Excellent overall: the issue #24 flow end-to-end, transitive cascades, targeted-mode non-widening, interactive approve and decline, probe-gating, AC-9.9 demotion with the became-fresh-before-popping skip, dry-run purity via `git status --porcelain`, corrupt-state degradation, and trigger metadata verified on all three surfaces. The stage-aware scripted provider and kicad-cli shim are a nice pattern. Gaps worth adding:

- The malformed-record crash above (blocker for the resilience claim).
- A mid-cascade stage failure: `--stage part-selection` where the queued `schematic` reconcile fails, asserting `ok:false` and that the next default run re-derives the stale set.
- `--from` mid-list failure (same shape).
- A `--dry-run` combined with `--from` (only `--stage` is covered).
- Mutating `process.env.PATH` in `beforeAll` is safe today because vitest isolates files in workers, but a comment noting that assumption would help.

## Security

- `create-state.json` contains only stage names, run ids, ISO timestamps, and sha256 hashes; nothing redactable, and it does not pass through the transcript redaction path anyway. No new key-handling surface.
- The `beforeCommit` seam cannot bypass any gate (fires post-gates, throw lands on commit-failed with rollback) and is not reachable from the agent's tool surface.
- The PATH shim exists only in tests. No injection risk from stage names: they are validated against the closed table before any use.

## Actionable summary

1. Validate per-record shape in `loadCreateState` (or use optional chaining in `classifyStages`) and add the shape-invalid test. This is the one code change I would ask for before merge.
2. Add the two #29 wrinkles (stale-shadowed-by-drift-probe, halt-vs-continue) to the rebase plan in the PR description.
3. Decide the fate of PR #31 explicitly; recommend closing it as superseded.
4. Warn when a targeted stage's upstream stages classify incomplete.
5. Optional hardening: `version` check on load, streaming hashes, probe-result caching ahead of #29, the extra failure-path tests.

Overall: high-quality work, clearly adversarially reviewed before submission, and the remaining issues are narrow and fixable. With item 1 addressed and the coordination items written down, this is mergeable.

pavvann:
Thanks for the thorough review — the reproduced TypeError especially. All actionable items addressed in 1c26dc5:

**1. Malformed-record crash (the blocker).** `loadCreateState` now validates each record's shape (`completedAt`/`runId` strings, `inputs`/`outputs` maps) and demotes malformed records to unrecorded with a warning naming them; valid sibling records survive. Your repro shape `{"version":1,"stages":{"schematic":{}}}` is now a test next to the existing envelope cases, asserting classification does not throw and falls back to the probe. Also took the optional `version !== 1` check while in there: an unrecognized version degrades the whole file with a warning instead of being silently misread.

**2. #29 wrinkles.** Both written into the PR description as a rebase checklist for whoever lands second: the drift-probe-shadows-stale classification issue (with the suggested resolution: prefer `stale` when a record exists and inputs mismatch) and the halt-vs-continue decision.

**3. PR #31.** Agreed that two completion-record stores would be harmful; noted in the description with a close-as-superseded recommendation. Your call to make.

**4. Upstream-incomplete warning.** Targeted modes (and their dry runs) now warn up front: `warning: upstream stage(s) firmware are incomplete; devplan may run against missing inputs`. Computed from graph ancestry, not list order, so it names only true ancestors.

**5. Optional items.** Taken now: version check (above), topo-order invariant comment on STAGES, symlink-behavior comment on walkFiles (with the do-not-fix-with-stat warning), PATH-shim worker-isolation comment, the misleading resume message after targeted-mode failures, em-dash cleanup in the docs prose, and D8 now documents both the within-run undeclared-write blind spot (with a do-not-fix-into-a-loop note for the rebase) and the self-edit one-cycle cost. Deferred as follow-ups, agreed they're not blockers: streaming hashes and probe-result caching ahead of #29.

**New tests:** malformed record shape, unsupported version, `--dry-run --from`, mid-cascade failure with next-run stale-set re-derivation, `--from` mid-list failure, and the upstream warning. Suite: 157 passed / 7 key-gated skips, typecheck clean, `openspec validate` green.

animesh-chouhan:
Heads up: #31 targets the same issue (#24) and implements the same `--stage`/`--from` flags, so these two PRs directly overlap. You already reference #31 in the description, but flagging here for visibility. The two state formats (`config.json` vs `create-state.json`) cannot coexist, so this needs a maintainer decision on which direction to take.

@pavvann, could you and @Prajwalraut29 sync on the two approaches so we converge on one before either merges?

pavvann:
Happy to sync. @Prajwalraut29, summarizing the concrete deltas between the two so we can converge (largely per the review above):

- **State**: #33 stores per-stage completion records in a committed `.copperhead/create-state.json`, with content hashes of every consumed and produced artifact; #31 stores completion in `config.json` with the brief hash as input. Separating pipeline state from user-facing config also keeps `config.json` hand-editable.
- **Invalidation**: #33 follows explicit produces→consumes graph edges (so `--from layout-draft` never re-runs firmware, and a BOM edit re-runs only its consumers); #31 invalidates by list order.
- **Verification**: #33 keeps the tail `check` run in every mode and gates records on the stage's completion probe, which is also what makes it composable with #29.

Given the review found #33 a strict superset on these axes, my suggestion is to converge on #33's format and close #31 as superseded — but genuinely open to counterpoints if there's a case #31 handles better. @Prajwalraut29, if you're up for it, the review also left concrete follow-ups unclaimed (streaming hashes for large `outputs/` files, probe-result caching ahead of #29) that would build directly on whichever PR lands. Final call is @animesh-chouhan's either way.

animesh-chouhan:
🤖 **Automated pr-review pass** (posted by Claude Code at a maintainer's request; this is a plain comment, not a formal GitHub review, and does not gate merge). A human review is still expected.

## Verdict: Approve with nits (clear one merge blocker first)

This PR does what it claims. The code is clean, well-documented, invariant-preserving, and thoroughly tested. The only blocker is mechanical: the branch now conflicts with `main`.

### Verified locally against PR head (not assumed)
- **Typecheck + build:** clean.
- **Full test suite (Linux):** **211 passed, 10 skipped, 0 failed.** The 10 skips are API-key-gated live tests. All 42 new offline tests pass (`test/stagestate.test.ts`, `test/create-stages.test.ts`), including the `beforeCommit` seam paths, AC-9.9 demotion, corrupt-state handling, and the issue #24 flow.
- **`openspec validate rerun-create-stages`:** valid.

### Invariant checks (all pass)
- **Verification-gated out — preserved.** The new `beforeCommit` seam (`src/agent/loop.ts:555-565`) fires strictly after the finish gate (ERC/DRC + ledger) and the changelog write, immediately before `commitAll`. It never runs on the refuse, dry-run, or rollback paths; a throw routes to `fail(..., 'commit-failed')` then snapshot rollback. All four paths are tested.
- **Spec-gated in — untouched.** Tool-list gating not modified.
- **`check` stays LLM-free and network-free — preserved.** `src/commands/check.ts` unchanged; no provider import.
- **Sync-obligations ledger — preserved.** `ledger.clear('changelog')` runs before the seam; commit-gating intact.
- **No sexp serialization / Secrets** — not touched.
- **Spec coherence — satisfied.** SPEC.md gains the AC-9 block and the OpenSpec change validates.

### Findings (all LOW / non-blocking)
1. **Dirty-tree warning skips the default self-heal path.** In `src/commands/create.ts` the warning is guarded by `if (planned.length && (opts.stage || opts.from))`. A flagless run that heals stale stages also commits per-stage with `git add -A`, so unrelated working-tree WIP can be swept into a re-run stage's commit with no warning: the same failure mode the warning exists to prevent. Suggest moving the `isDirty` check ahead of the mode guard.
2. **A grown `consumes` edge to a legitimately-absent artifact classifies `stale` permanently** (recorded `undefined` never equals current `'absent'`). Documented as intended, but such a stage never settles back to fresh.
3. **Perf:** `classify()` re-hashes every consumed artifact (full directory walks) after each changed stage in a cascade, O(stages x files) per step. Negligible at current scale; worth a note if artifact sets grow.
4. **Orphaned state file on post-write commit failure.** `create-state.json` is written inside `beforeCommit`, before `commitAll`; if the commit fails afterward, cleanup relies on `restore(snap)` and is untested (the throw test exercises the pre-write path). Low risk.

### Merge blocker (mechanical, not code quality)
The branch is **5 commits behind `main`** and conflicts in `src/cli.ts` and `src/commands/create.ts`, from the recently merged BOM-parser / supplier-BOM-export PRs (#44, #45), not from #29 (its hardened probes are already merged into this branch). **Needs a rebase before merge.** The `CONFLICTING` status is real, not stale.

### Note on overlap
This PR and #31 implement the same `--stage`/`--from` feature with incompatible state formats and cannot coexist; the author already recommends closing #31 as superseded. On code, spec-fit, and test coverage, this PR is the stronger candidate.

Prajwalraut29:
@animesh-chouhan  check claude response, i resolve the merged conflict

animesh-chouhan:
still has two conflicts, ready to merge this after they get resolved.

animesh-chouhan:
## Head-to-head (based on the real diffs, not the PR descriptions)

| | #31 | #33 |
| --- | --- | --- |
| Core mechanism works? | ❌ No: the stage-completion state it stores is write-only dead code, never read | ✅ Yes: completion records drive classification |
| Invalidation | List-order; re-runs unrelated branches | Graph-based via `consumes`/`produces` edges; only true descendants |
| Change detection | None | Content-hash diffing; byte-identical output invalidates nothing |
| Robustness | Non-atomic config writes + no parse guard, a corruption vector for all commands | Per-record validation; corrupt state preserved as `.corrupt` |
| Tests | Zero for the new code (~89 lines of branching) | 50 new offline tests, high coverage |
| SPEC/OpenSpec | ❌ Missing: violates CLAUDE.md's spec-coherence rule | ✅ Full SPEC.md §2.5/§3 + AC-9 + OpenSpec change |
| Verification gate | Preserved, but `--stage` exits 0 with no command-level gate | Preserved; `beforeCommit` seam fires strictly after the ERC/DRC gate, all paths tested |
| Reviewer verdict | Request changes | Mergeable (minor nits only) |

### The decisive finding on #31

Its headline feature is inert. The `stageCompletion` records it writes to `.copperhead/config.json` are never read anywhere (grep-confirmed). Re-runs actually happen only because a flag forces the loop, so the "durable, resumable audit trail" the PR advertises doesn't function. It also ships no tests and no spec artifacts, and introduces a config-corruption risk. The "6 failing tests are pre-existing" claim did check out, but that's beside the point since the new flags have no tests at all.

### #33's tradeoffs

Solid and invariant-preserving. The 2344 lines are mostly spec/docs/tests; net new logic is ~450 lines in a cleanly separated, testable `stagestate.ts`. The `beforeCommit` seam does not weaken the verification gate. Minor nits only: the default self-heal run can sweep unrelated WIP into a commit via `git add -A` (no dirty-tree warning on that path), and a perf cost from re-hashing on each cascade step. None blocking.

### Recommendation

Merge #33, close #31 as superseded, which is exactly what #33's author already proposed. #33 will need a rebase (currently conflicts with main) and coordination with #29.

pavvann:
> still has two conflicts, ready to merge this after they get resolved.

should be good to merge now :)

animesh-chouhan:
🤖 **Automated pr-review pass** (Claude Code `/pr-review`; not a human review, and not a merge-gating GitHub review). Reviewed at merge-base `origin/main` @9435fa5.

## Verdict: **Approve** (one low-severity nit)

This is a well-constructed change. The stage graph is genuinely data (per-artifact `consumes`/`produces`), invalidation cascades transitively in correct topological order, and the hardest edges (probe-gated records, changed-inputs-outrank-probe, walked-artifact rename detection, model-free/kicad-free dry-run) are all handled and tested. All repo invariants hold. I ran the suite on the branch: **250 passed / 10 key-gated skips**, typecheck clean, `openspec validate rerun-create-stages` passes.

## Verification performed

- `npm ci && npm run typecheck` — clean.
- `npm test` — 250 passed, 10 skipped (key-gated), against kicad-cli present. Matches the PR's claimed status.
- `openspec validate rerun-create-stages` — valid.
- Traced the `--stage part-selection` cascade by hand: `bom` change → schematic + outputs queued → schematic run re-queues layout-draft/firmware/devplan → queue re-sorts by stage index each pop, so `outputs` correctly runs *after* `layout-draft` (which produces `board`, an outputs input). No double-runs (`ran`/`queue.includes` guards), no backward cascade (`changedThisRun` tracks declared `produces` only).

## Invariant checks (all pass)

- **Spec-gated in** — `src/agent/tools.ts` untouched; edit/write tools stay absent until proposal validates. Not touched by this PR.
- **Verification-gated out** — the new `beforeCommit` seam ([loop.ts:555](src/agent/loop.ts#L555)) fires only on the success/commit path, after all gates and the changelog clear, immediately before `commitAll`; a throw routes to `commit-failed`. It never fires on refuse/dry-run/rollback (covered by three tests). The runner additionally halts with `ok:false` when a stage's post-run completion probe fails ([create.ts:453](src/commands/create.ts#L453)), so absent work never advances the pipeline.
- **`check`/`verify` LLM-free & network-free** — `src/commands/check.ts` untouched; `create --dry-run` resolves classification before any model/kicad-cli resolution ([cli.ts](src/cli.ts), [create.ts:320](src/commands/create.ts#L320)), and none of the `isComplete` probes shell out to kicad-cli.
- **No sexp serialization** — `src/kicad/sexp.ts` untouched; `stagestate.ts` only reads bytes for hashing.
- **Sync-obligations ledger** — untouched; `beforeCommit` is inserted strictly after the existing gate on the already-obligation-cleared success path.
- **Secrets** — `.copperhead/create-state.json` holds only stage names, run ids, ISO timestamps, and sha256 hashes ([stagestate.ts:154](src/memory/stagestate.ts#L154)); redaction paths untouched. `.gitignore` still carries `.env` and `.copperhead/runs/`; `create-state.json` is correctly *not* ignored (committed design state).
- **Spec coherence** — SPEC.md §2.5/§3 + AC-9.1–9.9 move with the delta specs; change validates.

## Findings

### 1. (Low / nit) Default self-healing re-run has no dirty-tree warning

[create.ts:332](src/commands/create.ts#L332) — the "working tree is dirty; uncommitted changes will be included in the re-run stage commit(s)" warning is gated on `opts.stage || opts.from`, so it is not shown for a plain `create` that heals a *stale* stage. Since stages commit with `git add -A`, a user who hand-edits `BOM.md` to reconcile while also holding unrelated WIP will have that WIP swept into the schematic re-run's commit with no warning — the exact hazard the targeted-mode warning guards against, now reachable from the default path too.

Failure scenario: `docs/BOM.md` edited + an unrelated `README.md` edit in the tree → `copperhead create` → schematic classifies stale and re-runs → its commit silently includes the `README.md` change.

This is consistent with create's long-standing "runs as if `--allow-dirty`" behavior, so it is a sharp edge rather than a regression. Consider extending the dirty warning to any run whose plan is non-empty (drop the `opts.stage || opts.from` guard), or noting the sweep in the default-mode plan log.

## Notes (not findings)

- The `spec` artifact bundles `SPEC.md` with `constraints.json`, so a downstream `record_constraint` marks architecture/part-selection stale on the next default run. Already called out and accepted in design.md Risks; fine to defer.
- `create` now passes `confirm: confirmTty` to the loop. `confirm` is only invoked under `ctx.interactive` ([tools.ts:192](src/agent/tools.ts#L192)), so autonomous runs are unaffected — and this actually fixes the interactive spec-approval gate, which previously auto-approved because `create` passed no `confirm`.
- `create --stage bom --from schematic` errors with "unknown stage bom" (bom is an artifact, not a stage) rather than the mutual-exclusion message, since name validation runs first. Behavior still satisfies AC-9.6 (exits non-zero, no run starts); the message just differs from the delta-spec scenario's phrasing.

animesh-chouhan:
hey, can you pleaes resolve the merge conflicts. will include this change in today's release @pavvann

pavvann:
> hey, can you pleaes resolve the merge conflicts. will include this change in today's release @pavvann

should be good to go now :)

pavvann:
@animesh-chouhan we should be now good to go. last review is literally a grammar fix lol

pavvann:
gm @animesh-chouhan lmk if this is still on the radar :)

PR #34, opened 2026-07-21
fix: Add --keep-on-fail debug flag: skip rollback/clean of the working tree on failed runs

### LINKED ISSUE #16: Add --keep-on-fail debug flag: skip rollback/clean of the working tree on failed runs
## Problem

Every failed run is hard-rolled-back: `fail()` (`src/agent/loop.ts:182-184`) calls `restore()` (`src/util/git.ts:45-51`), which runs `git reset --hard <snapshot>` followed by `git clean -fd -e .copperhead/runs`. That `clean -fd` deletes all untracked files the agent wrote (docs/, openspec/, KiCad files).

This is the right default — a failed run means the verification gates never passed, so the tree shouldn't keep unverified edits. But when debugging copperhead itself (or a failing stage), the rollback destroys exactly the evidence you want to inspect: the actual files the agent produced. The transcript records tool calls, but reading a whole design doc out of `transcript.jsonl` is much worse than opening the file. Two recent spec-seed runs failed on turn-budget exhaustion with ~90%-complete spec docs, deleted both times.

## Proposal

Add a `--keep-on-fail` (or `--no-rollback` / `--debug`) flag to `copperhead run` / `copperhead create`:

- On failure, **skip** `restore()` and leave the working tree exactly as the agent left it.
- Print a prominent warning with the snapshot ref and the manual recovery command, e.g.:
  `run failed; tree left dirty for inspection (--keep-on-fail). To roll back: git reset --hard <head> && git clean -fd -e .copperhead/runs`
- Do NOT commit anything — the failure semantics stay the same (no commit, obligations unmet); only the cleanup is skipped.
- Summary.md should record that rollback was skipped.

## Details / edge cases

- The next run's clean-tree check (`src/agent/loop.ts:118`) will refuse a dirty tree, which is good — it forces an explicit decision (`git reset --hard` or `--allow-dirty`) before re-running, so the flag can't silently compound state across runs.
- With `--allow-dirty` + `--keep-on-fail`, the stash-based snapshot (`git stash create`) still exists; the warning should print the stash ref too.
- Complements #15 (which proposes auto-stashing failed work by default): `--keep-on-fail` is the zero-friction debug version — files stay in place, no stash archaeology needed.

cneuralnetwork:
## What

Add `--keep-on-fail` to `copperhead do` and `copperhead create` so an unrecoverable failed run can leave its edits in place for debugging. Failure still exits non-zero, leaves obligations unmet, and never commits.

## Why

Default rollback is the safe behavior, but it removes the exact generated files needed to diagnose a failing stage. This closes [issue #16](https://github.com/chouhanindustries/copperhead/issues/16) while keeping rollback as the default.

## Design

- Thread the opt-in flag through [cli.ts](src/cli.ts), [create.ts](src/commands/create.ts), and the provider-agnostic loop in [loop.ts](src/agent/loop.ts).
- Skip restore only for unrecoverable failures when explicitly requested. Constraint refusals still restore.
- Record rollback status and recovery details in the transcript and `summary.md`.
- Print a shell-safe, audit-preserving recovery command, including the dirty-tree stash object when applicable.
- Refuse stale dirty state at create entry so retained partial artifacts cannot satisfy stage-completion markers on rerun.
- Reject `--dry-run --keep-on-fail` because their write guarantees conflict.

## Spec / OpenSpec

Updated `openspec/specs/SPEC.md` and the active `build-copperhead-phase-1` proposal, design, capability deltas, and task record. The change implements AC-3.11 and validates alongside the upstream `turn-budget-continue-and-loop-efficiency` change.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | Pass: 148 passed |
| Live-LLM (opt-in) | `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `COPPERHEAD_TEST_CODEX=1` | Skipped: credentials and opt-in unavailable |
| Docs | `ASTRO_TELEMETRY_DISABLED=1 npm run docs:build` | Pass |
| OpenSpec | `openspec validate build-copperhead-phase-1` | Pass |
| OpenSpec | `openspec validate turn-budget-continue-and-loop-efficiency` | Pass |

Focused regression run: 60/60 tests passed across keep-on-fail, create hardening, per-stage budgets, preflight, observability, and git safety.

- Sandbox variant used (`create` / `edit`): CLI help surface only. A live provider sandbox run was not performed because provider credentials were unavailable.
- Commands run and outcome:

```text
$ node dist/cli.js do --help
--keep-on-fail, --dry-run, and --max-turns were present.

$ node dist/cli.js create --help
--keep-on-fail and --interactive were present.
```

## Docs

Updated the README, CLI reference, guardrails documentation, source specification, and active Phase 1 OpenSpec artifacts. The docs clarify create rerun safety, dry-run incompatibility, refusal behavior, and manual recovery.

---

## Invariant checklist

- [x] **Spec-gated in**: structural tool omission is unchanged; gating tests pass.
- [x] **Verification-gated out**: ERC/DRC, repair, failure status, and no-commit behavior remain enforced.
- [x] **`check`/`verify` stays LLM-free and network-free**: the deterministic check module graph remains unchanged by this PR.
- [x] **No sexp serialization**: KiCad parsing and anchored text-edit behavior are unchanged.
- [x] **Sync-obligations ledger**: obligation tracking and commit refusal remain unchanged and covered by the offline suite.
- [x] **Secrets**: new transcript and summary fields continue through write-time redaction; recovery output contains only git refs.
- [x] **Spec coherence**: SPEC.md and active change artifacts moved together; both applicable OpenSpec validations pass.

Fixes chouhanindustries/copperhead#16

## Summary by CodeRabbit

* **New Features**
  * Added `--keep-on-fail` to `do` and `create` to preserve unrecoverable failures for inspection.
  * When enabled, Copperhead skips automatic rollback/commits and prints a manual recovery command; `--dry-run` is rejected when combined with `--keep-on-fail`.
* **Documentation**
  * Updated README and CLI/reference docs to detail kept-failure behavior, rerun/resume conditions, and stricter `create` dirty-tree preflight rules.
  * Documented rollback/skip reporting in the run summary.
* **Tests**
  * Expanded coverage for CLI flag forwarding, kept-failure outcomes, recovery-command output, and `create` refusal/resume scenarios (including CLI symlink smoke tests).


animesh-chouhan (commented):
# Review: PR #34, `--keep-on-fail` debug flag

## Overview

Clean, well-scoped implementation of issue #16. The flag is threaded `cli.ts` -> `RunOptions` -> `fail()` in `src/agent/loop.ts`, plus `create.ts` stage forwarding, and it touches exactly the right seam: only the cleanup step inside `fail()`, so every unrecoverable exit path (`provider-error`, `stalled`, `repair-cycles-exhausted`, `commit-failed`, `turn-budget-exhausted`) gets uniform behavior while the no-commit rule, exit code, verification gates, and open-obligations reporting stay untouched. Docs and OpenSpec artifacts are synced per the CLAUDE.md convention (SPEC.md AC-3.11 maps 1:1 onto the new delta-spec scenarios in `agent-core`, `cli-surface`, `safety-rails`, and `create-pipeline`). Nice work overall; one real safety gap in the `create` interaction, detailed below.

## What is done well

- **Right altitude for the skip.** Guarding only the `restore()` call inside `fail()` (loop.ts) means the flag cannot accidentally widen: `run-end`, `summary.md`, the outcome line, and the Synap-free failure return all still execute, consistent with the recent "survive a failed rollback" hardening (cee78f3).
- **Observability is first-class.** The `rollback-skipped` JSONL event, the `keepOnFail` field on `run-failed`, and the new `## Rollback` section in `summary.md` (transcript.ts) with status `completed | skipped | failed` are exactly what a debugging flag should leave behind. Bonus: the pre-existing rollback-failed path now also gets the recovery command recorded in `summary.md`, a genuine improvement over main.
- **`recoveryCommand()` quoting.** `shellQuote` in `src/util/git.ts` is correct POSIX single-quote escaping, and it has a dedicated unit test in `test/safety.test.ts`.
- **The `--allow-dirty` test is excellent.** `test/observability.test.ts` ("reports the stash ref and restores it with the printed recipe") parses the actual printed output, then executes the exact printed reset/clean/stash-apply sequence and asserts the user's edit survives while the agent's debris is gone. That test proves the recovery instructions work rather than just that they were printed.
- Both surfaces (transcript and summary) pass through `redactSecrets` at write time, so the new rollback section inherits AC-4.1 for free.

## Correctness

### Major: `create --keep-on-fail` can poison stage resumability

The next-run safety story in the docs is only true for `do`. `runCreate` (src/commands/create.ts, line 121) always passes `allowDirty: true`, and stage completion is inferred from repo state via existence/marker checks (`isComplete`: `BOM.md` exists, `outputs/` exists, `firmware/` exists, `SPEC.md` contains `## Budgets`, ...). Design D10 says resumability works "for free" precisely because a failed stage rolls back and leaves no partial artifacts. With `--keep-on-fail`:

- A re-run of `create` is **not** refused by the dirty-tree preflight (allowDirty is hardcoded per stage), contradicting the added claims in `docs/src/content/docs/concepts/guardrails.md` (line 31: "the next run refuses this dirty tree"), the `safety-rails` delta ("the dirty tree SHALL therefore still be refused by the next default run"), and SPEC.md §7.
- Worse, a kept partial artifact can make `isComplete` return true, so the pipeline silently resumes **past** a stage whose output failed verification. Example: the `firmware` stage fails after `mkdir firmware/`; re-running `create` skips the stage entirely.

Suggestions, in increasing order of effort:

1. At minimum, correct the docs/spec wording to scope the "next run refuses the dirty tree" guarantee to `do`, and make the `create` failure warning explicit: "recover the tree before re-running create; resume detection may treat this partial output as complete."
2. Better: have `runCreate` refuse to start (or require an explicit acknowledgment) when the tree is dirty at pipeline start, mirroring `do`'s preflight, since `allowDirty: true` was only ever meant for intra-pipeline state between stages, not stale state from a previous failed pipeline.
3. Add a `create-pipeline` scenario covering re-run-after-kept-failure; the new "Failed stage retained for debugging" scenario stops one step short of the dangerous part.

### Minor: printed recovery command is not the exact equivalent of `restore()`

The docstring on `recoveryCommand()` says it prints "the exact manual equivalent of restore()", but `restore()` (git.ts lines 93-128) first backs up `.copperhead/runs` outside the repo, precisely because the runs dir can be staged and `git reset --hard` deletes staged paths before `clean -e` can protect them; the comment there even names "while preserving failed work" as the scenario. The printed recipe (`git reset --hard <head> && git clean -fd -e .copperhead/runs && git stash apply <stash>`) has no such protection, so in the `commit-failed` + `--keep-on-fail` case (where `commitAll` already ran `git add -A`) the user's copy-paste recovery can destroy the very audit trail the flag exists to preserve. It is an edge case (the runs dir is normally gitignored per AC-4.3), but since the whole point is a trustworthy printed command, consider prefixing the recipe with something like `git reset -q -- .copperhead/runs` or adding a one-line caveat to the warning.

### Minor: `--dry-run --keep-on-fail` contradiction

Nothing rejects the combination. A dry run that hits `fail()` (for example a provider error mid-run) will now leave writes on disk, violating `--dry-run`'s "propose the diff and write nothing" contract (AC-3.9). Suggest ignoring `keepOnFail` when `dryRun` is set, or erroring on the combination in `cli.ts`.

### Minor: `filesTouched` on the kept path is both under-reporting and untested

`filesTouched: keepOnFail ? [...ctx.filesTouched] : []` (loop.ts) only reflects tool-mediated writes. On `turn-budget-exhausted` the loop already computes `changedFiles(repoRoot, snap.head)`, which is the accurate on-disk answer; on provider-error the set can be empty even though the tree is dirty (as in both new tests, where the scripted provider writes the file out of band, so this new line is never exercised with a non-empty set). Suggest using `changedFiles()` for the kept case and asserting it in a test, or dropping the change if callers do not need it.

### Nit: rollback-failed CLI branch still omits the recovery command

`manualRecovery` is computed unconditionally in `fail()`, and it lands in `summary.md` for the `failed` status, but the `else if (restoreError)` log branch still prints only the warning. Since the string is already in hand, printing it there too would make the pre-existing "rollback failed" path as recoverable as the new one.

### Nit: refusal path is untouched

`finish({outcome: 'refuse'})` still restores unconditionally (loop.ts line 362, and that `restore` is unguarded, pre-existing). This matches the spec wording ("unrecoverable failure"), but since a refusal can also discard partial edits worth inspecting, consider stating the exclusion explicitly in the CLI help or docs so nobody expects `--keep-on-fail` to cover it.

## Project conventions

- OpenSpec hygiene is exemplary: `proposal.md`, `design.md` D6, four delta specs, `tasks.md` 8.10, and SPEC.md AC-3.11 all move together, exactly what CLAUDE.md asks for.
- One asymmetry worth a deliberate note: `sync`'s resolve phase also runs the agent loop with snapshot/rollback but does not get the flag. Fine as scoped to issue #16, but a follow-up issue would keep the surface consistent.
- The `tasks.md` status-header rewrite folds this PR's local run counts ("115 tests green; 7 live tests skipped") into a note that also claims live verification this PR did not perform; harmless, but keep the header's claims attributable.

## Performance

Negligible. `recoveryCommand()` is pure string work, there are no extra git invocations on the success path, and the failure path actually does less work when the flag is set.

## Test coverage

Good breadth: default-rollback regression, kept-output happy path, allow-dirty round trip that executes the printed recipe, shell-quoting unit, and create-stage forwarding with a mocked loop. Gaps:

- `test/cli-help.test.ts` asserts regexes against the **source text** of `src/cli.ts`. That is brittle (an equivalent refactor like `keepOnFail: Boolean(opts.keepOnFail)` breaks it) and tests the implementation, not the behavior its own name promises. Prefer running the built CLI with `--help` via `execa`, or exporting the commander program and inspecting `command.options`.
- No test covers `commit-failed` + `keepOnFail` (the staged-index case) or the non-empty `filesTouched` return, per the minor above.
- No test covers the `create` re-run-after-kept-failure resume behavior, which is where the major issue hides.

## Security and safety of skipping rollback

The core invariants hold: the flag is opt-in, failure still exits non-zero, no commit is created, obligations stay open and are reported, and for `do` the dirty-tree preflight forces an explicit decision before the next run. Kept state can include a half-updated `.copperhead/constraints.json` or a leftover validated `openspec/changes/<id>` directory; a subsequent `--allow-dirty` run would read those as truth, which is inherent to a debug flag and acceptable given it is documented as debugging-only, but the guardrails page could name those two artifacts specifically. Secret hygiene is unaffected: the recovery command contains only git object ids, and both the transcript and summary are redacted at write time.

## Verdict

Approve direction; request changes for the `create` resumability gap (major) and the doc/spec claims that overstate the next-run refusal guarantee. The rest is polish.

cneuralnetwork:
Addressed review feedback in commit 7689daf. The update adds create-pipeline dirty-state preflight/resume protection, rejects --dry-run with --keep-on-fail, reports on-disk changed files for kept failures, preserves .copperhead/runs during printed recovery, prints recovery guidance when rollback itself fails, documents refusal behavior, replaces brittle CLI-help coverage, and adds regression tests/spec updates. Verification: typecheck passed; npm test 120 passed with 7 live LLM tests skipped.

animesh-chouhan:
🤖 **Automated pr-review pass** (Claude Code, not a human review — a formal review is not implied and merge gating is unaffected)

## Verdict: Request changes

The `--keep-on-fail` feature itself is well-built, well-specified, and well-tested, and it holds every repo invariant. But the same PR refactors the CLI entry point, and that refactor ships a regression that makes the **installed `copperhead` binary a silent no-op** on Linux/macOS. That one issue is a shipping blocker; the feature work is otherwise solid.

Verified locally on the PR branch: `npm run typecheck` ✓, `npm run build` ✓, `npm test` → 158 passed / 10 skipped (live-LLM) ✓, `openspec validate` on both changes ✓.

---

## Findings

### HIGH — the installed CLI entry-point guard never fires through an npm-linked bin (silent no-op)
[src/cli.ts:246](src/cli.ts#L246)

The refactor replaced the unconditional `program.parseAsync()` with a guarded `main()`:

```js
if (process.argv[1] && path.resolve(process.argv[1]) === fileURLToPath(import.meta.url)) {
  main().catch(...)
}
```

`path.resolve()` does **not** resolve symlinks, but Node resolves a module's real path for `import.meta.url`. npm installs the `copperhead` bin as a symlink (`node_modules/.bin/copperhead → ../copperhead/dist/cli.js`) on Linux/macOS, so at runtime `argv[1]` is the *symlink* path while `import.meta.url` is the *realpath* (`dist/cli.js`). They never match → `main()` is never called → the process exits 0 having done nothing.

Failure scenario (reproduced against the actual built `dist/cli.js`):
- `node dist/cli.js --help` → prints the usage banner (this is exactly how the PR's manual test log invoked it, which is why it looked fine).
- Same file invoked through a symlink (i.e. `copperhead --help`, `npx copperhead`, any global/local install on Unix) → prints nothing, exits 0.

So every real user invocation of the published binary is dead. Windows `.cmd` shims pass the real path and would still work, which makes this easy to miss.

Fix: resolve the symlink before comparing, e.g.
```js
import { realpathSync } from 'node:fs';
if (process.argv[1] && realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) { ... }
```
(verified: the `realpathSync` variant returns `true` under the same symlink.) A CLI smoke test that invokes the binary *through a symlink* (not `node dist/cli.js`) would have caught this and would guard against regressions.

Scope note: this entry refactor exists only to make `program`/`main` importable for the new `test/cli-help.test.ts`. A test-support change should not have altered production launch behavior — worth isolating.

### MEDIUM — new `create` clean-tree entry preflight rejects an untracked in-repo brief
[src/commands/create.ts:130](src/commands/create.ts#L130)

The entry guard refuses when `isDirty()` is true, and `git status --porcelain` counts untracked files. The natural workflow — write `brief.md` inside your project, run `copperhead create --brief brief.md` — now fails with "working tree is dirty; copperhead create refuses to infer completed stages…", even though the brief is *input*, not stale stage output. The new tests sidestep this by placing the brief in a separate temp dir (`mkdtemp(... 'ch-create-brief-')`), so CI doesn't exercise the common case.

This is partly by-design (the design/spec do call for a clean entry tree), but flagging an input file is a false positive. Options: exclude the resolved brief path from the dirtiness check; or only refuse when a known stage-marker path (`outputs/`, `firmware/`, `.copperhead/constraints.json`, `openspec/changes/<id>`) is dirty; or at minimum have the error name the brief and tell the user to commit it. Failure scenario: first-time user follows the README, gets an opaque refusal on their very first `create`.

### LOW — redundant double restore on an ordinary first-stage `create` failure
[src/commands/create.ts:186](src/commands/create.ts#L186)

When the first stage fails without `--keep-on-fail`, `runAgentLoop`'s `fail()` already hard-restores that stage's snapshot, and then `runCreate` calls `restore(pipelineStart)` again. Both are idempotent hard resets to the same (or equivalent) HEAD, so this is harmless — just two `reset --hard`/`clean` passes where one would do. Noted for awareness, not a blocker.

---

- **Spec-gated in** — no change to the tool list or its gating; `edit_file`/`write_file` remain structurally absent until a proposal validates. ✓
- **Verification-gated out** — `--keep-on-fail` only skips the post-*failure* rollback; ERC/DRC, the repair loop, the no-commit-on-failure rule, and the non-zero exit are all unchanged. `keepOnFail` is reachable only through `fail()`, which never commits. ✓
- **`check`/`verify` stays LLM-free / network-free** — untouched; `create.ts` only newly imports `git`/`preflight` helpers, no provider or network. ✓
- **No sexp serialization** — parser untouched. ✓
- **Sync-obligations ledger** — `fail()` still reports open obligations in `summary.md`; commit refusal path unchanged. ✓
- **Secrets** — new transcript/`summary.md` fields (`rollback.head`/`stash`/`recoveryCommand`) carry only git refs and a git command; write-time `sk-…` redaction path is unchanged. ✓
- **Constraint refusals still restore** — the refusal branch ([loop.ts:506](src/agent/loop.ts#L506)) restores unconditionally, ignoring `keepOnFail`, matching the spec. ✓
- **`--dry-run --keep-on-fail` rejected before any write** — enforced at the top of `runAgentLoop` ([loop.ts:150](src/agent/loop.ts#L150)). ✓
- **Spec coherence** — SPEC.md and all `build-copperhead-phase-1` artifacts moved together; both it and `turn-budget-continue-and-loop-efficiency` validate. ✓

## Bottom line

Land the HIGH fix (one-line `realpathSync`) plus a symlink-invoked smoke test before merge — without it the release binary does nothing. The MEDIUM is a real first-run UX papercut worth addressing in the same PR. Everything else, including the entire `--keep-on-fail` design, is clean.

Issue #13, opened 2026-07-21
Mouser cart CSV emitter for supplier BOM export (golden-file tested)


animesh-chouhan:
## Context

The planned `copperhead export bom --supplier mouser` (OpenSpec change `add-supplier-bom-export`, task 1.3) emits a Mouser cart-upload CSV. One emitter function, one golden file; see the sibling JLCPCB and DigiKey issues for the pattern.

## What to do

1. Implement one emitter function producing Mouser's cart upload CSV: MPN, manufacturer, quantity, customer reference (refdes group).
2. Quantities come from the shared order-quantity arithmetic (boards + spares); stub the interface if that issue has not merged.
3. Golden-file test from the fixture BOM, asserted byte-equal in CI.

## Definition of done (from the delta spec)

- Rows carry MPN, manufacturer, and computed quantity from BOM.md.
- MPN-less rows and `UNVERIFIED` rows are excluded and named in the warnings footer; `--include-unverified` behaves as specified in the change.
- Golden file committed and asserted.

## Notes

Offline, no API key. Familiarity with Mouser's upload template (column names, delimiter tolerance) is the valuable part.

Issue #11, opened 2026-07-21
JLCPCB assembly CSV emitter for supplier BOM export (golden-file tested)


animesh-chouhan:
## Context

The planned `copperhead export bom --supplier jlcpcb` (OpenSpec change `add-supplier-bom-export`, task 1.3) emits JLCPCB's assembly-service CSV. Each supplier format lives in exactly one emitter function with a golden output file in the test suite, so format churn stays a one-function fix. This issue is the JLCPCB emitter; DigiKey and Mouser are separate issues.

## What to do

1. Implement one emitter function producing JLCPCB's assembly CSV: Comment, Designator (grouped per line), Footprint, LCSC Part # (blank when BOM.md has no LCSC column; JLCPCB accepts blank with manual matching).
2. Input is the parsed BOM.md row shape plus computed quantities (see the shared parser and quantity arithmetic issues; you can stub their interfaces if they have not merged).
3. Golden-file test: emit from the fixture BOM, commit the expected CSV, assert byte equality.

## Definition of done (from the delta spec)

- Header and rows match JLCPCB's current assembly upload format, designators grouped per line.
- Rows lacking an MPN and rows flagged `UNVERIFIED` are excluded and named in a warnings footer (stderr + CSV comment).
- Golden file committed and asserted in CI.

## Notes

Offline, no API key. Real-world experience ordering JLCPCB assembly is the valuable contribution here: format details (encoding, header spelling, grouping) matter more than the TypeScript.

tanav29:
i have tried add this feature here 
https://github.com/chouhanindustries/copperhead/pull/36
pls check

Issue #7, opened 2026-07-21
Implement order-quantity arithmetic for supplier BOM export


animesh-chouhan:
## Context

The planned `copperhead export bom` command (OpenSpec change `add-supplier-bom-export`, task 1.2) computes per-line order quantities from board count and a spares percentage. The rules are fully specified; this issue is a pure function plus tests.

## Spec

- Base: `ceil(perBoardCount x boards x (1 + spares/100))`, with `--boards` defaulting to 1 and `--spares` defaulting to 10.
- Passive minimum: for lines whose footprint has prefix `R_`, `C_`, or `L_`, raise the result to `perBoardCount x boards + 2` when the percentage yields less (losing two 0402s to tweezers is normal).

## Definition of done (from the delta spec, these are the test cases)

- A line with 4 parts per board at `--boards 10 --spares 10` exports quantity 44.
- A passive line with 1 part per board at `--boards 1 --spares 10` exports quantity 3.
- Edge cases covered: zero spares, non-passive single part, fractional ceil results.

## Notes

Implement as a standalone function (suggested home: `src/kicad/bom-export.ts`) with vitest unit tests. Offline, no API key. Depends loosely on the shared BOM parser issue but can be built independently against the typed row shape.

Chirag6722:
Pretty sure this one's already done — `orderQuantity()` in `src/kicad/bom-export.ts` (from #45) matches the spec here almost exactly:

​```ts
export function orderQuantity(
  perBoardCount: number,
  boards: number,
  sparesPercent: number,
  isPassive: boolean,
): number {
  const base = perBoardCount * boards;
  const withSpares = Math.ceil((base * (100 + sparesPercent)) / 100 - 1e-9);
  return isPassive ? Math.max(withSpares, base + 2) : withSpares;
}
​```

Checked it against the test cases in the DoD and they hold up (4×10 boards @10% spares → 44, 1 passive part @1 board/10% spares → 3). It also handles a float rounding edge case that isn't even in the spec.

Mind closing this as resolved by #45, or is it staying open for a separate reason?

PR #35, opened 2026-07-21
fix(export): defuse spreadsheet formulas in supplier CSV fields

macayu17:
## Summary

#45 landed the full `add-supplier-bom-export` change, including the Mouser emitter this PR originally carried, so I have rebased this branch onto main and repointed it at the one review item from #35 that did not make it in: CSV formula injection in the shared `csvField`.

## What changes

BOM.md cells are agent-written, and the exports are routinely opened in a spreadsheet before being uploaded to a supplier. A cell starting with `=`, `+`, `-` or `@` is evaluated as a formula there, so `csvField` now prefixes those fields with an apostrophe on top of the existing RFC 4180 quoting.

The guard sits in the shared helper, so it covers the JLCPCB, DigiKey and Mouser formats at once. No real MPN, manufacturer, value or designator opens with one of those characters, so nothing orderable is altered and every existing golden expectation is untouched.

## Tests

One case added to the golden emitter block: a BOM row with `=cmd'/c calc'` as its MPN and `-Yageo` as its manufacturer comes out apostrophe-prefixed.

## Verification

- `npm run typecheck` passed
- `npx vitest run test/bom-export.test.ts`, 25 passed

### DIFF
--- a/src/kicad/bom-export.ts
+++ b/src/kicad/bom-export.ts
@@ -164,9 +164,17 @@ function naturalCompare(a: string, b: string): number {
   return a.localeCompare(b);
 }
 
-/** RFC-4180 field quoting: quote when the field holds a comma, quote, or newline. */
+/**
+ * RFC-4180 field quoting: quote when the field holds a comma, quote, or newline.
+ *
+ * Fields opening with `=`, `+`, `-`, or `@` are also prefixed with an
+ * apostrophe: BOM.md cells are agent-written, and a spreadsheet opened on the
+ * export before upload would evaluate such a cell as a formula. No real MPN,
+ * manufacturer, or designator starts with one, so nothing orderable is altered.
+ */
 function csvField(value: string): string {
-  return /[",\n\r]/.test(value) ? `"${value.replace(/"/g, '""')}"` : value;
+  const v = /^[=+\-@]/.test(value) ? `'${value}` : value;
+  return /[",\n\r]/.test(v) ? `"${v.replace(/"/g, '""')}"` : v;
 }
 
 const csvRow = (fields: string[]): string => fields.map(csvField).join(',');
--- a/test/bom-export.test.ts
+++ b/test/bom-export.test.ts
@@ -122,6 +122,16 @@ describe('supplier format emitters (golden, supplier-bom-export)', () => {
         'RC0603FR-0710KL,Yageo,3,R1\n',
     );
   });
+
+  it('defuses cells a spreadsheet would evaluate as a formula', () => {
+    const hostile = parseBom(
+      '| Refdes | Value | Footprint | MPN | Manufacturer |\n' +
+        '|---|---|---|---|---|\n' +
+        '| R1 | 10k | R_0603 | =cmd\'/c calc\' | -Yageo |\n',
+    );
+    const { csv } = buildExport(hostile, 'mouser', { boards: 1, spares: 10, includeUnverified: false });
+    expect(csv.split('\n')[1]).toBe("'=cmd'/c calc','-Yageo,3,R1");
+  });
 });
 
 describe('exclusion rules (supplier-bom-export)', () => {


macayu17:

macayu17:
recheck

animesh-chouhan (commented):
# Review: Mouser cart CSV emitter (PR #35)

## Overview

Clean, well-scoped implementation of the Mouser half of task 1.3 in `add-supplier-bom-export`: a pure emitter (`emitMouserCsv` in `src/kicad/bom-export.ts`) plus golden-file tests in `test/bom-export.test.ts` and two committed fixtures under `test/fixtures/bom-export/`. The function is deterministic, LLM-free, and network-free, which matches both the delta spec's requirement and the repo's `check`-adjacent safety posture. The PR description is exemplary: it states scope, format rationale, spec deviations, and stubbed interfaces explicitly.

## Blocker: file collision with PR #36

Both this PR and #36 (JLCPCB emitter) create `src/kicad/bom-export.ts` from scratch. Whichever merges second gets a hard conflict, and worse, the two designs diverge on every shared surface:

- **Input shape**: this PR takes pre-grouped `BomLine` (mpn, manufacturer, refdes[], quantity, unverified: boolean); #36 takes ungrouped `NormalizedBomRow` (per-refdes rows with `status`/`verification` strings, optional `lcsc`, an index signature) and does its own grouping and natural-sort inside the emitter.
- **Warnings shape**: `string[]` here vs structured `{ refdes, reason, message }[]` in #36.
- **Warning placement**: this PR keeps warnings out of the CSV (correct for Mouser, well argued in the PR body); #36 appends `# ...` comment lines inside the JLCPCB CSV. JLCPCB's upload parser has the same problem Mouser's does: a `#` line reads as a part row. The spec's "appended as CSV comments where the format permits" (delta spec line 28) is arguably satisfied by neither format permitting them.
- **Quoting policy**: minimal RFC 4180 quoting here vs quote-everything in #36. Both are valid CSV, but golden files from the two styles will look inconsistent.

Recommendation: pick one PR as the owner of the shared module skeleton (`BomLine`/`EmitOptions`/`EmitResult`, `csvField`, the exclusion/warning logic), land it first, and rebase the other onto it. This PR's shapes are the better foundation for cart-style formats (Mouser, DigiKey), but grouping/natural-sort has to live somewhere shared once #7's quantity arithmetic lands, otherwise JLCPCB and Mouser will group differently. The exclusion rules (MPN-less always out, UNVERIFIED gated by the flag) are literally duplicated logic between the two PRs today and should be one function.

## Correctness

- **Header and column set**: `Mfr. Part Number, Manufacturer, Quantity, Customer Part Number` matches the delta spec's "MPN, manufacturer, quantity, customer reference" for Mouser (delta spec line 6). Mouser's importer maps columns interactively, so exact header wording is low-risk. However, design D5 says formats are "version-stamped", and nothing here records which Mouser template revision or URL the header was copied from. Suggest a comment on `MOUSER_HEADER` citing the template source and date, so a future format break is diagnosable.
- **Escaping**: `csvField` (line 34 of the new file) is correct RFC 4180: quotes on `"` `,` CR LF, doubles embedded quotes. Good.
- **Space-joined refdes**: sensible and well justified; avoids quoting the customer-reference field entirely.
- **MPN check does not trim** (`if (!line.mpn)`): a whitespace-only MPN from a sloppy BOM.md cell passes the guard and emits a junk row. #36 trims all fields. Suggest `line.mpn.trim()` for the guard and the emitted value (same for manufacturer). Whether trimming belongs in the emitter or the upstream parser, decide it once with #36.
- **No duplicate-MPN merging**: two `BomLine`s with the same MPN emit two rows. The doc comment says input is "grouped by part" upstream, so this is a contract, not a bug, but the emitter cannot detect a violated contract. A cheap dev-time guard (or at least a warning on duplicate MPNs) would keep a bad upstream grouping from silently producing a cart with split quantities.
- **Quantity is unvalidated**: `String(line.quantity)` will happily emit `NaN`, `0`, `-1`, or `2.5`. Until #7 lands, a `Number.isInteger(q) && q > 0` guard with a warning would make the stub contract enforceable rather than aspirational.
- **Empty refdes array**: `ref` becomes `''` and the warning reads `": excluded, no MPN in BOM.md"`. #36 uses `<unnamed row>`. Nit, but worth aligning.
- **Warning wording when included**: `"...is UNVERIFIED, included (--include-unverified)"` reads as an instruction to pass a flag that was already passed. Suggest `included via --include-unverified` for that branch. Also, referencing the CLI flag string from a library function is a small layering leak; the command layer (task 2.1) could own the flag phrasing.

## Conventions

- Location `src/kicad/bom-export.ts` matches tasks.md 1.3 exactly, and the module doc comment correctly cites D5. Good spec traceability.
- Zero LLM, zero network, no writes: consistent with the export command's contract and the repo's invariants.
- `\n` line endings with a trailing newline: fine for Mouser, but see the Windows note under tests.

## Tests

Good coverage for the emitter's stated scope: golden byte-equality for both flag states, the MPN-rule-beats-flag case, and escaping. Two gaps and one platform risk:

- **CRLF risk on the golden files**: the tests assert byte equality against committed fixtures, and the repo has no `.gitattributes`. On a Windows checkout with `core.autocrlf=true`, the `.csv` fixtures check out with CRLF and both golden tests fail. Since the PR author develops on Windows (per the PR body's note about `test/safety.test.ts`), add a `.gitattributes` with `*.csv text eol=lf` (or `-text`) in this PR, or the golden tests will be flaky across contributor machines.
- **Empty input**: no test that `emitMouserCsv([])` yields header plus trailing newline and no warnings. One line, worth having since the command layer will hit it on an all-excluded BOM.
- **Newline-in-field**: `csvField` handles `\r\n` but no test exercises it. The escaping test covers comma and slash but not quotes-in-MPN or embedded newline.

## Performance

Non-issue at BOM scale (hundreds of lines, linear string work). No concerns.

## Security

- No injection surface into the toolchain: pure string formatting, no subprocess, no network, no file writes.
- **CSV formula injection**: fields starting with `=`, `+`, `-`, or `@` execute as formulas if a user opens the export in Excel before uploading, and BOM.md content is agent-written. Mouser's importer itself is unaffected, so this is defense-in-depth, but a one-line guard (prefix `'` or quote-and-space) or a documented decision not to guard would be worth adding once, in the shared `csvField`, for all three suppliers.

## What is done well

- The PR body documents every deviation and stub with rationale, exactly the openspec workflow this repo wants.
- Returning warnings to the caller instead of polluting the CSV is the right call for Mouser, and the argument generalizes: #36 should probably adopt it.
- `csvField` is minimal and correct; golden fixtures are small and readable; the fixture data reuses the existing open-key fixture parts, keeping the test universe coherent.
- Explicitly shaping `BomLine`/`EmitOptions`/`EmitResult` for sibling reuse shows the right instinct, it just needs #36 to actually adopt them.

macayu17:
Thanks, all the nits are addressed in 980c14c.

On the #36 collision: happy either way. If you want this PR to own the shared skeleton (`BomLine` / `EmitOptions` / `EmitResult` / `csvField` / the exclusion rules), merge it first and I'll help rebase #36's JLCPCB emitter onto these shapes. If you'd rather #36 lands first, say so and I'll rebase this one onto its `NormalizedBomRow` and structured-warning shapes instead.

Two things worth deciding once, whichever way it goes:

- Grouping and natural refdes sort belong in the shared layer, not per emitter, otherwise JLCPCB and Mouser will group differently once #7's arithmetic lands.
- CSV comment lines: I'd keep warnings out of the file for both suppliers.
  JLCPCB's uploader reads a `#` line as a part row the same way Mouser's does, so "where the format permits" is arguably satisfied by neither.

animesh-chouhan:
Heads-up: this looks like it overlaps with two other open PRs implementing the same `add-supplier-bom-export` change. All three add `src/kicad/bom-export.ts` and will conflict:

- #35 (this one) — Mouser emitter, closes #13
- #36 — JLCPCB emitter, fixes #11
- #45 — full `export bom` command with all three suppliers (jlcpcb/digikey/mouser) + CLI wiring + create-pipeline stage 6, fixes #42

#45 appears to subsume the per-supplier work here. Could you three sync up on whether these should be consolidated (or rebased so they don't collide on the same module) before we merge? Want to avoid duplicate effort and a painful three-way conflict.

macayu17:
#45 subsumes the Mouser emitter, so rather than leave a duplicate open I have rebased this branch onto main and cut it down to the single review item from here that #45 does not cover: the CSV formula-injection guard in the shared `csvField`.

It is now a 12-line diff against main, all three supplier formats covered by the one helper, no golden expectations changed. Happy to close instead if you would rather see it as a fresh PR.

Devesh36:

## Verdict: request changes

The CSV formula-defuse change is correct and well tested, but the PR fails the required `template` check and should not merge until the description matches the repo PR template.

### Metrics
```text
lines: +20 / -2 across 2 files  ·  src +10/-2, test +10
tests: +1 focused golden case for hostile `=` / `-` cells
ci: template FAIL; cla pass
```

### Invariants checked
- Spec-gated in: n/a
- Verification-gated out: n/a
- check/verify LLM-free: n/a
- No sexp serialization: n/a
- Secrets: pass (defuses spreadsheet formula injection in BOM export cells)

### Findings

**[medium] Required `template` CI check is failing**
- Where: PR description (GitHub Actions `template` job)
- Fails when: merge / required-check gating runs against the current PR body.
- Fix: Rewrite the PR body to follow the repo PR template sections and re-run the template workflow green.

**[low] Only `= + - @` prefixes are defused**
- Where: `src/kicad/bom-export.ts` (`csvField`)
- Fails when: a cell starts with tab/`\t` or whitespace-then-`=` that some spreadsheet engines also treat as formulas.
- Fix: Optionally also prefix/strip leading `\t` and whitespace-then-formula starters; add a regression test.

## Fix prompt

````text
You are fixing review findings on PR #35 in chouhanindustries/copperhead (branch feat/mouser-bom-emitter, based on main).
  gh pr checkout 35

FINDING 1 [medium] Required template CI check is failing
  Where: PR description (GitHub Actions template job)
  Fails when: required-check gating runs against the current PR body.
  Fix: Rewrite the PR body to follow the repo PR template; re-run template workflow green.

FINDING 2 [low] (optional) Only = + - @ prefixes are defused
  Where: src/kicad/bom-export.ts (csvField)
  Fails when: a cell starts with tab or whitespace-then-= treated as a formula.
  Fix: Also handle leading tab / whitespace-then-formula; add a regression test.

Then: npm run typecheck && npm run build && npm test
````

PR #36, opened 2026-07-21
Fixes #11: add isolated JLCPCB BOM emitter

### LINKED ISSUE #11: JLCPCB assembly CSV emitter for supplier BOM export (golden-file tested)

tanav29:
Implements the JLCPCB assembly CSV emitter for normalized BOM rows.

## Changes

- Adds src/kicad/bom-export.ts
- Emits the exact JLCPCB header:

- Groups compatible rows by value, footprint, and LCSC part number
- Sorts rows and designators naturally and deterministically
- Handles CSV escaping and blank LCSC values
- Excludes MPN-less and UNVERIFIED rows by default
- Supports includeUnverified
- Returns structured exclusion warnings and appends them as CSV comments
- Adds fixture and golden-file tests

## Scope

DigiKey, Mouser, CLI wiring, parser integration, quantity arithmetic, and create-pipeline integration
remain out of scope.

## Verification

- Focused JLCPCB tests pass
- Typecheck passes
- Full suite’s remaining failures are unrelated Windows/KiCad environment issues

### DIFF
--- a/openspec/changes/add-supplier-bom-export/tasks.md
+++ b/openspec/changes/add-supplier-bom-export/tasks.md
@@ -4,8 +4,8 @@
 
 - [ ] 1.1 Extract the BOM.md table parser from `src/memory/drift.ts` into a shared module and reuse it
 - [ ] 1.2 Implement quantity arithmetic (`ceil(qty × boards × (1 + spares/100))`, passive minimum `qty × boards + 2` for `R_`/`C_`/`L_` footprint prefixes) with unit tests
-- [ ] 1.3 Implement per-supplier emitters in `src/kicad/bom-export.ts`: JLCPCB assembly CSV, DigiKey cart CSV, Mouser cart CSV, each as one function
-- [ ] 1.4 Implement exclusion rules: MPN-less rows always excluded, `UNVERIFIED` rows excluded unless `--include-unverified`; warnings footer to stderr and CSV comments where the format permits
+- [ ] 1.3 Implement per-supplier emitters in `src/kicad/bom-export.ts`: JLCPCB assembly CSV, DigiKey cart CSV, Mouser cart CSV, each as one function (JLCPCB emitter implemented in PR #36)
+- [ ] 1.4 Implement exclusion rules: MPN-less rows always excluded, `UNVERIFIED` rows excluded unless `--include-unverified`; warnings footer to stderr and CSV comments where the format permits (JLCPCB rules implemented in PR #36; CSV warning comments are opt-in pending parser confirmation)
 
 ## 2. CLI wiring
 
@@ -15,8 +15,8 @@
 
 ## 3. Tests
 
-- [ ] 3.1 Golden-file tests per supplier format against the fixture BOM
-- [ ] 3.2 Exclusion and `--include-unverified` behavior tests
+- [ ] 3.1 Golden-file tests per supplier format against the fixture BOM (JLCPCB golden test implemented in PR #36)
+- [ ] 3.2 Exclusion and `--include-unverified` behavior tests (JLCPCB coverage implemented in PR #36)
 - [ ] 3.3 Drift-refusal test (edited BOM.md value → non-zero exit, drift hint)
 - [ ] 3.4 Network guard: no api.* connections during export
 
--- a/src/kicad/bom-export.ts
+++ b/src/kicad/bom-export.ts
@@ -16,6 +16,8 @@ export interface NormalizedBomRow {
 
 export interface JlcpcbBomOptions {
   includeUnverified?: boolean;
+  /** Include diagnostic comment lines in the CSV when the target accepts them. */
+  includeWarningsInCsv?: boolean;
 }
 
 export interface BomExportWarning {
@@ -48,18 +50,18 @@ export function emitJlcpcbBom(
     const refdes = row.refdes.trim();
     const mpn = row.mpn?.trim() ?? '';
     const mpnIsUnverifiedMarker = mpn.toUpperCase() === 'UNVERIFIED';
-    if (!mpn) {
-      warnings.push(warning(refdes, 'missing MPN'));
+    if (!mpn || mpnIsUnverifiedMarker) {
+      warnings.push(warning(refdes, mpnIsUnverifiedMarker ? 'invalid MPN (UNVERIFIED marker)' : 'missing MPN'));
       continue;
     }
 
-    const verification = (row.status ?? row.verification ?? (mpnIsUnverifiedMarker ? 'UNVERIFIED' : '')).trim().toUpperCase();
-    if (verification === 'UNVERIFIED' && !options.includeUnverified) {
+    const verification = (row.status || row.verification || '').trim().toUpperCase();
+    if (verification === 'UNVERIFIED') {
       warnings.push(warning(refdes, 'UNVERIFIED (use includeUnverified to include)'));
-      continue;
+      if (!options.includeUnverified) continue;
     }
 
-    const key = JSON.stringify([row.value, row.footprint, row.lcsc?.trim() ?? '']);
+    const key = JSON.stringify([row.value.trim(), row.footprint.trim(), row.lcsc?.trim() ?? '']);
     const group = groups.get(key);
     if (group) group.push(row);
     else groups.set(key, [row]);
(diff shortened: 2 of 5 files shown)


animesh-chouhan (commented):
# Review: PR #36, isolated JLCPCB BOM emitter

## Overview

Adds `src/kicad/bom-export.ts` with a single pure function `emitJlcpcbBom(rows, options)` that takes normalized BOM rows and returns the four-column JLCPCB assembly CSV (`Comment,Designator,Footprint,LCSC Part #`) plus structured exclusion warnings. Rows are grouped by (value, footprint, LCSC), designators are joined per line with natural sort, MPN-less and `UNVERIFIED` rows are excluded with `includeUnverified` as the opt-in, and warnings are appended as `#` comment lines. A JSON fixture, a committed golden CSV, and three tests round it out.

This matches the scope of issue #11 and task 1.3/1.4 (JLCPCB slice) of `openspec/changes/add-supplier-bom-export/tasks.md` well. I fetched the branch, ran `test/jlcpcb-bom.test.ts` (3/3 pass) and `tsc --noEmit` (clean).

## What is done well

- Clean isolation: one pure, deterministic function with no I/O, no LLM, no network. This keeps the repo's "check is LLM-free and network-free" posture trivially safe when CLI wiring lands later.
- Exclusion ordering is correct per the delta spec: the MPN check runs before the `UNVERIFIED` check, so MPN-less rows can never sneak in via `includeUnverified` (`supplier-bom-export/spec.md`, "never MPN-less rows").
- `JSON.stringify([...])` as the group key avoids the classic delimiter-collision bug of `value + '|' + footprint` keys.
- CSV quote escaping (`""` doubling, every data field quoted) is correct and covered by a dedicated test, including a comma inside a footprint field.
- Golden-file byte-equality test is exactly what the issue asked for, and warnings are structured data (`refdes`, `reason`, `message`), which will make the stderr footer in task 1.4 easy.
- Natural refdes sort (`R2` before `R10`) with a deterministic tiebreak is a nice touch JLCPCB reviewers will appreciate.

## Correctness

- **Major, real-world format risk:** warnings are appended to the CSV body as `# Excluded ...` lines (`emitJlcpcbBom`, the `lines.push(...warnings...)` block). CSV has no comment syntax, and JLCPCB's upload parser will most likely read each of these as a one-column data row, at best flagging "unrecognized" rows, at worst polluting the part-matching step. The delta spec says CSV comments "where the format permits", and it is not established that JLCPCB permits them. Suggestion: either verify against a real JLCPCB upload and note the result in the PR, or gate the comment footer behind an option (default off) so the CLI can decide, keeping warnings authoritative in the returned `warnings` array and future stderr output. The issue itself says format details matter more than the TypeScript here.
- **Minor:** the grouping key trims `lcsc` but not `value` or `footprint` (`const key = JSON.stringify([row.value, row.footprint, row.lcsc?.trim() ?? ''])`), and the emitted `Comment`/`Footprint` come from the untrimmed `first.value`/`first.footprint`. Rows differing only by whitespace will not group and will emit padded fields. Trim all three in both the key and the output for consistency.
- **Minor:** `row.status ?? row.verification` uses nullish coalescing, so an empty-string `status: ""` shadows a populated `verification: "UNVERIFIED"` and the row is exported as verified. If the alias is meant as a fallback while the shared parser settles, prefer `(row.status || row.verification || ...)` or normalize both before choosing.
- **Minor:** an MPN that is literally the string `UNVERIFIED` (the `mpnIsUnverifiedMarker` heuristic) is a placeholder, not a real part number, yet with `includeUnverified: true` the row is emitted. That leans against the spec's "never MPN-less rows" intent. Since MPN is not printed in the JLCPCB format the blast radius is small, but I would treat marker-MPN rows as MPN-less (always excluded) rather than as `UNVERIFIED`.
- **Edge case, minor:** duplicate refdes in the input (possible from a buggy upstream parse or a multi-unit symbol appearing twice) is not deduplicated, producing `"C2,C2"` in the Designator column. A `Set` on trimmed refdes before joining would make the emitter robust to that.
- **Nit:** lines are joined with `\n`. RFC 4180 specifies CRLF, and some fab-house parsers are picky. JLCPCB generally accepts LF, but since the golden file locks this in, it is worth confirming once against a real upload.

## Conventions

- File lands at `src/kicad/bom-export.ts`, exactly where task 1.3 says emitters live, and the test plus fixtures follow the existing `test/*.test.ts` and `test/fixtures/<name>/` layout with ESM `.js` import specifiers. Good.
- `tasks.md` in `openspec/changes/add-supplier-bom-export/` is untouched. The JLCPCB slices of 1.3, 1.4, 3.1, and 3.2 are now partially done; per CLAUDE.md the change artifacts should track reality. Since the tasks are only partially complete, a short annotation (for example "JLCPCB done, PR #36") beats leaving them unchecked and silent.
- **Nit:** `emitJlcpcbCsv` is an unused alias export of `emitJlcpcbBom`. Two names for one function invites drift in call sites; drop it until something needs it.
- **Nit:** the `[key: string]: unknown` index signature on `NormalizedBomRow` defeats excess-property checking. Understandable as a stub while the shared parser (task 1.1) is pending, but consider removing it once the real row shape exists so typos in field names fail to compile.

## Performance

- `firstRefdes` copies and sorts the whole group inside the outer sort comparator, and each group's designators are sorted again in the output map, so per-group sorting is done O(g log g) times per comparison. For BOM-sized inputs this is irrelevant in practice; if you want it tidy, precompute each group's sorted designator list once and sort groups by its first element. Nit only.

## Tests

- Good: golden byte-equality, `includeUnverified` behavior with warning assertions, quote and comma escaping, quantity ignored, blank and null LCSC, `status` field, and the marker-MPN case is exercised implicitly via `U1`.
- Missing, all minor: empty input (header-only output), duplicate refdes, the `verification` alias field (the fixture only uses `status`, so the alias path is untested), and whitespace-padded value/footprint grouping. One or two of these would harden the emitter cheaply.

## Security

- No I/O, no subprocess, no network: nothing to flag. CSV formula injection (`=`, `+`, `@` prefixes) is a theoretical concern for spreadsheet-opened exports, but the data is the user's own BOM destined for JLCPCB's uploader, and escaping would corrupt part values, so leaving it alone is the right call.

## Verdict

Solid, well-scoped contribution that does what issue #11 asked, with the right shape for the DigiKey/Mouser emitters to follow. The one thing I would resolve before merge is the `#` comment footer inside the CSV: confirm JLCPCB tolerates it or make it opt-in. The rest is minor polish.

tanav29:
Implemented the review improvements:

  - Warning comments are now opt-in via includeWarningsInCsv; default CSV is parser-safe.
  - UNVERIFIED rows remain in warnings even when included.
  - UNVERIFIED MPN markers are always excluded.
  - Trimmed value, footprint, and LCSC fields consistently.
  - Empty status now falls back to verification.
  - Duplicate/blank designators are removed.
  - Removed unused emitJlcpcbCsv alias.
  - Added coverage for all noted edge cases.
  - Annotated the OpenSpec task checklist.

  Files changed:

  - [bom-export.ts](C:\Users\Tanav Poswal\Dev\copperhead\src\kicad\bom-export.ts)
  - [jlcpcb-bom.test.ts](C:\Users\Tanav Poswal\Dev\copperhead\test\jlcpcb-bom.test.ts)
  - [expected.csv](C:\Users\Tanav Poswal\Dev\copperhead\test\fixtures\jlcpcb\expected.csv)
  - [tasks.md](C:\Users\Tanav Poswal\Dev\copperhead\openspec\changes\add-supplier-bom-export\tasks.md)

  Verification:

  - Focused JLCPCB tests: 7/7 passed.
  - TypeScript typecheck: passed.
  - Full suite: 111 passed, 5 unrelated pre-existing/environment-dependent failures.

animesh-chouhan:
Heads-up: this overlaps with two other open PRs implementing the same `add-supplier-bom-export` change. All three add `src/kicad/bom-export.ts` and will conflict:

- #35 — Mouser emitter, closes #13
- #36 (this one) — JLCPCB emitter, fixes #11
- #45 — full `export bom` command with all three suppliers (jlcpcb/digikey/mouser) + CLI wiring + create-pipeline stage 6, fixes #42

#45 appears to subsume the per-supplier work here. Could you three sync up on whether these should be consolidated (or rebased so they don't collide on the same module) before we merge? Want to avoid duplicate effort and a painful three-way conflict.

PR #37, opened 2026-07-21
feat(provider): add saved-login claude-code provider

Shiveshrane:
> Rebased onto v0.5.0 (async `makeProvider` + `CodexProvider`); merges cleanly. Mirrors `CodexProvider`: reasoning-only, `close()` cleanup, layered tool-disable, distinct name / no fallback. Addresses the automated review (findings 1-5).

## What

Adds a third LLM provider selected by `--model claude-code` (and `claude-code:<model-id>`) that drives Claude Code through the Claude Agent SDK and reuses its saved login, so a Claude subscription user runs copperhead with no `ANTHROPIC_API_KEY`.

## Why

The two existing providers (`gpt-5`, `claude`) each require a per-token API key. Users already logged into Claude Code have no way to point copperhead at that subscription. This closes that gap while keeping authentication entirely with the CLI.

## Design

The Agent SDK wants to run its own autonomous tool loop, which would bypass copperhead's safety orchestration (snapshot, ERC/DRC verify, repair, rollback, commit gate, obligations ledger), all of which live in [loop.ts](src/agent/loop.ts). So the provider is reasoning-only, mapped onto the existing single-turn `Provider.chat(messages, tools)` seam:

- Each `chat()` issues one `query()` in an isolated `cwd`, so the SDK executes nothing. "Executes nothing" is defended in four layers, each caught by the next: `tools: []` (verified against `@anthropic-ai/claude-agent-sdk@0.3.x` as "disable all built-in tools"), a `'*'`-led `disallowedTools`, a deny-all `canUseTool` handler (the Agent-SDK analog to Codex's read-only sandbox / `approvalPolicy:'never'`), and a runtime tripwire that throws if a `tool_use` block ever appears. The provider implements `Provider.close()` to remove its scratch cwd, which `loop.ts` calls in a `finally`.
- SDK option names are compile-checked via a local `QueryOptions` interface (an `import type` from the SDK is not possible: it is an undeclared optional dep, so that would couple `tsc` to its install, exactly the coupling `codex.ts` now carries). Parsed tool-call names are validated against the turn's `availableTools` catalog: an unknown or locked name is left as prose for the loop to nudge.
- copperhead's `availableTools(ctx)` are advertised to the model as a JSON tool protocol, and the reply is parsed back into `Turn.toolCalls`. [loop.ts](src/agent/loop.ts) stays the driver, so every safety gate applies identically to every provider, and spec-gating stays structural (the advertised tool set is recomputed each turn).
- Auth stays external: no key check, the credential is never read or logged. The SDK `env` is passed with `ANTHROPIC_API_KEY` and `OPENAI_API_KEY` stripped, so a saved-login run never silently bills a paid key. Errors re-throw with `status` intact (auth only on 401/403) so retry and backoff still work; missing-dependency and unauthenticated cases become actionable messages. The distinct provider name (`claude-code`) makes `otherProvider()` return null, so there is no fallback to a keyed provider.
- Routing: `makeProvider` matches `claude-code` before the `claude*` prefix.

Full rationale is in [design.md](openspec/changes/add-claude-code-provider/design.md) (decisions D1 to D6).

Packaging note for reviewers: `@anthropic-ai/claude-agent-sdk` is intentionally not declared in `package.json`. It peer-requires `@anthropic-ai/sdk >=0.93.0`, conflicting with the pinned `^0.39.0`, and a peer conflict fails `npm install` even for an `optionalDependency`. It is a documented manual install; the provider lazy-imports it and errors actionably when absent. Bumping the core SDK to satisfy the peer is a deferred alternative (happy to do that if you prefer it bundled).

## Spec / OpenSpec

- New OpenSpec change [add-claude-code-provider](openspec/changes/add-claude-code-provider/) (proposal, design, delta specs for `claude-code-provider` and `cli-surface`, tasks).
- [SPEC.md](openspec/specs/SPEC.md): AC-3.10 extended to include `--model claude-code`; new AC-3.11 (saved login: runs with no API key, credential untouched, actionable errors on missing-dep or unauthenticated).

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | 168 pass, 14 skip, 0 fail |
| Live-LLM (opt-in) | keyed on `CLAUDE_CODE_OAUTH_TOKEN` (+ SDK present) | skip (no token in CI) |

New tests: `test/claude-code-provider.test.ts` (22 cases, 1 self-skips when the SDK is installed) covers routing precedence (async `makeProvider`), tool-protocol advertising, bare and multiple fenced JSON parsing, malformed-JSON tolerance, `renderConversation` flattening plus `[result of <name>]` mapping, the `tool_use` tripwire, the deny-all `canUseTool`, the empty `tools` + wildcard `disallowedTools`, catalog rejection of an unknown tool name, `close()` removing the cwd, error-status preservation (including a 429 whose message mentions "oauth token" staying retryable), the missing-dependency error, no-fallback, env key-stripping, and cwd reuse. `AC-3.10` in `test/agent-integration.test.ts` now parameterizes `codex`, `gpt-5`, `claude`, and `claude-code`; the `claude-code` entry is gated on both the token and the SDK being resolvable.

Sandbox variant used (`create` / `edit`): n/a for the full CLI loop. This change was exercised with a direct provider smoke test against the real SDK and a real saved login (no `ANTHROPIC_API_KEY`), which is the runtime path this PR adds. A full `create`/`do` loop against the manual-tests sandbox is pending (needs `kicad-cli` plus a longer run) and can be added before merge if you want it.

Commands run and outcome:

```text
$ npm run build && CLAUDE_CODE_OAUTH_TOKEN=<saved-login> node cc-smoke.mjs
calling claude-code provider (real SDK)...
--- Turn ---
text: null
toolCalls: [ { "id": "cc-1", "name": "read_file", "args": { "path": "docs/BOM.md" } } ]
usage: {"inputTokens":438,"outputTokens":36}
OK: the saved-login provider round-tripped a turn.
```

This confirms end to end: authentication via the saved login with no API key, one `query()` round-trip, the JSON tool protocol parsed back into a `Turn.toolCalls`, and real token usage.

## Docs

[README.md](README.md) (saved-login section), the configuration reference [configuration.md](docs/src/content/docs/reference/configuration.md), [.env.example](.env.example) (`CLAUDE_CODE_OAUTH_TOKEN`), and the `resolveModel` doc comment in [config.ts](src/config.ts).

---

## Invariant checklist

- [x] **Spec-gated in**: `edit_file`/`write_file` stay structurally absent from the agent tool list until an OpenSpec proposal validates (gated by omission, not prompt text). Preserved: the provider advertises exactly `availableTools(ctx)` each turn and `dispatchTool` re-checks, so a locked tool is neither offered nor executable.
- [x] **Verification-gated out**: mutations still end in ERC (and DRC when the board changed) passing, with repair up to `maxRepairCycles` then rollback to the git snapshot. Preserved: `loop.ts` stays the driver for every provider, so `claude-code` inherits the same gates.
- [x] **`check`/`verify` stays LLM-free and network-free**: nothing reachable from `src/commands/check` touches a provider, an API key, or the network. The new provider is only reachable via `makeProvider` in the loop, not from `check`.
- [ ] **No sexp serialization**: N/A. This PR does not touch `src/kicad/`; the parser stays read-only.
- [ ] **Sync-obligations ledger**: N/A. This PR does not change the ledger, its hooks, or the commit gate.
- [x] **Secrets**: transcripts and summaries redact `sk-[A-Za-z0-9_-]+` at write time; keys live only in env vars. The provider reads no credential, and `.gitignore` is unchanged. Claude Code OAuth tokens (`sk-ant-oat01-...`) already match the redaction pattern.
- [ ] **Spec coherence**: SPEC.md and the change artifacts moved together in this PR. `openspec validate add-claude-code-provider` was not run locally (the `openspec` CLI is not installed here); please confirm it passes in CI.


Shiveshrane:

animesh-chouhan (commented):
# Review: feat(provider): add saved-login claude-code provider

## Overview

This PR adds a third provider, selected by `--model claude-code` / `claude-code:<id>`, that drives a locally-installed Claude Code through `@anthropic-ai/claude-agent-sdk` and reuses its saved login, so a subscription user runs copperhead with no `ANTHROPIC_API_KEY`. The core design move is sound: rather than letting the Agent SDK run its own agentic loop (which would bypass the snapshot / ERC-DRC verify / repair / rollback / commit-gate orchestration in `src/agent/loop.ts`), the provider is reasoning-only. Each `chat()` issues one `query()` with SDK tools disabled in an isolated `mkdtemp` cwd, advertises copperhead's `availableTools(ctx)` as a JSON text protocol, and parses the reply back into `Turn.toolCalls` so `loop.ts` stays the single driver. Routing, docs, `.env.example`, OpenSpec artifacts (proposal, design D1-D6, two delta specs, tasks), SPEC.md AC-3.10/AC-3.11, and an offline test suite are all included.

## What is done well

- The reasoning-only mapping (design D1) is the right call: forking the safety path per provider would have been unacceptable in a tool whose value proposition is verified, reversible edits. Keeping `loop.ts` as the driver means every gate applies identically to `claude-code`.
- The spec-gated-in invariant genuinely stays structural: the advertised tool set is `availableTools(ctx)` recomputed each turn, and I confirmed `dispatchTool` (`src/agent/tools.ts:519`) re-checks against `availableTools(ctx)` and refuses anything not in the set, so even a hallucinated `edit_file` JSON call is rejected. The delta spec scenario "Locked edit tools are not advertised" matches the code.
- No-fallback guarantee (D4) is correct and cheap: the distinct `name: 'claude-code'` makes `otherProvider()` in `src/agent/loop.ts` return `null`, so a rate-limited subscription run never silently continues on a billed API. The added comments in `makeProvider` and `otherProvider` document the ordering trap and the guarantee.
- Error handling is careful: the original SDK error is re-thrown so `status` survives for `withRetry`/`isRateLimit` (verified against `src/util/retry.ts`, which keys retryability on `status === 429`), while missing-dependency and auth failures become actionable messages.
- Credential hygiene is good: `src/agent/providers/claude-code.ts` never reads `CLAUDE_CODE_OAUTH_TOKEN` or any env var, the constructor has no key check (unlike the keyed providers, deliberately), and error messages never embed the token. As defense in depth, Claude Code OAuth tokens (`sk-ant-oat01-...`) already match the `sk-[A-Za-z0-9_-]+` write-time redaction pattern in `src/util/redact.ts`, so AC-4.1 holds even if one leaked into a transcript.
- The offline test suite (`test/claude-code-provider.test.ts`) is thorough for a fake-injected seam: routing precedence, protocol advertising, fenced and bare JSON parsing, malformed-JSON tolerance, 429 re-throw identity (`rejects.toBe(rateLimit)` is a nice touch), auth-error mapping, and a `skipIf` missing-dep case.
- OpenSpec discipline is exemplary: proposal, design with rejected alternatives, delta specs whose scenarios map onto the new AC-3.11, and a checked-off tasks.md.

## Correctness and safety concerns

### Major: "the SDK executes nothing" rests on an unverified `tools: []` option key

In `chat()` (`src/agent/providers/claude-code.ts`), the options object is a `Record<string, unknown>`, and the SDK is imported untyped via a non-literal specifier, so nothing type-checks the option names against the real SDK `Options`. If the installed SDK version does not recognize a top-level `tools` key (its documented surface historically uses `allowedTools` / `disallowedTools`), the empty allowlist is silently ignored and the only protection is `DISALLOWED_BUILTINS`, a hand-maintained denylist that already omits several built-ins (`BashOutput`, `KillShell`, `ExitPlanMode`, `SlashCommand`, `Skill`, `ListMcpResources`, `ReadMcpResource`) and will silently miss any tool added in a future SDK release. Since this is the load-bearing claim behind D1/D5 and AC-3.11:

- Verify against the pinned SDK version that `tools: []` is an honored option (the PR says a live round-trip was done, but a correct text reply does not prove the allowlist was applied).
- Add a runtime assertion in `chat()`: if any streamed message contains a `tool_use` block (or a `type: 'tool_use'`-bearing assistant message), throw rather than continue. That converts a silent safety regression in a future SDK into a loud failure.
- Consider also passing `permissionMode` or the SDK `env` option to further pin down the subprocess, and keep the denylist as belt-and-suspenders as intended.

### Minor: `isAuthError` message regex can eat a retryable error's status

`isAuthError` returns true for any error whose message matches `/unauthenticat|unauthoriz|not logged in|please log in|no.*credential|oauth token|setup-token/`, regardless of status. A 429 whose message happens to mention "oauth token" (plausible for subscription rate-limit messages) would be wrapped in a fresh `Error` with no `status`, defeating `withRetry` and contradicting the delta spec's "preserve the SDK error's status" requirement. Suggest: treat a present `status` as authoritative (401/403 auth, anything else re-throw untouched) and only apply the message heuristic when no status exists. The `no.*credential` fragment is also broad enough to match unrelated messages ("no such file: credentials.json").

### Minor: `ANTHROPIC_API_KEY` present alongside `claude-code` may bill the API key

D2 notes the SDK resolves credentials by its own precedence, including `ANTHROPIC_API_KEY` if set. So the headline "runs with no API key" is accurate, but the converse case (user has both a key and a token) may silently bill the API key rather than the subscription, which cuts against the spirit of "never falls back to a billed API". Worth either a one-line doc warning, or passing the SDK's `env` option with `ANTHROPIC_API_KEY` stripped for this provider.

### Minor: `ChatOpts.maxTokens` is silently dropped

`chat(messages, tools, _opts)` ignores `ChatOpts`. Today `loop.ts:299` calls `provider.chat(messages, tools)` with no opts, so nothing breaks, but the interface contract is quietly narrower for this provider. A comment saying maxTokens has no SDK equivalent (or mapping it if one exists) would prevent surprise if the loop starts passing opts.

## Performance

- Each `chat()` spawns a fresh Agent SDK `query()` (a full Claude Code CLI subprocess) plus a new `mkdtemp` directory. On a long `do` run this is one subprocess startup per turn and one leaked temp dir per turn (never removed). The design risk section acknowledges the per-turn session cost; please also clean up the temp dir in a `finally`, or create one directory per provider instance and reuse it.
- Full-history re-send per turn matches the other providers, fine.

## Tests

- Good offline coverage as listed above. Gaps worth closing cheaply:
  - `renderConversation` is only asserted to contain the user text. Add a test with an assistant tool call plus a `tool` result message and assert the `[assistant tool call]` JSON re-rendering and `[result of <name>]` mapping, since that flattening is the provider's memory.
  - No test for two fenced JSON blocks in one reply (the parser will emit two tool calls even though the protocol says "EXACTLY ONE"); pin the intended behavior either way.
  - The integration entry in `test/agent-integration.test.ts` gates only on `CLAUDE_CODE_OAUTH_TOKEN`; if the token is set but the optional SDK is not installed, the live test will fail rather than skip. Consider gating on both.
- Per the packaging choice (D3), the missing-dep test self-skips when the SDK is installed, which is the right shape.

## Conventions and docs

- Routing precedence (`claude-code` before the `claude*` prefix) is implemented, commented, tested, and documented consistently across `makeProvider`, `src/config.ts`, `.env.example`, README, and the configuration reference. SPEC.md and the delta specs were updated together, as CLAUDE.md requires.
- `check` remains untouched (still LLM-free and network-free), and `.gitignore`/transcript redaction obligations are unaffected.
- The undeclared-dependency decision (D3) is well argued and honestly surfaced in the PR body. The residual risk is that the local `QueryArgs`/`QueryMessage` structural types can drift from the real SDK with no compile-time signal; the opt-in live parity test is the only tripwire, so keep it healthy. Bumping `@anthropic-ai/sdk` to satisfy the peer and declaring the SDK as a real `optionalDependency` remains the better end state; fine to defer as stated.
- Nit: the new prose in README, `.env.example`, and `docs/src/content/docs/reference/configuration.md` leans heavily on em dashes; this repo's doc style prefers colons or commas in user-facing prose.
- Nit: tasks.md 2.5 says only `MODULE_NOT_FOUND` is wrapped, but `resolveQuery` wraps every import failure (including a present-but-broken SDK) as "install the optional dependency", which can mislead. Checking `err.code === 'ERR_MODULE_NOT_FOUND' || err.code === 'MODULE_NOT_FOUND'` and re-throwing otherwise would match the task and the spec scenario.

## Verdict

Solid, well-specified contribution that respects both structural invariants and handles credentials correctly. I would like the major item addressed before merge: verify the `tools: []` option is honored by the pinned SDK and add the no-tool-use runtime assertion, since the entire safety story for this provider hangs on the SDK executing nothing.

animesh-chouhan:
🤖 **Automated pr-review pass** (Claude Code `/pr-review`, not a human review). One reviewer's read; treat as input, not a merge gate.

# Review — PR #37 `feat(provider): add saved-login claude-code provider`

**Verdict: Request changes** — not for a correctness bug (the code is clean, well-tested, and the author has real-SDK evidence it works), but because it **no longer merges** and it **predates the `CodexProvider` that main has since merged to solve this exact problem**. Rebasing onto that pattern resolves most of the below.

## Does Claude actually support this pattern? (your explicit question)

**Yes.** Three independent confirmations:

1. **Saved-login auth is a real SDK path.** `CLAUDE_CODE_OAUTH_TOKEN` minted by `claude setup-token` is the supported way to run the Agent SDK against a Claude subscription with no `ANTHROPIC_API_KEY`. (A docs-only sub-agent pass didn't find the token in the pages it read, but it is a real, documented lever.)
2. **The reasoning-only mapping already exists in this repo.** Main merged `src/agent/providers/codex.ts` (`CodexProvider`) which does the identical thing for the Codex/ChatGPT saved login: a read-only backend, copperhead stays the sole dispatcher, optional SDK, distinct provider name, no fallback. This PR is the Claude analog of an already-accepted design.
3. **The author's own smoke test proves it end-to-end.** The PR description shows a real `query()` against the real SDK + a real saved login (no `ANTHROPIC_API_KEY`) round-tripping a `read_file` tool call with real usage `{in:438, out:36}`. That directly exercises auth, the JSON tool protocol, and usage extraction.

The **one** thing not confirmable from docs is that `tools: []` is the "disable all built-in tools" lever (docs point to `disallowedTools` / permissions, with `disallowedTools: ['*']` as the wildcard). The PR's `tool_use` tripwire makes this **fail safe** — worst case is a loud run-ending throw, never a silent unverified edit — so the safety invariant holds even if that option is a no-op. See Finding 3.

## Findings (most severe first)

**1. [Blocker for merge] Does not merge with current `main`; must rebase.** `main` has since made `makeProvider` **async** (`Promise<Provider>`) and added `CodexProvider`. A trial merge conflicts in `src/agent/loop.ts` (import + the whole `makeProvider` body), `SPEC.md`, `README.md`, `.env.example`, and `configuration.md`. The PR's synchronous `return new ClaudeCodeProvider(...)` branch and its SPEC §4.4 / model-routing edits need reworking against the async, codex-aware versions (e.g. §4.4 and the routing tables now need to list `codex` too).

**2. [Medium] Diverges from the merged `CodexProvider`, which solves the same problem more robustly.** Rebase should mirror it:
   - **Untyped SDK options.** `CodexProvider` does `import type { ThreadOptions, Usage } from '@openai/codex-sdk'` — erased at build, so the optional dep can be absent, **but option names are compile-checked**. This PR deliberately types options as `Record<string, unknown>` (`claude-code.ts:492`), so the load-bearing `tools: []` name is unchecked — exactly the thing at risk in Finding 3. Using `import type` from `@anthropic-ai/claude-agent-sdk` would catch it at build.
   - **Freeform tool protocol.** The PR asks the model to emit a fenced ```json block and regex-parses it (`parseToolCalls`). `CodexProvider` uses the SDK's structured `outputSchema`. The prose-JSON path is strictly less reliable (the design's own D1 trade-off); check whether the Agent SDK exposes a structured-output equivalent.
   - **Layered sandbox vs. single lever.** `CodexProvider` sets `sandboxMode:'read-only'`, `approvalPolicy:'never'`, `networkAccessEnabled:false`. This PR leans on `tools: []` + a denylist + tripwire. Check for an Agent-SDK `permissionMode` / `canUseTool` analog.

**3. [Medium — verify] `tools: []` as the primary "disable all tools" mechanism is unconfirmed.** `src/agent/providers/claude-code.ts:557`. Documented lever is `disallowedTools` (wildcard `['*']`); `DISALLOWED_BUILTINS` is a hand-maintained enumerated list that can't cover tools added in a future SDK version. **Mitigated** by the enumerated denylist and the `tool_use` tripwire (`claude-code.ts:572`), which fails safe. Action: confirm `tools: []` against the pinned SDK version, and consider adding `disallowedTools: ['*']`.

**4. [Low] Temp-dir leak.** `ensureCwd()` (`claude-code.ts:605`) creates one `mkdtemp` dir per provider instance and never removes it; the `Provider` interface exposes no `close()` and `loop.ts` wouldn't call it (it only calls `memory?.close()`). `CodexProvider` cleans its dir (`rm`, `ownsWorkingDirectory`). Result: one empty temp dir leaks per run. The code comment only claims to avoid a *per-turn* leak, so this is arguably by design, but it's a gap vs. codex.

**5. [Nit] "locally-installed Claude Code" is imprecise.** The SDK bundles its own native binary; what's reused is the saved login / OAuth token, not necessarily an installed `claude` CLI. `README.md` / `configuration.md` wording.

## Invariant checks (performed)

| Invariant | Result |
| --- | --- |
| **Spec-gated-in** (edit tools absent until proposal validates) | ✅ Preserved — advertises exactly `availableTools(ctx)` each turn; `dispatchTool` re-checks. Structural, not prompt-text. |
| **Verification-gated-out** (ERC/DRC + repair→rollback) | ✅ `loop.ts` stays the driver for all providers; `claude-code` inherits every gate. |
| **`check` LLM-free / network-free** | ✅ New provider reachable only via `makeProvider` in the loop, never from `src/commands/check`. |
| **No sexp serialization** | ✅ N/A — PR doesn't touch `src/kicad/`. |
| **Sync-obligations ledger** | ✅ N/A — untouched. |
| **Secrets redaction / env-only keys / `.gitignore`** | ✅ Provider reads no credential; strips `ANTHROPIC_API_KEY`/`OPENAI_API_KEY` from the SDK env; OAuth `sk-ant-oat01-…` matches the existing `sk-[A-Za-z0-9_-]+` redaction; `.gitignore` unchanged. |
| **No silent fallback to a paid API** | ✅ Distinct `name:'claude-code'` → `otherProvider()` returns `null`; verified in code and test. |
| **Spec coherence** | ✅ `openspec validate add-claude-code-provider` → **valid**. SPEC AC-3.10/3.11 + §4.4 + delta specs moved together. Caveat: §4.4/routing docs omit `codex` (branched pre-codex) — surfaces at rebase (Finding 1). |

## What I ran

- `openspec validate add-claude-code-provider` → **valid**
- `npm run typecheck` / `npm run build` (PR branch) → **pass**
- `npx vitest run test/claude-code-provider.test.ts` → **18/18 pass** (missing-dep path exercised; live integration self-skips — SDK not installed here)
- Trial merge vs `origin/main` → **conflicts** (Finding 1)

Note: budget refusal (AC-3.4) is **turn-count based** (`budget = maxTurns`), not token based — so even if usage extraction read the wrong field, AC-3.4 is unaffected; token counts feed transcript reporting only. The author's smoke test returned real non-zero usage, so the shape looks correct against the pinned SDK anyway.

Shiveshrane:
Thanks for the thorough review. Addressed in the force-pushed `1cfd822` (rebased onto v0.5.0, single commit, merges cleanly). Full offline suite is green: 168 pass / 14 skip / 0 fail; typecheck and build pass.

**1 (blocker: does not merge).** Rebased onto v0.5.0. The `claude-code` branch is integrated into the now-async `makeProvider` (matched before the `claude*` prefix, alongside `codex`, with an empty-override guard). SPEC §4.4, AC-3.10, the routing tables, `.env.example`, and the config reference now list `codex` and `claude-code` together. AC-3.10 parity is parameterized over `codex`, `gpt-5`, `claude`, and `claude-code`.

**2 (diverge from CodexProvider).**
- Untyped options: replaced `Record<string, unknown>` with a local `QueryOptions` interface, so the option names (`tools`, `disallowedTools`, ...) are compile-checked. I did not `import type` from the SDK on purpose: the SDK is an undeclared optional dep (peer conflict, see below), so importing its types would couple `tsc` to its install, which is the coupling `codex.ts` now carries (it is why v0.5.0 added `@openai/codex-sdk` to `devDependencies`). The local interface gives the compile-check without that cost, and keeps CI green with no `package.json` change.
- Sandbox vs single lever: added a deny-all `canUseTool` handler (`interrupt: true`), the Agent-SDK permission analog to Codex's `sandboxMode: 'read-only'` / `approvalPolicy: 'never'`, plus a `'*'` wildcard in `disallowedTools`. "The SDK executes nothing" is now four layers, each caught by the next: `tools: []`, wildcard deny list, `canUseTool` deny, and the runtime `tool_use` tripwire.
- Freeform protocol vs structured output: the Agent SDK's `query()` exposes no single-shot `outputSchema` equivalent to the Codex SDK's `TurnOptions.outputSchema` (it drives an agent loop, not a structured completion). I hardened the freeform path instead: parsed tool-call names are validated against the turn's `availableTools` catalog, so an unknown or locked name is left as prose for the loop to nudge rather than dispatched. Happy to add a validate-and-correction-retry turn if you want it closer to codex.

**3 (verify `tools: []`).** Confirmed against the pinned `@anthropic-ai/claude-agent-sdk@0.3.216` type declarations: `Options.tools` documents `[]` as "Disable all built-in tools". Added the `'*'` wildcard denylist as belt-and-suspenders; the tripwire remains the fail-safe if a future SDK ever ignores this.

**4 (temp-dir leak).** Implemented `Provider.close()` to `rm` the scratch cwd; `loop.ts` already calls `provider.close?.()` on every provider in a `finally`, so the dir does not outlive the run (mirrors `CodexProvider`'s `ownsWorkingDirectory` cleanup). Covered by a test.

**5 (wording nit).** README, config reference, and `.env.example` now describe reusing the saved login rather than a "locally-installed Claude Code" binary.

New offline tests cover `close()`, the deny-all `canUseTool`, the empty `tools` plus wildcard denylist, catalog rejection of an unknown tool name, and the async `makeProvider` routing. OpenSpec design/spec/tasks were updated to match (D1 four-layer defense, D5 cwd cleanup).

Note on packaging (unchanged, and deferred as you flagged): the SDK stays a documented manual install because its peer on `@anthropic-ai/sdk >=0.93.0` conflicts with the pinned `^0.39.0`, and a peer conflict fails `npm install` even for an `optionalDependency`. Bumping the core SDK to bundle it is the better end state; glad to do that as a follow-up if preferred.

animesh-chouhan:
Does Claude actually support this pattern?

Can you provide the run logs to prove that this pattern works?

Shiveshrane:
**Does Claude actually support this pattern? Yes, verified live against the real Claude Code SDK.** I ran the provider end to end over a saved login (no `ANTHROPIC_API_KEY`); the full input/output log is below.

What it proves, line by line:

- **Saved-login auth works.** `ANTHROPIC_API_KEY present: false`, `CLAUDE_CODE_OAUTH_TOKEN present: true`. The run authenticated on the subscription token alone, which is the whole point of the provider.
- **Reasoning-only holds on real hardware.** `SDK emitted a tool_use block: false`. copperhead advertises its tools as a text protocol in the system prompt; Claude replies with a plain fenced-JSON block (`{"tool":"read_file","args":{"path":"docs/BOM.md"}}`), not a native `tool_use`, so the SDK executes nothing and `loop.ts` stays the sole dispatcher.
- **The round-trip maps cleanly.** copperhead parsed that raw text into `Turn.toolCalls = [{name:"read_file", args:{path:"docs/BOM.md"}}]` with real `usage {in:450, out:56}`.

So the three claims the design rests on (saved-login auth, SDK-executes-nothing, text-protocol tool-calling) are confirmed against the real backend, not just mocked. This complements the offline suite (21/22, the parsing/gating/tripwire logic) and the SDK's own type docs, where `tools: []` is documented as "Disable all built-in tools".

Scope, to be precise: this is a provider-level single-turn proof (auth + reasoning-only + one tool round-trip). It is not a full `do`/`create` run (that needs `kicad-cli` and a longer loop); it demonstrates the provider seam and the invariant, not an end-to-end board edit.

```text
ANTHROPIC_API_KEY present: false  (pattern requires NONE)
CLAUDE_CODE_OAUTH_TOKEN present: true  (saved login)

================= INPUT: system prompt sent to the SDK =================
You are the reasoning backend inside copperhead. Use the tool protocol exactly.

# Tool protocol

You are the reasoning half of a tool-driven workflow; you cannot run anything yourself.
To take an action, reply with EXACTLY ONE JSON object and nothing else, wrapped in a
```json fenced code block:

{"tool": "<tool_name>", "args": { ... }}

Use only the tools listed below, with `args` matching the tool's JSON Schema. If you have
no tool to call and only want to say something, reply with plain prose and no JSON block.

## Available tools

### read_file
Read a file from the repo
Parameters (JSON Schema): {"type":"object","properties":{"path":{"type":"string"}},"required":["path"]}

================= INPUT: user prompt (flattened history) ==============
[user]
Read the bill of materials. Call read_file on "docs/BOM.md".

================= OUTPUT: raw assistant text from Claude ==============
{"tool": "read_file", "args": {"path": "docs/BOM.md"}}

================= copperhead parsed this into a Turn ==================
text: null
toolCalls: [
  { "id": "cc-1", "name": "read_file", "args": { "path": "docs/BOM.md" } }
]
usage: {"inputTokens":450,"outputTokens":56}

SDK emitted a tool_use block: false   (reasoning-only held; SDK executed nothing)
```

animesh-chouhan:
🤖 **Automated pr-review pass** (via the repo's `/pr-review` skill — not a human review, and not a merge-gating approval).

## Verdict: Approve with nits

The only thing blocking a clean merge is a rebase conflict with `main`. The code itself is solid, well-tested, and preserves every repo invariant.

Verified locally on the PR branch:
- `npm run typecheck` — pass
- full `vitest run` — **169 passed, 13 skipped, 0 failed**
- `test/claude-code-provider.test.ts` — 22/22 pass
- `openspec validate add-claude-code-provider` — **valid** (the PR body noted this couldn't run in the author's env; it passes here)

## Findings (ranked)

**1. (Merge-blocker, mechanical) Branch conflicts with `main`.** `main` has advanced past the stated v0.5.0 rebase (the BOM-export work merged), so `README.md` and `src/config.ts` now have adjacency conflicts — both branches inserted a new section/table after the same anchor. Merging as-is is impossible; a rebase is needed. Both conflicts are textual (added-vs-added), not semantic.

**2. (Low, residual risk — acknowledged) The safety mapping rests on SDK runtime behavior CI never exercises.** Every test injects a fake `queryFn`; the real `@anthropic-ai/claude-agent-sdk` is undeclared and absent, so the assumptions that enforce "the SDK executes nothing" — `tools: []` meaning *disable all*, a `tool_use` block arriving as `type:'assistant'`/`content[].type==='tool_use'`, usage at `result.usage.{input,output}_tokens`, and `env` replacing the subprocess environment — are validated only by the author's one manual happy-path smoke test. The adversarial paths (does `tools:[]` truly prevent execution? does the tripwire actually fire?) have no live coverage. Inherent to shipping against an undeclared optional dep and well-mitigated by the four-layer defense + tripwire; worth a tracking note for the eventual SDK bump, not a blocker.

**3. (Nit) `isAuthError` message heuristic can over-classify.** `src/agent/providers/claude-code.ts:829-840`: with no status present, the regex matches `oauth token`, so a status-less transient error mentioning that string would be treated as non-retryable auth. Status is correctly authoritative (a 429 stays retryable, with a test for it), so this only bites the narrow status-less case. Fine to leave.

## Invariant checks (performed and passed)

- **Spec-gated in ✅** — provider advertises exactly `availableTools(ctx)` each turn; `dispatchTool` re-checks; `parseToolCalls` drops any tool name not in the turn catalog. Structural, not prompt-text.
- **Verification-gated out ✅** — `loop.ts` stays the sole driver; reasoning-only provider adds no path around snapshot/ERC/DRC/repair/rollback. `close()` wired into the `finally` at `loop.ts:157`.
- **`check` LLM-free/network-free ✅** — `src/commands/check.ts` imports no provider or loop; new provider reachable only via `makeProvider`.
- **No sexp serialization ✅** — `src/kicad/` untouched.
- **Sync-obligations ledger ✅** — untouched.
- **Secrets ✅** — provider reads/logs no credential; billed keys stripped from the SDK `env`; the OAuth token (`sk-ant-oat01-…`) matches the existing `sk-[A-Za-z0-9_-]+` redaction pattern; `.gitignore` unchanged.
- **No silent fallback ✅** — distinct `name: 'claude-code'` makes `otherProvider()` return `null`; a rate-limited run retries then fails rather than switching to a billed provider.
- **Spec coherence ✅** — `SPEC.md` (AC-3.10 extended, AC-3.11 added, §4.4 note) moves with the new delta specs; `openspec validate` passes.

animesh-chouhan:
**Correction to finding #1 (from local verification):** a real trial merge into `origin/main` shows only **`README.md`** actually conflicts — `src/config.ts` auto-merges cleanly (`git merge` reports `Auto-merging src/config.ts` with no conflict). The rebase is still required, but its scope is just the README section. Also confirmed `npm run build` passes (exit 0) in addition to typecheck; all other verified results stand (169 pass / 13 skip / 0 fail, 22/22 provider tests, `openspec validate` valid).

animesh-chouhan:
Pushed a follow-up commit (`daf36f2`) directly to this branch (maintainer edit), since #56 just merged.

**Context:** #56 landed on `main` (bumps `@anthropic-ai/sdk` to `^0.113.0` and adds a `zod ^4` override), which satisfies `@anthropic-ai/claude-agent-sdk`'s peers. You'd already rebased this PR onto that `main`, so `@anthropic-ai/claude-agent-sdk` is now installed via #56's `optionalDependencies` declaration, but the docs here still described the old undeclared / manual-install / `--legacy-peer-deps` reality.

**The commit corrects that narrative** in README, the configuration reference, `.env.example`, SPEC.md §4.4, and the OpenSpec artifacts (proposal, design D3, tasks 1.1): they now describe the SDK as a declared optional dependency that a normal `npm install` includes. The missing-dependency behavior and its spec scenario are unchanged (the provider still lazy-imports and errors actionably under `--omit=optional`).

Verified on the rebased branch: `openspec validate add-claude-code-provider` valid, `npm run typecheck` clean, full suite **221 pass / 14 skip / 0 fail** (the `missing optional dependency` provider test now self-skips because the SDK resolves). Docs-only change, no source touched. Shout if you'd rather have squashed it into your commit.

Shiveshrane:
Thanks for the follow-up and for verifying it. Keeping it as a separate commit is totally fine with me.

animesh-chouhan:
🤖 **Automated pr-review pass** (2nd, via the repo's `/pr-review` skill — not a human review, not a merge-gating approval). Head `af44dee`.

## Verdict: Approve

The provider is unchanged from the first deep-review pass and remains solid. Since then the branch rebased onto `main` (picking up #56), the packaging docs were corrected, and the live-coverage gap flagged last time is now empirically closed. One low, non-blocking observation.

**Disclosure:** two of the three commits under review were authored this session by the maintainer's assistant (`daf36f2` packaging-doc fix, `af44dee` doc/help discoverability); treat those as self-review. The provider code and tests are the contributor's.

**Verified locally on `af44dee`:**
- `npm run typecheck` — pass
- `openspec validate add-claude-code-provider` — valid; `build-copperhead-phase-1` — valid
- full `vitest run` — **221 pass / 14 skip / 0 fail**
- **Live** saved-login smoke (real SDK, logged-in Claude Code, no `ANTHROPIC_API_KEY`): a reasoning turn returned text + real token usage, and a tool-protocol turn round-tripped `{"tool":"read_file",...}` into `Turn.toolCalls`. This exercises the exact path that had no coverage at the first review, closing that gap.

## Findings

**1. (Low, non-blocking) The missing-optional-dependency error path lost its CI coverage.** `test/claude-code-provider.test.ts:307` guards the "actionable install message" test with `it.skipIf(sdkInstalled)`. Now that #56 declared `@anthropic-ai/claude-agent-sdk` under `optionalDependencies`, it installs in every normal/CI install, so `sdkInstalled` is always true and that test always skips — `resolveQuery`'s `ERR_MODULE_NOT_FOUND` branch (`src/agent/providers/claude-code.ts:212`) is no longer exercised. Failure scenario: a regression in that branch (e.g. a wrong error-code match) ships undetected unless someone runs with `--omit=optional`. Behavior and its spec scenario remain correct; this is a coverage note. If you want it covered, a small refactor to inject an import that throws `{code:'ERR_MODULE_NOT_FOUND'}` would restore it.

## Invariant checks (performed, all pass)

- **Spec-gated in** — advertises exactly `availableTools(ctx)` per turn; `parseToolCalls` drops names not in the turn catalog; `dispatchTool` re-checks. Structural, not prompt-text.
- **Verification-gated out** — `loop.ts` stays the driver; reasoning-only mapping adds no path around snapshot/ERC/DRC/repair/rollback; `close()` wired into the `finally`. The four-layer "SDK executes nothing" defense + `tool_use` tripwire were confirmed inert in the live run (no tool executed).
- **`check` LLM-free/network-free** — `src/commands/check.ts` keeps its "must never import a provider" guard, has 0 provider/loop imports, never references `claude-code`.
- **No sexp serialization** — `src/kicad/` untouched.
- **Sync-obligations ledger** — untouched.
- **Secrets** — provider reads/logs no credential (the `CLAUDE_CODE_OAUTH_TOKEN` mentions are comments and the auth-hint string); billed keys stripped from the SDK env; OAuth token matches the `sk-` redaction pattern; `.gitignore` still lists `.env` and `.copperhead/runs/`.
- **No silent fallback** — distinct `name: 'claude-code'` makes `otherProvider()` return null; rate-limit stays retryable via preserved status.
- **Spec coherence** — SPEC.md (AC-3.10/3.11, §4.4) moves with the delta specs; both openspec changes validate; user-facing docs now consistently describe the declared optional dependency (no stale manual-install language).

animesh-chouhan:
🤖 **Automated pr-review pass** (3rd, via the repo's `/pr-review` skill — not a human review, and not a merge-gating approval). Head `6142ae9`.

**⚠️ Self-review disclosure:** the reviewer identity (`animesh-chouhan`) authored 3 of the 4 commits under review, including the only new commit since the last pass (`6142ae9`). This is self-review; I traced the new branch logic by hand and tested the PR head on a clean tree rather than taking it on trust.

## Verdict: Approve

Incremental over the earlier **Approve** (`af44dee`). One commit landed since — `6142ae9 test(mcp): cover the missing-SDK error path deterministically` — and it is a strict improvement: it closes the live-coverage gap flagged in prior passes and adds nothing risky. The PR is now `MERGEABLE`, `openspec validate add-claude-code-provider` passes, and the full suite is green at the PR head.

### What changed since the last pass (`af44dee` → `6142ae9`)
- Adds an injectable `ImportLike` importer as a 3rd constructor param, defaulting to `(specifier) => import(specifier)` — the specifier stays non-literal, so **invariant D3 (tsc never resolves the optional dep at build time) is preserved** and build/typecheck pass.
- Replaces the inline dynamic import in `resolveQuery()` with `this.importSdk(...)`; error-mapping logic is byte-for-byte identical.
- Adds 3 deterministic tests for the three previously-uncovered branches: `ERR_MODULE_NOT_FOUND` → actionable non-retryable install message; a non-module error (`ERR_DLOPEN_FAILED`) re-thrown unchanged (`.toBe(broken)` identity check); SDK-present-but-no-`query`-export → "did not export `query`". The old integration test now self-skips instead of being the only coverage.

### Metrics block
```
scope (delta since last pass): af44dee..6142ae9 = 2 files, src +13/-4, test +30/-0
full PR vs merge-base 40e76a4: +1089 / -29 across 18 files
  · src +395/-8 (claude-code.ts +367, loop.ts +13, config.ts +13/-6, cli.ts +2/-2)
  · test +361  (claude-code-provider.test.ts +343, agent-integration.test.ts +18)
  · docs +77/-18 · openspec+spec +254/-3
new vs net: 6 added files; src net-new surface = claude-code.ts (367 lines, the hotspot)
tests: +361 test lines vs +395 src lines (~0.91:1); suite at PR head = 224 pass / 14 skip / 0 fail (measured)
diff coverage (delta): 3/3 new branches in resolveQuery's catch exercised by asserting deterministic tests (manual map)
mergeable: MERGEABLE · openspec validate add-claude-code-provider: valid (measured) · CI: cla/template pass
```
Method: suite/typecheck/build run locally on a `git stash`-cleaned tree at `6142ae9` (my working tree carries unrelated WIP, stashed so it did not contaminate the result, then restored).

### Invariant checks (performed, all pass)
- **Spec-gated in** — delta does not touch tool advertising; provider still recomputes `availableTools(ctx)` each turn. ✅
- **Verification-gated out** — `loop.ts` untouched by the delta; every provider still driven through the same gates. ✅
- **`check`/`verify` LLM-free & network-free** — delta touches only the provider and its test, neither reachable from `check`. ✅
- **No sexp serialization / ledger** — N/A, untouched. ✅
- **Secrets** — new code reads/logs no credential; the injected importer carries only a module specifier. ✅
- **Spec coherence** — no spec artifacts in the delta; `openspec validate` now passes. ✅

### Findings

**low — CJS `MODULE_NOT_FOUND` half of the branch is untested.** `src/agent/providers/claude-code.ts` maps both `ERR_MODULE_NOT_FOUND` (ESM) and `MODULE_NOT_FOUND` (CJS) to the install message, but the new tests only inject `ERR_MODULE_NOT_FOUND`. The CJS variant is low-risk (this is an ESM package, so that code is defensive), but strictly it is an uncovered branch-half. Fix: add one case reusing `moduleNotFound` with `code: 'MODULE_NOT_FOUND'`, asserting the same message — a one-line addition that makes the branch fully covered.

No correctness, security, or invariant issues. The delta is a clean, well-targeted coverage improvement; the one nit is optional.

PR #43, opened 2026-07-21
Post PR-lint results as a comment

animesh-chouhan:
## What

The PR-lint check now posts its result as a comment on the PR, listing exactly which template sections or the invariant checklist are missing, instead of only writing to the Actions job log.

## Why

The findings were only visible by opening the failed check's Details in the Actions tab, so authors rarely saw what was wrong. A comment puts the specific gaps in front of them. Every open PR is currently from a fork, and a plain `pull_request` run gets a read-only token on forks, so it could not comment at all.

## Design

The trigger moves from `pull_request` to `pull_request_target` so the job holds a writable token in the base-repo context. This stays safe because the workflow never checks out or runs PR code: it reads the body from the `PR_BODY` env var and parses it, same as before. A file comment warns against adding a PR-head checkout. The comment carries a hidden `<!-- pr-lint-report -->` marker so later runs (on edit or synchronize) update the same comment in place rather than piling on new ones. The check still exits non-zero on an incomplete description, so merge gating is unchanged. See [pr-lint.yml](.github/workflows/pr-lint.yml).

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | `npm run typecheck` | n/a |
| Build | `npm run build` | n/a |
| Unit + offline | `npm test` | n/a |

CI-config-only change: no application code is touched. The YAML was validated as parseable and the node parsing/output logic was dry-run locally against sample bodies to confirm it detects missing sections and emits the comment payload.

- Commands run and outcome: n/a, CI-config-only. The workflow cannot be exercised from the CLI sandbox; `pull_request_target` runs the default-branch version, so it will be observable on PRs once merged.

```text
n/a
```

## Docs

None.

---

## Invariant checklist

- [ ] **Spec-gated in**: N/A, no agent tooling touched.
- [ ] **Verification-gated out**: N/A, no mutation path touched.
- [ ] **`check`/`verify` stays LLM-free and network-free**: N/A, CI workflow only.
- [ ] **No sexp serialization**: N/A, KiCad layer untouched.
- [ ] **Sync-obligations ledger**: N/A, loop untouched.
- [ ] **Secrets**: N/A, no transcript or key handling touched.

### DIFF
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -4,12 +4,18 @@ name: PR lint
 # the required sections and the invariant checklist must be present and non-empty.
 # It verifies presence, not truth. A reviewer (see the pr-review skill) still confirms
 # the content is honest and the ticked invariants actually hold.
+#
+# Uses pull_request_target so the job holds a writable token and can post its result
+# as a comment on fork PRs (a plain pull_request run gets a read-only token on forks).
+# This is safe here only because the workflow never checks out or runs PR code: it reads
+# the PR body from an env var and parses it. Do not add a checkout of the PR head.
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, edited, synchronize, reopened, ready_for_review]
 
 permissions:
   contents: read
+  pull-requests: write
 
 jobs:
   template:
@@ -18,11 +24,13 @@ jobs:
     if: ${{ !github.event.pull_request.draft && github.event.pull_request.user.login != 'dependabot[bot]' }}
     steps:
       - name: Check PR description
+        id: check
         env:
           # Passed via env, never interpolated into the script, so a crafted body cannot inject shell.
           PR_BODY: ${{ github.event.pull_request.body }}
         run: |
           node <<'EOF'
+          const fs = require("fs");
           const body = process.env.PR_BODY || "";
           // Drop HTML comments so the template's placeholder guidance never counts as content.
           const stripped = body.replace(/<!--[\s\S]*?-->/g, "");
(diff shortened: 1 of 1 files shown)


animesh-chouhan:
🤖 **Automated pr-review pass** (copperhead pr-review skill; not a human review)

**Verdict: approve with nits.** The core change is correct and the `pull_request_target` usage is safe (no checkout of PR code; body read via env). `npm test` was not run: no application code or tests are reachable from this CI-only change. All seven repo invariants are N/A here (no agent tooling, mutation path, `check` imports, KiCad parser, ledger, or secret handling touched).

### Findings

**1. No `concurrency` group leads to duplicate comments (medium)** — `.github/workflows/pr-lint.yml`
Idempotency relies on finding an existing marker comment before posting, but `edited` and `synchronize` fire independent, overlapping runs: both list comments, both find none, both POST. If an author pushes a commit and edits the description within a few seconds, two "PR lint" comments appear and later runs update only one. Fix: add `concurrency: { group: pr-lint-${{ github.event.pull_request.number }}, cancel-in-progress: true }`.

**2. Marker match not scoped to the bot (low)** — comment lookup
The lookup selects any comment whose body starts with the marker, regardless of author. A contributor could post such a comment and the workflow would PATCH theirs instead of its own. Cosmetic; scope the jq filter with `and (.user.login == "github-actions[bot]")`.

**3. Comment step has no guard against an empty report (low)**
If the check step fails before writing `GITHUB_OUTPUT`, the unconditional comment step calls `gh api ... -f body=""`, which returns 422 and masks the real error. Add `if: ${{ steps.check.outputs.report != '' }}`.

### Verified and passing
`pull_request_target` is safe (no PR-code checkout; body via env, not interpolated); the `GITHUB_OUTPUT` heredoc delimiter cannot be injected since the report is built from fixed strings only; `pull-requests: write` is the correct scope for PR conversation comments; the unchanged job name preserves required-status-check config; the check still exits non-zero on incomplete descriptions, so merge gating is unchanged.

Issue #4, opened 2026-07-21
Contribute an example design brief (simple, medium, or hard tier)


animesh-chouhan:
## Context

`examples/` contains design briefs in `simple/`, `medium/`, and `hard/` tiers that feed `copperhead create --brief`. More briefs mean more coverage of the pipeline and better demos. This is a pure-markdown contribution: no TypeScript, no toolchain.

## What to do

1. Read `examples/README.md` and one existing brief per tier to absorb the format and level of detail (constraints, budgets, unstated-decision traps).
2. Write a brief for a new board. Ideas: a battery sensor node, a small motor driver, an audio preamp, a USB-C power sink. Pick something you actually know.
3. Good briefs include explicit budgets (power, cost, size) and at least one deliberate trap: a constraint that a naive design would violate, so the agent has something to refuse or flag.
4. Open a PR adding the brief to the appropriate tier.

## Definition of done

A new brief merged under `examples/<tier>/`, consistent in format with its neighbors, containing at least one explicit budget and one trap, with a one-line entry in `examples/README.md`.

## Notes

You do not need to run `create` against your brief (that needs an API key); format and content review happens in the PR. Domain knowledge is the contribution here.

Issue #2, opened 2026-07-21
Add a saved-login Claude Code provider via the Claude Agent SDK


animesh-chouhan:
## Summary

PR #1 adds a saved-login Codex provider: `--model codex` drives the locally installed, ChatGPT-authenticated Codex CLI through the official SDK, so no OpenAI API key is needed. The same pattern should exist for Claude Code users via the **Claude Agent SDK** (`@anthropic-ai/claude-agent-sdk`, docs at https://code.claude.com/docs/en/agent-sdk). It drives the locally installed Claude Code and reuses its existing login, so a `--model claude-code` provider would let subscription users run copperhead without an `ANTHROPIC_API_KEY`.

## Proposed design

Mirror the decisions settled in PR #1:

- **Provider seam**: route `claude-code` (and `claude-code:<model-id>`) in `makeProvider`, implement `Provider.chat(messages, tools) -> Turn` on top of the SDK's query surface, and keep an injectable client seam for offline tests, same as `CodexProvider`.
- **Reasoning-only backend**: disable the SDK's built-in file-editing, bash, and web tools; run in an isolated working directory. All mutations flow through copperhead's capability-filtered tools, obligations ledger, ERC/DRC verification, snapshot, and commit gates.
- **Structural tool gating**: this may come out cleaner than the Codex version. Instead of mirroring the gate in a structured-output enum, the Agent SDK supports registering custom in-process tools, so copperhead can expose exactly the tools `availableTools(ctx)` returns each turn. The spec-gated-in invariant stays structural rather than schema-mirrored.
- **Auth stays external**: copperhead never reads, copies, or logs the Claude Code credential; the CLI owns authentication. A missing or unauthenticated install fails through the normal rollback path with an actionable error.
- **No silent fallback** to a paid API provider, matching the Codex provider's behavior.
- **Optional dependency**: lazy `import()` in `makeProvider` rather than a hard install for every user (same packaging question raised on PR #1 for `@openai/codex-sdk`).

## Scope

- OpenSpec change with proposal, design, delta specs, and tasks (same shape as `add-codex-cli-provider`).
- Offline provider tests (routing, tool mapping, gating, malformed output, error status preservation for the shared retry wrapper).
- Opt-in live provider-parity entry in `test/agent-integration.test.ts` (AC-3.10).
- Docs: README, docs site, `.env.example`, configuration reference.

## Blocked on

PR #1 landing first, since this reuses the provider seam, gating approach, and packaging decisions settled there.

PR #48, opened 2026-07-22
feat(agent): support reasoning models and Gemini via OpenAI provider

thegoodengineer:
## What

This PR adds support for reasoning models and Gemini by preserving non-standard or vendor-specific tool call properties (such as thinking signatures) across agent turns in the OpenAI provider.

## Why

OpenAI-compatible reasoning models (like Google Gemini via OPENAI_BASE_URL) return custom metadata in their tool calls (e.g. extra_content.google.thought_signature). If this metadata is stripped from subsequent assistant history messages, the API rejects the request on the next turn with a 400 Bad Request error. This PR preserves those fields through a generic extra bag on ToolCall.

## Design

Add a generic extra field to ToolCall in [types.ts](src/agent/types.ts). In [openai.ts](src/agent/providers/openai.ts), extraction excludes id, type, and function; everything else is captured into extra. On the outgoing side, extra is spread back onto each tool call object. Both spread sites carry a comment explaining why the field exists and what breaks without it.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | npm run typecheck | pass |
| Build | npm run build | pass |
| Unit + offline | npm test | pass (147 passed / 10 skipped) |
| Live-LLM (opt-in) | keyed on OPENAI_API_KEY + OPENAI_BASE_URL | pass (Gemini 3.5 Flash via AI Studio) |

- Sandbox variant used (create / edit): create
- Commands run and outcome:

Ran copperhead create against Gemini 3.5 Flash with OPENAI_BASE_URL pointing to generativelanguage.googleapis.com. The run completed 9 turns of spec-seed successfully, committing 5 files. The transcript at .copperhead/runs/2026-07-22T05-53-45-473Z/transcript.jsonl confirms extra.extra_content.google.thought_signature round-tripped on every assistant turn (turns 2 through 9). Without this patch, the run 400'd on turn 2.

## Docs

None required. The extra field is internal to the provider; no user-facing configuration changes.

---

## Invariant checklist

- [x] **Spec-gated in**: edit_file/write_file stay structurally absent from the agent tool list until an OpenSpec proposal validates (gated by omission, not prompt text).
- [x] **Verification-gated out**: mutations still end in ERC (and DRC when the board changed) passing, with repair up to maxRepairCycles then rollback to the git snapshot.
- [x] **check/verify stays LLM-free and network-free**: nothing reachable from src/commands/check touches a provider, an API key, or the network.
- [x] **No sexp serialization**: the src/kicad/ parser stays read-only; KiCad files are edited only via anchored exact-match text replace (no round-tripping).
- [x] **Secrets**: transcripts and summaries redact sk- secrets at write time; keys live only in env vars; .gitignore keeps .env and .copperhead/runs/.
- [x] **Spec coherence**: N/A, no spec-level behavior changed.

### DIFF
--- a/src/agent/providers/openai.ts
+++ b/src/agent/providers/openai.ts
@@ -37,6 +37,7 @@ export class OpenAIProvider implements Provider {
                       id: t.id,
                       type: 'function' as const,
                       function: { name: t.name, arguments: JSON.stringify(t.args) },
+                      ...(t.extra || {}),
                     })),
                   }
                 : {}),
@@ -55,11 +56,20 @@ export class OpenAIProvider implements Provider {
         : {}),
     });
     const choice = res.choices[0];
-    const toolCalls = ((choice?.message.tool_calls ?? []) as OpenAIToolCall[]).map((t) => ({
-      id: t.id,
-      name: t.function.name,
-      args: safeParse(t.function.arguments),
-    }));
+    const toolCalls = ((choice?.message.tool_calls ?? []) as any[]).map((t) => {
+      const extra: Record<string, any> = {};
+      for (const [k, v] of Object.entries(t)) {
+        if (k !== 'id' && k !== 'type' && k !== 'function') {
+          extra[k] = v;
+        }
+      }
+      return {
+        id: t.id,
+        name: t.function.name,
+        args: safeParse(t.function.arguments),
+        ...(Object.keys(extra).length ? { extra } : {}),
+      };
+    });
     return {
       text: choice?.message.content ?? null,
       toolCalls,
--- a/src/agent/providers/openai.ts
+++ b/src/agent/providers/openai.ts
@@ -32,6 +32,8 @@ export class OpenAIProvider implements Provider {
                       id: t.id,
                       type: 'function' as const,
                       function: { name: t.name, arguments: JSON.stringify(t.args) },
+                      // Preserve vendor-specific tool-call fields (e.g. Gemini thought signatures).
+                      // Dropping them makes the next turn's request 400.
                       ...(t.extra || {}),
                     })),
                   }
@@ -51,6 +53,9 @@ export class OpenAIProvider implements Provider {
         : {}),
     });
     const choice = res.choices[0];
+    // Capture any non-standard properties returned by the API (e.g. Gemini thought
+    // signatures) so they can be echoed back on subsequent turns. Dropping them
+    // causes reasoning-model backends to reject the follow-up request with 400.
     const toolCalls = ((choice?.message.tool_calls ?? []) as unknown as Record<string, unknown>[]).map((t: Record<string, unknown>) => {
       const extra: Record<string, unknown> = {};
       for (const [k, v] of Object.entries(t)) {
--- a/src/agent/providers/openai.ts
+++ b/src/agent/providers/openai.ts
@@ -1,10 +1,5 @@
 import type { ChatOpts, Msg, Provider, ToolSchema, Turn } from '../types.js';
 
-interface OpenAIToolCall {
-  id: string;
-  function: { name: string; arguments: string };
-}
-
 export class OpenAIProvider implements Provider {
   readonly name = 'openai';
 
@@ -56,17 +51,18 @@ export class OpenAIProvider implements Provider {
         : {}),
     });
     const choice = res.choices[0];
-    const toolCalls = ((choice?.message.tool_calls ?? []) as any[]).map((t) => {
-      const extra: Record<string, any> = {};
+    const toolCalls = ((choice?.message.tool_calls ?? []) as unknown as Record<string, unknown>[]).map((t: Record<string, unknown>) => {
+      const extra: Record<string, unknown> = {};
       for (const [k, v] of Object.entries(t)) {
         if (k !== 'id' && k !== 'type' && k !== 'function') {
           extra[k] = v;
         }
       }
+      const fn = t.function as { name: string; arguments: string };
       return {
-        id: t.id,
-        name: t.function.name,
-        args: safeParse(t.function.arguments),
+        id: t.id as string,
+        name: fn.name,
+        args: safeParse(fn.arguments),
         ...(Object.keys(extra).length ? { extra } : {}),
       };
     });
--- a/src/agent/providers/openai.ts
+++ b/src/agent/providers/openai.ts
@@ -1,4 +1,4 @@
-import type { ChatOpts, Msg, Provider, ToolSchema, Turn } from '../types.js';
+import type { ChatOpts, Msg, Provider, ToolSchema, Turn, ToolCall } from '../types.js';
 
 export class OpenAIProvider implements Provider {
   readonly name = 'openai';
@@ -28,14 +28,7 @@ export class OpenAIProvider implements Provider {
               content: m.content,
               ...(m.toolCalls?.length
                 ? {
-                    tool_calls: m.toolCalls.map((t) => ({
-                      id: t.id,
-                      type: 'function' as const,
-                      function: { name: t.name, arguments: JSON.stringify(t.args) },
-                      // Preserve vendor-specific tool-call fields (e.g. Gemini thought signatures).
-                      // Dropping them makes the next turn's request 400.
-                      ...(t.extra || {}),
-                    })),
+                    tool_calls: m.toolCalls.map(serializeToolCall),
                   }
                 : {}),
             };
@@ -56,21 +49,7 @@ export class OpenAIProvider implements Provider {
     // Capture any non-standard properties returned by the API (e.g. Gemini thought
     // signatures) so they can be echoed back on subsequent turns. Dropping them
     // causes reasoning-model backends to reject the follow-up request with 400.
-    const toolCalls = ((choice?.message.tool_calls ?? []) as unknown as Record<string, unknown>[]).map((t: Record<string, unknown>) => {
-      const extra: Record<string, unknown> = {};
-      for (const [k, v] of Object.entries(t)) {
-        if (k !== 'id' && k !== 'type' && k !== 'function') {
-          extra[k] = v;
-        }
-      }
-      const fn = t.function as { name: string; arguments: string };
-      return {
-        id: t.id as string,
-        name: fn.name,
-        args: safeParse(fn.arguments),
-        ...(Object.keys(extra).length ? { extra } : {}),
-      };
-    });
+    const toolCalls = ((choice?.message.tool_calls ?? []) as unknown as Record<string, unknown>[]).map(parseToolCall);
     return {
       text: choice?.message.content ?? null,
       toolCalls,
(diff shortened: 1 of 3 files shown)


thegoodengineer:

animesh-chouhan:
🤖 **Automated pr-review pass** (Claude Code `/pr-review`, not a human review — no merge gating implied)

## Verdict: Approve with nits

Small, well-scoped change (+16/−5, two files). Applies cleanly onto current `main`, `npm run typecheck` passes, and the full offline suite passes (**147 passed / 10 skipped**). No repo invariant is touched. Nits are about test evidence and type hygiene, not correctness.

### What it does
Adds an optional `extra?` bag to `ToolCall` and, in the OpenAI provider, captures any non-`id`/`type`/`function` fields from an incoming tool call, then re-spreads them onto the outgoing tool call next turn — so vendor metadata (e.g. Gemini "thought signatures") survives the round-trip and the follow-up request doesn't 400.

### Verification performed
- **Round-trip is sound.** Extraction excludes `id`/`type`/`function`, so the outgoing `...(t.extra || {})` spread can't clobber those keys. The bag is carried into the assistant message at `loop.ts:412`, so it genuinely reaches the next turn.
- **No OpenAI regression.** For a standard OpenAI response the bag is empty, and `...(Object.keys(extra).length ? { extra } : {})` omits the field — output identical to before.
- **Gemini is reachable.** `makeProvider` routes non-`claude`/non-`codex` models (`gemini-*`) to `OpenAIProvider`, and the OpenAI SDK defaults `baseURL` to `process.env.OPENAI_BASE_URL`. So a user can point it at Gemini's OpenAI-compat endpoint via env.

### Invariant checks (all pass — none touched by this PR)
- **Spec-gated in**: tool exposure unchanged. ✅
- **Verification-gated out**: repair/rollback path untouched. ✅
- **`check`/`verify` LLM-free & network-free**: `src/commands/check.ts` imports no provider (asserted at its line 13); this PR only edits provider/type code. ✅
- **No sexp serialization**: parser untouched. ✅
- **Sync-obligations ledger**: untouched. ✅
- **Secrets**: `extra` blobs reach transcript events, but `sk-…` redaction still runs at write time and signatures aren't secrets. ✅

### Findings (all low severity / nits)

1. **No test coverage for the new behavior.** No dedicated OpenAI-provider unit test exists, and none is added. A ~10-line test feeding a mocked tool call with an extra field (e.g. `thought_signature`) and asserting it survives into the next outgoing `tool_calls` payload would lock in the exact behavior. Failure scenario: a future refactor silently drops `extra` and nothing goes red.

2. **`any` where the codebase uses `unknown`.** `types.ts` declares `extra?: Record<string, any>` and the map casts `as any[]`; the rest of the repo uses `Record<string, unknown>`. `unknown` keeps type safety with no loss — the bag is only spread, never dereferenced.

3. **Dead interface left behind.** The cast changed from `as OpenAIToolCall[]` to `as any[]`, leaving the `OpenAIToolCall` interface (`openai.ts:3-6`) unused. `tsc` doesn't flag unused interfaces, so typecheck stays green, but it's stray code — delete it or keep using it as the cast target.

4. **Manual-test evidence is a placeholder — and it covers the one thing I can't verify.** The description's invariant checklist is entirely unchecked, the command block is truncated, and it names `--model gemini-3.5-flash`, which is not a real Gemini model id. I independently confirmed typecheck + offline tests pass, but the actual payoff (Gemini not 400-ing on turn two because the signature is echoed back at the tool-call level) rests on an unverified assumption about where Gemini's OpenAI-compat layer expects the signature, with no live run behind it. Worth one real `create` run against Gemini before merge, plus a one-line doc note that targeting Gemini requires `OPENAI_BASE_URL` (nothing in the repo surfaces that today).

None block merge. Recommend addressing **#1** and **#4** if a Gemini run is easy to produce; #2/#3 are cleanup.

animesh-chouhan:
🤖 **Automated pr-review pass — follow-up (merge requirements)**

Two things from the review above are hard requirements before this should merge, not optional nits:

### 1. Add a "why" comment at the `extra` sites
The two spread lines are load-bearing but read as innocent pass-through:

```ts
...(t.extra || {}),                              // outgoing (assistant history)
...(Object.keys(extra).length ? { extra } : {})  // incoming (from the API)
```

Nothing on the page tells a maintainer that `extra` is what keeps Gemini/reasoning-model tool calls valid. Without a comment, a future "cleanup" of the provider deletes it and silently reintroduces the exact `400 Bad Request` this PR fixes — and there's no test to catch it (see finding #1). Please add a one-line rationale at the extraction/merge sites, e.g.:

> `// Preserve vendor-specific tool-call fields (e.g. Gemini thought signatures). Dropping them makes the next turn's request 400.`

### 2. Live multi-turn verification against the Gemini backend is required
This PR is a claim about how a specific backend behaves *across turns*. Typecheck and the offline suite (both confirmed green locally: 147 passed / 10 skipped) cannot prove it — only a real multi-turn run can. Please attach a transcript from an actual `create`/`do` run against the Gemini OpenAI-compat endpoint (`OPENAI_BASE_URL` set accordingly) showing the second turn succeeds where it previously 400'd.

### 3. Fix the placeholder text in the PR description
The description's evidence is currently non-functional and should be corrected before the claims stand:
- The **Invariant checklist** is entirely unchecked (`- [ ]` for all seven items).
- The **Manual test log** command block is truncated and cites `--model gemini-3.5-flash`, which is not a real Gemini model id.
- The **Automated test status** table lists `Live-LLM (opt-in) … pass` keyed on `<ENV_VAR>` (literal placeholder). Given the points above, that "pass" is unverified — please replace it with the real env var and back it with the transcript from #2, or mark it not-run.

Happy to re-review once the comment, the checklist, and a real Gemini transcript are in.

thegoodengineer:
Hi @animesh-chouhan, thanks for the thorough review! I've addressed all three requirements:

1. **Why-comments**: Added explanatory comments at both `extra` spread sites in `openai.ts` (commit 8f1239c).
2. **Live Gemini evidence**: Updated the manual test log with real output from a 9-turn `copperhead create` run against Gemini 3.5 Flash. The transcript confirms `thought_signature` round-tripped on every turn.
3. **PR description**: Replaced all placeholder text, checked the invariant boxes, and cited the real model and env var names.

Happy to make any further changes if needed!

animesh-chouhan:
🤖 **Automated pr-review pass 2** (Claude Code `/pr-review`, not a human review — no merge gating implied)

## Verdict: Approve with one nit

Every item from pass 1 is addressed. Code is correct, typechecks, and the full offline suite is green (**149 passed / 10 skipped**, up from 147). One remaining nit about what the new test actually guards.

### Changes since pass 1 (all confirmed applied)
- `any` → `unknown`: `extra?: Record<string, unknown>` in `types.ts` and throughout the map. ✅
- Dead `OpenAIToolCall` interface removed from `openai.ts`. ✅
- Why-comments added at both the extraction and outgoing-spread sites, naming the 400 failure mode. ✅
- Test added (`test/toolcall-extra.test.ts`) — passes. ✅
- PR description rewritten — invariant checklist filled, plus a concrete manual-test log (Gemini run, transcript path, turns 2–9 round-trip, 400 on turn 2 without the patch). ✅

### Verification performed this pass
- Applied the 3-file diff onto current `main` — applies cleanly.
- `npm run typecheck` — pass.
- `npx vitest run test/toolcall-extra.test.ts` — 2 passed.
- `npm test` — 149 passed / 10 skipped, no regressions. Reverted cleanly afterward.

### Invariant checks (all pass — none touched)
Spec-gated in ✅ · Verification-gated out ✅ · `check`/`verify` LLM-free & network-free (`src/commands/check.ts` still imports no provider) ✅ · No sexp serialization ✅ · Sync-obligations ledger ✅ · Secrets/redaction ✅. Change is confined to the provider and a core type; no spec-level behavior changed, so no SPEC/artifact update is owed.

### Remaining finding (low — quality, not correctness)

**1. The new test re-implements the serialization instead of exercising the provider, so it doesn't guard the real code path.** `test/toolcall-extra.test.ts` hand-copies the `.map(...)` spread into the test body rather than importing `OpenAIProvider`. Failure scenario: someone deletes `...(t.extra || {})` from `openai.ts` — the exact regression the new comment warns against — and this test stays **green**, because it never runs `openai.ts`. The incoming extraction loop (the `Object.entries` filter) isn't exercised at all. Cheap fix: extract the two transforms into small exported pure helpers (e.g. `serializeToolCall` / `parseToolCall`) in `openai.ts` and have the test import *those*, so the assertions are pinned to the shipping code. As written the test documents intent but gives false regression confidence.

### Note on the live claim (not a finding)
I could not independently verify the Gemini transcript — `.copperhead/runs/` is gitignored (AC-4.3), so it isn't in the PR. But the described evidence is now specific and credible (named turns, the `extra_content.google.thought_signature` path, the turn-2 400 without the patch), which is what pass 1 asked for. Taken at face value; anyone wanting hard proof can request the transcript pasted (redacted). I did not attempt to confirm the "Gemini 3.5 Flash" model name — outside my knowledge cutoff and not something the diff depends on.

Recommend merging once the test is pointed at the real provider code (nit #1); everything else is resolved.

thegoodengineer:
Done! Extracted `serializeToolCall` and `parseToolCall` out as pure exported helpers in `openai.ts` and pointed the tests directly at them (commit 1be696b).

animesh-chouhan:
Can you attach the transcript of the run or just the plain log file: Ran copperhead create against Gemini 3.5 Flash with OPENAI_BASE_URL pointing to generativelanguage.googleapis.com. The run completed 9 turns of spec-seed successfully, committing 5 files. The transcript at .copperhead/runs/2026-07-22T05-53-45-473Z/transcript.jsonl confirms extra.extra_content.google.thought_signature round-tripped on every assistant turn (turns 2 through 9). Without this patch, the run 400'd on turn 2.

thegoodengineer:
Here is the transcript from the Gemini run showing the thought signatures round-tripping properly! 

[transcript.txt](https://github.com/user-attachments/files/30262811/transcript.txt)

animesh-chouhan:
🤖 **Automated pr-review pass 3** (Claude Code `/pr-review`, not a human review — no merge gating implied)

## Verdict: Approve

The last remaining nit is resolved. No outstanding findings. Mergeable.

### Change since pass 2
The two transforms are now exported pure functions — `serializeToolCall` and `parseToolCall` in `openai.ts` — and `chat` calls them (`m.toolCalls.map(serializeToolCall)`, `.map(parseToolCall)`). The test now **imports and exercises those real functions**, adding a third case that runs the incoming `parseToolCall` path directly:
- outgoing: `extra` survives serialization ✅
- outgoing: no `extra` → only `id`/`type`/`function` emitted ✅
- incoming: non-standard fields (`extra_content`, `some_other_field`) captured into `extra`, `id`/`type`/`function` excluded ✅

This closes pass 2's finding: deleting `...(t.extra || {})` from `serializeToolCall`, or breaking the extraction loop in `parseToolCall`, now turns the suite **red**. The test is pinned to shipping code.

### Verification performed this pass
- Applied the 3-file diff onto current `main` — applies cleanly.
- `npm run typecheck` — pass.
- `npx vitest run test/toolcall-extra.test.ts` — 3 passed.
- `npm test` — **150 passed / 10 skipped**, no regressions. Reverted cleanly afterward.

### Invariant checks (all pass — none touched)
Spec-gated in ✅ · Verification-gated out ✅ · `check`/`verify` LLM-free & network-free (`src/commands/check.ts` imports no provider) ✅ · No sexp serialization ✅ · Sync-obligations ledger ✅ · Secrets/redaction ✅. Confined to the provider + a core type; no spec-level behavior changed, so no SPEC/artifact update owed.

### Trivial note (not a finding)
`git apply` flagged one trailing-whitespace line in the test (blank line ~L57). Cosmetic only — the repo lints markdown solely, so CI won't fail. Optional cleanup.

**Recommendation: merge.** All three passes' findings are resolved. The Gemini live-run remains the author's attestation (transcript is gitignored, so not independently checkable), but the mechanism is correct, tested against the real code, and regression-guarded.

PR #62, opened 2026-07-23
feat(spice): parse simulation blocks and compile assertions

### LINKED ISSUE #10: Parse the ## Simulation block and SPICE assertion grammar
## Context

The planned SPICE verification gate (OpenSpec change `add-spice-verification-gate`, tasks 1.2 and 1.3) lets a subsystem opt into behavioral verification via a `## Simulation` block in SUBSYSTEMS.md. The block parser and the assertion-to-`.meas` compiler are pure parsing work, fully unit-testable without ngspice installed. Good fit if you enjoy writing small parsers.

## Spec

The block names: a netlist scope (hierarchical sheet reference or explicit net set), an analysis type (`op`, `dc`, `ac`, `tran`), optional `sources` lines for driven ports, and one or more assertions.

Assertion grammar (closed, nothing else is legal):
- Measurables: `V(<net>)`, `I(<refdes>)`, `corner(<net>)`
- Comparators: `between <a> and <b>`, `< <x>`, `> <x>`
- Numbers with SI suffixes (`25uA`, `3.3`, `10k`)

Raw ngspice `.control` blocks must be rejected with a parse error naming the offending line (reviewability and injection safety are the point of the closed grammar).

## Definition of done (from the delta spec)

- `V(VREF) between 3.25 and 3.35` compiles to a corresponding `.meas` directive with both bounds evaluated.
- A raw `.control` line fails with a parse error naming the line, without any ngspice invocation.
- Unit tests cover every measurable x comparator pair and SI-suffix parsing.

## Notes

Offline, no API key, no ngspice needed for this piece. The wrapper that actually runs ngspice is a separate task; this issue ends at "correct deck fragment out, or precise parse error".

raj921:
## What

Adds the parser and compiler for opt-in `## Simulation` blocks in `SUBSYSTEMS.md`.

The new pure KiCad-layer module parses sheet or net scopes, analysis type, driven sources, and assertions. It compiles the closed assertion grammar to ngspice `.meas` fragments and reports parse failures with the original Markdown line number.

## Why

The planned SPICE gate needs a safe input boundary before netlist export or ngspice execution can be added. Accepting arbitrary control text would make the gate difficult to review and unsafe to assemble into a deck.

Closes #10.

## Design

- The parser is line-oriented so every failure points back to the source line.
- It accepts one scope, one analysis, repeated `source` or `sources` entries, and one or more assertions.
- Fenced examples, HTML-commented blocks, non-field narrative, and lower-level Markdown sections are not treated as live Simulation input.
- Raw dot commands such as `.control` are rejected before any process layer is involved.
- SI values are parsed to normalized numbers while retaining the original safe SPICE literal for deck generation.
- `between` requires both bounds to omit units or use the same unit.
- Each assertion produces one value measure plus bound-margin measures. `between` emits separate lower and upper checks. The returned check metadata distinguishes inclusive bounds from strict `<` and `>` bounds.
- `corner(net)` is limited to AC analysis and compiles to the first 3 dB crossing.
- The module has no filesystem, subprocess, provider, or network dependency.

## Spec / OpenSpec

- Active change: `openspec/changes/add-spice-verification-gate`
- Completes tasks 1.2 and 1.3.
- Clarifies the line-oriented block fields in the `spice-verification` delta spec.
- `openspec validate add-spice-verification-gate` passes.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | partial locally: 242 pass, 7 fail because `kicad-cli` is not installed, 14 skipped |
| Focused parser/compiler | `npm test -- test/spice.test.ts` | pass, 40/40 |
| Markdown | `npm run lint:md` | pass, 106 files |
| Package | `npm pack --dry-run` | pass |
| OpenSpec | `npx --yes @fission-ai/openspec@latest validate add-spice-verification-gate --no-interactive` | pass |
| Live-LLM (opt-in) | API-key-gated suite | skipped, this change is pure and does not use a provider |

The seven full-suite failures are all downstream of the missing local `kicad-cli` prerequisite. The repository CI workflow installs KiCad before running the same suite.

New tests cover both scope forms, repeated sources, fenced and HTML-commented examples, trailing narrative, line-numbered validation errors, raw control rejection, SI suffixes, consistent units, the exact VREF range example, all nine measurable and comparator pairings, and the AC-only corner rule.

- Sandbox variant used (`create` / `edit`): `edit`
- Commands run and outcome:

```text
$ ./manual-tests/setup.sh edit
sandbox ready: manual-tests/runs/edit

$ node --input-type=module -e 'import { readFile } from "node:fs/promises"; import { parseSimulationBlocks, compileSpiceAssertion } from "./dist/kicad/spice.js"; const markdown = await readFile("manual-tests/runs/edit/docs/SUBSYSTEMS.md", "utf8"); const [block] = parseSimulationBlocks(markdown); if (!block) throw new Error("Simulation block not found"); console.log(`parsed ${block.scope.kind} scope, ${block.analysis} analysis, ${block.sources.length} source, ${block.assertions.length} assertions`); for (const [index, assertion] of block.assertions.entries()) console.log(compileSpiceAssertion(assertion, block.analysis, `assertion_${index + 1}`).deck); try { parseSimulationBlocks("## Simulation\nscope: nets VREF,GND\nanalysis: op\n.control\nassert: V(VREF) > 3\n"); } catch (error) { console.log(String(error)); }'
parsed sheet scope, op analysis, 1 source, 2 assertions
.meas op assertion_1_value FIND v(VREF)
.meas op assertion_1_lower_margin PARAM='assertion_1_value - 3.25'
.meas op assertion_1_upper_margin PARAM='3.35 - assertion_1_value'
.meas op assertion_2_value FIND i(R5)
.meas op assertion_2_margin PARAM='25uA - assertion_2_value'
SimulationParseError: line 4: raw ngspice statement ".control" is not allowed
```

The parser/compiler tasks are not wired into the CLI yet. That work starts at task 2.2, so the built module was exercised directly against the repository's edit sandbox.

## Docs

Updated the active SPICE delta spec with the accepted field syntax, inert Markdown rules, and unit consistency requirement. OpenSpec tasks 1.2 and 1.3 are marked complete. User-facing docs remain part of task 5.1 after the gate is wired.

---

## Invariant checklist

- [ ] **Spec-gated in**: N/A, agent edit tools are not touched.
- [ ] **Verification-gated out**: N/A, mutation and rollback paths are not touched.
- [x] **`check`/`verify` stays LLM-free and network-free**: the new parser/compiler is pure and does not import a provider, subprocess, or network path.
- [x] **No sexp serialization**: the new KiCad-layer parser reads Markdown strings only and never reads, rewrites, or serializes a KiCad s-expression.
- [ ] **Sync-obligations ledger**: N/A, ledger hooks and commit gates are not touched.
- [ ] **Secrets**: N/A, transcripts, environment handling, and ignore rules are not touched.
- [x] **Spec coherence**: the active delta spec and task list move with the implementation, and OpenSpec validation passes.

### DIFF
--- a/openspec/changes/add-spice-verification-gate/specs/spice-verification/spec.md
+++ b/openspec/changes/add-spice-verification-gate/specs/spice-verification/spec.md
@@ -3,7 +3,7 @@
 ## ADDED Requirements
 
 ### Requirement: Simulation opt-in via SUBSYSTEMS.md
-A subsystem SHALL be flagged for SPICE verification by a `## Simulation` block in SUBSYSTEMS.md. The line-oriented block SHALL contain one `scope: sheet <path>` or `scope: nets <net>, <net>` field, one `analysis: op|dc|ac|tran` field, optional repeated `source: <port>=<number>` lines for driven ports, and one or more `assert: <assertion>` lines. The `sources` field name SHALL be accepted as an alias for `source`. A scope that references sheets or nets absent from the schematic SHALL be reported as a failure by `check`.
+A subsystem SHALL be flagged for SPICE verification by a `## Simulation` block in SUBSYSTEMS.md. The line-oriented block SHALL contain one `scope: sheet <path>` or `scope: nets <net>, <net>` field, one `analysis: op|dc|ac|tran` field, optional repeated `source: <port>=<number>` lines for driven ports, and one or more `assert: <assertion>` lines. The `sources` field name SHALL be accepted as an alias for `source`. Fenced code examples and non-field narrative lines SHALL NOT be parsed as live Simulation input. A scope that references sheets or nets absent from the schematic SHALL be reported as a failure by `check`.
 
 #### Scenario: Flagged subsystem is picked up
 - **WHEN** SUBSYSTEMS.md contains a `## Simulation` block scoped to the divider sheet with an `op` analysis
@@ -14,7 +14,7 @@ A subsystem SHALL be flagged for SPICE verification by a `## Simulation` block i
 - **THEN** `check` exits non-zero naming the block and the missing net
 
 ### Requirement: Assertion grammar
-Assertions SHALL use the closed grammar: measurables `V(<net>)`, `I(<refdes>)`, and `corner(<net>)`; comparators `between <a> and <b>`, `< <x>`, `> <x>`; numbers with SI suffixes. The gate SHALL compile assertions to ngspice `.meas` directives and SHALL reject any other syntax with a parse error naming the line. Raw ngspice control blocks SHALL NOT be accepted.
+Assertions SHALL use the closed grammar: measurables `V(<net>)`, `I(<refdes>)`, and `corner(<net>)`; comparators `between <a> and <b>`, `< <x>`, `> <x>`; numbers with SI suffixes. Both `between` bounds SHALL omit units or use the same unit, compared case-insensitively. The gate SHALL compile assertions to ngspice `.meas` directives and SHALL reject any other syntax with a parse error naming the line. Raw ngspice control blocks SHALL NOT be accepted.
 
 #### Scenario: Valid assertion compiles
 - **WHEN** a block contains `V(VREF) between 3.25 and 3.35`
--- a/openspec/changes/add-spice-verification-gate/specs/spice-verification/spec.md
+++ b/openspec/changes/add-spice-verification-gate/specs/spice-verification/spec.md
@@ -3,7 +3,7 @@
 ## ADDED Requirements
 
 ### Requirement: Simulation opt-in via SUBSYSTEMS.md
-A subsystem SHALL be flagged for SPICE verification by a `## Simulation` block in SUBSYSTEMS.md that names the netlist scope (a hierarchical sheet reference or an explicit net set), the analysis type (`op`, `dc`, `ac`, or `tran`), optional `sources` lines for driven ports, and one or more assertions. A scope that references sheets or nets absent from the schematic SHALL be reported as a failure by `check`.
+A subsystem SHALL be flagged for SPICE verification by a `## Simulation` block in SUBSYSTEMS.md. The line-oriented block SHALL contain one `scope: sheet <path>` or `scope: nets <net>, <net>` field, one `analysis: op|dc|ac|tran` field, optional repeated `source: <port>=<number>` lines for driven ports, and one or more `assert: <assertion>` lines. The `sources` field name SHALL be accepted as an alias for `source`. A scope that references sheets or nets absent from the schematic SHALL be reported as a failure by `check`.
 
 #### Scenario: Flagged subsystem is picked up
 - **WHEN** SUBSYSTEMS.md contains a `## Simulation` block scoped to the divider sheet with an `op` analysis
--- a/openspec/changes/add-spice-verification-gate/tasks.md
+++ b/openspec/changes/add-spice-verification-gate/tasks.md
@@ -3,8 +3,8 @@
 ## 1. Wrapper and grammar
 
 - [ ] 1.1 Implement ngspice wrapper in `src/kicad/spice.ts` mirroring the kicad-cli wrapper: execa batch mode, version detection, install hint, per-run timeout from config
-- [ ] 1.2 Implement the `## Simulation` block parser (scope, analysis, sources, assertions) with parse errors naming lines; reject raw control blocks
-- [ ] 1.3 Implement the assertion grammar → `.meas` directive compiler with SI-suffix number parsing; unit tests for every measurable/comparator pair
+- [x] 1.2 Implement the `## Simulation` block parser (scope, analysis, sources, assertions) with parse errors naming lines; reject raw control blocks
+- [x] 1.3 Implement the assertion grammar → `.meas` directive compiler with SI-suffix number parsing; unit tests for every measurable/comparator pair
 
 ## 2. Netlist scoping and execution
 
--- a/src/kicad/spice.ts
+++ b/src/kicad/spice.ts
@@ -98,10 +98,8 @@ export function parseSpiceNumber(raw: string, line = 1): SpiceNumber {
 
   const base = Number(match[1]);
   const suffix = (match[2] ?? '').toLowerCase();
-  const multiplier = SI_MULTIPLIERS[suffix];
-  if (multiplier === undefined) {
-    throw new SimulationParseError(line, `unsupported SI suffix in "${raw}"`);
-  }
+  // The regex limits suffixes to keys in SI_MULTIPLIERS.
+  const multiplier = SI_MULTIPLIERS[suffix]!;
 
   const value = base * multiplier;
   if (!Number.isFinite(value)) {
@@ -143,11 +141,7 @@ export function parseSpiceAssertion(raw: string, line = 1): SimulationAssertion 
   if (betweenMatch) {
     const lower = parseSpiceNumber(betweenMatch[1]!, line);
     const upper = parseSpiceNumber(betweenMatch[2]!, line);
-    if (
-      lower.unit.length > 0 &&
-      upper.unit.length > 0 &&
-      lower.unit.toLowerCase() !== upper.unit.toLowerCase()
-    ) {
+    if (lower.unit.toLowerCase() !== upper.unit.toLowerCase()) {
       throw new SimulationParseError(line, 'between bounds use different units');
     }
     if (lower.value >= upper.value) {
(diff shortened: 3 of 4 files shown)


raj921:

animesh-chouhan:
🤖 **Automated pr-review pass** (not a human review; no formal approval implied)

## Verdict: approve with nits

PR #62 adds a pure, dependency-free parser + `.meas` compiler for opt-in `## Simulation` blocks. The core is solid: the deck-generation path is injection-safe (number and target literals that carry a `'`, space, or dot-command are rejected before they reach a `PARAM='...'` fragment), errors are line-numbered, and the grammar pairings are well covered (25/25 spice tests pass). All the concerns below are robustness/coverage on error paths and one latent format-in-format gap. Nothing is blocking, mainly because the module is **not wired into any command yet** (task 2.2+) — but the top finding should be resolved before it is.

### Metrics
```
lines: +578 / -3 (net +575) across 4 files · src +363/-0, test +212/-0, openspec +3/-3, docs 0
new vs net: 2 new files (spice.ts, spice.test.ts), 2 modified (openspec); 363 new src lines
tests: +212 test lines; spice suite 0 -> 25 (25/25 pass, measured on branch)
        full suite: 7 fail in BOTH base & head from missing local kicad-cli (unrelated to this PR)
diff coverage: 87.9% stmts / 79.4% branch of new src (measured, @vitest/coverage-v8 throwaway install)
        uncovered: spice.ts:103,108,128,151,160,231,246,259,275,281,292,300,319 (error/validation branches)
                   + HTML-comment handling 190-201
deps/audit: no package.json change
```
Method: numstat vs `git merge-base origin/main <head>`; coverage via `vitest run --coverage` on the branch; injection/robustness traced by hand-built payloads run against the compiled module.

### Invariant checks (all performed)
- **check LLM-free / network-free**: PASS — spice.ts has zero imports (no provider, subprocess, or network); nothing in `src/` imports it yet.
- **No sexp serialization**: PASS — reads Markdown strings only; never touches a KiCad s-expression.
- **Spec-gated in / Verification-gated out / ledger / secrets**: N/A — agent tool list, mutation/rollback paths, ledger hooks, and secret handling are untouched (confirmed).
- **Spec coherence**: PASS — delta spec + tasks move with the code; `openspec validate add-spice-verification-gate` = valid.

---

### Findings

**1. [low–medium] Any non-field line inside/around a `## Simulation` section is a hard parse error — including fenced doc examples (format-in-format).** `src/kicad/spice.ts:231`
`parseSimulationBlocks` scans the whole document for a `## Simulation` heading and then requires **every** non-empty, non-comment line until the next h1/h2 to match `field: value`. Traced payloads (against the built module):
- Prose after the assertions → `line 6: expected "field: value", got "This subsystem powers the ADC..."`
- A fenced ```` ``` ```` code block *documenting* the syntax → the inner `## Simulation` is parsed as a live block and the closing fence throws `line 8: expected "field: value", got "```"`
- An `### Details` h3 subheading (h3 does not break the block; the break regex is `^#{1,2}\s+`) and a `- bullet` line → both throw.

Failure scenario: once this is wired into `check` (task 2.2), a SUBSYSTEMS.md whose `## Simulation` section contains one sentence of explanation, or a design doc that shows the block syntax inside a fenced example, turns a clean design into a spurious `check` failure with a confusing message. This is the classic "format inside the format" case for parser code. Latent today (not wired), which is why it's not higher.
Fix options: track fenced-code state and skip fenced regions in the heading scan; ignore/collect non-field lines as narrative instead of throwing; or make the block break on any non-`field:` line rather than consuming to the next h1/h2. Add a test with a fenced example and with trailing prose.

**2. [low] The HTML-comment feature is a new branch with zero test coverage.** `src/kicad/spice.ts:190-201`
The `<!-- ... -->` single- and multi-line comment handling (lines 190-201, incl. the `inComment` state machine) is entirely uncovered — no test opens, spans, or closes a comment. Multi-line comments and the "`-->` with trailing text on the same line is dropped" edge are unverified.
Fix: add a parse test with an inline `<!-- x -->` line and a multi-line comment spanning several field lines, asserting the surrounding fields still parse.

**3. [low] Uncovered validation/error branches — the highest-risk untested surface.** `src/kicad/spice.ts`
No test reaches: invalid measurable target (128), between-bounds unit mismatch (151), invalid comparator (160), unsupported analysis (246), unknown field (259), bad scope (275), empty sheet scope (281), bad net scope (292), malformed `source:` (300), invalid measure name (319), number out-of-range (108). These are error paths, so low individually, but they are exactly where a wrong line number or a swallowed case would hide. Recommend a compact table-driven "rejects and names the line" test covering each.

**4. [nit] Dead branch: unsupported-SI-suffix check is unreachable.** `src/kicad/spice.ts:103`
`parseSpiceNumber`'s regex only captures a suffix from `(meg|[tgkmunpf])`, all of which are keys of `SI_MULTIPLIERS`, so `multiplier === undefined` (line 102-103) can never be true. Harmless defensively, but it reads as reachable validation and inflates the uncovered count. Either drop it or add a comment noting it's a guard.

**5. [nit] Between-bounds unit mismatch is only enforced when both bounds carry a unit.** `src/kicad/spice.ts:148-152`
`between 3.25 and 3.35V` (one bound unit-less) passes the guard. Cosmetic given the numeric comparison still applies; worth a one-line note or tightening if unit consistency is meant to be strict.

*Nothing found on the security-critical path: deck literals are constrained to `[digits . e +/-][a-z]*` with no quotes/spaces, and `SAFE_TARGET`/`SAFE_MEASURE_NAME` gate every interpolated identifier — the `.control`-rejection and quote-injection payloads I tried were all rejected.*

raj921:
hey i have Addressed all review feedback in 07e95b9.

Fenced examples, HTML comments, narrative text, and subheadings are now handled safely. I also tightened unit validation, removed the unreachable suffix branch, and added coverage for the validation/error paths.

Focused tests are now 40/40 passing. Typecheck, build, Markdown lint, and OpenSpec validation also pass.

Ready for another  review

animesh-chouhan:
🤖 **Automated pr-review pass — follow-up** (not a human review; no formal approval implied)

Re-review after commit `07e95b9 fix(spice): harden simulation block parsing`, following up on my [earlier pass](https://github.com/chouhanindustries/copperhead/pull/62#issuecomment-5061581523). Reporting only the delta.

## Verdict: approve

The new commit resolves **all five** findings from the prior pass, adds targeted regression tests for each, and moves the spec in step. Re-ran on the branch: 40/40 spice tests pass (was 25), measured diff coverage **99.3% stmts / 94.4% branch** of `src/kicad/spice.ts` (was 87.9% / 79.4%). The security-critical path is intact — `.control` is still rejected after the rewrite, and injection payloads remain blocked.

### Prior findings — status
| # | Prior finding | Status | Evidence |
|---|---|---|---|
| 1 | Non-field lines (prose / fenced examples / h3) hard-fail the parser | **Resolved** | `parseSimulationBlocks` now tracks `` ``` ``/`~~~` fences (doc- and body-level) and drops non-`field:`/non-`.` narrative instead of throwing; break widened to any `#{1,6}` heading. New tests: "ignores fenced examples and narrative", document-level HTML comment. Traced by hand: a fenced `## Simulation` example and a following narrative line are ignored; only the live block parses. |
| 2 | HTML-comment branch had zero coverage | **Resolved** | New tests cover single-line, multi-line, and document-level `<!-- -->` comments. |
| 3 | Validation/error branches untested | **Resolved** | New table-driven tests hit unsupported analysis, unknown field, bad/empty scope, invalid net, malformed source, empty field value, unsafe measure name, and number overflow. Coverage now 99.3%. |
| 4 | Unreachable unsupported-SI-suffix guard | **Resolved** | Removed; replaced with `SI_MULTIPLIERS[suffix]!` and a comment. |
| 5 | Between-unit mismatch only checked when both units set | **Resolved** | Guard is now unconditional (`lower.unit !== upper.unit`, case-insensitive); `V(OUT) between 3.25 and 3.35V` now throws, with a spec line and test to match. |

### Delta metrics
```
this commit: src/kicad/spice.ts +54/-18, test/spice.test.ts +196/-0, spec.md +2/-2
full PR (base b9a7de7 -> head 07e95b9): +811/-4 across 4 files
        · src +399/-0, test +408/-0, openspec +4/-4
tests: spice suite 25 -> 40 (40/40 pass, measured on branch)
diff coverage: 99.3% stmts / 94.4% branch of new src (measured, @vitest/coverage-v8)
        only uncovered: spice.ts:245-246 (toggleFence "different-char/shorter fence" branch)
mergeable: MERGEABLE · CI: cla + template checks passing
```
Method: numstat vs `git merge-base origin/main <head>`; coverage via `vitest run --coverage` on the branch; fence/comment/control payloads traced by hand against the compiled module.

### Invariant checks (re-performed, all pass)
- **check LLM-free / network-free**: PASS — `spice.ts` still has zero imports; nothing in `src/` imports it yet.
- **No sexp serialization**: PASS — Markdown strings only.
- **Spec-gated in / verification-gated out / ledger / secrets**: N/A — untouched.
- **Spec coherence**: PASS — delta spec updated for the fence/narrative and between-unit rules; `openspec validate add-spice-verification-gate` = valid.

### Residual (new, low — non-blocking)
- **[low] A narrative line shaped like `Word:` inside a `## Simulation` section is still consumed as a field.** `src/kicad/spice.ts:227` — the body filter admits any `^[A-Za-z]+\s*:` line, so a prose line such as `Note: keep the reference near the ADC.` parses as field `note` and throws `line 4: unknown Simulation field "note"` (traced by hand). Much narrower than the original finding (plain prose and fenced examples are now correctly ignored), and the error is line-numbered, so this is an edge case, not a defect on the intended path. Optional fix: treat a leading token that is not one of the known field names as narrative rather than an unknown-field error, or document that field-shaped lines are reserved inside the block.
- **[nit] One uncovered branch**: `toggleFence` line 245-246 (a fence marker of a different char / shorter length appearing while another fence is open) has no test. I verified it by hand — a `~~~` inside a ```` ``` ```` block is correctly treated as content — so this is a coverage nit, not a bug.

Nothing here blocks merge. Nice, thorough hardening pass.

PR #64, opened 2026-07-23
fix(create): harden the claude-code pipeline (drift gates, KiCad/git resilience, observability)

animesh-chouhan:
## What

Hardens the `claude-code` `copperhead create` pipeline end to end: fixes the drift
gate false-positives that trapped the finish loop, makes the KiCad/git commit path
survive KiCad 10 and rollbacks, adds symbol verification against the real KiCad
libraries, and adds whole-run observability (heartbeat, per-stage cost table,
printed resume point). This PR carries the two earlier pipeline commits plus the
new hardening commit as one change.

> Update: a follow-up commit addresses the automated `/pr-review` findings F1 to F6
> (restore the SPEC.md §1.3 structural edit-tool lock, gate resume-commit on
> ERC-clean, route the drift BOM read through header-name resolution, keep
> footprint case, harden the temp-dir sweep, and add retry/resume tests that
> surfaced and fixed an `isManagedPath` trailing-slash bug). See PIPELINE-ISSUES.md
> Issues 18 to 20.

## Why

Driving `copperhead create --brief brief.md --model claude-code` to completion on
the USB-C power breakout brief surfaced a batch of blockers and token sinks: a
correct schematic could not clear the finish gate (drift misread auxiliary tables,
positional columns, and byte-exact value encodings), the commit aborted on KiCad's
git-backed `.history/`, a rollback wiped the scaffold and cascaded the retry,
leaked temp dirs filled the disk, and a malformed tool call was silently dropped so
the model misdiagnosed a broken tool. Each is logged with root cause and fix in
[PIPELINE-ISSUES.md](PIPELINE-ISSUES.md) (Issues 6 to 17; the I1 to I13 identifiers
reference `../copperhead-test/run-logs/ISSUES-FOUND.md`).

## Design

Grouped by concern (full detail in [PIPELINE-ISSUES.md](PIPELINE-ISSUES.md)):

- **Drift gate correctness** ([drift.ts](src/memory/drift.ts),
  [bom-table.ts](src/memory/bom-table.ts)): `parseCanonicalTables` keeps only the
  Refdes/Pin contract table (ignores a quiescent-current roll-up or net legend);
  `parsePinoutRows` resolves Refdes/Pin/Net by header name, not a fixed index;
  `normalizeValue`/`foldEncodings` compare values and footprints semantically (fold
  `>=`, `ohm`, `u`, the inch mark, smart quotes, case, and spacing) so only a real
  value difference is flagged.
- **Symbol verification** ([symlib.ts](src/kicad/symlib.ts), `verify_symbols` in
  [tools.ts](src/agent/tools.ts)): reads the installed `.kicad_sym` libraries and
  reports pins that diverge from the canonical part plus lib_ids that do not exist
  in this KiCad version. Verification (not blind auto-splice) because KiCad renames
  symbols across versions. The tool is read-only (`requiresUnlock: false`).
- **KiCad/git resilience** ([git.ts](src/util/git.ts),
  [create.ts](src/commands/create.ts)): `ensureIgnored` appends `.history/` to
  `.gitignore` before every `git add -A`; the KiCad project is re-scaffolded per
  stage attempt so a rollback does not cascade; `sweepStaleTempDirs` reclaims leaked
  scratch dirs (age-gated); `assertDiskSpace` preflights free space.
- **Loop/provider** ([loop.ts](src/agent/loop.ts),
  [claude-code.ts](src/agent/providers/claude-code.ts),
  [types.ts](src/agent/types.ts)): `detectMalformedCall` returns a re-emit nudge via
  a new `Turn.nudge` when a turn names a catalog tool but parses no call; the turn
  timeout budget resets on a productive turn; a session/usage limit is detected as a
  schedulable pause; opt-in session resume sends only the new-message delta.
- **Observability** ([render.ts](src/agent/render.ts),
  [response-cache.ts](src/agent/response-cache.ts)): a periodic
  elapsed-plus-streamed heartbeat, a per-stage cost table with a TOTAL, and a
  printed resume command on any stop.

Safety-gate invariants are preserved: `verify_symbols` is read-only and the KiCad
parser stays read-only (symlib only reads `.kicad_sym`); the drift changes are pure
parsing and comparison reachable from `check` with no provider or network use; the
commit path still refuses while obligations are open and still rolls back to the git
snapshot on a failed verify.

## Spec / OpenSpec

No SPEC.md acceptance criteria change: these are bug fixes and robustness on the
existing create pipeline, plus one read-only inspection tool (`verify_symbols`) and
observability output. No `edit_file`/`write_file` gating, ERC/DRC gating, or
obligations semantics change. An OpenSpec change artifact for `verify_symbols` as a
first-class capability is a reasonable follow-up if the team wants it specced.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Build | `npm run build` | pass (exit 0) |
| Unit + offline | `npm test` | pass (288 passed, 14 skipped, 29 files) |
| Live-LLM (opt-in) | keyed on `ANTHROPIC_API_KEY` | not run (no key in this environment) |

New or changed tests: `test/corruption-guard.test.ts`, `test/symlib.test.ts`,
`test/tmp-sweep.test.ts`, `test/session-limit-disk.test.ts`,
`test/observability.test.ts`, plus additions to `test/bom-table.test.ts` (canonical
table isolation, header-name columns, semantic value equality) and
`test/claude-code-provider.test.ts` (malformed-call nudge). No known pre-existing
failures. `markdownlint-cli2 PIPELINE-ISSUES.md`: 0 errors.

- Sandbox variant used (`create` / `edit`): `create`
- Commands run and outcome:

These fixes were derived from full end-to-end `create` runs on the USB-C power
breakout brief, captured in `../copperhead-test/run-logs/create-run-03..06.log`,
which is where each Issue was observed. In this session I rebuilt `dist/` and
smoke-checked the built CLI; a fresh full end-to-end `create` on the fixed build
(live provider plus a local `kicad-cli`) is recommended before merge to confirm the
finish gate now clears on a correct schematic.

```text
$ npm run build
> tsc -p tsconfig.json   (exit 0)

$ node dist/cli.js --help
Usage: copperhead [options] [command]
Cursor for circuit boards: an AI agent for real KiCad repositories
...   (exit 0)

$ npm test
Test Files  27 passed | 1 skipped (28)
     Tests  284 passed | 14 skipped (298)
```

## Docs

- [PIPELINE-ISSUES.md](PIPELINE-ISSUES.md): new Issues 6 to 17 with root cause and
  fix for every change in this PR.
- Extensive doc comments added at each new function and tool (`verify_symbols`,
  `normalizeValue`/`foldEncodings`, `parsePinoutRows`, `ensureIgnored`,
  `sweepStaleTempDirs`, `sessionLimit`, `assertDiskSpace`, `detectMalformedCall`,
  the heartbeat and cost-table helpers).

---

## Invariant checklist

- [x] **Spec-gated in**: preserved structurally (corrected after review F1). The loop
  advertises `availableTools(ctx)`, so `edit_file`/`write_file` are omitted from the
  advertised tool list until a proposal validates, per SPEC.md §1.3; `dispatchTool`
  is the live gate on top. `verify_symbols` is a read-only inspection tool.
- [x] **Verification-gated out**: preserved. The commit/rollback path still ends in
  ERC (and DRC when the board changed) with repair then rollback to the snapshot;
  the I6/I7 fixes only make that path robust to KiCad `.history/` and a wiped
  scaffold.
- [x] **`check`/`verify` stays LLM-free and network-free**: the drift and bom-table
  changes reachable from `src/commands/check` are pure parsing and comparison; no
  provider, key, or network is introduced.
- [x] **No sexp serialization**: preserved. `src/kicad/symlib.ts` only reads
  `.kicad_sym`; KiCad files are still edited only via anchored text replace.
- [ ] **Sync-obligations ledger**: N/A, not touched.
- [ ] **Secrets**: N/A, redaction and `.gitignore` for `.env`/`.copperhead/runs/`
  unchanged (this PR adds `.history/` to project `.gitignore`, unrelated to secrets).
- [x] **Spec coherence**: no spec-level behavior changed, so SPEC.md and change
  artifacts did not need to move; see Spec / OpenSpec above.


animesh-chouhan:
## Goal that drove this work

From `run-logs/GOAL-drive-to-clean-run.md` in the companion `copperhead-test` repo: the objective was to drive the `create` pipeline to a clean full 8-stage run, fixing tool bugs and documenting only new findings.

<details>
<summary>GOAL-drive-to-clean-run.md</summary>

```markdown
GOAL: Drive copperhead's create pipeline to a clean, complete run, fixing tool
bugs as needed and documenting only NEW findings. Work in one continuous session
— carry what each attempt teaches into the next.

COMMAND (resumes past already-committed stages automatically):
  copperhead create --brief brief.md --model claude-code

Log each attempt to run-logs/create-run-NN.log (increment NN; currently at 02).

DONE = the create pipeline completes all 8 stages (spec-seed → architecture →
part-selection → schematic → … → final), exiting 0 with the stage committed.
It is currently wedged at stage 4 (schematic, 4/8). Nothing short of a clean
full run counts — keep going.

BEFORE THE FIRST RUN:
- Confirm the global `copperhead` resolves to the dev build
  (readlink -f "$(which copperhead)" → .../copperhead/dist/cli.js). If a stale
  global install shadows it, rebuild/relink first.

EACH ATTEMPT:
1. Run the command; capture full output to the log.
2. Succeeded → done, report.
3. Failed or hung:
   - Watch the known resource leak (I8) before rerunning: check /tmp temp-dir
     buildup and .history/ growth; clean up if it's filling the disk, else the
     next run fails for the wrong reason.
   - Diagnose from the log + .copperhead/runs/*/transcript.jsonl.
   - Apply the smallest fix that unblocks progress — in ../copperhead if it's a
     tool bug (rebuild after). Record what changed and why.
   - Rerun, carrying the diagnosis forward.

DOCUMENT — only NEW findings, in the existing files/format:
- I1–I9 in run-logs/ISSUES-FOUND.md are already written. Add new issues as I10+
  using the same legend (BLOCKER/DEFECT/INEFFICIENCY/NOTE) and fields
  (Where / Symptom / Suggested / Status). A recurrence of I1–I9 → reference it,
  don't duplicate.
- Add new efficiency/robustness ideas to run-logs/RECOMMENDATIONS.md using its
  P0–P3 scheme and section structure; cross-reference the issue number.
- Keep both grounded in actual logs/transcripts/source — no speculation.

STOP if: a clean full run (success), OR the same root cause recurs 3× with no
viable fix (report as a blocker and stop).

FINAL REPORT: outcome, which stages now pass, new issues (I10+) and
recommendations added, and any fixes applied to ../copperhead.

```

</details>

animesh-chouhan:
## Issues found (I1 to I13)

The grounded root-cause log this PR fixes, from `run-logs/ISSUES-FOUND.md` (companion `copperhead-test` repo). Cross-referenced from [PIPELINE-ISSUES.md](https://github.com/chouhanindustries/copperhead/blob/fix/claude-code-pipeline/PIPELINE-ISSUES.md) Issues 6 to 17.

<details>
<summary>ISSUES-FOUND.md</summary>

# Issues & inefficiencies — `copperhead create --brief brief.md --model claude-code`

Live run on the USB-C power breakout brief. Dev build linked via `npm link`
(global `copperhead` → `../copperhead/dist/cli.js`), model `claude-code`.
Full logs: `run-logs/create-run-01.log` and `.copperhead/runs/*`.

Legend: **BLOCKER** stalls the pipeline · **DEFECT** wrong/corrupt output that
still completes · **INEFFICIENCY** wasted time/tokens · **NOTE** observation.

Pipeline reached: stages 1–3 committed (spec-seed, architecture, part-selection);
stage 4 (schematic) reached ERC-clean + drift-clean with all 9 symbols, then
failed at the **commit** step (I6) and its retry cascaded (I7).

---

## I1 — Hung SDK turns cost ~5 min each (INEFFICIENCY, recovered)

**Where:** spec-seed turn 7; schematic turns 3 and 8 (multiple times each).
**Symptom:** `turn exceeded 300000ms; aborted the hung call and retrying (n/3)`.
The turn watchdog (`turnTimeoutMs`, default 300000) fired, tore down the hung
`claude-code` SDK subprocess, and retried. Recovery worked, but each hang is a
~5-min dead wait. Hangs cluster on **large-output turns** (schematic capture,
19k–45k out tokens).

**Notes:** `maxTurnTimeouts` (`loop.ts:289`) is a **cumulative budget of 3 for
the whole stage run**, not per-turn — a long stage that hangs on a few
independent big turns can exhaust it and hard-fail even though each hang was
transient. Two things kept the run alive: the response cache (a watchdog-
*abandoned* turn whose SDK call still finishes in the background writes its
result to `.copperhead/llm-cache/`, so the retry replays it free — observed
"llm-cache: replayed a cached response (hit #1)"), and stage-level retry
replaying completed turns free.

**Suggested:** shorter initial deadline + exponential backoff; make
`maxTurnTimeouts` per-turn (or reset on any productive turn).

**Mitigation (applied):** raised `turnTimeoutMs` default 300000 → **600000**
(`src/config.ts:48`). The largest observed capture turns (a full `lib_symbols`
+ instances edit, ~40k out tokens) legitimately run several minutes; the old
5-min deadline killed those mid-flight and, because the watchdog budget is spent
per stage, could fail a stage that was only slow rather than hung. 10 min clears
the largest observed turns while still catching a genuinely stuck subprocess.
The deeper backoff/per-turn-budget redesign is still open.

**Status:** not a bug; latency cost documented; timeout floor raised.

---

## I2 — Ω (ohm) corrupted to U+FFFD (`�`) in DECISIONS.md (DEFECT, upstream SDK)

**Where:** spec-seed `record_decision` → `docs/DECISIONS.md`.
**Symptom:** the CC-resistor decision reads `Rd = 5.1k� ± 5% to GND` on disk. The
brief calls the CC decision "the single thing most likely to be wrong," so a
mangled value in the exact doc it asks for is a real quality hit.

**Scope:** corrupted in `docs/DECISIONS.md`, the spec-seed `summary.md`, and both
transcripts (the `record_decision` arg). **Clean** in `docs/SPEC.md` (`5.1kΩ`,
bytes `ce a9`), `constraints.json`, and `openspec/specs/cc-termination/spec.md` —
all authored by `write_file` in the *same* run/provider.

**Root cause:** the transcript's tool-call arg already holds `ef bf bd` (U+FFFD)
where `ce a9` should be — it arrived corrupted *from the SDK*; copperhead's write
path is innocent. An identical Ω a few turns earlier (SPEC.md) survived intact —
the signature of a **UTF-8 multibyte char split across a streaming chunk boundary
and decoded per-chunk** inside `@anthropic-ai/claude-agent-sdk`. copperhead only
sees already-decoded strings, so it cannot losslessly recover it.

**Fix:** correct fix is upstream (stream-aware UTF-8 decode). In copperhead only
mitigations exist (reject a turn whose args contain U+FFFD and re-emit; or steer
machine-recorded fields to ASCII "5.1 kohm"). Both are workarounds.

**Mitigation (applied):** `corruptionError()` in `src/agent/tools.ts` rejects any
`write_file`/`edit_file`/`record_decision` call whose content-bearing args
contain U+FFFD, returning a message that tells the model to re-emit the value
(the corruption is nondeterministic — it depends where a stream chunk split — so
the retry almost always lands clean). U+FFFD never appears in a legitimate PCB
doc, so the guard has no false positives. Unit-tested in
`test/corruption-guard.test.ts`. The root decode bug remains upstream.

**Status:** upstream SDK defect; documented; in-copperhead guard applied.

---

## I3 — finish gate reveals unmet obligations one at a time (INEFFICIENCY)

**Where:** every stage's finish; sharp on schematic turns 8–9.
**Symptom:** `finish` → `Refusing to finish: 1 obligation(s) unmet` (names ONE)
→ fix → `finish` again → a *different* single unmet obligation → fix → `finish`…
Each attempt is a full turn (fresh subprocess + whole-history re-send on
claude-code), so a 3-obligation backlog burns ~3 turns.

**Suggested:** return **all** currently-unmet obligations in one response so the
model clears the backlog in a single turn. Compounds with per-turn full-history
re-send. Low-risk, high-leverage.

**Re-investigation:** the `finish` handler (`src/agent/tools.ts`) **already
aggregates** — it collects ERC/DRC status *and* every open obligation (except
`changelog`) into one `problems[]` and returns them together as "cannot finish
yet: …"; there is no loop/index/early-return that surfaces only one. The literal
string "Refusing to finish: N obligation(s) unmet" does not exist anywhere in
the source. The one-at-a-time *effect* observed live is real but has a different
cause: obligations become visible **as earlier ones clear** (you cannot get a
drift obligation until a schematic edit opens it; you cannot clear ERC until you
run it), so each finish attempt legitimately surfaces the next newly-open
obligation. That is a property of the do-loop workflow, not a truncation bug.

**Status:** not a code defect — current `finish` already returns all
simultaneously-open obligations at once. No change made. The residual cost is the
per-turn full-history re-send on claude-code (see I1), not the finish contract.

---

## I4 — schematic captured as full-block rewrites, not incremental (INEFFICIENCY)

**Where:** schematic turns 5–8.
**Symptom:** the model placed batch 1, hit a grid-alignment error, and **rewrote
the entire inserted block** to fix it — then again for a coordinate error, then a
`replace_all` across 10 occurrences. Cumulative out-tokens climbed
22k→35k→38k→45k→67k across these rewrites. One geometry slip re-emits everything.

**Suggested:** matches the repo's own PIPELINE-EFFICIENCY.md #8 — push
symbol-by-symbol capture with `run_erc` after each part so a mistake is local.
On claude-code (no cross-turn prompt cache) each rewrite is fully re-billed, so
this is the dominant cost driver of the whole pipeline.

**Fix (applied):** the schematic-stage prompt (`src/commands/create.ts`) now
directs the model to work **one part at a time** — add a symbol, `run_erc`, fix,
next — explicitly noting that small incremental edits keep a geometry/grid slip
local instead of forcing a full-block rewrite. (Prompt change only; behavioural
improvement, not enforced by a gate.)

**Status:** fixed in the stage prompt; effect to be confirmed in the fresh run.

---

## I5 — drift checker misreads *any* auxiliary table in BOM.md / PINOUT.md as parts/pins (DEFECT + INEFFICIENCY, fixable)

**Where:** `src/memory/drift.ts` via `src/memory/bom-table.ts:parseMarkdownTables`.
**Symptom (live):** schematic turns 9–10. Parts table already drift-clean, but
`check_drift` reported bogus `PINOUT.md: claims "Net:Meaning exists" but actual
is "Net:Meaning not in schematic"` and flagged BOM.md's quiescent-current
roll-up rows as parts "not in schematic". The model had to **convert its second
tables into bullet lists** to appease the checker — extra turns, zero design
change.

**Root cause:** `parseMarkdownTables(md)` parses **every** pipe-table in the doc
and concatenates all rows; `isHeader` only drops rows containing a literal
`refdes`/`pin` cell. A *second* table (`| Item | Current |`, `| Net | Meaning |`)
is neither a header nor recognised as non-parts data, so each of its rows is
treated as a part/pin absent from the schematic. BOM.md/PINOUT.md are thereby
silently constrained to exactly one table, which the prompt never states.

**Fix:** select only the canonical table — parse rows of the table whose header
matches the fixed contract (`Refdes | Value | Footprint | MPN | Rationale` for
BOM; `Pin`/`…Net` for PINOUT), ignore other tables. Unit-testable in isolation
(BOM.md with a second summary table → only real parts compared). `parseBomTable`
(used by `export bom`) benefits identically.

**Status:** reported; fix planned (safe to fix + test without a pipeline re-run).

---

## I6 — `git add -A` aborts on KiCad's git-backed `.history/` local-history dir (BLOCKER) — FIX APPLIED

**Where:** `src/util/git.ts` `commitAll` (and `preserveFailedRun`).
**Symptom:** schematic stage did all its work (ERC clean, drift clean, PINOUT
written, 9 symbols) then the run **failed at commit**:
```
run failed: commit failed: Command failed with exit code 128: git add -A
error: '.history/' does not have a commit checked out
fatal: adding files failed
commit-failed · ERC clean · 30m17s
```

**Root cause:** KiCad ≥9/10 writes a **git-backed local-history directory**
(`.history/`, with its own nested `.git`; the dir's own README says "KiCad Local
History Directory") into the project the first time `kicad-cli` touches it.
copperhead runs `kicad-cli` for ERC/DRC, so `.history/` appears mid-run; left
untracked it's a nested repo with an unborn HEAD, so `git add -A` in the parent
aborts (exit 128). This would break the commit step of **every** KiCad-touching
stage (schematic, layout, outputs) on any KiCad-10 machine — a general blocker,
not just this environment.

**Fix (applied):** added `ensureIgnored(repo, ['.history/'])` in `git.ts`,
called before every `git add -A` in `commitAll` and `preserveFailedRun`. It
appends `.history/` to the repo's root `.gitignore` (idempotent, best-effort,
never throws), so the nested history repo is skipped instead of aborting the
add. Local history is never a project artifact, so ignoring it is also correct.

**Status:** fixed in source (`src/util/git.ts`); pending rebuild + re-run to
confirm the commit succeeds.

---

## I7 — stage retry doesn't re-scaffold after a rollback wipes it (BLOCKER, cascade) — FIX APPLIED

**Where:** `src/commands/create.ts` `runCreate`.
**Symptom:** after I6's commit failure, the failure path (`restore()`:
`git reset --hard` + `git clean -fd`) deleted the **still-untracked scaffold**
(`.copperhead/config.json` + the empty `.kicad_sch/.kicad_pcb/.kicad_pro`). The
schematic bootstrap runs **once per stage** (outside the attempt loop), so the
retry ran against a missing config.json and cascaded — the model searched for a
schematic that no longer existed, spending turns on ENOENT errors.

**Fix (applied):** `bootstrapKicadProject` is now called at the top of **each
attempt** in the retry loop (`src/commands/create.ts:237-240`), not just once per
stage (idempotent — a no-op when the project exists; re-creates it after a
rollback). Note: fixing I6 alone prevents this cascade for the observed case (no
commit failure → no rollback → no wipe), but I7 is real defense-in-depth for any
*other* stage failure that rolls back.

**Status:** fixed in source (`src/commands/create.ts`); pending rebuild + re-run
to confirm a rolled-back retry re-scaffolds cleanly.

---

## I8 — copperhead leaks `/tmp` temp dirs and grows `.history/`, filling the disk (BLOCKER, resource leak)

**Where:** `src/kicad/cli.ts` (`mkdtemp(tmpdir(), 'copperhead-')` per ERC/DRC/
export), claude-code provider (`copperhead-cc-*` scratch cwds), + KiCad's
unbounded `.history/` git store.
**Symptom:** ~30 min into the run the whole filesystem hit **100%** (`/` at
100G/105G), and every command + file write began failing with `ENOSPC` — halting
the session. copperhead's kicad-cli wrapper creates a fresh `mkdtemp` per check
and removes it in a `finally`, but a watchdog-killed turn or an aborted stage can
skip that cleanup, so `/tmp/copperhead-*` dirs accumulate; KiCad's `.history/`
also grows on every project touch and is never pruned.

**Suggested:** (a) reuse a single per-run temp dir for kicad-cli instead of one
per call, and clean it in the run's outermost `finally`; (b) sweep stale
`copperhead-*` tmp dirs on startup; (c) since `.history/` is now gitignored
(I6), also prune/ cap it, or disable KiCad local history for headless runs.

**Fix (applied, part b):** `sweepStaleTempDirs()` in `src/util/tmp.ts`, called at
the top of `runCreate` (`src/commands/create.ts`), removes leaked `copperhead-*`
scratch dirs left in the OS temp dir by earlier runs whose `finally` was skipped.
It is **age-gated** (default 2h) so a concurrent run's fresh dirs are never
touched, and best-effort (never throws / never blocks a run). This breaks the
**cross-run accumulation** that actually filled the disk — a crashed run's leak
is reclaimed on the very next invocation. Covers all five prefixes
(`copperhead-`, `-validate-`, `-runs-`, `-cc-`, `-codex-`) in one sweep.
Tested in `test/tmp-sweep.test.ts`.

**Still open (parts a, c):** (a) reuse a single per-run kicad-cli temp dir
instead of one `mkdtemp` per ERC/DRC call (reduces intra-run churn; the happy
path already cleans up, so lower priority); (c) prune/cap KiCad's `.history/`
store (now gitignored per I6 but still grows on disk). Both documented for a
follow-up change.

**Status:** cross-run leak fixed + tested (part b); intra-run reuse and
`.history` pruning (parts a/c) still open.

---

## I9 — schematic symbols are LLM-authored, not pulled from KiCad's libraries (DEFECT-risk + INEFFICIENCY)

**Where:** `src/commands/create.ts:89` (schematic-stage prompt) → the model's
anchored `edit_file` writes into `usb-c-power-breakout.kicad_sch`.
**Symptom:** the stage prompt tells the model to *"add lib_symbols, symbols, and
connectivity by anchored edits"* — so the model **hand-draws the entire
`lib_symbols` block**: every pin, pin number, pin name, electrical type, and the
graphic geometry. In this run it authored 19 symbol definitions (7 top-level
lib symbols: `Connector:USB_C_Receptacle_USB2.0`, `Device:R`, `Device:C`,
`Device:Polyfuse`, `Device:LED`, `Connector:Screw_Terminal_01x02`,
`Connector:Conn_02x02_Odd_Even`, plus their unit sub-symbols).

**Why this is wrong-data risk:** every one of those lib_ids is a **real,
canonical KiCad standard-library part** (`Device:R`, `Connector:USB_C_...`), but
the geometry and pin table under it are **invented by the LLM**, not copied from
the installed `.kicad_sym` library. The lib_id therefore *claims* to be the
canonical part while the actual pins can silently diverge — a
`USB_C_Receptacle_USB2.0` with a wrong or missing shield/CC/SBU pin, a
`Polyfuse` with swapped pins, etc. ERC only checks the net graph *as drawn*, so a
symbol that is internally consistent but wrong vs. the real part passes clean.
This is the highest-consequence class of error: it looks authoritative and no
gate catches it.

**Why this is also inefficiency:** because the model is authoring geometry, it
also spends turns iterating on **grid alignment and pin coordinates** (the churn
in I4) — work it would never do if it were inserting a verified library symbol
and only placing instances + wiring. The two issues share a root cause: the model
is doing the symbol *librarian's* job as well as the *schematic capture* job.

**Root cause:** copperhead never resolves symbols from the KiCad symbol
libraries that ship with the `kicad-cli` it already invokes for ERC/DRC. There is
no tool to fetch a real `(symbol …)` definition by lib_id and insert it; the
bootstrap emits an empty `(lib_symbols)` (`src/kicad/bootstrap.ts:44`) and the
prompt delegates the rest to the model.

**Suggested:** add a symbol-resolution tool — given a lib_id, read the matching
`(symbol …)` block from the system symbol libraries (the `.kicad_sym` files whose
search paths `kicad-cli` knows) and splice the **verified** definition into
`lib_symbols`; restrict the model to *placing* instances and *wiring*. This kills
the divergence risk outright and removes the geometry-churn turns of I4 in one
change. Where a part has no stock-library symbol, fall back to authoring — but
that should be the exception, flagged, not the default path.

**Fix (applied — verification, not blind auto-splice):** added
`src/kicad/symlib.ts` and a `verify_symbols` tool (`src/agent/tools.ts`). It
reads the installed `.kicad_sym` libraries (env `KICAD*_SYMBOL_DIR` → standard
install dirs), resolves each schematic `lib_symbols` entry by its lib_id
(following `extends`), and reports pins that diverge from the real part (count,
name, electrical type) and lib_ids that don't exist in the installed KiCad (with
the closest real names). The stage prompt now tells the model to run it and
reconcile every finding. **Why verify rather than auto-splice:** KiCad renames
symbols across versions — run against KiCad 10 this exact test schematic's
`Connector:USB_C_Receptacle_USB2.0` **does not exist** (it's now
`…_USB2.0_14P`/`_16P`), so a blind by-lib_id splicer would fail on the most
important part of the brief. The checker surfaces the real name instead of
guessing; the 5 self-authored passives (R/C/LED/Polyfuse/Screw_Terminal) verify
clean, and only the two genuinely-wrong connector lib_ids are flagged. Tested in
`test/symlib.test.ts`; smoke-run against the live schematic confirms the drift.

**Still open:** the full auto-resolver (splice verified geometry, restrict the
model to placement) remains the ideal end state — `verify_symbols` closes the
**wrong-data** risk now and de-risks that follow-up by pinning down the
name-normalization problem first.

**Status:** verification tool applied + tested; full auto-splice resolver still a
documented follow-up.

---

## I10 — malformed fenced tool call is silently dropped, model misdiagnoses it as a tool bug (INEFFICIENCY + DEFECT-in-artifact)

**Where:** `src/agent/providers/claude-code.ts` — `parseToolCalls` / `scanJsonObject`
(the D1-tolerant fenced-JSON tool-call extractor). Live: run-03 spec-seed turns
12–16.

**Symptom:** the model emitted a batch `resolve_affected` call as a ```json
fenced block whose JSON was **one closing brace short** (`opens=7, closes=6`;
tail `…schematic."}]}` where the outer object needed `]}}`). The tolerant parser
could not brace-balance the outer `{"tool":…}` object, retried from the next
`{`, matched the inner `args` object (`{"resolutions":[…]}` — self-balanced but
has **no `tool` key**), `toToolCall` rejected it, and the turn dispatched **zero
tool calls with no error surfaced**. The model saw no result, retried the same
malformed batch once more (same silent no-op), then concluded *"the batch
`resolutions[]` form is a tool bug"*, fell back to single-item calls (which work
fine — the handler is correct, see below), and **wrote a false bug NOTE into the
committed stage summary**: *"resolve_affected batch resolutions[] form silently
no-ops."* That false claim is now permanent in `summary.md` / the git-committed
transcript.

**Not the handler:** `resolve_affected`'s batch branch (`src/agent/tools.ts:527`)
is correct — later in the same stage the model fired **4** `resolve_affected`
calls in one turn (native tool-call form) and all 4 resolved. The batch *form*
was never the problem; the two failed attempts were malformed JSON that never
reached the handler.

**Root cause (two compounding):** (1) the model produced unbalanced-brace JSON —
a nondeterministic model-output defect; (2) copperhead's tolerant parser (by D1
design: "unparseable output degrades to prose rather than throwing") gives the
model **no signal** that a clearly-intended tool call failed to parse. The
near-miss is machine-recognizable — a fenced ```json block / brace-run that
literally begins `{"tool":"<known-tool-name>"` but yields no dispatched call —
yet nothing nudges, so the model can't tell "my JSON was malformed" from "this
tool is broken," and it guesses wrong in a way that pollutes a committed artifact.

**Suggested:** in the extractor, detect the near-miss — a candidate span (or
fenced ```json region) that contains `"tool"` naming a tool in the catalog but
produces zero dispatched calls (unbalanced braces, or a parsed object missing
`tool`) — and return a one-line nudge to the model ("a tool call looks malformed
(likely unbalanced JSON / missing a closing brace); re-emit it") instead of
silently dropping it. Keeps D1's no-throw contract; adds a signal exactly where
the current silence misleads. Would have saved ~2 turns here and prevented the
false NOTE. Cross-ref RECOMMENDATIONS §5.5.

**Status:** documented; **not blocking** (model self-recovered via single-item
calls). Fix not applied — it is a parser-robustness improvement, not a pipeline
blocker, and the goal run was in progress; deferred to avoid destabilizing it.

---

## I11 — drift check does exact-string value compare, so equivalent encodings (≥/>=, inch "/in, µ/u) whack-a-mole the model (INEFFICIENCY, robustness)

**Where:** `src/memory/drift.ts:67` — `if (value !== undefined && value !== sym.value)`
compares the BOM.md `Value` cell against the schematic symbol's `value` property
with **raw `!==`**, no normalization. Live: run-03 schematic turns 5–8.

**Symptom:** after the schematic reached ERC-clean + verify_symbols-clean, the
finish gate's `check_drift` kept surfacing value mismatches that are
**semantically identical**, and the model burned turns editing one side then the
other, flipping the mismatch back and forth:
- `F1` value `Ihold≥3A, Vmax≥6V` (U+2265) vs `Ihold>=3A, Vmax>=6V` (ASCII) —
  flagged in **both directions** across turns (BOM has `≥`/sch has `>=`, then
  BOM has `>=`/sch has `≥`) as the model normalized whichever side drift last
  named.
- `J3` value `2x2 0.1" header` (inch double-quote) vs `2x2 0.1in header`.
- Compounded by **I2 mojibake**: on one write the schematic `≥` was mangled to
  `?`/U+FFFD (`Ihold?3A`), so the model chased a corruption *and* an encoding
  mismatch at once.

The model reached "all gates green" (its own `check_drift` returned clean), tried
to `finish`, and the finish-time re-check re-flagged a residual encoding
difference — several finish→fix→finish cycles, each a full claude-code turn
(fresh subprocess + whole-history resend, see I1), tens of k tokens of pure
bookkeeping churn on a design that was already electrically correct.

**Root cause:** the `value` field is free text with several equally-valid
renderings (`≥`/`>=`, `≤`/`<=`, `Ω`/`ohm`, `µ`/`u`/`μ`, `"`/`in`, smart quotes).
An exact `!==` treats each rendering as a distinct value, so BOM.md and the
`.kicad_sch` symbol property must be **byte-identical**, which the prompt never
states and which I2 can spontaneously break.

**Suggested:** normalize both operands before comparing — fold the common
equivalences (`≥`→`>=`, `≤`→`<=`, `Ω`→`ohm`, `µ`/`μ`→`u`, `”`/`“`/`"`→`"`, and
optionally collapse the inch mark) and compare case/space-insensitively; only
flag a *semantic* difference. Kills the whole whack-a-mole. Same normalization
should apply to the footprint compare (`drift.ts:70`) and `parseBomTable`. This
is adjacent to I5 (both live in the drift path) but distinct: I5 is *which table*
is parsed; I11 is *how a value is compared* once parsed.

**Status:** reported; fix planned (safe to fix + unit-test in isolation without a
pipeline re-run). Not blocking — the model can eventually force byte-identical
strings — but a large, avoidable token/turn sink. Cross-ref RECOMMENDATIONS §3.4.

---

## I12 — PINOUT drift reads the Net column by fixed index, so a 3-column pinout table false-flags every pin as "NC" → infinite finish loop (BLOCKER) — FIX APPLIED

**Where:** `src/memory/drift.ts:91` — `const [ref, pinNumber, , net] = row.cells`
read the net from **column index 3**. Live: run-03 schematic turns 9–22.

**Symptom:** the schematic reached ERC-clean + verify_symbols-clean (all 9
symbols matched the real KiCad library) with a correct `docs/PINOUT.md`, but
`check_drift` at the **finish gate** reported **29 mismatches, every one**
`PINOUT.md: claims "<ref>:<pin> net NC" but actual is "<ref>:<pin> net <real>"`
— e.g. `J1:A4 net NC` vs `J1:A4 net VBUS` — for a doc whose A4 row plainly reads
`VBUS`. The model could not reconcile a drift it could not see, rewrote
PINOUT.md between formats, and **looped finish→fix→finish from turn 9 to turn 22
(~163k output tokens, ~half the whole pipeline's budget) without ever
committing.** It would have run to the 40-turn cap and hard-failed the stage. The
model even narrated the confusion: *"the transcript has become a tangle of
phantom narration."* An apparent check_drift-clean vs finish-fails contradiction
seen in the log was just the model thrashing PINOUT.md between a pin table (29
false mismatches) and prose (no pin rows → clean) across turns.

**Root cause:** the scaffold's canonical PINOUT contract is **5 columns**
(`Refdes | Pin | Name | Net | Notes`, net at index 3 — `src/memory/scaffold.ts:43`),
and the drift reader hard-codes that index. But the pipeline never scaffolds a
PINOUT.md template nor states the contract in the stage prompt, so the model
authored a perfectly reasonable **3-column** `Refdes | Pin | Net` table (net at
index 2). `parseCanonicalRows` accepts it (its header has a Pin/Refdes cell), but
the destructure skips index 2 and reads index 3 = `undefined` → `claimed = 'NC'`,
which never matches the real net. Off-by-one against an unstated column contract.
The BOM path was clean (0 mismatches) — BOM.md happens to match its 5-column
scaffold — so PINOUT was the sole blocker.

**Fix (applied):** column resolution is now **by header name, not position**.
Added `parseCanonicalTables` (keeps each canonical table's header) and
`parsePinoutRows` (resolves Refdes/Pin/Net by matching the header cells, and
strips markdown backticks so `` `VBUS` `` compares equal to the schematic's plain
`VBUS`) in `src/memory/bom-table.ts`; `drift.ts` uses it. Now both
`Refdes|Pin|Net` and `Refdes|Pin|Name|Net|Notes` parse correctly, and a
supporting `Net|Role` legend table (no Pin column) is ignored. Verified against
the live on-disk schematic + PINOUT.md: **29 → 0 mismatches**. Unit-tested (4
cases) in `test/bom-table.test.ts`. Distinct from I5 (which is *which table* is
parsed); I12 is *how a column is located within the parsed table*.

**Status:** fixed in source + tested; drove the run's restart (below). Follow-up
worth doing: the schematic stage should either scaffold a PINOUT.md template or
state the column contract, and drift error text should say "PINOUT.md needs a
Net column" rather than a per-pin `NC` that misdirects. Cross-ref
RECOMMENDATIONS §3.5.

---

## I13 — claude-code session limit fails the run as a generic provider-error; recovery is cheap via llm-cache (NOTE, external + robustness)

**Where:** run-04 layout-draft stage (5/8), turn 9. Provider claude-code.

**Symptom:** mid-layout the run died with
`run failed: provider error: Claude Code returned an error result: You've hit
your session limit · resets 1:40pm (Asia/Kolkata)` and exited 1. This is the
Claude subscription's **usage/session cap**, an external time-based limit — not a
copperhead bug and not a defect in the design produced. The run had reached
layout turn 9 (~177k out tokens) with footprints placed and tracks routed.

**What copperhead did right:** `preserveFailedRun` stashed all uncommitted work
(`stash@{0}: copperhead failed run 2026-07-23T05-28-50-097Z`) before the tree was
cleaned — the schematic, partial PCB, PINOUT.md, and openspec proposals were
recoverable, not lost (I6 mechanism working). The completed turns were also
written to `.copperhead/llm-cache/` (54 entries), so **re-running after the reset
replays them at ~0 tokens** and resumes where it left off — the recovery is cheap
by design.

**Two robustness gaps this exposed:**
1. A session-limit error is **transient and time-scheduled** (it names the exact
   reset time), yet copperhead classifies it as a generic `provider-error` and
   hard-fails. It could instead (a) detect this specific error, (b) surface the
   reset time and the one-line resume command, and optionally (c) wait until the
   reset and auto-resume — since the llm-cache makes resumption nearly free. As
   shipped, the operator sees a generic error and must know to just re-run later.
2. The schematic stage had been **resumed-past uncommitted** (its correct work
   sat untracked on disk after I12 unblocked it), so the layout failure's
   working-tree cleanup wiped the schematic from the tree (recovered only via the
   stash). Resuming past a stage on `isComplete` should not leave prior work
   uncommitted and exposed to a later stage's rollback — either commit it when
   resuming past present-but-uncommitted work, or (better) commit each verified
   artifact as soon as it passes (RECOMMENDATIONS §2.1). On the next run the
   schematic file is absent, so the stage **re-runs and commits properly** — the
   durable path.

**Status:** external limit, documented as a NOTE; no code change. Recovery is to
re-run after the reset — the llm-cache replays completed turns. The two
robustness gaps are cross-referenced to RECOMMENDATIONS §2.1 / §2.4.


</details>

animesh-chouhan:
## Recommendations (P0 to P3)

The efficiency/robustness backlog from `run-logs/RECOMMENDATIONS.md` (companion `copperhead-test` repo). Items implemented here are noted in the PR; the rest are documented follow-ups.

<details>
<summary>RECOMMENDATIONS.md</summary>

# What can be made better / more efficient — `copperhead create` end-to-end

Analysis from driving `copperhead create --brief brief.md --model claude-code`
to completion on the USB-C power breakout brief (this session). Grounded in the
run logs (`run-logs/create-run-0*.log`), transcripts (`.copperhead/runs/*`), and
the source. Cross-references issues in `ISSUES-FOUND.md` (I1–I8).

Priorities: **P0** blocks or corrupts a run · **P1** large cost/latency win ·
**P2** robustness/ergonomics · **P3** polish.

## Implementation status (all landed in `../copperhead`)

Every recommendation below is now implemented and covered by the test suite
(`npm test` → 284 passing). Highlights of what changed, by area:

- **1.1** claude-code session resume — implemented as an **opt-in**
  (`COPPERHEAD_CC_SESSION_RESUME=1`, and only when the response cache is off,
  since the cache replays turns a resumed session never saw). Default behaviour
  (full-history resend) is unchanged, so the tested cache-recovery path is safe.
- **1.2** schematic prompt now states the net-label-≠-wire and 1.27 mm-grid rules.
  **1.3/1.5** already done. **1.4** protocol note: always send populated `args`.
- **2.1** a changelog/bookkeeping failure no longer rolls back the verified design
  commit. **2.2/4.1** a hung turn's `claude` subprocess is torn down via the SDK
  `abortController` (no more orphan). **2.3** already done. **2.4** a session/usage
  limit is a distinct `session-limit` exit path (reset time + resume hint), and a
  resumed stage's uncommitted work is committed so a later rollback can't wipe it.
- **3.1/3.3** already done. **3.2** `run_erc` warns on a zero-symbol schematic.
  **3.4** value/footprint/net compares fold encoding equivalences (≥/>=, Ω/ohm,
  µ/u, quotes, inch). **3.5** BOM columns resolve by header name; a PINOUT table
  with no Net column gets one explicit message instead of per-pin false NCs.
- **4.1** disk-space preflight (`COPPERHEAD_MIN_FREE_MB`), bounded `.history/`
  prune, plus the subprocess kill above and the pre-existing tmp sweep.
- **5.1/5.2/5.3/5.5** already done. **5.4** each KiCad stage renders sch+pcb SVG
  into its run dir's `artifacts/`. **5.6** cumulative live totals per stage end +
  an end-of-run `.copperhead/runs/REPORT.md` and `report.json`.

---

## 1. Token / cost efficiency

### 1.1 [P1] Stop re-sending the whole conversation every turn (claude-code)
The claude-code provider spawns a fresh SDK subprocess per turn (`maxTurns: 1`)
and flattens the **entire** prior conversation into the prompt each call
(`renderConversation`). Cost is ~quadratic in turns. The schematic stage spent
**67k output tokens over ~10 turns**, most of it re-billed history.
- **Fix:** use the Agent SDK's session resume (continue one session across
  turns) instead of replaying history. This is the single biggest lever for
  claude-code. The keyed `AnthropicProvider` already solves the equivalent
  problem with 3 `cache_control: ephemeral` breakpoints (0.1× reads) — claude-code
  has no such hook because copperhead doesn't own the API call.
- **Corollary:** for dollar-cost-sensitive runs, `--model claude` (prompt caching
  already active) is materially cheaper than `claude-code` today.

### 1.2 [P1] Incremental schematic capture instead of full-block rewrites (I4)
The model placed a batch, hit a grid error, and **re-emitted the entire block**
three times (22k→35k→38k→45k out tokens). One geometry slip re-bills everything.
- **Fix:** stage prompt should mandate symbol-by-symbol capture with `run_erc`
  after each part, so a mistake is local and cheap. Pair with a KiCad in-prompt
  note: *a net label on a pin names the net but is not a connection without a
  wire* (the run burned an ERC cycle rediscovering this), and *place component
  origins on the 1.27 mm grid* (the run burned 2–3 turns on off-grid pins).

### 1.3 [P1] Return ALL unmet obligations from `finish` at once (I3)
`finish` reveals one unmet obligation per call, forcing finish→fix→finish
ping-pong; each attempt is a full turn (subprocess + history resend). A
3-obligation backlog ≈ 3 wasted turns per stage.
- **Fix:** have `finish` return the complete list of currently-unmet obligations
  so the model clears them in one turn.

### 1.4 [P2] Trim the ceremonial empty-args first call
The model routinely opens a stage with `{"tool":"read_file"}` (no args) and eats
an error+turn. A one-line protocol note ("always send a populated `args` object")
removes a round-trip per stage.

### 1.5 [P2] Cache survives, but make replay deterministic by design
The llm-cache (`.copperhead/llm-cache/`) is what made recovery cheap this session
— on the resumed run, the whole schematic stage replayed at **0 tokens** (61 cache
hits). Its value depends on byte-identical conversation prefixes. Bootstrap
already uses deterministic UUIDs (good). Keep that discipline everywhere (no
timestamps/random in anything fed back into the prompt) or cache hit-rate silently
degrades.

---

## 2. Reliability / resilience

### 2.1 [P0] Never let a peripheral failure destroy completed design work (I6+I7)
The schematic stage finished ERC-clean, then the **commit** aborted on KiCad's
git-backed `.history/` (I6), and the rollback's `git clean -fd` deleted the
uncommitted populated files (I7). Two fixes applied this session; the general
principle:
- Git operations must be robust to nested repos / tool-created dirs
  (`ensureIgnored` — done).
- Recovery scaffolding must be re-established per attempt, not once per stage
  (per-attempt `bootstrapKicadProject` — done).
- **Further:** commit the verified design artifact *as soon as ERC/DRC passes*,
  before the doc/obligation bookkeeping, so the expensive KiCad work is durable
  the instant it's correct and only cheap bookkeeping is ever at rollback risk.

### 2.2 [P1] Watchdog: per-turn budget + adaptive deadline (I1)
`maxTurnTimeouts` is a **cumulative** budget of 3 for the whole stage; sporadic
hangs on independent large turns accumulate into a hard failure. And the flat
5-min deadline means every hang is a 5-min dead wait.
- **Fix:** make the timeout budget per-turn (or reset on any productive turn);
  use a shorter initial deadline with backoff; detect subprocess exit / stalled
  output earlier than the wall-clock deadline. On claude-code specifically, an
  abandoned turn leaves an **orphaned `claude` subprocess still running** — kill
  the process group on timeout, don't just drop the promise (also fixes I8).

### 2.3 [P2] Surface the preserved-work stash on failure
`preserveFailedRun` stashes uncommitted work before rollback but says nothing.
The failure message should name it: *"work saved in stash `copperhead failed run
<id>` — re-run to resume, or `git stash apply <sha>`."* Right now the safety net
is invisible.

---

### 2.4 [P2] Treat a session/usage limit as a schedulable pause, not a hard error (I13)
run-04 died mid-layout with `provider error: ... You've hit your session limit ·
resets 1:40pm`, exiting 1 like any other provider failure. But this error is
**transient and names its own reset time**, and the completed turns are already
in `.copperhead/llm-cache/`, so re-running after the reset replays them at ~0
tokens and resumes in place.
- **Fix:** detect the session-limit result distinctly (it is not a 429 and not a
  code bug); print the reset time + the exact resume command (§5.3); optionally
  sleep until the reset and auto-resume. Also: when resuming past a stage via
  `isComplete` whose work is present-but-uncommitted, commit it — otherwise a
  later stage's rollback wipes it from the tree (recovered only via the failure
  stash this run). Pairs with §2.1 (commit verified artifacts eagerly).

## 3. Correctness / gates

### 3.1 [P1] Drift checker must target the canonical table only (I5)
Any second markdown table in BOM.md/PINOUT.md (a quiescent roll-up, a summary)
has its rows misread as parts/pins "not in schematic," teaching the model to
degrade good docs into bullet lists to dodge the parser. Select rows only from
the table whose header matches the fixed contract; ignore other tables.

### 3.4 [P1] Normalize equivalent value encodings before drift-comparing (I11)
`check_drift` compares the BOM.md value cell against the schematic symbol value
with raw `!==` (`drift.ts:67`). Semantically-identical renderings — `≥`/`>=`,
`≤`/`<=`, `Ω`/`ohm`, `µ`/`μ`/`u`, the inch mark `"`/`in`, smart quotes — are
flagged as drift, so the two artifacts must be byte-identical. This session the
schematic reached ERC-clean + verify-clean, then the model spent ~4 turns
(tens of k tokens) flip-flopping `F1`'s `Ihold≥3A` vs `Ihold>=3A` and `J3`'s
`0.1"` vs `0.1in` across BOM.md and the `.kicad_sch` — a design already correct,
churning on encoding. I2 mojibake made it worse by mangling `≥`→`?` on one write.
- **Fix:** fold the common equivalences on both operands and compare
  case/space-insensitively; flag only a *semantic* value difference. Apply to the
  footprint compare and `parseBomTable` too. Removes a large, avoidable
  finish→fix→finish token sink with zero design downside.

### 3.5 [P0] Resolve doc-table columns by header name, and scaffold/state the PINOUT contract (I12)
`check_drift` read the PINOUT Net column by a **fixed index (3)** matching the
5-column scaffold (`Refdes|Pin|Name|Net|Notes`). The model wrote a reasonable
3-column `Refdes|Pin|Net`, so every pin read net=`undefined`=`NC` and drift
reported 29 false mismatches on a correct doc — an **unwinnable finish loop that
burned ~163k tokens (half the pipeline) and would have hard-failed the stage.**
- **Fix (done):** resolve Refdes/Pin/Net by header name (`parsePinoutRows`),
  tolerant of the optional Name/Notes columns, stripping backticks. 29→0.
- **Also do:** (a) the schematic stage should **scaffold a PINOUT.md template**
  with the canonical header (like part-selection seeds BOM.md) or state the
  column contract in the prompt, so the doc format is never guessed; (b) when a
  doc table is missing an expected column, the drift message should say so
  ("PINOUT.md has no Net column") instead of emitting a per-pin `NC` mismatch
  that misdirects the model into rewriting correct data. General principle:
  never compare against a positional column contract the author was never told.

### 3.2 [P2] ERC "clean" on a zero-symbol schematic is a false green
An empty schematic passes ERC (0 violations) and empty-sheet drift, so a
premature `finish` looks verified. The stage contract already checks symbols>0;
`run_erc` itself should additionally warn on a zero-symbol schematic so no gate
can be satisfied by the empty starting state.

### 3.3 [P2] Guard machine-recorded fields against mojibake (I2)
A UTF-8 multibyte char (Ω) split across an SDK stream chunk became U+FFFD in
DECISIONS.md while surviving in write_file docs. Root fix is upstream, but
copperhead could detect U+FFFD in tool-call args and reject/re-emit the turn, or
prefer ASCII in machine-recorded fields.

---

## 4. Resource management

### 4.1 [P0] Bound temp + local-history growth (I8)
A 30-min run filled a 105 GB disk to 100%, halting the session. kicad-cli temp
dirs are cleaned per-call, but watchdog-orphaned SDK subprocesses keep writing to
`copperhead-cc-*` cwds, and KiCad's `.history/` grows unbounded on every project
touch.
- **Fix:** kill orphaned subprocess groups on timeout (see 2.2); reuse one temp
  dir per run rather than one per kicad-cli call; sweep stale `copperhead-*` tmp
  dirs at startup; prune/cap `.history/` (now gitignored) or disable KiCad local
  history for headless runs (`--define` / env, if kicad-cli supports it).
- **Add a preflight disk check:** refuse to start (or warn) when free space is
  below a threshold, with an actionable message, instead of failing mid-run with
  an opaque `ENOSPC`.

---

## 5. Observability / ergonomics

### 5.1 [P2] Distinguish "slow" from "hung" in the status line
A 44k-char turn looks identical to a hang for up to 5 minutes. Emit a heartbeat
(tokens streamed / elapsed) so an operator can tell a working large turn from a
dead one.

### 5.2 [P3] Per-stage cost summary at the end
The run prints per-stage wall/turns/tokens inline. A final table (stage → wall,
turns, out-tokens, cache-hit%) would make the expensive stages obvious and track
the effect of the fixes above across runs.

### 5.5 [P2] Nudge on a malformed/near-miss tool call instead of silent-dropping it (I10)
The claude-code tool-call extractor is tolerant by design (D1): unparseable
output degrades to prose with no dispatched call. But when the model emits a
fenced ```json block that clearly *intends* a call — it literally begins
`{"tool":"<known-tool>"` — and the JSON is malformed (this session: a batch
`resolve_affected` call one closing brace short, `opens=7/closes=6`), the parser
silently drops it. The model gets **no signal**, misreads "no result" as "this
tool is broken," and — worst — wrote that false conclusion into the committed
stage summary ("resolve_affected batch silently no-ops"), which is simply wrong
(the handler is correct; the JSON was malformed).
- **Fix:** when a brace-run / fenced ```json region contains `"tool":"<name>"`
  for a catalog tool but yields zero dispatched calls (unbalanced braces, or a
  parsed object missing `tool`), return a one-line nudge ("a tool call looks
  malformed — likely unbalanced JSON / missing a closing brace; re-emit it").
  Preserves the no-throw contract; adds a signal precisely where silence
  currently misleads. Saves the wasted retry turns and keeps false bug reports
  out of permanent artifacts.

### 5.6 [P1] Whole-run observability: total time/tokens + an end-of-run per-step report
Each `copperhead create` invocation advances exactly one stage and writes a rich
`.copperhead/runs/<id>/summary.md` for *that stage* (Duration, Tokens in/out,
Turns, per-turn `2/4717` breakdown — all already present). But a design is built
over **8 separate invocations**, and nothing aggregates them: there is no single
place that says what the *whole* pipeline cost. An operator wanting "how long / how
many tokens did this board take, and which stage dominated?" has to hand-collect 8
`summary.md` files and add them up. §5.2 asked for a final per-stage table; this is
that, made durable and pipeline-wide.
- **Fix (two parts):**
  1. **Live totals.** At the end of every stage, print a running cumulative line —
     `pipeline so far: 5/8 stages · 41m12s · 112.4k out tokens · 63% cache hits` —
     computed by summing the sibling `summary.md` files under `.copperhead/runs/`
     for this repo/board. Cheap, deterministic, no extra model calls. On the final
     stage it becomes the grand total.
  2. **End-of-run report artifact.** When the pipeline reaches the last stage (or on
     demand via `copperhead report`), write `.copperhead/runs/REPORT.md` (and a
     machine-readable `report.json`) with one row per stage:

     | stage | wall | turns | in | out | cache hit% | repairs | outcome |
     |---|---|---|---|---|---|---|---|
     | spec-seed | 4m01s | 7 | 12 | 5.4k | 0% | 0 | ok |
     | … | … | … | … | … | … | … | … |
     | schematic | 18m | 12 | 15 | 67k | 51% | 1 | ok |
     | **total** | **1h07m** | **58** | **…** | **214k** | **51%** | **2** | **ok** |

     plus a slowest-stage / most-expensive-stage callout so the bottleneck is
     obvious at a glance. Source every field from the existing `summary.md` stats
     (Duration, Turns, `Tokens: N in / N out`, repair cycles, cache-hit counts from
     the `llm-cache: replayed …` log lines) — no new instrumentation needed, just
     aggregation.
- **Why P1, not P3:** this is the only artifact that makes the effect of §1.1/§1.2
  (the big token levers) *measurable across runs* — without a per-stage roll-up you
  cannot tell whether a fix actually moved cost. It turns the efficiency work from
  anecdote into a tracked number. Keep the `report.json` schema stable so successive
  boards' costs can be diffed/trended over time.

### 5.3 [P2] Log the exact resume point on any stop
When a stage stops for a human, print the single command to resume and which
stage it will resume at, so the operator never guesses.

### 5.4 [P1] Render schematic + PCB screenshots as run artifacts, and vision-check them (I7, 3.2)
Every text/ERC/drift gate can be satisfied by a design that is visibly wrong or
even **empty**, and nothing in the run ever *looks* at the board. Rendering the
schematic and layout to an image per stage closes that gap cheaply.
- **Concrete evidence, this repo:** the committed `usb-c-power-breakout.kicad_sch`
  on disk holds only `lib_symbols` *definitions* and **zero placed instances**;
  `usb-c-power-breakout.kicad_pcb` is a bare 30×20 mm `Edge.Cuts` rectangle with
  no footprints or nets. Both render as an empty A4 title block. This is the
  post-rollback scaffold from I7 (the populated 9-symbol schematic was wiped by
  `git clean -fd`), yet ERC is "clean" and drift is "clean" on it — the exact
  false green of 3.2. A one-line screenshot check reveals it instantly; no text
  gate does.
- **Fix:** after each KiCad stage, emit `kicad-cli sch export svg` and
  `kicad-cli pcb export svg` (or `pdf`) into the run dir
  (`.copperhead/runs/<id>/artifacts/`), and reference them from `summary.md`.
  This is fast, deterministic, and needs no extra tokens.
- **Stronger (optional):** feed those images to a vision pass (the model already
  in the loop) as a final per-stage acceptance check — "does this schematic show
  the parts BOM.md claims, wired as PINOUT.md claims?" Catches
  symbol-count/label/connectivity errors that pin-level text gates miss, and is
  the natural home for the "analyze the render for correctness" step. Gate it
  behind symbols>0 so an empty sheet fails *before* the vision call.

---

## Summary of highest-leverage changes

1. **Session resume for claude-code** (1.1) — kills the quadratic history resend.
2. **Incremental schematic capture** (1.2) — the dominant cost driver.
3. **Commit verified artifacts before bookkeeping** (2.1) — makes expensive work
   durable the instant it's correct.
4. **Bound temp/history + preflight disk check** (4.1) — prevents the hard
   ENOSPC halt.
5. **Canonical-table drift matching** (3.1) — stops the checker forcing doc
   degradation.


</details>

animesh-chouhan:
🤖 **Automated pr-review pass** (not a human review). Generated by the `/pr-review` skill.

## Verdict: request changes

Strong, well documented hardening work with genuine test coverage for most of the new units. One change silently alters a load-bearing repo invariant (and the PR's own checklist misreports it), and one new resume path can commit an ERC-unverified schematic. Both should be resolved or explicitly signed off before merge; the rest are low-severity.

## Metrics

```
lines: +4257 / -114 (net +4143) across 41 files
  · src +2072/-111 · test +859/-3 · docs +529 · spec +0 · run-logs +797
  (method: git diff --numstat BASE...head, BASE=b9a7de7 = merge-base with origin/main)
new vs net: 22 added files; 5 new src files (recovery, response-cache, bootstrap, symlib, tmp ≈ 773 new src lines); rest are run-logs/docs/tests
tests: +859 test lines; suite (head, measured `npm test`): 284 passed / 14 skipped / 0 fail
  base suite not re-run (offline); PR body claims all green, no pre-existing failures
diff coverage: MANUAL mapping (coverage-v8 not run). Most new exported symbols have unit tests
  (symlib, tmp, recovery+withTimeout+parseDiagnosis, CachingProvider, sessionLimit, assertDiskSpace,
   corruptionError, bom-table parsers, scanJsonObject/detectMalformedCall).
  Uncovered: git.ts ensureIgnored; loop.ts timeout-retry branch; loop.ts session-limit exit path;
   create.ts retry/resume/re-scaffold/writeRunReport branches (see F3).
CI: `gh pr checks 64` could not be fetched (network error to api.github.com) — NOT measured.
mergeable: MERGEABLE (per gh pr view).
```

## Invariant checks performed

- **Verification-gated-out**: primary path confirmed intact. A stage advances only on `stageDone` = `outcome==='success'` AND `isComplete()`, and `finish` refuses unless `lastErc.ok` (+`lastDrc.ok` when the board changed). See F2 for a resume-path exception.
- **No sexp serialization**: confirmed. `symlib.ts` only reads `.kicad_sym`; KiCad files still edited via anchored text replace only.
- **check/verify LLM-free and network-free**: confirmed. The drift/bom-table changes reachable from `check` are pure parsing/comparison; no provider, key, or network introduced.
- **Sync-obligations ledger**: unchanged; commit still refuses on open obligations.
- **Secrets**: no secret value committed. The one match in `pipeline-run-logs/transcripts/…18-47…jsonl` is a search result showing the variable name `CLAUDE_CODE_OAUTH_TOKEN=` with an empty value, not a token. `.gitignore` for `.env`/`.copperhead/runs/` unchanged.
- **Spec-gated-in**: **VIOLATED** by this PR, see F1.

## Findings

### F1 (high) — `edit_file`/`write_file` are now advertised in the tool catalog every turn, contradicting SPEC.md §1.3

[src/agent/loop.ts](src/agent/loop.ts) (the `const tools = TOOLS.map((t) => t.schema)` change, was `availableTools(ctx).map(...)`).

SPEC.md §1.3 invariant 1 is explicit: the edit tools are "hard-locked, not prompt-discouraged, but **unavailable in the tool list** until the proposal validates." This PR advertises every tool (including `edit_file`/`write_file`) to the model on every turn, and re-gates them by (a) tool-description prose ("Requires a validated change proposal first…") and (b) a runtime `dispatchTool` check. The review skill's step-3 rubric names this exact anti-pattern: "Reject any change that exposes mutation tools unconditionally or gates them by prompt text instead of by omission."

- **What is still safe**: `dispatchTool` → `availableTools(ctx).find(...)` still returns "tool not available (edit tools unlock after the proposal validates)" for a locked edit, so no mutation actually executes before unlock. The security property (no edit before a validated proposal) is preserved.
- **What is wrong**: the *mechanism* the spec mandates (structural omission from the advertised list) is gone, SPEC.md §1.3 was not updated, no OpenSpec artifact records the change (`spec +0`), and the PR's invariant checklist states "Spec-gated in: unchanged … edit_file/write_file gating is untouched by this PR", which is inaccurate.
- **Why the change was made** (legitimate): a turn that unlocked (`validate_change`) and edited in the *same* reply had its edit dropped, because `toToolCall` rejects any call whose name is not in the advertised catalog, and the catalog was computed before the turn. That is a real bug worth fixing.
- **Fix**: either (a) restore structural gating and instead let a same-turn `validate_change` re-expand the catalog before parsing later calls in that reply (or advertise edit tools only once unlock is pending), or (b) if the team accepts advertise-always + dispatch-gate as the new model, update SPEC.md §1.3 and add an OpenSpec delta so code and spec move together, and correct the PR checklist.

### F2 (medium) — `commitResumedStage` commits an ERC-unverified schematic on resume and advances the pipeline

[src/commands/create.ts:186-203](src/commands/create.ts#L186-L203) (new in this PR) together with schematic `isComplete` at [create.ts:78-92](src/commands/create.ts#L78-L92).

Schematic completion is judged by *symbols exist* + *BOM/PINOUT drift-clean*; it never runs ERC (the ERC-blindness predates this PR). New in this PR: when a resumed stage is `isComplete` but dirty, `commitResumedStage` commits it (`commitAll`, no ERC re-run) so a later rollback can't wipe it.

- **Failure scenario**: a prior invocation is hard-killed (watchdog SIGKILL, session limit, power loss) after symbols are placed and BOM/PINOUT drift has gone clean, but while ERC is still failing (unconnected pins mid-capture) and before `finish`/commit. On resume, `isComplete` returns true, `commitResumedStage` commits the ERC-failing schematic, and the pipeline advances to layout against unverified work. This is the verification-gated-out invariant being bypassed on the resume path.
- **Fix**: run `run_erc` inside `commitResumedStage` (or fold an ERC-ok check into the schematic `isComplete`) before trusting/committing resumed schematic work; refuse to commit-and-advance if ERC does not pass.

### F3 (medium) — the new retry/resume/re-scaffold branches are essentially untested

`test/create-stage-turns.test.ts` covers only the doc-stage happy path plus a single schematic halt where `diagnose()` aborts because the un-mocked OpenAI provider throws for lack of a key (a fragile implicit dependency). Not exercised anywhere:

- a `'retry'` verdict driving a successful second attempt (the whole auto-retry loop body, [create.ts:559-643](src/commands/create.ts#L559-L643));
- `commitResumedStage` — neither the managed-commit branch nor the foreign-path bail-out (this is F2's path and the "don't sweep user WIP" guard);
- re-scaffold-after-rollback inside the retry loop;
- `writeRunReport` content / `report.json` schema;
- cost accumulation across multiple attempts of one stage;
- [git.ts](src/util/git.ts) `ensureIgnored` (the `.history/` gitignore fix, core to the I6/I7 reliability claim);
- the loop's timeout-retry branch and session-limit exit path (`sessionLimit()` is unit-tested; the loop wiring that turns it into a `session-limit` exit is not).

These are the highest-risk new paths (control flow with `turn--`, commits, rollbacks). Add a scripted-provider integration test that drives one retry to success and one `commitResumedStage` of a managed-only dirty tree.

### F4 (low) — 2h stale-temp cutoff can delete a live concurrent run's scratch dir

[src/util/tmp.ts:71-108](src/util/tmp.ts#L71) + [create.ts:519](src/commands/create.ts#L519). `sweepStaleTempDirs` removes any `copperhead-*` temp dir older than 2h, on the stated assumption that "a run is minutes, not hours." This PR raises `turnTimeoutMs` to 10 min and adds per-stage retries across 8 stages, so multi-hour runs are now realistic. A second `copperhead` process starting while a first has run >2h could `rm -rf` the first's live provider scratch dir (`copperhead-cc-*`, created once and reused). Narrow (concurrent runs + >2h + stale mtime), but the safety comment no longer matches the configured timeouts. Consider a lock/pid marker or an "in-use" sentinel rather than a pure age heuristic.

### F5 (low) — drift BOM check reads columns positionally while `parseBomTable` resolves by header name

[src/memory/drift.ts](src/memory/drift.ts) still does `const [ref, value, footprint] = row.cells` on `parseCanonicalRows`, whereas [bom-table.ts](src/memory/bom-table.ts) `parseBomTable` now resolves by header name. A BOM.md that reorders columns is read correctly by `export bom` but produces false drift in `check`. Narrow because the scaffold fixes BOM column order, but the two readers should agree; route the drift BOM read through header-name resolution too.

### F6 (low) — a few minor/observability items

- `normalizeValue` lowercases footprints, so a case-only footprint difference is silently not flagged as drift ([bom-table.ts](src/memory/bom-table.ts)). Footprint library refs are case-sensitive; consider not lowercasing the footprint comparison.
- `CachingProvider.keyFor` hashes `this.name` (the provider family, e.g. `claude-code`), not the concrete model id ([response-cache.ts](src/agent/response-cache.ts)). Re-running the same repo after switching `claude-code:opus` → `claude-code:sonnet` replays the other model's cached turns. Cache is disposable, so low, but include the resolved model id in the key.
- Per-stage cost table under-reports: each retry's `diagnose()` makes a real model call whose tokens are never folded into the stage cost ([create.ts](src/commands/create.ts)). Observability only.
- `resumeCommand` emits an absolute `--repo` but keeps the original (possibly relative) `--brief` path; resuming from a different cwd would miss it.
- Pre-existing (not this diff, but multiplied by the new retry loop): `restore()` runs `git clean -fd`, which deletes untracked user files that `git stash create` never captured. Each of the 8 stages' failures is now a rollback opportunity. Worth a follow-up to snapshot untracked state.

## Things confirmed correct (so a clean bill is distinguishable from an unexamined one)

- The tool-call parser rewrite (`scanJsonObject`) is string-aware and correctly keeps a `write_file` call intact when its `content` holds nested ``` fences: the exact adversarial "format inside the format" case, and it is tested. The malformed-near-miss nudge is bounded by the consecutive-stall counter (`nudges = 0` resets on any tool call, AC-15.27 preserved).
- `bootstrapKicadProject` is a no-op when the configured schematic exists and only writes empty files guarded by `!existsSync`, so per-attempt re-scaffold cannot overwrite a committed stage's work.
- Auto-retry is bounded (`attempt > maxStageRetries`, forced non-negative int in config); a `'retry'` verdict never bypasses verification; diagnosis fails safe to `abort` on any error/timeout.
- Cost table has no double-counting (single emission per stop; all-resumed case guarded against NaN); the resume command's arguments match the CLI surface.
- Session-resume is off by default and mutually exclusive with the cache (guarded in loop.ts), and the `sentCount` high-water mark is not advanced on a thrown turn.
- Disk-space threshold parsing is safe (unset → default, `=0` disables cleanly, unknown platform skips).

animesh-chouhan:
Thanks for the review. All six findings addressed in `460bab2` (detail in PIPELINE-ISSUES.md Issues 18 to 20):

- **F1 (high)** — Restored the SPEC.md §1.3 structural lock: the loop advertises `availableTools(ctx)` again, so `edit_file`/`write_file` are omitted from the advertised list until a proposal validates; `dispatchTool` stays the live gate as defense in depth. Also corrected the invariant checklist, which had misreported this. Chose to honor the spec rather than weaken it (the same-turn batch was worth one turn, not the structural guarantee).
- **F2 (medium)** — Schematic `isComplete` now also requires `runErc(...).ok`, so a resume can never treat an ERC-failing schematic as complete and `commitResumedStage` never commits unverified work.
- **F3 (medium)** — Added `test/create-resilience.test.ts`: one retry-to-success (guidance prepended, cost accumulated across attempts) and both `commitResumedStage` branches. This surfaced a real bug: `config.docs` defaults to `docs/`, so `isManagedPath` built `docs//` and read every doc as **foreign**, meaning `commitResumedStage` never actually committed its own work. Fixed by normalizing the trailing slash.
- **F4 (low)** — The reused claude-code provider cwd is now `utimes`-touched each turn, so a multi-hour run's only long-lived temp dir never goes stale and a concurrent run's age-gated sweep cannot delete it.
- **F5 (low)** — Drift's BOM read now routes through `parseBomTable` (header-name resolution), so it and `export bom` agree on a reordered table.
- **F6 (low)** — Footprint compare keeps case (`normalizeFootprint`); the cache key includes the resolved model id; the diagnosis call's tokens fold into the stage cost; the resume command uses an absolute `--brief` path.

The one pre-existing item (rollback's `git clean -fd` can delete untracked user files) is left for a dedicated follow-up as suggested.

animesh-chouhan:
🤖 **Automated pr-review pass (follow-up)** (not a human review). Generated by the `/pr-review` skill.

## Verdict: approve

Follow-up to my [earlier pass](https://github.com/chouhanindustries/copperhead/pull/64#issuecomment-5062371787). This reviews only the new commit `460bab2` ("address pr-review F1-F6"). All six findings from the prior pass are verified resolved in the actual diff, the fix even surfaced and fixed a real latent bug, and the suite is green. No new findings.

## Delta metrics

```
commit reviewed: 460bab2 (only change since the prior pass at 7140262)
delta size: src +91/-26 · test +203/-0 · docs +87/-9  (method: git diff --numstat 7140262 460bab2)
PR totals now: +4608 / -119 across 42 files (gh pr view)
tests: suite 288 passed / 14 skipped / 0 fail (measured `npm test`); was 284/14/0 → +4 tests
new test: test/create-resilience.test.ts (191 lines, 12 assertions, deterministic — no network/Date.now/Math.random)
CI: gh pr checks still not fetchable offline — NOT measured.
mergeable: MERGEABLE.
```

## Findings from the prior pass — all resolved (verified in the diff, not taken on trust)

- **F1 (high, invariant) — RESOLVED.** [loop.ts](src/agent/loop.ts) is back to `availableTools(ctx).map((t) => t.schema)`, so `edit_file`/`write_file` are omitted from the advertised list until `editsUnlocked`; `dispatchTool` re-checks the same `availableTools(ctx)` as defense in depth. SPEC.md §1.3's structural-omission guarantee is honored again, and the invariant checklist wording was corrected. The right call: the same-turn propose→validate→edit batch was worth one turn, not the structural guarantee.
- **F2 (medium, verification gate) — RESOLVED.** Schematic `isComplete` ([create.ts:78-100](src/commands/create.ts#L78)) now ends in `return (await runErc(p)).ok`, so symbols + drift-clean is no longer enough: a resume can't treat an ERC-failing schematic as complete, and `commitResumedStage` can't commit unverified work. `check.ts` does not import the create STAGES, so this ERC call stays off the LLM-free `check` path.
- **F3 (medium, coverage) — RESOLVED, and it paid off.** New `test/create-resilience.test.ts` drives a retry-to-success (guidance prepended, cost accumulated across attempts) and both `commitResumedStage` branches with a scripted loop + scripted diagnosis. Writing it surfaced a genuine bug: `config.docs` defaults to `docs/`, so `isManagedPath` built `startsWith('docs//')` and read every doc as *foreign*, meaning `commitResumedStage` never actually committed its own work. Fixed by trimming the trailing slash ([create.ts:174](src/commands/create.ts#L174)). Exactly the kind of defect an untested branch hides.
- **F4 (low) — RESOLVED.** `ensureCwd` now `utimes`-touches the reused provider scratch dir each turn ([claude-code.ts:272-282](src/agent/providers/claude-code.ts#L272)), so the one long-lived temp dir a run holds never goes stale and a concurrent run's age-gated sweep can't delete it mid-run.
- **F5 (low) — RESOLVED.** Drift's BOM read now routes through `parseBomTable` (header-name resolution) in [drift.ts](src/memory/drift.ts), so `check` and `export bom` agree on a reordered table. Behavior on empty value/footprint cells is preserved (parseBomTable maps `''`→`undefined`, and the `!== undefined` guards still skip them), so no false-positive regression.
- **F6 (low, x4) — RESOLVED.** Case-preserving `normalizeFootprint` (directly unit-tested); cache key includes the resolved model id; the diagnosis call's tokens fold into the stage cost (`?? 0`-guarded for the throw path); the resume command emits an absolute `--brief`.

## Invariant re-checks (on the delta)

- Spec-gated-in: restored (F1) — structural omission confirmed.
- Verification-gated-out: strengthened on the resume path (F2).
- check/verify LLM-free and network-free: intact — the new `runErc` in `isComplete` is a kicad-cli subprocess reachable only from the create pipeline, not from `check.ts`.
- No sexp serialization, sync-obligations ledger, secrets: untouched by this commit.

## The one deferred item

The pre-existing rollback behavior (`restore()`'s `git clean -fd` can delete untracked user files that `git stash create` never captured) is explicitly left for a dedicated follow-up, as suggested in the prior pass. Reasonable to keep out of this PR; worth its own issue so it isn't lost.

Nice, disciplined turnaround. From my side this is good to merge.

PR #67, opened 2026-07-24
feat: add support for Cursor Agent CLI as a saved-login provider

### LINKED ISSUE #54: Add Cursor CLI provider (--model cursor)
### Summary

Add a **Cursor Agent CLI provider** so users can run Copperhead’s agent loop (`do`, `create`, `sync` resolve) using their **local Cursor CLI authentication** (`agent login`), without needing an OpenAI or Anthropic API key.

This mirrors the existing **Codex CLI provider** (`--model codex`), where an external CLI acts as a **reasoning-only backend** and Copperhead remains the sole dispatcher for file edits, KiCad verification, and git commits.

### Motivation

Today, Copperhead supports:

- Direct API providers: `gpt-5` / OpenAI, `claude` / Anthropic (require API keys)
- Local CLI provider: `codex` (uses `codex login` + `@openai/codex-sdk`)

Users who already have **Cursor CLI** set up (`agent` on PATH, authenticated via `agent login`) cannot use that subscription/login for Copperhead runs today. They must obtain separate model API keys even though Cursor Agent is available locally.

**Use case:** run `copperhead do` / `create` with `--model cursor` using existing Cursor CLI auth, same workflow as `--model codex`.

### Proposed solution

Add a new provider namespace:

```bash
copperhead do "rename net KEY_DAH to KEY_DASH" --model cursor
copperhead create --brief brief.md --model cursor
copperhead do "..." --model cursor:composer-2.5   # explicit model override (if supported)
```

**Implementation approach (follow Codex prior art):**

| Area | Reference |
|---|---|
| Provider implementation | `src/agent/providers/codex.ts` |
| Provider routing | `src/agent/loop.ts` → `makeProvider()` |
| Model resolution | `src/config.ts` → `resolveModel()` |
| OpenSpec change template | `openspec/changes/add-codex-cli-provider/` |
| Offline tests | `test/codex-provider.test.ts` |

**Design constraints (non-negotiable):**

1. **Spec-gated in:** `edit_file` / `write_file` stay structurally absent until an OpenSpec proposal validates.
2. **Verification-gated out:** ERC/DRC must pass before success/commit; failures roll back to the git snapshot.
3. **Reasoning-only Cursor:** Cursor must not mutate the target hardware repo directly. All mutations flow through Copperhead’s capability-filtered tools, obligations ledger, and verification gates.
4. **`check` stays LLM-free:** no provider/network imports on the `check` path.
5. **Structured tool calls:** returned tool names must be validated against the current Copperhead tool catalog (same pattern as Codex structured output).

### Authentication

Support at minimum:

- [ ] **Saved login:** `agent login` session (primary UX, analogous to `codex login`)
- [ ] **Explicit key fallback:** `CURSOR_API_KEY` env var / `--api-key` passthrough (optional, confirm in design)

Optional env var for non-standard installs:

- `COPPERHEAD_CURSOR_PATH` — path override when `agent` is not on PATH (mirrors `COPPERHEAD_CODEX_PATH`)

### Scope

**In scope:**

- New `CursorProvider` implementing the existing `Provider` interface
- Route `cursor` and `cursor:<model-id>` in `makeProvider()`
- Update `resolveModel()` so `--model cursor` works without OpenAI/Anthropic keys
- Offline unit tests with mocked CLI/SDK (no network)
- Opt-in live integration test flag (e.g. `COPPERHEAD_TEST_CURSOR=1`), mirroring Codex
- Docs: README, `.env.example`, `openspec/specs/SPEC.md` §4.4
- OpenSpec change: `add-cursor-cli-provider` (proposal, design, delta specs, tasks) before implementation

**Out of scope:**

- Replacing or removing existing providers
- Cursor editing KiCad files or docs directly
- MCP server wrapper (`add-mcp-server` is separate)
- Live viewer / Phase 2 features

### Acceptance criteria (draft)

- [ ] `makeProvider('cursor')` returns a Cursor provider without `OPENAI_API_KEY` or `ANTHROPIC_API_KEY`
- [ ] Cursor thread/session runs read-only / plan-mode; Copperhead alone dispatches mutations
- [ ] Structured tool calls validated against the current tool catalog; invalid calls get one corrective retry (Codex parity)
- [ ] Missing CLI or auth produces an actionable error (e.g. point to `agent login` / `agent status`)
- [ ] `npm test` offline suite stays green; new provider tests added
- [ ] Manual test: `copperhead do` on `manual-tests/runs/edit` sandbox passes with `--model cursor`
- [ ] Provider parity: at least AC-3.1-style rename test passes live with `--model cursor` (opt-in suite)
- [ ] `check` / `verify` behavior unchanged (LLM-free, network-free)

### Open questions

1. **CLI vs SDK:** subprocess to `agent --print --output-format json` vs `@cursor/sdk` — which is preferred for multi-turn structured tool calling?
2. **Model namespace:** is `cursor:<model-id>` the right override syntax (matching `codex:<model-id>`)?
3. **Auth precedence:** saved `agent login` only, or also `CURSOR_API_KEY` when both are present?
4. **Peer dependency:** should `@cursor/sdk` be optional/lazy-loaded like `@openai/codex-sdk`?

Devesh36:
## What

Adds `--model cursor` and `cursor:<model-id>` so copperhead runs the agent loop through the local Cursor Agent CLI (`agent login`) without OpenAI or Anthropic API keys.

## Why

Closes [#54](https://github.com/chouhanindustries/copperhead/issues/54). Users with Cursor CLI auth should get the same saved-login pattern as `codex` and `claude-code`: an external CLI as a reasoning-only backend while copperhead keeps spec-gating, ERC/DRC, drift checks, and git gates.

## Design

- New [cursor.ts](src/agent/providers/cursor.ts): subprocess to `agent` (override via `COPPERHEAD_CURSOR_PATH`, default `agent` on PATH) with `--mode plan`, `--sandbox enabled`, `--print --output-format json`, `--trust`, isolated temp workspace, and `--resume` for multi-turn sessions. See [design.md](openspec/changes/add-cursor-cli-provider/design.md).
- Shared JSON tool protocol in [tool-protocol.ts](src/agent/providers/tool-protocol.ts), extracted from [claude-code.ts](src/agent/providers/claude-code.ts) without behavior change for claude-code.
- Tripwire: fail the run if CLI JSON output indicates native tool execution or file mutation (claude-code-style enforcement, not Codex sandbox).
- Subprocess env strips `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, and `CURSOR_API_KEY` so saved login is not replaced by billed keys. Phase 1: no `CURSOR_API_KEY` passthrough.
- Routing in [loop.ts](src/agent/loop.ts) `makeProvider()`; distinct `name = 'cursor'` so keyed failover does not apply.

## Spec / OpenSpec

- Change: [add-cursor-cli-provider](openspec/changes/add-cursor-cli-provider/) (proposal, design, delta specs, tasks).
- [SPEC.md](openspec/specs/SPEC.md): section 4.4 cursor provider; AC-3.10 includes `--model cursor`; AC-3.12 cursor saved-login criterion.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npx vitest run test/cursor-provider.test.ts test/claude-code-provider.test.ts` | pass (38 passed, 1 skipped) |
| Live-LLM (opt-in) | `COPPERHEAD_TEST_CURSOR=1` | n/a (not run on author machine) |

Full `npm test` not claimed green here: several tests require `kicad-cli` (not installed on author machine due to disk space).

- Sandbox variant used (`create` / `edit`): `edit` (not run end-to-end)
- Commands run and outcome:

```text
$ npm run typecheck
(pass)

$ npm run build
(pass)

$ npx vitest run test/cursor-provider.test.ts test/claude-code-provider.test.ts
38 passed, 1 skipped

$ agent status
Logged in as deveshrathod047@gmail.com

$ node -e "import('./dist/agent/loop.js').then(m => m.makeProvider('cursor').then(p => console.log(p.name)))"
cursor
```

End-to-end `copperhead do ... --model cursor` deferred until `kicad-cli` is available (KiCad install blocked by insufficient disk space).

## Docs

- [README.md](README.md) (Saved login Cursor Agent section)
- [.env.example](.env.example) (`COPPERHEAD_CURSOR_PATH`, routing comments)
- [configuration.md](docs/src/content/docs/reference/configuration.md), [cli.md](docs/src/content/docs/reference/cli.md)
- [manual-tests/README.md](manual-tests/README.md)

---

## Invariant checklist

- [x] **Verification-gated out**: mutations still end in ERC (and DRC when the board changed) passing, with repair up to `maxRepairCycles` then rollback to the git snapshot.
- [x] **`check`/`verify` stays LLM-free and network-free**: nothing reachable from [check.ts](src/commands/check.ts) touches a provider, an API key, or the network.
- [x] **No sexp serialization**: the `src/kicad/` parser stays read-only; KiCad files are edited only via anchored exact-match text replace (no round-tripping).
- [x] **Spec coherence**: [SPEC.md](openspec/specs/SPEC.md) and the change artifacts (`proposal.md`, `design.md`, delta specs, `tasks.md`) moved together.

## Summary by CodeRabbit

* **New Features**
  * Added Cursor Agent CLI as a model backend via `cursor` / `cursor:<id>` using saved-login authentication.
  * Runs Cursor in plan mode with sandboxed execution, supports session resume, and extracts structured tool calls (reasoning-only enforced).
  * Added `COPPERHEAD_CURSOR_PATH` to locate the `agent`/`cursor-agent` binary when not on `PATH`.
* **Documentation**
  * Updated README and CLI/config docs with routing rules, saved-login setup steps, and clarified `check` behavior.
* **Tests**
  * Added Cursor provider unit tests and opt-in integration coverage.
* **Chores**
  * Ignored macOS `.DS_Store` files.


Devesh36:

animesh-chouhan:
End-to-end copperhead do ... --model cursor deferred until kicad-cli is available (KiCad install blocked by insufficient disk space).

this test is required since it would affect all users using cursor down the line.

Devesh36:
> End-to-end copperhead do ... --model cursor deferred until kicad-cli is available (KiCad install blocked by insufficient disk space).
> 
> this test is required since it would affect all users using cursor down the line.

wip

Devesh36:
> End-to-end copperhead do ... --model cursor deferred until kicad-cli is available (KiCad install blocked by insufficient disk space).
> 
> this test is required since it would affect all users using cursor down the line.

Updated: kicad-cli is installed now. Ran full `do --model cursor` on the manual edit sandbox (ERC/drift/finish/commit) and `COPPERHEAD_TEST_CURSOR=1` integration (AC-3.1/3.4/3.6) — all pass.

animesh-chouhan:
@Devesh36 can you share the logs?

Devesh36:
RUN  v3.2.7 /Users/deveshrathod/OpenSource/circuits/copperhead

 ✓ test/agent-integration.test.ts (16 tests | 13 skipped) 638073ms
   ✓ agent loop with cursor > AC-3.1: net rename propagates, ERC clean, one commit, surgical diff (AC-3.7)  457399ms
   ✓ agent loop with cursor > AC-3.4: budget-violating pullup is refused, citing the budget  100561ms
   ✓ agent loop with cursor > AC-3.6: a structurally forced failure rolls back to a byte-identical tree  80111ms

 Test Files  1 passed (1)
      Tests  3 passed | 13 skipped (16)
   Start at  15:37:48
   Duration  638.59s (transform 158ms, setup 0ms, collect 257ms, tests 638.07s, environment 0ms, prepare 47ms)

Screenshot 2026-07-25 at 10 55 19 AM

Devesh36:
Screenshot 2026-07-25 at 10 57 35 AM
Screenshot 2026-07-25 at 10 57 50 AM
Screenshot 2026-07-25 at 10 58 00 AM

AniketR10:
🤖 **Automated pr-review pass** (Claude Code `/pr-review`). Not a human review, and not a formal approval — findings below were each traced by hand and, where marked *confirmed*, reproduced locally.

## Verdict: request changes

Solid, well-shaped provider: the `tool-protocol.ts` extraction is a genuine no-behavior-change move, the env allowlist is stricter than `claude-code`'s, `name = 'cursor'` correctly disables keyed failover, and the spec artifacts moved with the code. Two high findings block: the provider resumes its CLI session unconditionally in exactly the configuration the codebase documents as incorrect, and the real subprocess path fails hard on realistically-sized KiCad files.

## Metrics

```
lines: +801 / -240 (net +561) across 23 files · src +473/-221, test +141/-0, openspec +153/-2, docs/other +34/-17
new vs net: 11 added files, 12 modified; ~302 genuinely new src lines (cursor.ts 283 + 19 in loop/cli/config)
             tool-protocol.ts (+171) is a verbatim move out of claude-code.ts (-216) — bodies byte-identical
tests: +141 test lines; suite 320/13/9 -> 329/16/9 (pass/skip/fail, base -> head)
diff coverage: 81.6% of changed src lines (315/386, measured); cursor.ts alone 69.5% (157/226)
deps/audit: no package.json change; package-lock.json has 12 unrelated deletions (L6) — no audit delta run
CI: all checks pass (CodeRabbit, cla, template) · mergeable: MERGEABLE
```

**How each number was obtained.** `BASE=887346b` (`git merge-base upstream/main <head>`). Sizes from `git diff --numstat $BASE...head`. Suite totals from `npx vitest run` on the head worktree and on a `$BASE` worktree — the 9 failures are **identical before and after** and are environmental here (`kicad-cli` not on PATH, `@openai/codex-sdk` not installed), not regressions. Diff coverage measured with a throwaway `@vitest/coverage-v8@3.2.7` (`--coverage.reportOnFailure`, `--coverage.include='src/**'`), intersecting the per-statement line hits with added lines from `git diff -U0 $BASE...head -- 'src/*'`.

**Uncovered changed src lines** (reconciles with L4): `cursor.ts:106-122` (`close()`), `182-183`, `189-191`, `197-208` (stdout fallback + `is_error`), `222-226` (subtype tripwire), `230-256` (**`defaultCursorRun` — the entire real subprocess path**), `258-271` (`isAuthError`/`authHint`), `282`; `cli.ts:128,206` (help strings only).

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| Spec-gated in (`edit_file`/`write_file` absent until proposal validates) | ✅ cursor takes its catalog from the loop's `tools` argument; `availableTools`/`requiresUnlock` untouched |
| Verification-gated out (ERC/DRC, repair, rollback) | ✅ `loop.ts` changes are routing + one comment; no verify/repair/rollback path touched |
| `check`/`verify` LLM-free and network-free | ✅ `src/commands/check.ts` imports unchanged; nothing reachable from it touches a provider |
| No sexp serialization | ✅ `src/kicad/` untouched |
| Sync-obligations ledger | ✅ untouched |
| Secrets (`sk-` redaction, env-only keys, `.gitignore`) | ✅ `subprocessEnv()` allowlist verified by test; redaction unchanged; `.gitignore` keeps `.env` + `.copperhead/runs/` — one caveat in H2 |
| Spec coherence (SPEC.md + change artifacts move together) | ✅ SPEC.md §4.4, §4.5, AC-3.10, AC-3.12 moved with proposal/design/delta specs/tasks. ⚠️ `openspec validate add-cursor-cli-provider` **not run** — the `openspec` CLI is not installed in this environment |

---

## Findings

### H1 · high · cursor resumes its CLI session unconditionally while the response cache is on — the exact combination the codebase documents as incorrect

`src/agent/providers/cursor.ts:65` · `src/agent/loop.ts:105-112`

`makeProvider(model, sessionResume)` computes `sessionResume` at `loop.ts:221` as `COPPERHEAD_CC_SESSION_RESUME === '1' && !config.llmCache`, with the reason spelled out at `loop.ts:217-220` and again at `claude-code.ts:121-129`:

> *"deliberately mutually exclusive with the response cache: the cache replays turns the resumed session never saw, so mixing them would desync the session."*

The cursor branch **drops that argument**: `return new CursorProvider(cursorModel)`. `CursorProvider` has no `sessionResume` flag at all — `const resume = this.sessionId` (line 65) resumes as soon as a session id exists. Meanwhile `config.llmCache` defaults to `true` (`src/config.ts:59`), so a `do --model cursor` run gets resume **and** the cache, with no way to turn resume off.

**Failure scenario.** `CachingProvider.chat` (`response-cache.ts:51-61`) returns on a hit *without* calling `inner.chat`, so a cached turn advances neither `this.sessionId` nor `this.sentCount`. Turn *k* is a real call → session `s1`, `sentCount = 8`. Turn *k+1* hits the cache (same message list as an earlier run — routine on a re-run after the documented "wait for the reset, then re-run" session-limit path, or a repair cycle that revisits an identical state). Turn *k+2* is a miss: it sends `--resume s1` with `renderDelta(messages, 8)`, and `renderDelta` **skips assistant messages by design** (`tool-protocol.ts:39-46`). So the resumed Cursor session receives `[result of edit_file] ...` for a tool call it never made and an assistant turn it never produced. The model then reasons over a history that does not match its own — inside the edit gates.

**Fix.** Mirror claude-code exactly — add the flag and thread it through:

```ts
// cursor.ts
constructor(
  private readonly model?: string,
  private readonly runFn: CursorRunLike = defaultCursorRun,
  private readonly sessionResume = false,
) {}
// ...
const resume = this.sessionResume ? this.sessionId : undefined;
// ...and guard the high-water mark the same way:
if (this.sessionResume) this.sentCount = messages.length;

// loop.ts
return new CursorProvider(cursorModel, undefined, sessionResume);
```

Regression test: drive a `CursorProvider` wrapped in `CachingProvider`, replay one cached turn between two real turns, and assert the second real call's `args.resume` is `undefined` (or that its prompt contains the intervening assistant tool call).

---

### H2 · high · the whole prompt goes in one argv element — a tool result over 128 KiB kills the run with `spawn E2BIG`

`src/agent/providers/cursor.ts:232, 247-254`

`defaultCursorRun` builds `fullPrompt = systemPrompt + '\n\n---\n\n' + prompt` and pushes it as a single positional argument. Linux caps one argv element at `MAX_ARG_STRLEN` = 128 KiB regardless of total `ARG_MAX`; macOS caps the total near 1 MB. `toolReadFile` (`src/agent/filetools.ts:6-22`) returns whole files with no cap, and nothing between it and the provider truncates.

**Confirmed, reproduced end-to-end** through `CursorProvider.chat` + `defaultCursorRun` against a stub binary, with a 150 KB `read_file` result in the conversation:

```
Cursor CLI provider failed: Command failed with E2BIG: /…/stub-agent.sh --print --output-format json …
```

Measured single-arg threshold on this Linux box: 127 KB ok, 128 KB `E2BIG`. Real `.kicad_pcb` files are routinely 200 KB–2 MB. The author's green E2E run used the fixture board (1 KB `.kicad_pcb`, 10 KB `.kicad_sch`) on macOS, which is why this did not surface. Note this also compounds H1: whenever `sessionId` is absent, `renderConversation` re-renders the **full** history every turn, so the argv grows monotonically and blows the limit well before a 40-turn budget even with modest files. The failure is an opaque `E2BIG` with no hint about size.

Same fix closes a secondary issue: argv is world-readable via `/proc/<pid>/cmdline` on Linux, so today the entire conversation — including any file contents the agent read — is visible to every local user for the life of each turn. `claude-code` and `codex` both go through an SDK and do not have this exposure.

**Fix.** Feed the prompt on stdin instead of argv (`cursor-agent`/`agent` reads the prompt from stdin in `--print` mode):

```ts
const { stdout } = await execa(bin, cmdArgs, {   // cmdArgs no longer ends with fullPrompt
  input: fullPrompt,
  env: args.env ?? subprocessEnv(),
  cancelSignal: args.signal,
  reject: true,
  maxBuffer: 50 * 1024 * 1024,
});
```

Regression test: the probe above — a `Msg[]` carrying a 150 KB tool result, run through `defaultCursorRun` against a stub script, asserting it returns rather than throwing `E2BIG`.

---

### M1 · medium · the "not logged in" half of the spec's own scenario is unreachable

`src/agent/providers/cursor.ts:258-263`

```ts
const status = (err as {...})?.status ?? (err as {...})?.exitCode;
if (typeof status === 'number') return status === 401 || status === 403;
const m = ((err as Error)?.message ?? '').toLowerCase();
return /unauthenticat|unauthoriz|not logged in|…|agent login/.test(m);
```

A subprocess never returns HTTP 401/403 — it returns an exit code. **Confirmed**: execa's nonzero-exit error carries `exitCode: 1`, `status: undefined`, `code: undefined`, so line 261 returns `false` and the message regex below it is dead code for every real CLI failure. `authHint()` (lines 266-271, also uncovered) can therefore never fire.

**Failure scenario.** User has not run `agent login`. The CLI exits 1 printing "not logged in". Confirmed output:

```
Cursor CLI provider failed: Command failed with exit code 1: agent --print
not logged in, run agent login
```

instead of the actionable hint. That fails the delta spec's own scenario (`specs/cursor-cli-provider/spec.md`: *"WHEN `agent` is not found **or the user is not logged in** THEN the run fails with an actionable message referencing `agent login` / `agent status`"*) and the AC-3.12 clause "unauthenticated … fail with an actionable message". The `agent`-not-found half works (ENOENT leaves `exitCode` undefined) and is the only half tested.

**Fix.** Check the message whenever no HTTP status is present, rather than treating `exitCode` as one:

```ts
function isAuthError(err: unknown): boolean {
  const status = (err as { status?: number; statusCode?: number })?.status
    ?? (err as { statusCode?: number })?.statusCode;
  if (status === 401 || status === 403) return true;
  const m = ((err as Error)?.message ?? '').toLowerCase();
  return /unauthenticat|unauthoriz|not logged in|please log in|login required|agent login/.test(m);
}
```

Regression test: a `runFn` rejecting with `Object.assign(new Error('… exit code 1\nnot logged in, run agent login'), { exitCode: 1 })`, asserting `chat()` rejects with `/agent login/` and `/agent status/`.

---

### M2 · medium · stdout that isn't one JSON object per line degrades to a silent empty turn

`src/agent/providers/cursor.ts:170-211`

`parseCursorStdout` splits on `\n` and `JSON.parse`s each line, `continue`-ing on failure; if nothing parses, the fallback loop (195-209) fails identically and it returns `{ text: '', sessionId: undefined }` with no error. `parseToolCalls('')` then yields `{ text: null, toolCalls: [] }` — an empty turn the loop reads as a stall.

**Confirmed** against two realistic payloads:

```
pretty-printed result object -> {"text":"","usage":{…}}     // no sessionId either
plain/banner stdout          -> {"text":"","usage":{…}}
```

**Failure scenario.** A `cursor-agent` release starts pretty-printing `--output-format json`, or prefixes a banner line. Every turn returns empty, `sessionId` is never captured (so `renderConversation` re-sends the full history each turn — see H2), the run burns its whole turn budget on stall nudges, and the raw stdout that would explain it appears nowhere in the error or transcript. The single-line shape is asserted only by `design.md`'s 2026-07-24 spike, and no test covers stdout that fails to parse.

**Fix.** Fail loudly instead of returning empty, and try the whole buffer as one object first:

```ts
if (!text) {
  try {
    const whole = JSON.parse(stdout) as Record<string, unknown>;   // pretty-printed single object
    assertNoNativeMutation(whole);
    if (typeof whole.result === 'string') { text = whole.result; /* …session_id… */ }
  } catch { /* fall through */ }
}
if (!text && stdout.trim()) {
  throw new Error(`cursor: could not parse Cursor Agent output as JSON — raw stdout: ${stdout.slice(0, 500)}`);
}
```

---

### M3 · medium · `.DS_Store` is committed to the repo and not gitignored

`.DS_Store` (new file, 8196 bytes, binary)

An 8 KB macOS Finder metadata blob unrelated to the PR, added at the repo root. `.gitignore` has no `.DS_Store` entry, so once merged it stays in history and will keep re-dirtying every macOS contributor's tree. It also records the author's local directory listing and view state.

**Fix.** `git rm --cached .DS_Store` and add `.DS_Store` to `.gitignore` in the same commit.

---

### L1 · low · every cursor turn reports zero tokens

`src/agent/providers/cursor.ts:211`

`parseCursorStdout` hardcodes `usage: { inputTokens: 0, outputTokens: 0 }`. `design.md` documents the cause ("official JSON schema does not expose token usage"), and it is honest rather than wrong — but nothing surfaces it, so `RunStats`, the per-turn `perTurn` array (`loop.ts:514-516`), and `create`'s per-stage cost table (SPEC §5.2, whose stated purpose is "so the expensive stages are obvious across runs") all silently read 0 for cursor runs, indistinguishable from a cached replay. Add a one-line note to the summary or the docs table stating that cursor runs do not report token usage.

### L2 · low · the subtype tripwire substring-matches an unbounded field

`src/agent/providers/cursor.ts:221-226`

`/tool|shell|write|edit|patch|mutation/.test(subtype)` on any non-`result` line hard-fails the run with "reasoning-only invariant violated". **Confirmed** false positive: `{"type":"assistant","subtype":"text_edit_suggestion"}` throws. Failing closed is the right side to err on for a safety gate, so this is low — but the message misattributes a benign informational event as an invariant breach, and the branch is uncovered by tests. Consider matching whole tokens (`/(^|_)(tool|shell|write|edit|patch|mutation)(_|$)/`) and naming the raw line in the error.

### L3 · low · the extraction dropped every explanatory comment

`src/agent/providers/tool-protocol.ts`

I verified the move is behavior-preserving (comment-stripped bodies diff to nothing but signature lines) — but all the "why" went with it: the D1 tolerance rationale, the `#I10` malformed-call rationale, and most importantly the comment explaining *why* `parseToolCalls` scans braces instead of matching ``` fences ("a fence regex truncates the JSON at the first inner fence … the model then assumes it wrote a file it never did"). That comment is the guardrail against someone re-simplifying it back into the bug it documents. Carry the block comments over to the new file.

### L4 · low · the real subprocess path is exercised only through injected fakes

`src/agent/providers/cursor.ts:230-256` (and `close()` 106-122, `isAuthError`/`authHint` 258-271, the stdout fallback 197-208)

Every test injects `runFn`, so `defaultCursorRun` — the binary name, the flag vector, `cancelSignal`, `maxBuffer`, and `parseCursorStdout` against real output — is unverified by the suite. Note that `--sandbox enabled` appears in D1 but **not** in the spike command line recorded in `design.md`; a flag rename or a changed default in a future `cursor-agent` release breaks every run with nothing in CI to catch it. `close()` (temp-workspace cleanup, the thing that guards the I8 disk-fill halt) is entirely untested too. A stub script on `COPPERHEAD_CURSOR_PATH` that echoes its argv and a canned result line would cover the arg vector, `close()`, and the parse path deterministically and offline — that is how I reproduced H2 and M1.

### L5 · low · doc nits

- `README.md` — the quick-start says `agent login status`; the correct command (used everywhere else, including `configuration.md` and `authHint()`) is `agent status`.
- `docs/…/configuration.md` — the routing table says "matched top to bottom" but lists `claude-code`, `cursor`, `codex`, while `makeProvider` matches `codex` first. Harmless today (the prefixes are disjoint) but the table claims to describe match order.

### L6 · low · unrelated `package-lock.json` churn

12 deletions removing `"libc": ["glibc"|"musl"]` from optional platform packages — an artifact of resolving the lockfile with an older npm than the one that wrote it. Unrelated to this PR, and npm uses `libc` to pick optional binaries on musl, so dropping it can change install resolution on Alpine. Revert this file.

---

*Reviewer disclosure: none of the commits under review are mine. No earlier automated pr-review pass exists on this PR; the prior comments are CodeRabbit's summary, CI/CLA bots, and the maintainer's E2E request (which the author has since answered with a live `--model cursor` run — that run is real evidence the arg vector works on macOS with the fixture, and is what makes H2's Linux/large-file case the remaining gap).*

Devesh36:
Addressed @AniketR10's pr-review findings in ee522b0:

**High**
- **H1** — Cursor `sessionResume` now mirrors claude-code: off by default, threaded from `makeProvider`, mutually exclusive with the response cache. Resume/`sentCount` only advance when the flag is on.
- **H2** — Prompt is fed on **stdin** (not argv) so large `read_file` results cannot hit Linux `E2BIG` / `MAX_ARG_STRLEN`. Covered by a stub-binary subprocess test with a 150 KiB payload.

**Medium**
- **M1** — `isAuthError` no longer treats `exitCode` as HTTP status; "not logged in" CLI exits get the `agent login` / `agent status` hint.
- **M2** — Unparseable stdout fails loudly; pretty-printed single JSON objects are accepted.
- **M3** — Removed committed `.DS_Store` and added it to `.gitignore`.

**Low (also done)**
- Windows `USERPROFILE`/`HOMEDRIVE`/`HOMEPATH` in the env allowlist
- Whole-token subtype tripwire + raw subtype in the error
- Restored explanatory comments in `tool-protocol.ts`
- Doc nits (`agent status`, routing table order, cursor token-usage note)
- Reverted unrelated `package-lock.json` libc churn
- Extra offline coverage: close(), auth exitCode path, defaultCursorRun argv/stdin

`npm run typecheck` + `npx vitest run test/cursor-provider.test.ts test/claude-code-provider.test.ts` → 47 passed, 1 skipped.

PR #84, opened 2026-07-26
fix(redact): redact base64 bearer tokens whole, match scheme case-insensitively

### LINKED ISSUE #83: redactSecrets leaves the tail of a base64 bearer token in the transcript
### Summary

`redactSecrets()` in `src/util/redact.ts` is the write-time guard for transcripts and summaries (AC-4.1) — `Transcript.event()` and `Transcript.writeSummary()` both run every line through it before it lands in `.copperhead/runs/`. Three gaps let a secret through, and the first one is the bad kind: it redacts **part** of the token and writes the rest to disk.

The file's own comment says the patterns are "deliberately broad: losing a few characters of log fidelity beats leaking a key" — these are cases where that intent isn't met.

### 1. A base64 bearer token is only partially redacted

`/Bearer\s+[A-Za-z0-9._-]{16,}/` doesn't include `+`, `/` or `=`, which are exactly the characters standard base64 uses. The match stops at the first one and the remainder is written verbatim:

```
in : Bearer ABCDEFGHIJKLMNOP+SECRETTAILabcdef/MORESECRET=
out: [REDACTED]+SECRETTAILabcdef/MORESECRET=
```

The surviving tail is `+SECRETTAILabcdef/MORESECRET=`. A partial redaction is worse than none — it looks handled in review while the secret is still on disk.

### 2. `Bearer` is matched case-sensitively

HTTP auth schemes are case-insensitive (RFC 7235 §2.1), and real clients and logs emit `bearer` and `BEARER`:

```
'authorization: bearer abcdefghijklmnopqrst'  -> unchanged
'BEARER abcdefghijklmnopqrst'                 -> unchanged
```

Every other pattern in the file is anchored to a literal prefix where case doesn't vary, so this is the only one affected.

### 3. `npm_` tokens with dashes are missed

`/npm_[A-Za-z0-9]{36,}/` excludes `-`, but npm also issues UUID-shaped tokens:

```
npm_0123456789abcdef-0123-4567-89ab-cdef01234567   -> unchanged
```

The `github_pat_` pattern next to it already allows `_`, so allowing `-` here is consistent with the file's existing approach.

### Reproduction

```js
import { redactSecrets } from './src/util/redact.js';

redactSecrets('Bearer ABCDEFGHIJKLMNOP+SECRETTAILabcdef/MORESECRET=');
// '[REDACTED]+SECRETTAILabcdef/MORESECRET='   ← tail leaks

redactSecrets('authorization: bearer abcdefghijklmnopqrst');
// unchanged

redactSecrets('npm_0123456789abcdef-0123-4567-89ab-cdef01234567');
// unchanged
```

### What I checked and did *not* report

- `ghp_ ghs_ gho_ ghu_ ghr_` are all covered — `[pousr]` happens to include every current GitHub prefix letter. No gap.
- `sk-` (OpenAI/Anthropic/OpenRouter) and JWT-style bearers are correctly redacted.
- `SYNAP_API_KEY` is absent from the patterns, but `src/memory/synap.ts` has no caller yet, so nothing can reach a transcript through it today. Not filing that.
- AWS/Google/Slack/GitLab keys aren't matched either, but copperhead never handles those, so that reads as out of scope rather than a defect.

The existing tests in `test/safety.test.ts` cover the `sk-`, `npm_`, `ghp_` and `github_pat_` cases; none exercise base64, lowercase `bearer`, or dashed npm tokens.

DPS0340:
## What

`redactSecrets()` in [redact.ts](src/util/redact.ts) now redacts a base64 bearer token in full instead of leaving its tail in the run transcript, matches the `Bearer` scheme case-insensitively, and covers npm tokens that contain dashes.

## Why

Closes #83.

`redactSecrets` is the write-time guard for `.copperhead/runs/`: both `Transcript.event()` and `Transcript.writeSummary()` route every line through it. Three shapes got past it, and the first is the dangerous kind because it redacted part of the token and wrote the rest to disk:

1. **Partial redaction of base64 bearer tokens.** The character class `[A-Za-z0-9._-]` excludes `+`, `/` and `=`, which is exactly what standard base64 uses, so the match stopped at the first one:

   ```
   in : Bearer ABCDEFGHIJKLMNOP+SECRETTAILabcdef/MORESECRET=
   out: [REDACTED]+SECRETTAILabcdef/MORESECRET=
   ```

   A partial redaction is worse than none: it reads as handled in review while the secret is still on disk.

2. **Case-sensitive scheme match.** HTTP auth schemes are case-insensitive (RFC 7235 section 2.1) and real clients log `bearer` and `BEARER`. Both passed through untouched.

3. **npm tokens with dashes.** npm also issues UUID-shaped tokens, and `npm_0123456789abcdef-0123-4567-89ab-cdef01234567` was not matched.

The file's own comment states the patterns are "deliberately broad: losing a few characters of log fidelity beats leaking a key". These are cases where that intent was not met.

## Design

Three small changes to the pattern list, no change to `redactSecrets`'s shape or call sites:

- the bearer character class gains `+/=` and the pattern gains the `i` flag
- the npm class gains `-`, matching how the neighbouring `github_pat_` pattern already allows `_`

The `i` flag is applied only to the bearer pattern. Every other entry is anchored to a literal lowercase prefix (`sk-`, `npm_`, `gh*_`, `github_pat_`) where the casing does not vary, so widening those would only add false-positive risk.

Deliberately not changed, to keep the diff honest about what it fixes:

- `ghp_ ghs_ gho_ ghu_ ghr_` are all already covered: `[pousr]` happens to contain every current GitHub prefix letter.
- `SYNAP_API_KEY` is absent from the patterns, but [synap.ts](src/memory/synap.ts) has no caller yet, so nothing can reach a transcript through it today.
- AWS, Google, Slack and GitLab key shapes are not matched, but copperhead never handles those credentials, so widening to them looked like scope creep. Happy to add if you would rather the net be wider.

## Spec / OpenSpec

No spec-level behavior change. This restores the intent of the existing AC-4.1 secret-redaction rail rather than altering it, so SPEC.md and the change artifacts are untouched.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | 324 pass, 14 skip, 8 pre-existing failures (see below) |

New tests, all in [safety.test.ts](test/safety.test.ts) under `secret redaction (AC-4.1)`:

- a base64 bearer token redacts whole, asserting the tail substrings `SECRETTAIL` and `MORESECRET` are absent rather than merely that `[REDACTED]` appears (the old behavior would have passed a "contains [REDACTED]" assertion)
- `Bearer` / `bearer` / `BEARER` / `BeArEr` all redact
- a dashed npm token redacts
- a guard-rail test: ordinary prose (`the bearer of this note`, `Bearer token missing`, `a npm_ package name`, `see the github_pat docs`) comes back byte-identical, so the widened character classes cannot start eating readable summary text

All tokens in the tests are synthetic and correctly shaped but never valid, matching the convention already in this file.

**Revert check.** With the new tests in place I restored the original `PATTERNS` array: exactly the 3 behavioral tests fail and the prose guard-rail passes under both implementations, so it is genuinely guarding rather than riding along.

n/a with a reason: this change touches one pure string function and its unit tests. It does not alter CLI behavior, the agent loop, providers, or the KiCad layer, so there is no runtime surface to exercise in the sandbox beyond what the unit tests already cover. The redaction path itself is exercised end to end by the pre-existing "transcript and summary are redacted at write time" test, which writes and reads back a real transcript on disk and still passes.

## Docs

None needed: no user-facing behavior, configuration, or CLI surface changes.

---

## Invariant checklist

- [x] **Secrets**: transcripts and summaries redact `sk-[A-Za-z0-9_-]+` at write time; keys live only in env vars; `.gitignore` keeps `.env` and `.copperhead/runs/`. (This PR strengthens exactly this rail; the `sk-` pattern is unchanged.)
- [ ] N/A **Spec-gated in**: does not touch the agent tool list.
- [ ] N/A **Verification-gated out**: does not touch the mutation or repair path.
- [ ] N/A **`check`/`verify` stays LLM-free and network-free**: nothing under `src/commands/check` is touched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secret redaction for bearer tokens regardless of capitalization.
  * Expanded redaction coverage for base64-formatted bearer tokens and dashed npm tokens.
  * Prevented partial secrets from remaining visible after redaction.
  * Preserved ordinary text that only resembles a secret token.

### DIFF
--- a/src/util/redact.ts
+++ b/src/util/redact.ts
@@ -5,10 +5,14 @@
  */
 const PATTERNS: RegExp[] = [
   /sk-[A-Za-z0-9_-]+/g,
-  /Bearer\s+[A-Za-z0-9._-]{16,}/g,
+  // Auth schemes are case-insensitive (RFC 7235 §2.1), and the token charset has
+  // to include base64's +, / and = or the match stops at the first one and the
+  // tail of the secret gets written out verbatim.
+  /Bearer\s+[A-Za-z0-9._+/=-]{16,}/gi,
   // Registry and forge tokens: a transcript that quotes a publish command or a
   // failing CI log can carry these just as easily as a model API key.
-  /npm_[A-Za-z0-9]{36,}/g,
+  // npm also issues UUID-shaped tokens, so the charset allows dashes.
+  /npm_[A-Za-z0-9-]{36,}/g,
   /gh[pousr]_[A-Za-z0-9]{36,}/g,
   /github_pat_[A-Za-z0-9_]{22,}/g,
 ];
--- a/test/safety.test.ts
+++ b/test/safety.test.ts
@@ -49,6 +49,44 @@ describe('secret redaction (AC-4.1)', () => {
     expect(out).toBe('[REDACTED] [REDACTED] [REDACTED]');
   });
 
+  it('redacts a base64 bearer token whole, leaving no tail', () => {
+    // Synthetic token: correct shape, never valid. Standard base64 uses +, /
+    // and =; a charset that stops at the first one writes the rest to disk.
+    const input = 'Bearer ABCDEFGHIJKLMNOP+SECRETTAILabcdef/MORESECRET=';
+    const out = redactSecrets(input);
+
+    expect(out).toBe('[REDACTED]');
+    expect(out).not.toContain('SECRETTAIL');
+    expect(out).not.toContain('MORESECRET');
+  });
+
+  it('redacts bearer tokens regardless of scheme casing', () => {
+    // HTTP auth schemes are case-insensitive (RFC 7235 §2.1).
+    for (const scheme of ['Bearer', 'bearer', 'BEARER', 'BeArEr']) {
+      const out = redactSecrets(`authorization: ${scheme} abcdefghijklmnopqrst`);
+      expect(out, scheme).toBe('authorization: [REDACTED]');
+    }
+  });
+
+  it('redacts npm tokens that contain dashes', () => {
+    // npm also issues UUID-shaped tokens. Synthetic, never valid.
+    const out = redactSecrets('npm_0123456789abcdef-0123-4567-89ab-cdef01234567');
+    expect(out).toBe('[REDACTED]');
+  });
+
+  it('leaves ordinary prose alone', () => {
+    // The patterns are deliberately broad, but not so broad that a run summary
+    // loses readable text.
+    for (const text of [
+      'the bearer of this note',
+      'Bearer token missing',
+      'a npm_ package name',
+      'see the github_pat docs',
+    ]) {
+      expect(redactSecrets(text), text).toBe(text);
+    }
+  });
+
   it('transcript and summary are redacted at write time', async () => {
     const dir = await mkdtemp(path.join(tmpdir(), 'ch-'));
     const t = new Transcript(dir);


DPS0340:

DPS0340:

animesh-chouhan:
**Automated pr-review pass** (Claude Code, not a human review; does not affect merge gating)

## Verdict: approve with one nit

The PR does exactly what it claims: the three redaction gaps from #83 are fixed, the tests genuinely guard the fix, and no repo invariant is touched. Every claim in the PR body that could be checked locally was verified, including the revert check. One low finding survives adversarial probing: a percent-encoded bearer token still hits the same partial-redaction failure class this PR argues is "worse than none".

## Metrics

```
lines: +44 / -2 (net +42) across 2 files  ·  src +6/-2, test +38/-0  (git diff --numstat 6dcca30...c9f681a)
new vs net: 0 new files, 2 modified; 6 added src lines, of which 2 are executable (the regex literals), 4 comments
tests: +38 test lines, +4 tests; suite 345/17/0 -> 349/17/0 (pass/skip/fail, measured locally at base and head; the 8 kicad-cli failures in the author's environment do not reproduce here since kicad-cli is on PATH)
diff coverage: 100% of changed executable src lines (2/2: redact.ts lines 11, 15), measured via throwaway @vitest/coverage-v8 install
deps/audit: n/a, package.json untouched
merge state: MERGEABLE; branch is behind main tip (merge base 6dcca30, main at 6d7631b) but conflict-free
CI: all 3 checks pass (CodeRabbit, cla, template)
```

Revert check independently reproduced: with base `redact.ts` and the new tests, exactly the 3 behavioral tests fail and the prose guard-rail passes under both implementations.

## Findings

### low: percent-encoded bearer tokens still redact partially

[src/util/redact.ts:11](https://github.com/chouhanindustries/copperhead/blob/c9f681a/src/util/redact.ts#L11). The widened charset covers raw base64 but not percent-encoding, so a URL-encoded token stops matching at the first `%` and the tail is written to disk, the exact failure class item 1 of #83 describes:

```
in : Bearer ABCDEFGHIJKLMNOP%2FSECRETTAIL
out: [REDACTED]%2FSECRETTAIL
```

Failure scenario: a transcript quotes a curl command, proxy log, or URL-encoded header value carrying the token. This is admittedly rarer than raw base64 (encoding a header value is unusual), so low rather than medium, and arguably a follow-up given the PR's deliberate scoping. Fix is one character plus a repro test: add `%` to the class (`[A-Za-z0-9._+/=%-]`; the hex digits after `%` are already covered), and assert `redactSecrets('Bearer ABCDEFGHIJKLMNOP%2FSECRETTAIL')` is `'[REDACTED]'`.

### Non-blocking observations

- The `i` flag plus `/` in the charset means lowercase `bearer` followed by a 16+ char path now redacts (`could not open bearer /home/u/.copperhead/runs/x.jsonl` loses the path). Consistent with the file's broad-by-design intent and the guard-rail test protects common prose; noting it so the trade-off is on record.
- Pre-existing, not this diff: the `sk-` pattern is unanchored with no minimum length, so `tsk-list` becomes `t[REDACTED]`. Worth a `\b` or `{20,}` in a follow-up.
- Legacy bare-UUID npm tokens (no `npm_` prefix) remain unmatched; they cannot reasonably be pattern-matched, just noting the boundary.

## Verification performed

- Adversarial probes against the new patterns: JSON-stringified transcript shape, base64url, newline between scheme and token, percent-encoding, prose false-positive candidates. All behave except the percent-encoding case above.
- New tests checked for determinism: no network, no clock, no randomness, all assert; the casing test correctly passes the scheme as the assertion message.
- `npm run typecheck` and full `npm test` run locally at head (349/17/0) and base (345/17/0).
- Invariant checks, all pass or not applicable exactly as the PR's checklist states: spec-gated-in (agent tool list untouched), verification-gated-out (mutation/repair path untouched), `check` LLM-free (nothing under `src/commands/check` touched), no sexp serialization (`src/kicad/` untouched), sync-obligations ledger (untouched), secrets rail (strengthened; `sk-` pattern and `.gitignore` unchanged).
- Spec coherence: SPEC.md pins only the `sk-` pattern (lines 387, 477), which is unchanged, so no artifact updates are required.
- Call-site grounding: both write-time surfaces ([transcript.ts:83](https://github.com/chouhanindustries/copperhead/blob/c9f681a/src/agent/transcript.ts#L83) and [transcript.ts:125](https://github.com/chouhanindustries/copperhead/blob/c9f681a/src/agent/transcript.ts#L125)) route through `redactSecrets`; the JSON-stringified event shape redacts correctly since `JSON.stringify` does not escape `+`, `/` or `=`.

aniket-intervue:
_Automated `pr-review` pass (not a human review). No prior automated pass found on this PR; CodeRabbit's summary comment is separate._

## Verdict: approve with nits

The core fix is real and I verified all three claims by running the base and PR patterns side by side against constructed payloads. The headline bug is genuine and was the dangerous kind: at base, `Bearer ABCDEFGHIJKLMNOP+SECRETTAILabcdef/MORESECRET=` redacted to `[REDACTED]+SECRETTAILabcdef/MORESECRET=`, writing the tail to disk. It now redacts whole. Case-insensitive scheme and dashed npm tokens likewise confirmed fixed. Test quality is good: the negative assertions (`not.toContain('SECRETTAIL')`) would have caught the old behavior where a naive "contains [REDACTED]" assertion would not, and the prose guard-rail test is the right instinct for a widened charset.

Findings are all low. Nothing blocks merge; the first two are worth a reply because one is a factual error in the PR's own reasoning.

```
lines: +44 / -2 (net +42) across 2 files  ·  src +6/-2, test +38, docs 0, spec 0
new vs net: 0 new files, 2 modified; 6 new src lines (2 executable, 4 comment)
tests: +38 test lines (ratio ~6:1 vs src); suite 345/17/0 -> 349/17/0 (pass/skip/fail)
        measured locally in clean worktrees at $BASE and at head, kicad-cli 9.0.9 present
diff coverage: 100% of executable changed src lines (2/2: redact.ts:11 and :15)
        measured: @vitest/coverage-v8@3.2.7 installed --no-save, coverage-final.json
        intersected with the per-line diff; the other 4 added lines are comments
deps/audit: n/a, package.json and package-lock.json untouched
CI: 3 checks green (CodeRabbit, cla, template). The `test` job NEVER RAN for this head:
        ci.yml run for c9f681a has conclusion `action_required` (fork PR awaiting
        maintainer workflow approval). See finding 3.
branch: 10 commits behind main, still MERGEABLE (no conflicts)
ReDoS: safe. `Bearer\s+[...]{16,}` is unambiguous (\s does not intersect the class),
        so it is linear: 200k-char adversarial input redacts in ~4 ms.
charset: trailing `-` in `[A-Za-z0-9._+/=-]` is a literal, not a range (verified: `?` is
        not consumed). Base64url `-`/`_` and JWT dots are all covered.
```

---

## Findings

### 1. low · The reason given for omitting a `SYNAP_API_KEY` pattern is factually wrong

The description states:

> `SYNAP_API_KEY` is absent from the patterns, but [synap.ts](src/memory/synap.ts) has no caller yet, so nothing can reach a transcript through it today.

`synap.ts` **does** have a caller, on this PR's own base:

- [src/agent/loop.ts:24](src/agent/loop.ts#L24) — `import { openSynapMemory, ... } from '../memory/synap.js'`
- [src/agent/loop.ts:278](src/agent/loop.ts#L278) — `transcript.event('synap-recall', ...)`
- [src/agent/loop.ts:297](src/agent/loop.ts#L297) / [:301](src/agent/loop.ts#L301) — `transcript.event('synap-record', ...)` / `transcript.event('synap-record-failed', { error: message })`

That last one is a live path from a **third-party SDK error string** into the transcript. It is gated on `SYNAP_API_KEY` being set (`synapEnabled()`, synap.ts:79), so it is opt-in and inactive by default, and it does route through `redactSecrets` via `transcript.event` — so a key is redacted **iff its shape matches an existing pattern**.

I am not claiming a leak: I have not shown that the Synap SDK echoes the credential in an error message, and if it does so as an `Authorization: Bearer …` header then this PR's own widening now catches it. The issue is narrower and still worth fixing: the decision to skip the pattern rests on a premise that does not hold, so it was not actually evaluated. Suggested fix is to the description, not necessarily the code: replace the claim with the accurate one ("wired into the loop but opt-in; the only transcript path is an SDK error message, which redactSecrets already covers for Bearer-shaped output"), or add a pattern if Synap's key shape is known and matches nothing.

### 2. low · The partial-redaction class is fixed for `Bearer` only; the same structure remains in the other four patterns

The PR's thesis is that a charset narrower than the token alphabet produces a partial redaction, and that a partial is worse than none. Correct, and it is fixed for `Bearer`. But the same shape is still reachable in the four untouched patterns. Constructed against the **PR's own** pattern list:

```
npm_ + 36 alnum + _SECRETTAIL123      -> [REDACTED]_SECRETTAIL123
ghp_ + 36 alnum + _SECRETTAIL123      -> [REDACTED]_SECRETTAIL123
github_pat_ + 22 + -SECRETTAIL123     -> [REDACTED]-SECRETTAIL123
sk- + 20 alnum + .SECRETTAIL123       -> [REDACTED].SECRETTAIL123
```

Each requires a token whose body contains a character outside its charset, and no currently-known npm / GitHub / OpenAI format does that — so this is hardening, not a live leak, which is why it is low. It is worth raising because the PR itself notes npm changed shape (the UUID case), so the alphabet-enumeration approach will keep drifting. A terminator-anchored match closes the whole class rather than one instance:

```ts
/npm_\S{36,}/g          // or: /npm_[^\s"']{36,}/g
```

Same for `gh[pousr]_` and `github_pat_`. If you would rather keep enumerated charsets, then a one-line comment saying "charsets are exhaustive for these three vendors' current formats" would at least make the assumption explicit.

Note the PR description says the npm change matches "how the neighbouring `github_pat_` pattern already allows `_`" — but the change adds `-`, not `_`. `npm_` still stops at an underscore.

### 3. low · The CI test suite never ran for this PR

`gh api .../commits/c9f681a/check-runs` returns only `template`, `template`, `cla`. The `ci.yml` run for this SHA has conclusion **`action_required`** — GitHub is holding workflow approval because this is a fork PR from a first-time contributor. So the `test` job (which installs KiCad and runs the suite) has not executed against this change.

That matters more than usual here: this is a security guard, and the author's own local run reported 8 failures, which they attributed to environment. That attribution is correct — I ran the suite in a clean worktree at this head with kicad-cli 9.0.9 present and got **349 passed / 17 skipped / 0 failed**, and **345/17/0** at the merge base, a delta of exactly the 4 new tests. So the change is green; nothing in CI records that. A maintainer approving the workflow run would close this.

### 4. low · New false positive: "bearer" followed by a path is now redacted

Combining the `i` flag with `/` in the charset means lowercase `bearer` before a path now matches:

```
in : bearer /usr/local/lib/node_modules/copperhead
base: bearer /usr/local/lib/node_modules/copperhead
head: [REDACTED]
```

Transcripts contain paths routinely, though the trigger needs the literal word `bearer` immediately before 16+ path characters, which is unlikely in a PCB design transcript. [redact.ts:3-4](src/util/redact.ts#L3-L4) explicitly prefers losing log fidelity to leaking a key, so this is arguably intended, and the PR's guard-rail test shows the risk was considered. Flagging only so the widening is a recorded choice rather than a surprise.

For balance, a case I initially suspected and then refuted: `Bearer authentication-scheme` is redacted, but it was **already** redacted at base (`-` was in the old charset too), so it is not a regression from this PR.

---

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| **Secrets** | **pass, strengthened** — this is the AC-4.1 rail. `sk-` pattern unchanged; the two changed patterns are strictly wider than base for every input I tested; both call sites (`Transcript.event()` at transcript.ts:83, `Transcript.writeSummary()` at :125) still route through `redactSecrets`. Residual gaps in findings 1 and 2. |
| **Spec-gated in** | **pass, N/A** — no change to the agent tool list or `availableTools`. |
| **Verification-gated out** | **pass, N/A** — mutation, repair, and rollback paths untouched. |
| **`check`/`verify` LLM-free and network-free** | **pass, N/A** — nothing under `src/commands/check` is touched; `redact.ts` is a pure string function with no imports. |
| **No sexp serialization** | **pass, N/A** — `src/kicad/` untouched. |
| **Sync-obligations ledger** | **pass, N/A** — untouched. |
| **Spec coherence** | **pass** — agree with the PR that this restores AC-4.1's stated intent rather than changing it, so SPEC.md and the change artifacts correctly stay put. |

## Notes

- **Tests are deterministic and assert properly**: no network, no `Date.now()`, no `Math.random()`, no order dependence, no non-asserting test. The `expect(out, scheme)` message argument is valid vitest and makes the loop failures legible. Verified by running `test/safety.test.ts` in isolation: 22 passed.
- **No mock-only concern**: this is a pure function, and the end-to-end path is genuinely covered by the pre-existing "transcript and summary are redacted at write time" test, which writes and reads a real transcript from disk. The PR's "manual test log: n/a" justification is reasonable for this change.
- **Scope is tight** — 2 files, one function's pattern list, no drive-by edits. The description's "deliberately not changed" section is a good habit; finding 1 is the one entry in it that does not hold up.
- **Cross-PR interaction worth knowing about**: open PR #113 adds a REPL session log that re-implements redaction inline with a `sk-`-only regex instead of calling `redactSecrets` ([repl.ts:321](src/commands/repl.ts#L321) on that branch). If both merge, the new persisted log will not benefit from the fixes in this PR. I raised it on #113; noting it here so the two are not reconciled independently.
- Authorship: single commit by `DPS0340`; none by this reviewer, so no self-review caveat applies.

PR #86, opened 2026-07-26
fix(safety): resolve symlinks before the repo-root containment check

### LINKED ISSUE #85: Path sandbox (AC-4.2) is lexical only: a symlink in the repo lets read/write/edit escape the root
### Summary

`resolveInRepo()` in `src/util/paths.ts` enforces AC-4.2 ("a tool call with a path outside the repo root is rejected") using `path.resolve` + a prefix check. `path.resolve` is **purely lexical** — it does not consult the filesystem — so a symlink inside the repo that points outside is resolved to a path that still *starts with* the repo root, passes the check, and the subsequent `fs` call follows the link.

```ts
export function resolveInRepo(repoRoot: string, p: string): string {
  const abs = path.resolve(repoRoot, p);   // lexical only
  const root = path.resolve(repoRoot);
  if (abs !== root && !abs.startsWith(root + path.sep)) {
    throw new SandboxError(p);
  }
  return abs;
}
```

The comment above it says "All file tools must go through this", and they do — which is why every one of them inherits the gap.

### Reproduction

I ran this against the **real exported tool functions** (compiled `src/agent/filetools.ts`), not a reimplementation. Setup is a temp repo containing `link -> ../outside`, with `outside/secret.txt` holding `TOPSECRET`:

```
READ via symlink  -> "TOPSECRET"                <-- escaped
WRITE via symlink -> wrote link/planted.txt     <-- escaped, file created outside the repo
EDIT via symlink  -> edited link/secret.txt     <-- escaped, secret.txt now "OVERWRITTEN"
```

So all three of `toolReadFile`, `toolWriteFile` and `toolEditFile` operate outside the repo root. The write path is the more serious one: `toolWriteFile` refuses to overwrite an existing file, but it will happily *create* one anywhere the process can write, and `toolEditFile` overwrites through the link.

The lexical cases the existing tests cover all still behave correctly:

```
'../secret/key.txt'            -> BLOCKED
'/etc/passwd'                  -> BLOCKED
'./link/../../secret/key.txt'  -> BLOCKED
'link/key.txt'                 -> ALLOWED   <-- the gap
```

`test/safety.test.ts` covers `../../etc/hosts`-style traversal and the shared-prefix sibling case (`/repo` vs `/repo-evil`), but nothing involving a symlink.

### Why it's reachable

The model chooses these paths. A `.history/` or vendored directory containing a symlink, a KiCad library symlinked into the project (a normal thing to do — shared symbol/footprint libs), or any repo the user points copperhead at that already has one, is enough. The agent doesn't need to create the link, only to walk into it.

### Expected

A path that resolves *through* a symlink to a location outside the repo root is rejected, the same as a lexical `../` escape.

### Suggested approach

Resolve the real path before the containment check, with the caveat that the target of a *write* may not exist yet — so the check needs to fall back to the nearest existing ancestor for new files. Something like: `realpath` the deepest existing prefix of the path, then re-apply the same containment test. Both `repoRoot` and the candidate need `realpath` treatment, or a repo that itself lives under a symlinked path (`/tmp` on macOS is `/private/tmp`) would start failing.

I have a fix along those lines ready and will open a PR referencing this issue.

DPS0340:
## What

`resolveInRepo()` in [paths.ts](src/util/paths.ts) now resolves symlinks before deciding whether a path is inside the repo, so `read_file`, `write_file` and `edit_file` can no longer reach outside the repo root through a symlink that lives inside it.

## Why

Closes #85.

`resolveInRepo` enforces AC-4.2 with `path.resolve` plus a prefix check. `path.resolve` is purely lexical: it never touches the filesystem. A symlink inside the repo pointing outside it therefore resolves to a path that still starts with the repo root, passes the check, and the following `fs` call follows the link out of the sandbox.

I reproduced this against the real exported tool functions from [filetools.ts](src/agent/filetools.ts), compiled and driven directly, not a reimplementation of the check. Setup is a temp repo containing `link -> ../outside`, with `outside/secret.txt` holding `TOPSECRET`:

```text
READ via symlink  -> "TOPSECRET"                <-- escaped
WRITE via symlink -> wrote link/planted.txt     <-- escaped, file created outside the repo
EDIT via symlink  -> edited link/secret.txt     <-- escaped, secret.txt now "OVERWRITTEN"
```

After the fix, the same probe:

```text
READ blocked: path escapes repo root: link/secret.txt
WRITE blocked: path escapes repo root: link/planted.txt
EDIT blocked: path escapes repo root: link/secret.txt
```

The write path is the more serious half. `toolWriteFile` refuses to overwrite an existing file, but it will create one anywhere the process can write, and `toolEditFile` overwrites straight through the link.

This is reachable without the agent creating anything: a symlinked KiCad symbol or footprint library is a normal project layout, and any repo the user points copperhead at may already contain one. The model only has to walk into it.

## Design

The lexical check is kept exactly as it was, and a realpath-based check is added after it. Two things made this more than a one-line change:

**A write target may not exist yet.** `toolWriteFile` creates new files, so a plain `realpathSync` would throw ENOENT on every legitimate write. `realpathDeepest()` walks up to the deepest ancestor that does exist, resolves that, then re-attaches the remaining segments. Those trailing segments have already been lexically normalised by `path.resolve`, so they cannot smuggle a `..` back in.

**The repo root itself can sit behind a symlink.** On macOS `/tmp` is `/private/tmp`, so resolving only the candidate path would make every path under a temp repo look external and fail the whole suite. Both sides get the same treatment.

`resolveInRepo` still returns the lexical path, so no caller sees a changed value: the realpath is used only for the decision.

One trade-off worth flagging: this is TOCTOU-adjacent. The check resolves the path and the `fs` call re-resolves it, so a link swapped between the two would still be followed. Closing that properly needs `O_NOFOLLOW`-style handling at each call site, which is a much larger change. Given the threat model here is a model wandering into a pre-existing symlink rather than an attacker racing the process, this felt like the right scope, but say the word if you would rather go further.

## Spec / OpenSpec

No spec-level behavior change. AC-4.2 already requires that a tool call with a path outside the repo root is rejected; this closes a hole in the existing implementation rather than changing what the rail promises, so SPEC.md and the change artifacts stay as they are.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | 323 pass, 14 skip, 8 pre-existing failures (see below) |

New tests, all in [safety.test.ts](test/safety.test.ts) under `path sandbox (AC-4.2)`:

- **the escape**: read of an existing file through the link, write of a file that does not exist yet through the link, and the link itself, all rejected
- **guard rail 1**, real paths still work: an existing file, a file that does not exist yet, a directory that does not exist yet, and `.` (the root). This is what pins both the "write target may not exist" and "macOS temp dir is itself symlinked" behaviors
- **guard rail 2**, a symlink that stays inside the repo is still allowed, since symlinked library dirs are a normal KiCad setup and must not start failing

**Revert check.** With the new tests in place I restored the original `paths.ts`: exactly 1 test fails, the escape test. Both guard rails pass under the old and the new implementation, which is the point, they prove containment was not achieved by simply refusing more things.

Sandbox variant used: none. Exercised the shipped tool functions directly instead, which is a closer test of this change than a CLI run: the escape happens inside `toolReadFile` / `toolWriteFile` / `toolEditFile`, and driving them by hand lets the probe assert on the file outside the repo rather than on agent output.

Commands run and outcome:

```text
$ npx tsc src/agent/filetools.ts src/util/paths.ts --outDir /tmp/tscout --module esnext --target es2022 --moduleResolution bundler
$ node /tmp/symprobe.mjs        # before the fix
READ via symlink  -> "TOPSECRET"   <-- ESCAPED
WRITE via symlink -> wrote link/planted.txt   <-- ESCAPED
EDIT via symlink  -> edited link/secret.txt   <-- ESCAPED

$ node /tmp/symprobe.mjs        # after the fix
READ blocked: path escapes repo root: link/secret.txt
WRITE blocked: path escapes repo root: link/planted.txt
EDIT blocked: path escapes repo root: link/secret.txt
```

The probe builds its own temp repo, writes `outside/secret.txt`, symlinks it in as `link`, and reads the outside file back afterwards to confirm whether it was actually modified. Happy to commit it as a fixture if that is useful.

## Docs

None needed: no user-facing behavior, configuration, or CLI surface changes. The refusal message and `SandboxError` shape are unchanged, so a caller that already handles a traversal rejection handles this one identically.

---

## Invariant checklist

- [x] **Spec-gated in**: unchanged, the tool list is untouched. `resolveInRepo` gates paths, not tool availability.
- [x] **`check`/`verify` stays LLM-free and network-free**: `paths.ts` has no provider, key, or network access, and this change adds only `node:fs`'s `realpathSync`.
- [ ] N/A **Secrets**: no change to redaction or key handling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened sandboxed repository path protection for file and directory operations when symbolic links are involved, including dangling and chained links.
  * Enhanced validation to prevent escaping the allowed area even after resolving symlinks, while still permitting safe in-repo targets and repo-relative paths.
  * Allowed legitimate paths when intermediate targets or the leaf file do not yet exist.
* **Tests**
  * Expanded safety test coverage with additional symlink scenarios to verify correct blocking/allowing behavior for both read and write cases.

### DIFF
--- a/src/util/paths.ts
+++ b/src/util/paths.ts
@@ -1,3 +1,4 @@
+import { realpathSync } from 'node:fs';
 import path from 'node:path';
 
 export class SandboxError extends Error {
@@ -8,8 +9,37 @@ export class SandboxError extends Error {
 }
 
 /**
+ * Resolve symlinks as far as the path actually exists.
+ *
+ * A write target legitimately does not exist yet, so we walk up to the deepest
+ * existing ancestor, resolve that, and re-attach the remaining segments. Those
+ * trailing segments are already lexically normalised by `path.resolve`, so they
+ * cannot contain `..`.
+ */
+function realpathDeepest(abs: string): string {
+  let head = abs;
+  const tail: string[] = [];
+  for (;;) {
+    try {
+      return path.join(realpathSync(head), ...tail);
+    } catch {
+      const parent = path.dirname(head);
+      // Reached the filesystem root without finding anything that exists.
+      if (parent === head) return abs;
+      tail.unshift(path.basename(head));
+      head = parent;
+    }
+  }
+}
+
+/**
  * Resolve a repo-relative path and reject anything that escapes the repo root
  * (AC-4.2). All file tools must go through this.
+ *
+ * The containment check runs on the symlink-resolved path: `path.resolve` is
+ * purely lexical, so a symlink inside the repo pointing outside it would resolve
+ * to something that still starts with the repo root and slip through, and the
+ * following `fs` call would then follow the link out of the sandbox.
  */
 export function resolveInRepo(repoRoot: string, p: string): string {
   const abs = path.resolve(repoRoot, p);
@@ -17,6 +47,15 @@ export function resolveInRepo(repoRoot: string, p: string): string {
   if (abs !== root && !abs.startsWith(root + path.sep)) {
     throw new SandboxError(p);
   }
+
+  // The repo root itself may sit under a symlink (on macOS /tmp is /private/tmp),
+  // so both sides have to be resolved or every path under it would look external.
+  const realRoot = realpathDeepest(root);
+  const realAbs = realpathDeepest(abs);
+  if (realAbs !== realRoot && !realAbs.startsWith(realRoot + path.sep)) {
+    throw new SandboxError(p);
+  }
+
   return abs;
 }
 
--- a/src/util/paths.ts
+++ b/src/util/paths.ts
@@ -1,4 +1,4 @@
-import { realpathSync } from 'node:fs';
+import { lstatSync, readlinkSync, realpathSync } from 'node:fs';
 import path from 'node:path';
 
 export class SandboxError extends Error {
@@ -15,14 +15,40 @@ export class SandboxError extends Error {
  * existing ancestor, resolve that, and re-attach the remaining segments. Those
  * trailing segments are already lexically normalised by `path.resolve`, so they
  * cannot contain `..`.
+ *
+ * A *dangling* symlink is the exception: `realpathSync` fails on it exactly as
+ * it does on a path that simply doesn't exist, so treating the two the same way
+ * would rebuild the link under the repo root and let a write follow it out of
+ * the sandbox. `lstat` describes the link itself, so it still succeeds where
+ * `realpath` gave up — that's how the two cases are told apart. When we find
+ * one, resolve its target manually and keep checking from there.
  */
 function realpathDeepest(abs: string): string {
   let head = abs;
   const tail: string[] = [];
+  // A dangling link can point at another dangling link; bound the chase so a
+  // cycle cannot spin here.
+  let hops = 0;
   for (;;) {
     try {
       return path.join(realpathSync(head), ...tail);
     } catch {
+      // A link whose target is missing: follow it lexically so containment is
+      // judged on where it points, not on where the link happens to live.
+      let link: string | null = null;
+      try {
+        if (hops < 32 && lstatSync(head).isSymbolicLink()) {
+          link = path.resolve(path.dirname(head), readlinkSync(head));
+        }
+      } catch {
+        // head does not exist at all; fall through to the parent walk.
+      }
+      if (link !== null) {
+        hops++;
+        head = link;
+        continue;
+      }
+
       const parent = path.dirname(head);
       // Reached the filesystem root without finding anything that exists.
       if (parent === head) return abs;
--- a/test/safety.test.ts
+++ b/test/safety.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { mkdir, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises';
+import { mkdir, mkdtemp, readFile, rm, symlink, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
 import { resolveInRepo, SandboxError, isKicadFile } from '../src/util/paths.js';
@@ -24,6 +24,53 @@ describe('path sandbox (AC-4.2)', () => {
 
   it('does not treat sibling dirs with a shared prefix as inside', () => {
     expect(() => resolveInRepo('/repo', '../repo-evil/x')).toThrow(SandboxError);
+  });
+
+  it('rejects a path that leaves the repo through a symlink (AC-4.2)', async () => {
+    // path.resolve is lexical, so "link/secret.txt" stays prefixed with the repo
+    // root even though the link points outside it.
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-symlink-'));
+    const repo = path.join(base, 'repo');
+    const outside = path.join(base, 'outside');
+    await mkdir(repo, { recursive: true });
+    await mkdir(outside, { recursive: true });
+    await writeFile(path.join(outside, 'secret.txt'), 'TOPSECRET\n', 'utf8');
+    await symlink(outside, path.join(repo, 'link'));
+
+    // Read of an existing file through the link.
+    expect(() => resolveInRepo(repo, 'link/secret.txt')).toThrow(SandboxError);
+    // Write of a file that does not exist yet, in a directory reached through it.
+    expect(() => resolveInRepo(repo, 'link/planted.txt')).toThrow(SandboxError);
+    // The link itself.
+    expect(() => resolveInRepo(repo, 'link')).toThrow(SandboxError);
+  });
+
+  it('still allows real paths, including ones that do not exist yet', async () => {
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-symlink-ok-'));
+    const repo = path.join(base, 'repo');
+    await mkdir(path.join(repo, 'sub'), { recursive: true });
+    await writeFile(path.join(repo, 'sub', 'here.txt'), 'ok\n', 'utf8');
+
+    // An existing file, a file that does not exist yet, a directory that does
+    // not exist yet, and the root itself. On macOS the temp dir is itself behind
+    // a symlink (/tmp -> /private/tmp), so this also pins that resolving the
+    // root doesn't make everything under it look external.
+    expect(resolveInRepo(repo, 'sub/here.txt')).toEqual(path.join(repo, 'sub', 'here.txt'));
+    expect(resolveInRepo(repo, 'sub/new.txt')).toEqual(path.join(repo, 'sub', 'new.txt'));
+    expect(resolveInRepo(repo, 'brand/new/deep.txt')).toEqual(path.join(repo, 'brand', 'new', 'deep.txt'));
+    expect(resolveInRepo(repo, '.')).toEqual(repo);
+  });
+
+  it('allows a symlink that stays inside the repo', async () => {
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-symlink-in-'));
+    const repo = path.join(base, 'repo');
+    await mkdir(path.join(repo, 'real'), { recursive: true });
+    await writeFile(path.join(repo, 'real', 'lib.txt'), 'shared\n', 'utf8');
+    await symlink(path.join(repo, 'real'), path.join(repo, 'alias'));
+
+    // Symlinked library dirs are normal in KiCad projects; only escaping ones
+    // should be refused.
+    expect(() => resolveInRepo(repo, 'alias/lib.txt')).not.toThrow();
   });
 });
 
--- a/test/safety.test.ts
+++ b/test/safety.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest';
 import { mkdir, mkdtemp, readFile, rm, symlink, writeFile } from 'node:fs/promises';
+import { existsSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
 import { resolveInRepo, SandboxError, isKicadFile } from '../src/util/paths.js';
@@ -71,6 +72,46 @@ describe('path sandbox (AC-4.2)', () => {
     // Symlinked library dirs are normal in KiCad projects; only escaping ones
     // should be refused.
     expect(() => resolveInRepo(repo, 'alias/lib.txt')).not.toThrow();
+  });
+
+  it('rejects a DANGLING symlink that points outside the repo (AC-4.2)', async () => {
+    // realpath fails on a dangling link exactly as it does on a path that
+    // simply does not exist. Treating them the same rebuilds the link under the
+    // repo root, and a later write follows it out of the sandbox.
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-dangling-'));
+    const repo = path.join(base, 'repo');
+    const outside = path.join(base, 'outside');
+    await mkdir(repo, { recursive: true });
+    await mkdir(outside, { recursive: true });
+    // Target deliberately does NOT exist.
+    await symlink(path.join(outside, 'planted.txt'), path.join(repo, 'link'));
+
+    expect(() => resolveInRepo(repo, 'link')).toThrow(SandboxError);
+    await expect(toolWriteFile(repo, 'link', 'escaped\n')).rejects.toThrow(SandboxError);
+    expect(existsSync(path.join(outside, 'planted.txt'))).toBe(false);
+  });
+
+  it('rejects a chain of dangling symlinks that ends outside the repo', async () => {
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-dangling-chain-'));
+    const repo = path.join(base, 'repo');
+    const outside = path.join(base, 'outside');
+    await mkdir(repo, { recursive: true });
+    await mkdir(outside, { recursive: true });
+    await symlink(path.join(outside, 'x.txt'), path.join(repo, 'hop2'));
+    await symlink(path.join(repo, 'hop2'), path.join(repo, 'hop1'));
+
+    expect(() => resolveInRepo(repo, 'hop1')).toThrow(SandboxError);
+  });
+
+  it('still allows a dangling symlink whose target is inside the repo', async () => {
+    // The guard rail: a link to a file the agent is about to create must keep
+    // working, otherwise the fix would just refuse more things.
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-dangling-ok-'));
+    const repo = path.join(base, 'repo');
+    await mkdir(path.join(repo, 'sub'), { recursive: true });
+    await symlink(path.join(repo, 'sub', 'future.txt'), path.join(repo, 'pending'));
+
+    expect(() => resolveInRepo(repo, 'pending')).not.toThrow();
   });
 });
 


DPS0340:

DPS0340:

DPS0340:
Confirmed — this was a real hole in my own fix, and a serious one. Fixed in `a6f7a55`.

I reproduced it against the compiled tools before changing anything. A repo containing `link -> ../outside/planted.txt` where the target does **not** exist:

```text
resolveInRepo ALLOWED -> /tmp/dangle/repo/link
toolWriteFile -> wrote link
  outside/planted.txt exists? true
  content: "ESCAPED\n"
```

So the write went straight through the sandbox and created a file outside the repo. Exactly as described.

**Why it happened.** `realpathSync` fails identically for "this path doesn't exist yet" and "this is a link whose target doesn't exist". `realpathDeepest` was built for the first case — a write target legitimately not existing — and swallowed both, rebuilding the link's own path under the repo root.

**The fix.** `lstat` describes the link itself, so it still succeeds exactly where `realpath` gave up — that is what tells the two cases apart. On a dangling link, resolve its target with `readlink` and keep checking from there, so containment is judged on *where it points*, not on where the link happens to sit. Chain-following is bounded at 32 hops so a cycle of dangling links can't spin.

After the fix, same probe:

```text
resolveInRepo BLOCKED: path escapes repo root: link
toolWriteFile blocked: path escapes repo root: link
```

**Three tests added**, including a guard rail, because the risk with this kind of fix is achieving containment by simply refusing more:

- a dangling link pointing outside → rejected, and the test asserts the outside file was **not** created
- a *chain* of dangling links (`hop1 -> hop2 -> outside/x.txt`) → rejected
- a dangling link whose target is **inside** the repo → still allowed. This matters: a link to a file the agent is about to create has to keep working.

Verification:

- 24 tests in `safety.test.ts` pass.
- Reverting just the dangling-link handling fails **only the two escape tests**; the inside-repo guard rail passes under both implementations.
- Full suite 326 pass (up from 324), `tsc --noEmit` clean. The 8 failures are the pre-existing `KicadCliMissingError` ones, unchanged.

Thanks for this one — I'd tested the *existing*-symlink escape thoroughly and never thought to point a link at something that wasn't there.

DPS0340:
Good call — taken in the commit above.

You're right that the guard rail was only asserting *resolution*, which is the weaker half. The whole reason `realpathDeepest` walks up to the deepest existing ancestor is so a write to a not-yet-existing target works, and I wasn't actually exercising that. Verified the behaviour first rather than just pasting the suggestion:

```
write -> "wrote pending"
sub/future.txt exists? true
content: "inside\n"
```

Added the write plus assertions on both existence and content. 24 tests pass; reverting the dangling-link handling still fails only the two escape tests, so the guard rail continues to pass under both implementations — which is what makes it a guard rail rather than another assertion of the fix.

AniketR10:
@DPS0340 thank you for your contribution, reviewing it shortly

animesh-chouhan:
🤖 **Automated pr-review pass** (Claude Code). Not a human review.

## Verdict: request changes

The direction is right and this is a strict improvement over `main`: every symlink escape that exists today is closed for the straightforward cases, and the guard rails (writes to not-yet-existing targets, symlinked repo roots, in-repo symlinks) hold. The prior review round's dangling-symlink finding is genuinely fixed; I re-derived it and chains of up to 32 dangling links are now blocked.

But the rail the PR claims to close is still open in three ways, and I confirmed two of them by actually creating a file outside the repo and reading a file outside the repo through the shipped tool functions. All three share one root cause: **`..` inside a symlink body is collapsed lexically, while the kernel applies it after resolving the preceding component.** Neither `path.resolve` nor Node's JS `fs.realpathSync` is safe to decide containment with, because both collapse `..` before any symlink is followed.

## Metrics

```
lines: +154 / -1 (net +153) across 2 files  ·  src +65/-0, test +89/-1, docs 0, spec 0
new vs net: 0 new files, 2 modified; 65 new src lines (all in src/util/paths.ts)
tests: +89 test lines (test:src 1.37:1); suite 345/17/0 -> 351/17/0 (pass/skip/fail, base -> head)
diff coverage: 100% of changed src lines (51/51 statements, measured); branches 17/18,
               uncovered: src/util/paths.ts:54
deps/audit: n/a (package.json unchanged)
```

Method: sizes from `git diff --numstat 6dcca30..HEAD` against the merge base `6dcca30`. Suite totals from `npm test` run on both the merge base and the PR head in a clean worktree. Coverage measured with `@vitest/coverage-v8@3.0.8` (installed `--no-save` to match `vitest ^3.0.8`), `npx vitest run test/safety.test.ts --coverage --coverage.include='src/util/paths.ts' --coverage.reporter=json`, intersected with the changed lines.

A note on the PR's own test table: it reports 8 pre-existing failures from `kicad-cli not found on PATH`. With `kicad-cli` installed there are zero failures at both the base and the head, so the suite is fully green on both sides. Nothing to fix, just recording the real numbers.

## Findings

### 1. high: a symlink body containing `..` reads any file outside the repo

`src/util/paths.ts:34`. Node's `fs.realpathSync` is the JavaScript implementation, and it collapses `..` lexically rather than applying it to the symlink-resolved prefix. `realpathSync.native` (the OS `realpath(3)`) does not.

Failure scenario, in a repo that contains a relative symlink to a shared directory, which is an ordinary KiCad library layout:

```
repo/sub  -> ../outside          (a normal relative symlink out to a shared dir)
repo/link -> sub/../outside/secret.txt
outside/secret.txt = "TOPSECRET"
```

The kernel resolves `repo/link` to `outside/secret.txt`. Node's JS `realpathSync` instead resolves the body to `repo/outside/secret.txt`, which does not exist, so it throws `ENOENT` and control falls into the dangling-link branch, which reaches the same lexical answer and concludes the path is contained. Measured against the compiled tool functions:

```
resolveInRepo(repo, 'link') -> ALLOWED
toolReadFile(repo, 'link')  -> "TOPSECRET\n"     <-- read from outside the repo
```

`toolEditFile` follows the same path and would overwrite the outside file.

### 2. high: a dangling symlink whose body contains `..` writes outside the repo

`src/util/paths.ts:41`. The dangling-link fallback does `path.resolve(path.dirname(head), readlinkSync(head))`, and `path.resolve` collapses `..` for the same reason.

```
repo/sub  -> ../outside
repo/link -> sub/../planted.txt   (target does not exist)
```

Lexically `sub/..` cancels back to the repo root, so the candidate looks like `repo/planted.txt` and is allowed. The kernel resolves `sub` to `outside/` first, so `..` lands in the repo's **parent**. Measured:

```
resolveInRepo(repo, 'link')            -> ALLOWED
toolWriteFile(repo, 'link', 'ESCAPED') -> wrote link
<base>/planted.txt exists              -> true      <-- created outside the repo
```

`toolWriteFile`'s existing-file guard does not help: `existsSync` follows the dangling link, sees nothing, and proceeds to create the file.

### 3. high: the 32-hop bound fails open, so a long dangling chain escapes

`src/util/paths.ts:40` and `src/util/paths.ts:54`. When `hops < 32` stops being true, `link` stays `null` and control falls through to the parent walk, which rebuilds the path under the repo root and **allows** it. The kernel follows 40 links, so a chain of 33 is still fully traversable by the subsequent `fs` call. Measured across the boundary:

```
chain=31: resolveInRepo=blocked, write=SandboxError, escaped=false
chain=32: resolveInRepo=blocked, write=SandboxError, escaped=false
chain=33: resolveInRepo=ALLOWED, write=ok,           escaped=true
```

A safety bound in a security rail has to refuse, never pass. The same applies to the `if (parent === head) return abs` fallback at line 54: returning the unresolved path there is also fail-open, and it is the one branch in the diff no test covers.

### 4. low: the in-repo guard rail asserts resolution, not the resolved value

`test/safety.test.ts`, the `allows a symlink that stays inside the repo` case asserts only `not.toThrow()`. An implementation that returned a wrong-but-contained path would still pass. Asserting the returned value (as the neighbouring `still allows real paths` test does) would pin it.

## Suggested fix

One change addresses all three: resolve the path the way the kernel does, one component at a time from the root, applying `..` to the already-resolved prefix, and fail closed on the hop bound. Missing components are simply appended, which preserves the write-target-does-not-exist behavior.

```ts
function realpathDeepest(abs: string): string | null {
  const root = path.parse(abs).root;
  let out = root;
  const queue = abs.slice(root.length).split(path.sep).filter(Boolean);
  let hops = 0;
  while (queue.length > 0) {
    const seg = queue.shift()!;
    if (seg === '.') continue;
    if (seg === '..') { out = path.dirname(out); continue; }   // applied to the RESOLVED prefix
    const next = path.join(out, seg);
    let isLink = false;
    try { isLink = lstatSync(next).isSymbolicLink(); }
    catch { out = next; continue; }                            // does not exist yet
    if (!isLink) { out = next; continue; }
    if (++hops > 40) return null;                              // fail closed, like the kernel's ELOOP
    const target = readlinkSync(next);
    if (path.isAbsolute(target)) out = path.parse(target).root;
    queue.unshift(...target.split(path.sep).filter(Boolean));
  }
  return out;
}
```

with the caller refusing on `null`:

```ts
const realRoot = realpathDeepest(root);
const realAbs = realpathDeepest(abs);
if (realRoot === null || realAbs === null) throw new SandboxError(p);
```

I applied exactly this in a scratch worktree and ran it: the three regression tests below pass, all 24 existing `safety.test.ts` cases pass unchanged (including both guard rails and the in-repo dangling-link case), and the full suite is 354 pass / 17 skip / 0 fail. Note this also removes the dependency on `fs.realpathSync` entirely, so the JS-versus-native discrepancy in finding 1 stops mattering.

## Regression tests

These three fail on the PR head as-is (`3 failed`) and pass with the fix above. Suggested as `test/safety.test.ts` additions:

```ts
it('rejects a dangling link whose body walks out through a symlinked dir', async () => {
  const base = realpathSync(await mkdtemp(path.join(tmpdir(), 'ch-dotdot-')));
  const repo = path.join(base, 'repo');
  await mkdir(repo);
  await mkdir(path.join(base, 'outside'));
  await symlink('../outside', path.join(repo, 'sub'));
  // Lexically "sub/.." cancels to the repo root; the kernel resolves `sub`
  // first, so ".." lands in the parent of the repo.
  await symlink('sub/../planted.txt', path.join(repo, 'link'));

  expect(() => resolveInRepo(repo, 'link')).toThrow(SandboxError);
  await expect(toolWriteFile(repo, 'link', 'ESCAPED\n')).rejects.toThrow(SandboxError);
  expect(existsSync(path.join(base, 'planted.txt'))).toBe(false);
});

it('rejects reading an existing outside file through a ".." link body', async () => {
  const base = realpathSync(await mkdtemp(path.join(tmpdir(), 'ch-dotdot-read-')));
  const repo = path.join(base, 'repo');
  const outside = path.join(base, 'outside');
  await mkdir(repo);
  await mkdir(outside);
  await writeFile(path.join(outside, 'secret.txt'), 'TOPSECRET\n', 'utf8');
  await symlink('../outside', path.join(repo, 'sub'));
  await symlink('sub/../outside/secret.txt', path.join(repo, 'link'));

  expect(() => resolveInRepo(repo, 'link')).toThrow(SandboxError);
  await expect(toolReadFile(repo, 'link')).rejects.toThrow(SandboxError);
});

it('fails closed on a dangling chain longer than the hop bound', async () => {
  const base = realpathSync(await mkdtemp(path.join(tmpdir(), 'ch-hops-')));
  const repo = path.join(base, 'repo');
  const outside = path.join(base, 'outside');
  await mkdir(repo);
  await mkdir(outside);
  const N = 33; // one past the 32-hop bound; the kernel still follows 40
  await symlink(path.join(outside, 'x.txt'), path.join(repo, `hop${N}`));
  for (let i = N - 1; i >= 1; i--) {
    await symlink(path.join(repo, `hop${i + 1}`), path.join(repo, `hop${i}`));
  }

  expect(() => resolveInRepo(repo, 'hop1')).toThrow(SandboxError);
  await expect(toolWriteFile(repo, 'hop1', 'ESCAPED\n')).rejects.toThrow(SandboxError);
  expect(existsSync(path.join(outside, 'x.txt'))).toBe(false);
});
```

They need `realpathSync` and `toolReadFile` added to the existing imports. `realpathSync` on the temp dir is what keeps the assertions honest on macOS, where `/tmp` is itself a symlink.

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| Spec-gated in (mutation tools structurally absent until a proposal validates) | pass, `src/agent/tools.ts` untouched; this gates paths, not tool availability |
| Verification-gated out (ERC/DRC, repair cycles, rollback) | pass, mutation and repair path untouched |
| `check`/`verify` stays LLM-free and network-free | pass, `paths.ts` adds only `node:fs`; grepped for provider, key and network references in `check.ts` and `paths.ts`, none |
| No sexp serialization | pass, `src/kicad/` untouched |
| Sync-obligations ledger | pass, untouched |
| Secrets and redaction, `.gitignore` entries | pass, untouched |
| Spec coherence | pass, AC-4.2 already promises this rail, so no SPEC.md or change-artifact movement is required; no `openspec/` files changed, so `openspec validate` was not applicable |
| New tests deterministic, no network, no wall-clock, all assert | pass |
| CI and mergeability | pass, all three checks green (`CodeRabbit`, `cla`, `template`), `mergeable: MERGEABLE` |

The AC-4.2 rail itself is the one that does not hold yet, per findings 1 to 3.

## Scope and process notes

Scope is clean: two files, no unrelated changes, no new dependencies, no public API change. The test-to-source ratio is good and the revert check described in the PR body (restoring the old `paths.ts` and confirming exactly the escape test fails) is the right instinct and worth keeping in the follow-up.

This is the first automated pass on this PR. The human review round that preceded it caught the dangling-symlink hole and the weak guard rail, and both are genuinely addressed in `2c635cd`; the findings above are separate from those and were not previously raised.

aniket-intervue:
🤖 **Automated `/pr-review` pass** (Claude Code). Not a human review — findings below were reproduced locally against the PR head (`2c635cd`), but a maintainer should still sign off.

## Verdict: request changes — one residual escape, fixable in ~4 lines

This is a real fix for a real hole, and the write-up is unusually good: the "write target may not exist yet" and "repo root itself is symlinked" cases are both correctly identified, and the dangling-link follow-up (`2c635cd`) closes the sharpest version of the second-order bug. The containment argument holds for every shape I probed **except one**: a dangling symlink chain of 33–40 hops still resolves to a path under the repo root and lets `write_file` create a file outside it. I reproduced it end-to-end and verified a fix.

```
lines: +154 / -1 (net +153) across 2 files  ·  src +65/-0, test +89/-1, docs 0, spec 0
new vs net: 0 new files, 2 modified; 65 new src lines (all in src/util/paths.ts)
tests: +89 test lines vs +65 src lines (1.37:1); suite 345/17/0 -> 351/17/0 (pass/skip/fail, base -> head)
diff coverage: 100% (65/65 changed src lines executed); 1 uncovered branch: src/util/paths.ts:54
deps/audit: package.json unchanged — n/a
CI: 3/3 pass (CodeRabbit, cla, template) · MERGEABLE · branch is 10 commits behind main
```

**How each number was obtained.** Sizes from `git diff --numstat $BASE..pr-86` with `BASE=$(git merge-base upstream/main pr-86)` = `6dcca30`; no added files (`--diff-filter=A` empty). Suite totals from `npx vitest run` on `$BASE` and on the PR head in clean worktrees — both green here (the 8 `kicad-cli not found` failures in the author's log do not reproduce in this environment; `kicad-cli` is on PATH). Diff coverage measured with `@vitest/coverage-v8@3.2.7`, `--coverage.include='src/util/paths.ts'`, intersecting the per-statement JSON with the diff's added lines: 51/51 statements executed, so every changed source line is exercised; the only gap is one branch, listed as finding 3. `npm test` was not run against a live provider (no keys set) — the LLM-gated integration tests were skipped, as expected for a change that touches neither.

---

## Findings

### 1. HIGH — a 33–40-hop dangling symlink chain still escapes the sandbox

**Claim.** [paths.ts:40](src/util/paths.ts#L40) — when the hop budget runs out, the code falls through to the parent walk instead of failing closed, which rebuilds the *link's own* path under the repo root. That is exactly the bug `2c635cd` fixed for short chains, reintroduced at the budget boundary.

```ts
if (hops < 32 && lstatSync(head).isSymbolicLink()) {   // budget exhausted -> link stays null
  ...
}
if (link !== null) { hops++; head = link; continue; }
const parent = path.dirname(head);
tail.unshift(path.basename(head));                      // <-- rebuilds under the repo root
```

**Failure scenario (reproduced).** A repo containing `hop1 -> hop2 -> … -> hop34 -> ../outside/planted.txt`, all dangling. The chase stops at `hop33`, the parent walk rebuilds `<repo>/hop33`, containment passes, and `write_file` follows the chain out. Linux's `ELOOP` limit is 40, so the whole window 33 ≤ N ≤ 40 is exploitable; at 41 the kernel saves us by accident.

```
CHAIN N=32: blocked(SandboxError)
CHAIN N=33: ALLOWED  ESCAPED-WRITE      <-- outside/planted.txt created, content "X\n"
CHAIN N=34: ALLOWED  ESCAPED-WRITE
CHAIN N=39: ALLOWED  ESCAPED-WRITE
CHAIN N=40: ALLOWED  ESCAPED-WRITE
CHAIN N=41: ALLOWED  write-threw(ELOOP: too many symbolic links)
```

**Fix (verified: 351 pass / 17 skip / 0 fail with it applied, and all 24 existing `safety.test.ts` tests still green).** Fail closed on an unfollowable link:

```ts
      let link: string | null = null;
      try {
        if (lstatSync(head).isSymbolicLink()) {
          // Budget exhausted on a link we cannot follow: fail closed rather than
          // rebuilding the link's own path under the repo root.
          if (hops >= 32) throw new SandboxError(abs);
          link = path.resolve(path.dirname(head), readlinkSync(head));
        }
      } catch (e) {
        if (e instanceof SandboxError) throw e;
        // head does not exist at all; fall through to the parent walk.
      }
```

**Regression test** (fails on `2c635cd`, passes with the fix — mirrors the existing dangling-chain test):

```ts
  it('rejects a long dangling symlink chain that ends outside the repo', async () => {
    const base = await mkdtemp(path.join(tmpdir(), 'ch-dangling-long-'));
    const repo = path.join(base, 'repo');
    const outside = path.join(base, 'outside');
    await mkdir(repo, { recursive: true });
    await mkdir(outside, { recursive: true });
    // 34 hops: past realpathDeepest's 32-hop budget but under the kernel's
    // 40-link ELOOP limit, so the fs call still follows it all the way out.
    const N = 34;
    await symlink(path.join(outside, 'planted.txt'), path.join(repo, `hop${N}`));
    for (let k = N - 1; k >= 1; k--) {
      await symlink(path.join(repo, `hop${k + 1}`), path.join(repo, `hop${k}`));
    }

    expect(() => resolveInRepo(repo, 'hop1')).toThrow(SandboxError);
    await expect(toolWriteFile(repo, 'hop1', 'escaped\n')).rejects.toThrow(SandboxError);
    expect(existsSync(path.join(outside, 'planted.txt'))).toBe(false);
  });
```

Yes, a 33-link dangling chain is contrived — but so is the threat model the PR already accepts (a pre-existing symlink the model walks into), and the whole point of the second commit was that "rebuild it under the repo root" is the wrong default. Fail-closed is the cheaper invariant to hold.

### 2. MEDIUM — `search` is the fourth file tool and still walks out of the repo (pre-existing, but in this PR's scope)

**Claim.** [filetools.ts:100-134](src/agent/filetools.ts#L100-L134) — `toolSearch` never calls `resolveInRepo`. It walks from `repoRoot` with `readdir` + `stat` (which follows symlinks), so the same symlink this PR now blocks for `read_file` still lets `search` read and return file contents from outside the repo.

**Failure scenario (reproduced against the PR head).** Repo containing `lib -> ../outside`, with `outside/secret.txt` holding `sk-TOPSECRET-outside-the-repo`:

```
toolSearch(repo, 'TOPSECRET') => [{"file":"lib/secret.txt","line":1,"text":"sk-TOPSECRET-outside-the-repo"}]
```

The content leaves the sandbox and lands in the model's context. Note the PR's own premise — "a symlinked KiCad symbol or footprint library is a normal project layout" — makes this at least as reachable as the read path it does fix, and `search` is the tool a model reaches for first.

**Fix.** In `walk`, skip entries whose real path leaves the root: `lstat` the entry and, when it is a symlink, run it through `resolveInRepo(repoRoot, path.relative(repoRoot, abs))` inside a `try`/`continue` — or more simply reuse `realpathDeepest` and drop anything not under `realpathDeepest(repoRoot)`. That also fixes the related nuisance: an in-repo symlink loop (`d/self -> .`) currently makes `toolSearch` reject with a raw `ELOOP` instead of skipping the cycle (verified).

This is not a regression from this PR, so it is legitimate to split into a follow-up issue — but the PR description claims the class is closed ("can no longer reach outside the repo root through a symlink that lives inside it"), and that sentence is only true for three of the four file tools. Worth either fixing here or narrowing the claim.

### 3. LOW — the filesystem-root fallback in `realpathDeepest` fails open (and is the one uncovered branch)

[paths.ts:54](src/util/paths.ts#L54) — `if (parent === head) return abs;` returns the *lexical* path when the walk reaches the filesystem root without finding anything that exists, which then trivially passes the containment check. This is the branch coverage flagged (`54:b0`, the only uncovered one in the file). It is unreachable in practice, since `/` always resolves, so I am not claiming a live bug — but given finding 1 has the same shape, `throw new SandboxError(abs)` here costs nothing and makes "unknown ⇒ refuse" the uniform rule for the whole function.

### 4. LOW — the in-repo dangling-link guard rail asserts less than the PR conversation says it does

[safety.test.ts:106-115](test/safety.test.ts#L106-L115) — `still allows a dangling symlink whose target is inside the repo` only asserts `not.toThrow()`. The [04:12 UTC comment](https://github.com/chouhanindustries/copperhead/pull/86) says "Added the write plus assertions on both existence and content", but no such assertion exists at the pushed head (`2c635cd`, the last commit) — it looks like that edit was verified locally and never pushed, possibly lost in the rebase from `a6f7a55`.

I checked the behavior directly rather than assuming, and **it is correct**: `toolWriteFile(repo, 'pending', …)` returns `wrote pending` and `repo/sub/future.txt` is created inside the repo. So this is a test-strength gap, not a bug — but it is precisely the guard rail that keeps a future tightening (including the fix in finding 1) from silently breaking legitimate writes, so it should assert the write, not just the resolution:

```ts
    await expect(toolWriteFile(repo, 'pending', 'inside\n')).resolves.toContain('wrote');
    expect(await readFile(path.join(repo, 'sub', 'future.txt'), 'utf8')).toBe('inside\n');
```

---

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| **Spec-gated in** (`edit_file`/`write_file` structurally absent until a proposal validates) | **pass** — `src/agent/tools.ts` untouched; `resolveInRepo` gates paths, not tool availability |
| **Verification-gated out** (ERC/DRC, repair, rollback) | **pass** — mutation/repair path untouched |
| **`check`/`verify` LLM-free and network-free** | **pass** — `paths.ts` adds only `node:fs` (`lstatSync`, `readlinkSync`, `realpathSync`); no provider, key, or network reachable from `src/commands/check` |
| **No sexp serialization** | **pass** — `src/kicad/` untouched |
| **Sync-obligations ledger / commit gating** | **pass** — untouched |
| **Secrets** (`sk-` redaction at write time, `.gitignore`) | **pass for this diff** — redaction and `.gitignore` unchanged. Flagged separately: finding 2 lets `search` pull an `sk-` secret *from outside the repo* into context, where redaction is not the control that applies |
| **Spec coherence** | **pass** — AC-4.2 ([SPEC.md:456](openspec/specs/SPEC.md#L456)) already requires rejecting out-of-root paths; this closes a hole in the implementation rather than changing the promise. No planning artifacts touched, so `openspec validate` was not required and was not run |

**Scope:** clean — 2 files, both on the stated fix. No scope creep. Typecheck passes (`tsc --noEmit`, exit 0). New tests are deterministic: no network, no `Date.now()`/`Math.random()`, no order dependence, each builds its own `mkdtemp` fixture and asserts.

**Also verified as correctly handled** (probed, no finding): read/write/edit through an existing outward symlink; a deep non-existent path under an outward symlink (`lib/a/b/c.txt`); a *relative* dangling symlink (`../outside/x.txt`); a leaf symlink inside an existing in-repo directory pointing out; a repo root that is itself a symlink (returns the lexical path under the link, as intended for the macOS `/tmp` case); in-repo symlinks and in-repo dangling symlinks still allowed.

**On TOCTOU:** agreed with the author's framing — closing it properly needs `O_NOFOLLOW` at each call site and is a much larger change. Out of scope here, and the residual risk is correctly stated in the description.

Fix finding 1 and this is a clear approve.

AniketR10:
@DPS0340 plz fix the reviews shortcomings and merge conflicts

PR #88, opened 2026-07-26
fix(search): keep toolSearch inside the repo and survive symlink loops

### LINKED ISSUE #87: toolSearch bypasses the sandbox: reads through symlinks and dies on a symlink loop
### Summary

`toolSearch()` in [filetools.ts](src/agent/filetools.ts) walks the repo with `readdir` + `stat` and never calls `resolveInRepo`. Two consequences, both reproduced against the real exported function:

1. **It reads outside the repo through a symlink.** `stat` follows links, so a symlinked directory inside the repo is descended into and its contents are returned to the model.
2. **A symlink loop kills the search entirely** with `ELOOP`, so `search` stops working for the whole repo, not just the affected subtree.

This is a sibling of #85 but a genuinely separate code path: `toolSearch` is the one file tool that does not go through `resolveInRepo`, so fixing the sandbox helper does not cover it.

### Reproduction 1: reading outside the repo

Temp repo with `link -> ../outside`, where `outside/leak.txt` contains `needle SECRET outside`:

```
matches: 3
  b.txt 1 "needle two"
  link/leak.txt 1 "needle SECRET outside"     <-- outside the repo
  sub/a.txt 1 "needle here"
```

The match is even reported with a path (`link/leak.txt`) that looks repo-relative, so neither the model nor a human reading the transcript can tell the content came from outside the sandbox.

### Reproduction 2: symlink loop

Temp repo containing `sub/loop -> <repo root>`:

```
ERROR: ELOOP: too many symbolic links encountered, stat
       '/tmp/srchprobe/repo/sub/loop/sub/loop/sub/loop/sub/loop/sub/loop/sub/l...'
```

`dispatchTool` catches this and hands the model `error: ELOOP: ...`, so the run does not crash — but `search` is unusable for the rest of the run, and the error text gives no hint that a symlink is the cause. Note the walk also has no visited-inode set, so a loop is traversed until the OS refuses rather than being detected.

### Why it's reachable

Same as #85: the agent does not need to create the link. A symlinked KiCad symbol or footprint library is an ordinary project layout, and `.history/`, vendored directories, or any user repo may already contain one.

### Expected

- `toolSearch` does not descend into a directory that resolves outside the repo root, and does not return matches from files outside it.
- A symlink loop does not take the whole `search` tool down.

### Notes on the current walk

Two smaller things in the same function, mentioned for context rather than as separate reports:

- `SKIP_DIRS` is matched on the bare entry name, so it correctly skips `.git` at any depth.
- The `maxMatches` guard is checked before `readdir` and after each entry, so a large repo terminates fine. Only the symlink cases misbehave.

DPS0340:
## What

`toolSearch()` in [filetools.ts](src/agent/filetools.ts) no longer returns matches from files outside the repo root, and no longer dies with `ELOOP` when the repo contains a symlink loop.

## Why

Closes #87.

`toolSearch` is the one file tool that does not go through `resolveInRepo`: it walks with `readdir` + `stat` directly. `stat` follows symlinks, so two things went wrong. Both reproduced against the real exported function, not a reimplementation.

**1. It read outside the repo.** Temp repo with `link -> ../outside`, where `outside/leak.txt` contains `needle SECRET outside`:

```text
matches: 3
  b.txt 1 "needle two"
  link/leak.txt 1 "needle SECRET outside"     <-- outside the repo
  sub/a.txt 1 "needle here"
```

The match is reported as `link/leak.txt`, which reads as a repo-relative path, so neither the model nor a human reading the transcript can tell the content came from outside the sandbox.

**2. A symlink loop took the whole tool down.** Temp repo containing `sub/loop -> <repo root>`:

```text
ERROR: ELOOP: too many symbolic links encountered, stat
       '.../repo/sub/loop/sub/loop/sub/loop/sub/loop/sub/l...'
```

`dispatchTool` catches this so the run does not crash, but `search` is unusable for the rest of the run and the error text gives no hint that a symlink caused it. The walk had no visited set, so a loop was traversed until the OS refused rather than being detected.

This is a sibling of #85 but a separate code path: fixing `resolveInRepo` does not cover `toolSearch`, because `toolSearch` never calls it.

It is reachable without the agent creating anything. A symlinked KiCad symbol or footprint library is an ordinary project layout, and `.history/`, vendored directories, or any user repo may already contain one.

## Design

Three changes inside `toolSearch`, all local to the walk:

- **`lstat` before `stat`.** `lstat` describes the entry itself; `stat` describes the link target and hides that an entry is a link at all. A symlink whose real path lands outside the repo is skipped.
- **`insideRoot()` helper.** Resolves a path with `realpath` and applies the same containment rule `resolveInRepo` uses. The repo root is resolved once up front, because the root itself can sit behind a symlink (on macOS `/tmp` is `/private/tmp`), and comparing a resolved child against an unresolved root would reject everything.
- **A `seenDirs` set of real paths.** A directory whose real path was already walked is skipped, which terminates loops by construction instead of relying on the OS to throw.

Every `stat`/`realpath` call is `.catch(() => null)` with an explicit skip, so an entry that vanishes mid-walk or a broken symlink is stepped over rather than failing the search. That matches how the function already treats unreadable files.

A symlink that stays inside the repo is still followed: those are normal in KiCad projects and must keep working.

## Spec / OpenSpec

No spec-level behavior change. This brings `toolSearch` in line with the containment rule AC-4.2 already states for tool paths, so SPEC.md and the change artifacts stay as they are.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | 323 pass, 14 skip, 8 pre-existing failures (see below) |

New tests, in [safety.test.ts](test/safety.test.ts) under `file tools`:

- **the escape**: asserts the exact match list is `['b.txt', 'sub/a.txt']` and that no match text contains `SECRET`. Asserting the full list matters here, since the old behavior returned a plausible-looking extra path rather than an obvious error
- **the loop**: a repo containing `sub/loop -> <root>` returns its single real match instead of throwing
- **guard rail**: a symlink that stays inside the repo is still followed, so symlinked library dirs keep working

**Revert check.** With the new tests in place I restored the original `filetools.ts`: exactly the 2 behavioral tests fail, and the guard rail passes under both implementations, so it is genuinely guarding rather than riding along.

Sandbox variant used: none. Drove the shipped `toolSearch` directly, which is a closer test of this change than a CLI run: the escape and the loop both happen inside the walk, and calling it by hand lets the probe assert on the returned match list rather than on agent output.

Commands run and outcome:

```text
$ npx tsc src/agent/filetools.ts src/util/paths.ts --outDir /tmp/tscout --module esnext --target es2022 --moduleResolution bundler

$ node /tmp/srch2.mjs     # escape probe, before
matches: 3
  b.txt 1 "needle two"
  link/leak.txt 1 "needle SECRET outside"
  sub/a.txt 1 "needle here"

$ node /tmp/srch.mjs      # loop probe, before
ERROR: ELOOP: too many symbolic links encountered, stat '.../sub/loop/sub/loop/sub/l...'

$ node /tmp/srch2.mjs     # escape probe, after
matches: 2
  b.txt 1 "needle two"
  sub/a.txt 1 "needle here"

$ node /tmp/srch.mjs      # loop probe, after
matches: 1
  sub/a.txt 1
```

## Docs

None needed: the `search` tool's schema, description, and output format are unchanged. A repo with no symlinks behaves identically.

---

## Invariant checklist

- [x] **Spec-gated in**: unchanged. `search` has `requiresUnlock: false` before and after; this PR changes what it may read, not when it exists.
- [x] **`check`/`verify` stays LLM-free and network-free**: this adds only `node:fs/promises` calls (`lstat`, `realpath`), no provider, key, or network access.
- [ ] N/A **Secrets**: no change to redaction or key handling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository file search safety using symlink-aware traversal.
  * Prevented searches from accessing content outside the repository root.
  * Added safeguards to skip broken/vanished links and avoid symlink cycle failures, while preserving correct matches for valid in-repo links.
  * Reduced duplicate results and fixed glob behavior when content is reachable via both real paths and aliases.
* **Tests**
  * Expanded symlink-focused test coverage for escaping, cycles, deduplication, alias reachability, and dangling symlink handling.

### DIFF
--- a/src/agent/filetools.ts
+++ b/src/agent/filetools.ts
@@ -111,7 +111,9 @@ export async function toolSearch(
   // pointing at an ancestor is otherwise descended until the OS throws ELOOP,
   // which takes down search for the whole run.
   const realRoot = await realpath(repoRoot).catch(() => path.resolve(repoRoot));
-  const seenDirs = new Set<string>();
+  // Seeded with the root: a link pointing back at it would otherwise be walked
+  // once more, duplicating every root-level match under the link's path.
+  const seenDirs = new Set<string>([realRoot]);
 
   /** The real path of `abs`, or null when it resolves outside the repo. */
   async function insideRoot(abs: string): Promise<string | null> {
--- a/src/agent/filetools.ts
+++ b/src/agent/filetools.ts
@@ -111,9 +111,11 @@ export async function toolSearch(
   // pointing at an ancestor is otherwise descended until the OS throws ELOOP,
   // which takes down search for the whole run.
   const realRoot = await realpath(repoRoot).catch(() => path.resolve(repoRoot));
-  // Seeded with the root: a link pointing back at it would otherwise be walked
-  // once more, duplicating every root-level match under the link's path.
-  const seenDirs = new Set<string>([realRoot]);
+  // Loop protection is scoped to the *current path*, not the whole walk. A cycle
+  // always means revisiting an ancestor, so an ancestor set terminates the walk
+  // while still visiting siblings: a global set would also drop non-cyclic
+  // aliases, hiding a real directory behind an in-repo symlink pointing at it.
+  const ancestors = new Set<string>([realRoot]);
 
   /** The real path of `abs`, or null when it resolves outside the repo. */
   async function insideRoot(abs: string): Promise<string | null> {
@@ -140,9 +142,10 @@ export async function toolSearch(
       if (st.isDirectory()) {
         const real = await insideRoot(abs);
         if (real === null) continue;
-        if (seenDirs.has(real)) continue; // already walked: a symlink loop
-        seenDirs.add(real);
+        if (ancestors.has(real)) continue; // on the current path already: a loop
+        ancestors.add(real);
         await walk(abs);
+        ancestors.delete(real);
       } else if (st.size < 5_000_000 && (!globRe || globRe.test(rel))) {
         let text: string;
         try {
--- a/src/agent/filetools.ts
+++ b/src/agent/filetools.ts
@@ -1,4 +1,4 @@
-import { readFile, writeFile, mkdir, readdir, stat } from 'node:fs/promises';
+import { readFile, writeFile, mkdir, readdir, lstat, realpath, stat } from 'node:fs/promises';
 import { existsSync } from 'node:fs';
 import path from 'node:path';
 import { resolveInRepo, isKicadFile } from '../util/paths.js';
@@ -106,14 +106,40 @@ export async function toolSearch(
   const re = new RegExp(pattern);
   const globRe = glob ? globToRegex(glob) : null;
   const matches: SearchMatch[] = [];
+  // The walk follows symlinks (stat does), so it needs the same containment rule
+  // the other file tools get from resolveInRepo, plus loop protection: a link
+  // pointing at an ancestor is otherwise descended until the OS throws ELOOP,
+  // which takes down search for the whole run.
+  const realRoot = await realpath(repoRoot).catch(() => path.resolve(repoRoot));
+  const seenDirs = new Set<string>();
+
+  /** The real path of `abs`, or null when it resolves outside the repo. */
+  async function insideRoot(abs: string): Promise<string | null> {
+    const real = await realpath(abs).catch(() => null);
+    if (real === null) return null;
+    if (real !== realRoot && !real.startsWith(realRoot + path.sep)) return null;
+    return real;
+  }
+
   async function walk(dir: string): Promise<void> {
     if (matches.length >= maxMatches) return;
     for (const entry of await readdir(dir)) {
       if (SKIP_DIRS.has(entry)) continue;
       const abs = path.join(dir, entry);
       const rel = path.relative(repoRoot, abs);
-      const st = await stat(abs);
+      // lstat describes the entry itself; stat would describe the link target
+      // and hide that this is a link at all.
+      const link = await lstat(abs).catch(() => null);
+      if (link === null) continue; // vanished mid-walk
+      if (link.isSymbolicLink() && (await insideRoot(abs)) === null) continue; // escapes the repo
+
+      const st = await stat(abs).catch(() => null);
+      if (st === null) continue; // broken link or race
       if (st.isDirectory()) {
+        const real = await insideRoot(abs);
+        if (real === null) continue;
+        if (seenDirs.has(real)) continue; // already walked: a symlink loop
+        seenDirs.add(real);
         await walk(abs);
       } else if (st.size < 5_000_000 && (!globRe || globRe.test(rel))) {
         let text: string;
--- a/src/agent/filetools.ts
+++ b/src/agent/filetools.ts
@@ -135,17 +135,20 @@ export async function toolSearch(
       // and hide that this is a link at all.
       const link = await lstat(abs).catch(() => null);
       if (link === null) continue; // vanished mid-walk
-      if (link.isSymbolicLink() && (await insideRoot(abs)) === null) continue; // escapes the repo
+      // Resolved once and reused below: containment and loop detection are the
+      // same question about the same entry, and realpath is a syscall.
+      const real = link.isSymbolicLink() ? await insideRoot(abs) : null;
+      if (link.isSymbolicLink() && real === null) continue; // escapes the repo
 
       const st = await stat(abs).catch(() => null);
       if (st === null) continue; // broken link or race
       if (st.isDirectory()) {
-        const real = await insideRoot(abs);
-        if (real === null) continue;
-        if (ancestors.has(real)) continue; // on the current path already: a loop
-        ancestors.add(real);
+        const realDir = real ?? (await insideRoot(abs));
+        if (realDir === null) continue;
+        if (ancestors.has(realDir)) continue; // on the current path already: a loop
+        ancestors.add(realDir);
         await walk(abs);
-        ancestors.delete(real);
+        ancestors.delete(realDir);
       } else if (st.size < 5_000_000 && (!globRe || globRe.test(rel))) {
         let text: string;
         try {
--- a/test/safety.test.ts
+++ b/test/safety.test.ts
@@ -161,6 +161,23 @@ describe('file tools', () => {
     expect(matches[0]!.file).toBe(path.join('sub', 'a.txt'));
   });
 
+  it('search does not report a file twice through a link back to the repo root', async () => {
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-search-dup-'));
+    const repo = path.join(base, 'repo');
+    await mkdir(path.join(repo, 'sub'), { recursive: true });
+    // A match at the ROOT is what exposes this: the loop test above only has a
+    // file under sub/, so re-walking the root produced nothing to duplicate.
+    await writeFile(path.join(repo, 'a.txt'), 'needle A\n', 'utf8');
+    await writeFile(path.join(repo, 'sub', 'b.txt'), 'needle B\n', 'utf8');
+    await symlink(repo, path.join(repo, 'sub', 'back'));
+
+    const matches = await toolSearch(repo, 'needle');
+
+    // Without seeding seenDirs with the root, this returns a third match at
+    // sub/back/a.txt — the same file reported under the link's path.
+    expect(matches.map((m) => m.file).sort()).toEqual(['a.txt', path.join('sub', 'b.txt')]);
+  });
+
   it('search still follows a symlink that stays inside the repo', async () => {
     const base = await mkdtemp(path.join(tmpdir(), 'ch-search-in-'));
     const repo = path.join(base, 'repo');
--- a/test/safety.test.ts
+++ b/test/safety.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { mkdir, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises';
+import { mkdir, mkdtemp, readFile, rm, symlink, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
 import { resolveInRepo, SandboxError, isKicadFile } from '../src/util/paths.js';
@@ -125,6 +125,55 @@ describe('file tools', () => {
     const mdOnly = await toolSearch(dir, 'KEY_DAH', '**/*.md');
     expect(mdOnly).toHaveLength(1);
     expect(mdOnly[0]!.file).toBe('a.md');
+  });
+
+  it('search does not read through a symlink that leaves the repo', async () => {
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-search-esc-'));
+    const repo = path.join(base, 'repo');
+    const outside = path.join(base, 'outside');
+    await mkdir(path.join(repo, 'sub'), { recursive: true });
+    await mkdir(outside, { recursive: true });
+    await writeFile(path.join(repo, 'b.txt'), 'needle two\n', 'utf8');
+    await writeFile(path.join(repo, 'sub', 'a.txt'), 'needle here\n', 'utf8');
+    await writeFile(path.join(outside, 'leak.txt'), 'needle SECRET outside\n', 'utf8');
+    await symlink(outside, path.join(repo, 'link'));
+
+    const matches = await toolSearch(repo, 'needle');
+
+    // The escaping match used to come back as "link/leak.txt", which reads as a
+    // repo-relative path and hides that the content was outside the sandbox.
+    expect(matches.map((m) => m.file).sort()).toEqual(['b.txt', path.join('sub', 'a.txt')]);
+    expect(matches.some((m) => m.text.includes('SECRET'))).toBe(false);
+  });
+
+  it('search survives a symlink loop instead of dying with ELOOP', async () => {
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-search-loop-'));
+    const repo = path.join(base, 'repo');
+    await mkdir(path.join(repo, 'sub'), { recursive: true });
+    await writeFile(path.join(repo, 'sub', 'a.txt'), 'needle here\n', 'utf8');
+    await symlink(repo, path.join(repo, 'sub', 'loop'));
+
+    // Previously this walked loop/sub/loop/sub/... until the OS threw, taking
+    // the search tool down for the rest of the run.
+    const matches = await toolSearch(repo, 'needle');
+
+    expect(matches).toHaveLength(1);
+    expect(matches[0]!.file).toBe(path.join('sub', 'a.txt'));
+  });
+
+  it('search still follows a symlink that stays inside the repo', async () => {
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-search-in-'));
+    const repo = path.join(base, 'repo');
+    await mkdir(path.join(repo, 'real'), { recursive: true });
+    await writeFile(path.join(repo, 'real', 'lib.txt'), 'needle inside\n', 'utf8');
+    await symlink(path.join(repo, 'real'), path.join(repo, 'alias'));
+
+    const matches = await toolSearch(repo, 'needle');
+
+    // Symlinked library dirs are a normal KiCad layout: the file is reachable
+    // by both names, and neither traversal is an escape.
+    expect(matches.length).toBeGreaterThanOrEqual(1);
+    expect(matches.every((m) => m.text.includes('needle inside'))).toBe(true);
   });
 
   it('isKicadFile covers the design formats', () => {
--- a/test/safety.test.ts
+++ b/test/safety.test.ts
@@ -183,14 +183,37 @@ describe('file tools', () => {
     const repo = path.join(base, 'repo');
     await mkdir(path.join(repo, 'real'), { recursive: true });
     await writeFile(path.join(repo, 'real', 'lib.txt'), 'needle inside\n', 'utf8');
-    await symlink(path.join(repo, 'real'), path.join(repo, 'alias'));
+    // A symlinked *file*, not a directory: a directory alias is subject to
+    // readdir ordering, so only a file link pins "the link was followed"
+    // deterministically.
+    await symlink(path.join(repo, 'real', 'lib.txt'), path.join(repo, 'alias.txt'));
 
     const matches = await toolSearch(repo, 'needle');
 
-    // Symlinked library dirs are a normal KiCad layout: the file is reachable
-    // by both names, and neither traversal is an escape.
-    expect(matches.length).toBeGreaterThanOrEqual(1);
-    expect(matches.every((m) => m.text.includes('needle inside'))).toBe(true);
+    // Symlinked library paths are a normal KiCad layout: the file is reachable
+    // by both names, and neither traversal is an escape. Asserting the exact
+    // set is what makes this a guard rail — refusing to follow the link drops
+    // alias.txt and fails here.
+    expect(matches.map((m) => m.file).sort()).toEqual(['alias.txt', path.join('real', 'lib.txt')]);
+  });
+
+  it('search reports a directory reachable by both its real name and an alias', async () => {
+    const base = await mkdtemp(path.join(tmpdir(), 'ch-search-alias-'));
+    const repo = path.join(base, 'repo');
+    await mkdir(path.join(repo, 'zreal'), { recursive: true });
+    await writeFile(path.join(repo, 'zreal', 'lib.txt'), 'needle inside\n', 'utf8');
+    // Sorts before 'zreal', so readdir yields the alias first: with walk-global
+    // loop detection the real directory is pruned and a path-anchored glob
+    // finds nothing.
+    await symlink(path.join(repo, 'zreal'), path.join(repo, 'aalias'));
+
+    expect((await toolSearch(repo, 'needle')).map((m) => m.file).sort()).toEqual([
+      path.join('aalias', 'lib.txt'),
+      path.join('zreal', 'lib.txt'),
+    ]);
+    expect((await toolSearch(repo, 'needle', 'zreal/**')).map((m) => m.file)).toEqual([
+      path.join('zreal', 'lib.txt'),
+    ]);
   });
 
   it('isKicadFile covers the design formats', () => {
--- a/test/safety.test.ts
+++ b/test/safety.test.ts
@@ -216,6 +216,17 @@ describe('file tools', () => {
     ]);
   });
 
+  it('search steps over a dangling symlink instead of failing the walk', async () => {
+    const repo = await mkdtemp(path.join(tmpdir(), 'ch-search-dangle-'));
+    await writeFile(path.join(repo, 'a.txt'), 'needle A\n', 'utf8');
+    await symlink(path.join(repo, 'gone'), path.join(repo, 'dangle'));
+
+    // A broken link is an ordinary thing to find in a checkout. The arm that
+    // actually saves the walk is realpath's in insideRoot: it fails first and
+    // drops the entry, so stat is never reached with a dangling target.
+    expect((await toolSearch(repo, 'needle')).map((m) => m.file)).toEqual(['a.txt']);
+  });
+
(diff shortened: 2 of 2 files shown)


DPS0340:

DPS0340:

DPS0340:
Confirmed and fixed in `ebe904c` — thank you, this was a real defect and my own test suite missed it.

I reproduced it before changing anything. A repo with a match **at the root** plus `sub/back -> <root>`:

```text
matches: 3
   a.txt 1
   sub/b.txt 1
   sub/back/a.txt 1     <- same file, reported again under the link
```

Two real files, three results. After seeding `seenDirs` with `realRoot`:

```text
matches: 2
   a.txt 1
   sub/b.txt 1
```

**Why my existing loop test didn't catch it** is the part worth noting: that fixture only has a file under `sub/`, so re-walking the root found nothing to duplicate and the count stayed at 1. The bug needs a match *at the root* to become visible. Added a separate regression test that places one there and asserts the exact match list, with a comment explaining why the fixture differs from the neighbouring one.

Verification:

- 22 tests in `safety.test.ts` pass.
- Reverting just the `seenDirs` seeding fails **only** the new test, so it pins this fix specifically.
- Re-ran my original escape and ELOOP probes against the compiled tool — both still behave correctly, so the fix didn't regress the containment or loop-termination behaviour.
- Full suite: 324 pass, `tsc --noEmit` clean. The 8 failures are the pre-existing `KicadCliMissingError` ones from `kicad-cli` not being on PATH in my environment, unchanged before and after.

animesh-chouhan:
🤖 **Automated `/pr-review` pass** (Claude Code). Not a human review, and not a merge-gating review: findings below were each reproduced against the branch at `ebe904c` before being reported.

## Verdict: request changes (narrow)

The core of this PR is right and the two bugs it targets are genuinely fixed. One change introduced alongside the fix, the *global* `seenDirs` set, over-prunes the walk and silently loses matches in a repo with an ordinary in-repo directory symlink. It is a ~4 line fix that keeps all four new tests green, verified below. Everything else is nits.

Prior passes: no earlier automated pr-review comment on this PR. CodeRabbit's one actionable comment (seed `seenDirs` with the root) was addressed in `ebe904c` with a regression test; I re-checked it and it is correctly fixed. No commit under review is mine.

```
lines: +97 / -3 (net +94) across 2 files  ·  src +30/-2, test +67/-1, docs 0, spec 0
new vs net: 0 new files, 2 modified; 30 new src lines (20 executable) in one function
tests: +67 test lines, +4 tests; suite 328/14/0 -> 332/14/0 (pass/skip/fail, base -> head)
diff coverage: 100% of executable added src lines (20/20) executed; 4 added *branches* uncovered
  uncovered: filetools.ts:121, :135, :139, :142 (all the .catch(() => null) arms)
deps/audit: n/a, package.json unchanged
CI: the `CI / test` job never ran on this PR (fork PR, workflow approval pending)
```

Method for each number: sizes from `git diff --numstat $(git merge-base origin/main ebe904c)...ebe904c` (base `887346b`); no added files per `--diff-filter=A`. Suite counts from `npx vitest run` at `887346b` and at `ebe904c` in a clean worktree, with `kicad-cli` on PATH, so the 8 `KicadCliMissingError` failures reported in the PR body do not reproduce here: both runs are fully green. Diff coverage measured with `@vitest/coverage-v8` (`--coverage.reporter=json` over the full suite), intersecting `statementMap`/`branchMap` hit counts with the added line numbers from `git diff -U0`. `npm run typecheck` clean on head. Only `cla` and `template` check runs exist on `ebe904c` (`gh api .../commits/ebe904cd/check-runs`), so `npm test` has never run in project CI on this branch. I ran it instead, hence the base -> head numbers above.

---

## Findings

### 1. medium: a global `seenDirs` hides a real directory behind an in-repo symlink alias, and a path-anchored glob then finds nothing

[src/agent/filetools.ts:116](src/agent/filetools.ts#L116) and [:143](src/agent/filetools.ts#L143). `seenDirs` is keyed on real paths and never unwound, so it does not only break cycles: it also de-duplicates *non-cyclic* aliases. When a directory is reachable both by its real name and by an in-repo symlink, whichever `readdir` yields first wins and the other name is dropped from the results entirely.

Failure scenario, a normal KiCad layout with `symbols -> vendor/kicad-symbols`. Reproduced against the exported `toolSearch` at `ebe904c`, fixture `zreal/lib.txt` containing `needle inside` plus `aalias -> zreal` (the alias sorts first in this directory's `readdir` order):

```text
head  no-glob:        ["aalias/lib.txt"]
head  glob zreal/**:  []                                <-- the file exists and matches
base  no-glob:        ["aalias/lib.txt","zreal/lib.txt"]
base  glob zreal/**:  ["zreal/lib.txt"]
```

So `search {"pattern":"needle","glob":"zreal/**"}` returns `no matches` for a file that exists, matches, and is inside the repo. This is a regression against `main`, and it is order-dependent: with the alias named `zzalias` instead, the real path wins and the glob works. The agent's habitual glob is `**/*.ext`, which is unaffected, so this is a non-default path, but the tool's own description advertises glob filtering and a path-anchored one is the natural way to scope a search to a subtree.

Fix: scope loop detection to the current traversal path rather than the whole walk. A cycle always means revisiting an ancestor, so an ancestor set is sufficient for termination and does not prune siblings:

```ts
  const ancestors = new Set<string>([realRoot]);
  ...
        if (ancestors.has(real)) continue; // on the current path already: a symlink loop
        ancestors.add(real);
        await walk(abs);
        ancestors.delete(real);
```

Verified: with that change, all 22 tests in `test/safety.test.ts` still pass, including the ELOOP test and the `sub/back -> root` duplication test added in `ebe904c`, and the probe above returns `["aalias/lib.txt","zreal/lib.txt"]` with `glob zreal/**` correctly yielding `["zreal/lib.txt"]`.

If the de-duplication is wanted on purpose rather than as a side effect of loop protection, that is a defensible choice, but then it should be stated in the PR body and pinned by a test that asserts *which* path survives. Today it is neither, and it depends on `readdir` order.

### 2. low: the "guard rail" test cannot fail, so nothing pins in-repo symlink following

[test/safety.test.ts:181](test/safety.test.ts#L181). The fixture has `real/lib.txt` and `alias -> real`, and asserts only `matches.length >= 1` plus that every match contains the needle. `real/lib.txt` satisfies both on its own, so the assertion holds whether or not the alias is ever traversed.

Demonstrated: replacing the containment check with `if (link.isSymbolicLink()) continue;`, that is, refusing to follow *any* symlink, leaves all 22 tests in `test/safety.test.ts` passing. The PR body reads "the guard rail passes under both implementations, so it is genuinely guarding rather than riding along"; the opposite follows, a check that cannot fail is not guarding anything.

Fix: use a symlinked *file*, which is not affected by the directory de-duplication and so is deterministic regardless of finding 1:

```ts
    await symlink(path.join(repo, 'real', 'lib.txt'), path.join(repo, 'alias.txt'));
    const matches = await toolSearch(repo, 'needle');
    expect(matches.map((m) => m.file).sort()).toEqual(['alias.txt', path.join('real', 'lib.txt')]);
```

Verified against `ebe904c`: returns `["alias.txt","real/lib.txt"]`, and drops to one entry under the skip-all-symlinks mutant, so it discriminates.

### 3. low: the new `.catch(() => null)` arms are the only uncovered branches, and one of them is a third bug fix that is neither claimed nor tested

[src/agent/filetools.ts:121](src/agent/filetools.ts#L121), [:135](src/agent/filetools.ts#L135), [:139](src/agent/filetools.ts#L139), [:142](src/agent/filetools.ts#L142) are the four added branches with zero hits in the coverage run. The realistic trigger for all of them is a dangling symlink, which no fixture creates.

That matters more than a routine coverage nit, because the behavior change there is real and unmentioned. Measured with a repo containing `a.txt` plus `dangle -> does-not-exist`:

```text
base:  THREW: ENOENT: no such file or directory, stat '.../dangle'
head:  RESULT: ["a.txt"]
```

On `main` a single broken symlink anywhere in the repo aborts the entire search. This PR quietly fixes that as well, which is good, and it is exactly the sort of thing that regresses later without a test. Add one:

```ts
  it('search steps over a dangling symlink instead of failing the walk', async () => {
    const repo = await mkdtemp(path.join(tmpdir(), 'ch-search-dangle-'));
    await writeFile(path.join(repo, 'a.txt'), 'needle A\n', 'utf8');
    await symlink(path.join(repo, 'gone'), path.join(repo, 'dangle'));
    expect((await toolSearch(repo, 'needle')).map((m) => m.file)).toEqual(['a.txt']);
  });
```

`:142` (a non-symlink directory failing containment) stays unreachable outside a race; that one is fine to leave uncovered.

### 4. low (nit): `insideRoot` runs `realpath` twice for every symlinked directory

[src/agent/filetools.ts:136](src/agent/filetools.ts#L136) computes the real path to decide containment and discards it, then [:141](src/agent/filetools.ts#L141) computes it again for the same entry. Hoisting the first result and reusing it removes one `realpath` syscall per symlinked directory and makes the two containment checks obviously the same check.

### 5. low: CI never ran the suite on this branch

Only `cla` and `template` produced check runs for `ebe904c`; the `CI / test` job in `.github/workflows/ci.yml` (which is what runs `npm test` with real `kicad-cli`) has not executed, because this is a fork PR awaiting workflow approval. Nothing for the author to do; a maintainer needs to approve the run before merge. I ran the equivalent locally and it is green (332 pass, 14 skip, 0 fail), so this is a process gap rather than a risk signal.

---

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| Spec-gated in | pass. `src/agent/tools.ts` is untouched; `search` is `requiresUnlock: false` before and after, and no mutation tool's presence or gating changed. |
| Verification-gated out | pass, not applicable. The ERC/DRC, repair, and rollback paths are untouched. |
| `check`/`verify` LLM-free and network-free | pass. The only new imports are `lstat` and `realpath` from `node:fs/promises`. No provider, key, or network reference is added. |
| No sexp serialization | pass, not applicable. `src/kicad/` is untouched. |
| Sync-obligations ledger | pass, not applicable. Hooks and commit gating are untouched. |
| Secrets | pass. Redaction and `.gitignore` unchanged. The `needle SECRET outside` fixture string is a test sentinel, not a key, and does not match the `sk-[A-Za-z0-9_-]+` pattern. |
| Spec coherence | pass. No spec-level behavior change: this brings `toolSearch` into line with the containment rule AC-4.2 already states. No planning artifact changed, so `openspec validate` was not required and was not run. |
| Test determinism | pass. The four new tests use `mkdtemp` fixtures, no network, no clock, no RNG, and no cross-test ordering. The escape, loop, and duplication tests assert exact match lists; only the guard rail under-asserts (finding 2). |

## Scope and context

Scope is clean: 30 source lines, all inside `toolSearch`'s walk, plus tests. No scope creep.

One thing worth recording for whoever picks up #85: `resolveInRepo` in [src/util/paths.ts:15](src/util/paths.ts#L15) is still purely lexical, so `read_file` on `link/leak.txt` in the PR's own escape fixture still returns `"SECRET outside\n"` at `ebe904c`. That is correctly out of scope here, and the PR says so, but it means the containment story is only half closed: `search` no longer *surfaces* an escaping path, while `read_file` will still follow one the model learns about by other means.

DPS0340:
Finding 1 is a real regression and you are right about finding 2 — my "guard rail" claim was wrong. Both fixed, plus 3 and 4. Pushed as `b07dd1e` and `3310356`.

## 1. Global `seenDirs` over-prunes — fixed

I reproduced it before changing anything, and against `main` to confirm it is a regression rather than pre-existing:

```
head  no-glob:        ["aalias/lib.txt"]
head  glob zreal/**:  []
base  no-glob:        ["aalias/lib.txt","zreal/lib.txt"]
base  glob zreal/**:  ["zreal/lib.txt"]
```

Took your fix: loop detection is now an ancestor set scoped to the current traversal path, added on descent and removed on return. Your framing is the part that makes it obviously correct — a cycle always means revisiting an *ancestor*, so an ancestor set is sufficient for termination and pruning siblings was never required for it.

I did not take the "document the de-duplication and pin which path wins" option. Which name survives depends on `readdir` order, so it is not a behaviour worth freezing in a test.

## 2. The guard rail could not fail — you are right, and I should have checked

The PR body claimed the test "passes under both implementations, so it is genuinely guarding rather than riding along". I asserted that instead of testing it. Running your check: with `if (link.isSymbolicLink()) continue;` all 22 tests passed, so the test was pinning nothing.

Fixed the way you suggested, with a symlinked *file* so it does not depend on `readdir` order, asserting the exact match set. Now verified in both directions rather than claimed:

```
refuse all symlinks      -> 2 failed | 21 passed   (this test + the new alias test)
revert to the global set -> 1 failed | 22 passed   (the alias test)
clean                    -> 24 passed
```

## 3. Dangling symlink — pinned, with one correction

Confirmed the behaviour change is real and was unmentioned:

```
base:  THREW: ENOENT: no such file or directory, stat '.../dangle'
head:  RESULT: ["a.txt"]
```

Test added. One correction to the finding, which I only caught by testing rather than reading: the arm that saves the walk is **not** `stat`'s. Removing `stat(abs).catch(() => null)` leaves all 24 tests green, because a dangling link fails `realpath` inside `insideRoot` first and is dropped there. Removing *that* catch fails the new test. The test comment names the real reason.

## 4. Double `realpath` — fixed

Resolved once per entry and reused. As you say, it also makes it visible that containment and loop detection are asking the same question.

## 5. CI on fork PRs

Understood, nothing for me to do. Locally: `npm run typecheck` clean, `test/safety.test.ts` 24/24, full suite 326 pass / 14 skip. My 8 failures are `kicad-cli` absence in my environment and are identical before and after this change.

---

One thing I want to name, because it changed how I will write PR descriptions. Finding 2 was not a test-quality nit — I made a factual claim about my own test and it was false, and it took ten seconds to disprove. Asserting a test is a guard rail is worthless next to reverting the source and watching it fail. I have started doing that on my other open PRs.

animesh-chouhan:
🤖 **Automated `/pr-review` pass** (Claude Code). Not a human review, and not a merge-gating review. This is a **follow-up pass**: the [earlier automated pass](https://github.com/chouhanindustries/copperhead/pull/88#issuecomment-5082237490) reviewed `ebe904c`; this one reviews `5c06133` and reports only what moved since. No commit under review is mine.

## Verdict: approve with nits

All four code findings from the previous pass are fixed, and each fix is pinned by a test that I confirmed **discriminates** (reverting the source fails exactly that test, and nothing else). I looked for new defects in the reworked walk and found none. The only open items are documentation and process: the PR body still describes the design that was replaced, and project CI still has not run the suite on this branch.

The branch was rebased between the two passes (merge base moved from `887346b` to `6dcca30`), so the suite totals below are not comparable to the previous pass's.

```
lines: +137 / -3 (net +134) across 2 files  ·  src +36/-2, test +101/-1, docs 0, spec 0
new vs net: 0 new files, 2 modified; 36 new src lines (19 executable statements) in one function
tests: +101 test lines, +6 tests; suite 345/17/0 -> 351/17/0 (pass/skip/fail, base -> head)
diff coverage: 100% of added executable src statements (19/19); 17/20 added branch arms hit (measured)
  uncovered: filetools.ts:137, :144, :147 (race-only arms), plus the :113 realpath fallback (untracked by v8)
deps/audit: n/a, package.json unchanged
CI: `CI / test` has still never run on this branch (fork PR, workflow approval pending); `cla` and `template` pass
base: mergeable, but 16 commits behind origin/main
```

Method. Sizes from `git diff --numstat 6dcca30...5c06133`; no added files per `--diff-filter=A`. Suite counts from `npx vitest run` in clean worktrees at `6dcca30` and at `5c06133`, with `kicad-cli` on PATH, so the `KicadCliMissingError` failures reported in the PR body do not reproduce here: both runs are fully green. Diff coverage measured with `@vitest/coverage-v8` (`--coverage.reporter=json` over the full suite), intersecting `statementMap` and `branchMap` hit counts with the added line ranges from `git diff -U0` (109-127, 134-144, 146-149, 151). `npx tsc --noEmit` clean on head. Check runs read from `gh api .../commits/5c061337/check-runs`.

---

## Previous findings: all four resolved

I did not take the author's word for any of these. For each, I mutated `src/agent/filetools.ts` on the head commit and re-ran `test/safety.test.ts` (24 tests, clean = 24 passed):

| Prior finding | Fix | Mutant applied | Result |
| --- | --- | --- | --- |
| 1 medium, global `seenDirs` over-prunes aliases | ancestor set, added on descent and removed on return ([filetools.ts:118](src/agent/filetools.ts#L118), [:149-151](src/agent/filetools.ts#L149-L151)) | delete `ancestors.delete(realDir)` (back to a global set) | 1 failed: *"reports a directory reachable by both its real name and an alias"* |
| 2 low, guard rail could not fail | rewritten around a symlinked **file** with an exact match-set assertion ([safety.test.ts:184](test/safety.test.ts#L184)) | `if (link.isSymbolicLink()) continue;` | 2 failed: the guard rail plus the alias test |
| 3 low, dangling symlink unclaimed and untested | regression test added ([safety.test.ts:216](test/safety.test.ts#L216)) | drop `.catch(() => null)` from `realpath` in `insideRoot` | 1 failed: *"steps over a dangling symlink instead of failing the walk"* |
| 4 low, double `realpath` | resolved once and reused via `real ?? (await insideRoot(abs))` ([filetools.ts:140](src/agent/filetools.ts#L140), [:146](src/agent/filetools.ts#L146)) | n/a (non-behavioral) | n/a |

The earlier `seenDirs`-seeding fix is also still pinned: removing `realRoot` from the seed fails exactly *"does not report a file twice through a link back to the repo root"*, and nothing else.

The author's correction on finding 3 is right and mine was wrong: the arm that saves the walk is `realpath`'s inside `insideRoot`, not `stat`'s. A dangling link fails containment before `stat` is ever reached, which is why `:144` remains uncovered. The test comment says so.

## New checks on the reworked walk: no defects found

The ancestor set is the part that changed shape, so I traced termination and containment against fresh hostile layouts rather than re-reading the tests:

- **Sibling cycle** `x/lx -> y`, `y/ly -> x`: terminates by hand-trace (each descent hits an ancestor at depth 2), and no match is lost.
- **Nested cycle** `a/b -> a`: terminates on the ancestor check at [:148](src/agent/filetools.ts#L148).
- **File symlink escaping the repo** (`peek.txt -> ../secret.txt`): base returns `["a.txt","peek.txt"]`, head returns `["a.txt"]`. The PR's escape fixture only covers a *directory* escape, but this is the same branch at [:141](src/agent/filetools.ts#L141), so it is covered.
- **Repo root itself behind a symlink**: returns `["sub/a.txt"]` on both base and head, so the `realpath(repoRoot)` hoist does not break the macOS `/tmp` case it was written for.
- **Containment rule vs `resolveInRepo`**: [`insideRoot`](src/agent/filetools.ts#L124) applies `real !== realRoot && !real.startsWith(realRoot + path.sep)`, character-for-character the rule in [paths.ts:18](src/util/paths.ts#L18). Consistent, so a future change to one is visibly a change to both.
- **Test determinism**: the alias test's assertions do not depend on `readdir` order (the no-glob result is sorted, the glob result is a single element), so the mutant catches it whichever entry comes first. All six new tests use `mkdtemp`, no network, no clock, no RNG, no cross-test ordering, and all assert. Not cleaning up the temp dirs matches the surrounding `file tools` tests.

---

## Findings

### 1. low: the PR body still describes the design that was replaced

The **Design** section says *"A `seenDirs` set of real paths. A directory whose real path was already walked is skipped"*. That is the exact behavior finding 1 rejected; the code now uses a per-path ancestor set, and the difference is the whole point of `138dc80`. The **Testing** section also still carries the claim the author has since retracted in a comment (*"the guard rail passes under both implementations, so it is genuinely guarding"*), and its status table (*"323 pass, 14 skip, 8 pre-existing failures"*, *"three new tests"*) predates two commits and three added tests.

Failure scenario: this repo squash-merges, so the PR body becomes the permanent commit message. Whoever reads `git log` for *why* loop detection is scoped to the traversal path gets the rejected rationale instead of the accepted one.

Fix: update the Design bullet to describe the ancestor set and why a global set was wrong, drop the retracted guard-rail sentence, and refresh the test table to 6 new tests and the current suite totals. No code change.

### 2. low: three added branch arms are still unexercised, all race-only

[filetools.ts:137](src/agent/filetools.ts#L137) (`lstat` returns null), [:144](src/agent/filetools.ts#L144) (`stat` returns null), [:147](src/agent/filetools.ts#L147) (a non-symlink directory failing containment). Measured at zero hits across the full suite. Each requires an entry to vanish or change type between `readdir` and the next syscall, which no fixture can create deterministically. Leaving them uncovered is correct; recording it so the number in the metrics block reconciles. The [:113](src/agent/filetools.ts#L113) `realpath(repoRoot)` fallback is also unexercised (v8 does not emit a branch for it) and is likewise unreachable while `repoRoot` exists.

No fix requested.

### 3. low (nit): `insideRoot` runs a `realpath` for every non-symlink directory, purely for bookkeeping

[filetools.ts:146](src/agent/filetools.ts#L146). For an entry that is not a symlink, containment is already guaranteed by its parent, so the `realpath` is needed only to produce the key for `ancestors`. That key is derivable without a syscall: thread the current directory's real path into `walk` and use `path.join(dirReal, entry)`.

Measured cost on a symlink-heavy probe (12 nested directories, two aliases each): base 2.0s, head 3.7s. Ordinary repos pay one extra `realpath` per directory, which is not much, but the syscall is avoidable and the threaded version also makes it explicit that only symlinks need checking.

```ts
  async function walk(dir: string, dirReal: string): Promise<void> {
    ...
        const realDir = real ?? path.join(dirReal, entry);
        ...
        await walk(abs, realDir);
  }
  await walk(repoRoot, realRoot);
```

### 4. low: an unreadable directory still takes down the whole search (pre-existing, not a regression)

[filetools.ts:130](src/agent/filetools.ts#L130). `readdir` is the one call in the walk without a `.catch`. Measured on a repo containing `a.txt` plus a `locked/` directory at mode `0o000`:

```text
base:  THREW: EACCES: permission denied, scandir '.../locked'
head:  THREW: EACCES: permission denied, scandir '.../locked'
```

Identical before and after, so this PR neither causes nor worsens it, and #87 is about symlinks. Raising it because it is the same failure shape the PR exists to remove: one bad entry makes `search` useless for the rest of the run, and the loop and dangling-link cases are now handled while this one is not. If the author wants to close the class in this PR it is one line (`for (const entry of await readdir(dir).catch(() => []))`) plus a fixture; equally fine as a follow-up issue.

### 5. low: project CI has still never run the suite on this branch

Only `cla` and `template` produced check runs for `5c06133`. The `CI / test` job in `.github/workflows/ci.yml` needs maintainer approval for a fork PR. Nothing for the author to do. I ran the equivalent locally: 351 pass, 17 skip, 0 fail, and `tsc --noEmit` clean.

---

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| Spec-gated in | pass. Only `src/agent/filetools.ts` and `test/safety.test.ts` changed; `src/agent/tools.ts` is untouched, `search` stays `requiresUnlock: false`, and `getTools` still filters on `!t.requiresUnlock \|\| ctx.editsUnlocked`. |
| Verification-gated out | pass, not applicable. ERC/DRC, repair, and rollback are untouched. |
| `check`/`verify` LLM-free and network-free | pass. The only added imports are `lstat` and `realpath` from `node:fs/promises`. No provider, key, or network reference. |
| No sexp serialization | pass, not applicable. `src/kicad/` is untouched. |
| Sync-obligations ledger | pass, not applicable. Hooks and commit gating are untouched. |
| Secrets | pass. Redaction and `.gitignore` unchanged. The `needle SECRET outside` fixture string is a test sentinel and does not match `sk-[A-Za-z0-9_-]+`. |
| Spec coherence | pass. No spec-level behavior change: this brings `toolSearch` in line with the containment rule AC-4.2 already states. No planning artifact changed, so `openspec validate` was not required and was not run. |
| Test determinism | pass. See the determinism note above. |

## Scope

Clean. 36 source lines, all inside `toolSearch`'s walk, plus tests. No scope creep across four commits.

Still open for whoever picks up #85, unchanged since the last pass: `resolveInRepo` in [src/util/paths.ts:15](src/util/paths.ts#L15) is purely lexical, so `read_file` on `link/leak.txt` in this PR's own escape fixture still returns the outside content. Correctly out of scope here, and the PR says so.

PR #90, opened 2026-07-26
fix(preflight): treat an empty COPPERHEAD_MIN_FREE_MB as unset

### LINKED ISSUE #89: An empty COPPERHEAD_MIN_FREE_MB silently disables the disk preflight (Number('') is 0)
### Summary

The `COPPERHEAD_MIN_FREE_MB` override in [create.ts](src/commands/create.ts) parses with bare `Number()`, and `Number('')` is `0`, not `NaN`. So an **empty or blank** value sets the threshold to 0 bytes, which silently disables the disk preflight that Issue 16 added:

```ts
const minFreeMb = Number(process.env.COPPERHEAD_MIN_FREE_MB);
const minFree = Number.isFinite(minFreeMb) && minFreeMb >= 0 ? minFreeMb * 1024 * 1024 : DEFAULT_MIN_FREE_BYTES;
```

`assertDiskSpace` then returns immediately, because `free >= 0` is always true.

### Actual results

| `COPPERHEAD_MIN_FREE_MB` | `Number()` | resulting threshold |
| --- | --- | --- |
| unset | `NaN` | 2.0 GiB (correct) |
| `abc` | `NaN` | 2.0 GiB (correct) |
| `500` | `500` | 0.488 GiB (correct) |
| **`` (empty)** | **`0`** | **0 GiB — check disabled** |
| **`" "` (blank)** | **`0`** | **0 GiB — check disabled** |

The failure mode is the one Issue 16 exists to prevent: the run starts on a nearly-full disk and dies partway through with a raw `ENOSPC` from whichever write loses, after doing expensive work. Because the intent of `Number.isFinite(...) && >= 0` is clearly "fall back on a bad value", an empty value reads as a bad value that isn't caught.

### Why an empty value is easy to produce

This repo's own `.env` loader creates exactly that shape. All three of these lines in `.env` set the variable to `''`:

```
COPPERHEAD_MIN_FREE_MB=
COPPERHEAD_MIN_FREE_MB=
COPPERHEAD_MIN_FREE_MB=""
```

A commented-out example that gets uncommented but not filled in, or a CI job with `COPPERHEAD_MIN_FREE_MB: ${{ vars.SOMETHING_UNSET }}`, lands in the same place. An operator who does this is trying to *keep* the default, and instead turns the guard off.

### Two smaller things in the same expression

- **`0` is accepted as a deliberate value.** That may be intentional (an explicit opt-out), but it is indistinguishable from the empty-value accident above, so there is currently no way to tell "I meant to disable it" from "my env var is empty".
- **`Infinity` falls back to the default** because `Number.isFinite(Infinity)` is `false`. That is arguably the wrong direction: `COPPERHEAD_MIN_FREE_MB=Infinity` reads as "never run", but silently becomes 2 GiB. Minor, and I would not change it without a word from you.

### Expected

An empty or whitespace-only value is treated as unset, i.e. it falls back to `DEFAULT_MIN_FREE_BYTES`, the same as `abc` already does.

### Reproduction

```js
Number('');    // 0     -> threshold becomes 0 bytes, assertDiskSpace returns immediately
Number(' ');   // 0     -> same
Number('abc'); // NaN   -> correctly falls back to 2 GiB
```

DPS0340:
## What

An empty or whitespace-only `COPPERHEAD_MIN_FREE_MB` is now treated as unset (falls back to the 2 GiB default) instead of becoming a 0-byte threshold that silently disables the disk preflight.

## Why

Closes #89.

The override parsed with a bare `Number()`, and `Number('')` is `0`, not `NaN`:

```ts
const minFreeMb = Number(process.env.COPPERHEAD_MIN_FREE_MB);
const minFree = Number.isFinite(minFreeMb) && minFreeMb >= 0 ? minFreeMb * 1024 * 1024 : DEFAULT_MIN_FREE_BYTES;
```

`assertDiskSpace` then returns immediately, since `free >= 0` is always true.

| `COPPERHEAD_MIN_FREE_MB` | `Number()` | threshold |
| --- | --- | --- |
| unset | `NaN` | 2.0 GiB (correct) |
| `abc` | `NaN` | 2.0 GiB (correct) |
| `500` | `500` | 0.488 GiB (correct) |
| (empty) | `0` | **0 GiB, check disabled** |
| `" "` (blank) | `0` | **0 GiB, check disabled** |

The failure mode is exactly the one Issue 16 exists to prevent: the run starts on a nearly-full disk and dies partway through with a raw `ENOSPC` from whichever write loses, after doing expensive work. Since the intent of `Number.isFinite(...) && >= 0` is clearly "fall back on a bad value", an empty value reads as a bad value that was not caught.

An empty value is easy to produce, including from this repo's own `.env` loader. All of these set it to `''`:

```text
COPPERHEAD_MIN_FREE_MB=
COPPERHEAD_MIN_FREE_MB=""
```

So does a CI job with `COPPERHEAD_MIN_FREE_MB: ${{ vars.SOMETHING_UNSET }}`. An operator doing this is trying to keep the default and instead turns the guard off.

## Design

The parse moves out of `runCreate` and into an exported `resolveMinFreeBytes()` in [preflight.ts](src/util/preflight.ts), next to `assertDiskSpace` and `DEFAULT_MIN_FREE_BYTES`. Two reasons: it is the only piece of this that is worth testing directly, and it belongs with the other preflight knobs rather than inline in an 8-stage pipeline function.

The only behavioral change is the added `raw.trim() === ''` check. Everything else is preserved deliberately:

- an explicit `0` still opts out. That is now distinguishable from the empty-value accident, which is the point: an operator can still turn the check off on purpose, and a blank variable no longer does it by mistake
- `abc`, `-1`, `NaN` still fall back, as before
- `Infinity` still falls back. I noticed this reads slightly wrong (`COPPERHEAD_MIN_FREE_MB=Infinity` means "never run" but becomes 2 GiB) and deliberately did not change it: it is a separate judgement call and not what this issue is about. Happy to follow up if you want it to refuse instead.

`create.ts` loses its now-unused `DEFAULT_MIN_FREE_BYTES` import.

## Spec / OpenSpec

No spec-level behavior change. This restores the intended behavior of the existing 4.1 disk preflight rather than altering what it promises, so SPEC.md and the change artifacts stay as they are.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | 325 pass, 14 skip, 8 pre-existing failures (see below) |

New tests, in [session-limit-disk.test.ts](test/session-limit-disk.test.ts) under `resolveMinFreeBytes`:

- unset falls back to the default
- **empty and blank values (`''`, `' '`, `'   '`, tab, newline) fall back**, which is the regression
- unparseable values (`abc`, `-1`, `Infinity`, `NaN`, `1,000`) fall back
- a valid value converts MiB to bytes, including one with surrounding whitespace and a fractional value
- an explicit `0` is still honoured as an opt-out

**Revert check.** With the tests in place I restored the original parse: exactly 1 test fails, the empty/blank case. The other four pass under both implementations, which is what I wanted, they confirm the fix did not achieve correctness by simply rejecting more inputs.

n/a with a reason: the change is a pure parsing function plus the one call site that feeds it. Exercising it at runtime would mean filling a real disk to below the threshold, which the existing `assertDiskSpace` tests already cover synthetically by passing thresholds directly (0 and `Number.MAX_SAFE_INTEGER`). The behavior this PR changes is entirely upstream of that call, and is covered directly by the new unit tests.

For what it is worth, I did verify the empty-value path end to end at the value level rather than assuming it:

```text
Number('')      -> 0      (threshold 0 bytes, assertDiskSpace returns immediately)
Number(' ')     -> 0      (same)
Number('abc')   -> NaN    (correctly falls back to 2 GiB)
```

and confirmed against this repo's own `.env` parser that `COPPERHEAD_MIN_FREE_MB=` and `COPPERHEAD_MIN_FREE_MB=""` both yield `''`.

## Docs

None needed. The documented behavior (`COPPERHEAD_MIN_FREE_MB=500`, default 2 GiB) is unchanged; this makes the implementation match it. The remedy text in `assertDiskSpace` already describes the override correctly.

---

## Invariant checklist

- [ ] N/A **Spec-gated in**: the agent tool list is untouched.
- [x] **`check`/`verify` stays LLM-free and network-free**: `preflight.ts` has no provider, key, or network access, and this change adds none.
- [ ] N/A **Secrets**: no change to redaction or key handling. The env var carries no secret.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved disk-space threshold handling for the create command.
  * Blank, invalid, negative, or unset values now safely use the default threshold.
  * Valid decimal values are converted accurately, while `0` can explicitly disable the minimum threshold.
* **Tests**
  * Added coverage for minimum free-disk threshold resolution across unset, whitespace, invalid, decimal, and `0` inputs.

### DIFF
--- a/src/commands/create.ts
+++ b/src/commands/create.ts
@@ -16,7 +16,7 @@ import type { RunMetaInput } from '../agent/runmeta.js';
 import { fmtDuration, fmtTokens, type ProgressRenderer } from '../agent/render.js';
 import { openspecInit } from '../openspec/cli.js';
 import { sweepStaleTempDirs, pruneHistoryDir } from '../util/tmp.js';
-import { assertDiskSpace, DEFAULT_MIN_FREE_BYTES } from '../util/preflight.js';
+import { assertDiskSpace, resolveMinFreeBytes } from '../util/preflight.js';
 import { runCheck } from './check.js';
 import { emitCreateJlcpcbBom } from './export.js';
 
@@ -525,9 +525,7 @@ export async function runCreate(opts: CreateOptions): Promise<{ ok: boolean; com
   // KiCad local history and can otherwise fill the disk mid-stage, failing with
   // an opaque ENOSPC only after doing expensive work. Threshold overridable via
   // COPPERHEAD_MIN_FREE_MB; an unknown reading (unsupported platform) skips it.
-  const minFreeMb = Number(process.env.COPPERHEAD_MIN_FREE_MB);
-  const minFree = Number.isFinite(minFreeMb) && minFreeMb >= 0 ? minFreeMb * 1024 * 1024 : DEFAULT_MIN_FREE_BYTES;
-  await assertDiskSpace(opts.repoRoot, minFree);
+  await assertDiskSpace(opts.repoRoot, resolveMinFreeBytes(process.env.COPPERHEAD_MIN_FREE_MB));
   // Reclaim scratch dirs leaked by earlier runs whose cleanup was skipped (a
   // watchdog SIGKILL or hard abort bypasses the per-call `finally`). Age-gated,
   // so a concurrent run's fresh dirs are never touched; best-effort, so it never
--- a/src/util/preflight.ts
+++ b/src/util/preflight.ts
@@ -31,6 +31,25 @@ export const DEFAULT_MIN_FREE_BYTES = 2 * 1024 * 1024 * 1024;
 const gib = (n: number): string => `${(n / 1024 / 1024 / 1024).toFixed(1)} GiB`;
 
 /**
+ * Resolve the free-space threshold from a `COPPERHEAD_MIN_FREE_MB` value.
+ *
+ * An empty or whitespace-only value is treated as unset, not as zero:
+ * `Number('')` is `0`, so parsing the raw value would turn an env var that was
+ * declared but never filled in (a blank line in `.env`, an unset CI variable)
+ * into a 0-byte threshold, silently disabling the check this module exists for.
+ * Anything else unparseable, negative, or non-finite also falls back.
+ */
+export function resolveMinFreeBytes(
+  raw: string | undefined,
+  fallback = DEFAULT_MIN_FREE_BYTES,
+): number {
+  if (raw === undefined || raw.trim() === '') return fallback;
+  const mb = Number(raw);
+  if (!Number.isFinite(mb) || mb < 0) return fallback;
+  return mb * 1024 * 1024;
+}
+
+/**
  * Free bytes available to this (unprivileged) user on the filesystem holding
  * `dir`, or null when the platform/Node build cannot report it — callers treat
  * null as "unknown" and skip the check rather than blocking a legitimate run.
--- a/src/util/preflight.ts
+++ b/src/util/preflight.ts
@@ -37,15 +37,13 @@ const gib = (n: number): string => `${(n / 1024 / 1024 / 1024).toFixed(1)} GiB`;
  * `Number('')` is `0`, so parsing the raw value would turn an env var that was
  * declared but never filled in (a blank line in `.env`, an unset CI variable)
  * into a 0-byte threshold, silently disabling the check this module exists for.
- * Anything else unparseable, negative, or non-finite also falls back.
+ * Anything else unparseable, negative, or non-finite also falls back to
+ * `DEFAULT_MIN_FREE_BYTES`, which is exported for callers that need it.
  */
-export function resolveMinFreeBytes(
-  raw: string | undefined,
-  fallback = DEFAULT_MIN_FREE_BYTES,
-): number {
-  if (raw === undefined || raw.trim() === '') return fallback;
+export function resolveMinFreeBytes(raw: string | undefined): number {
+  if (raw === undefined || raw.trim() === '') return DEFAULT_MIN_FREE_BYTES;
   const mb = Number(raw);
-  if (!Number.isFinite(mb) || mb < 0) return fallback;
+  if (!Number.isFinite(mb) || mb < 0) return DEFAULT_MIN_FREE_BYTES;
   return mb * 1024 * 1024;
 }
 
--- a/test/session-limit-disk.test.ts
+++ b/test/session-limit-disk.test.ts
@@ -3,7 +3,13 @@ import { mkdtemp, rm } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
 import { sessionLimit, isRateLimit } from '../src/util/retry.js';
-import { assertDiskSpace, freeDiskBytes, PreflightError } from '../src/util/preflight.js';
+import {
+  assertDiskSpace,
+  DEFAULT_MIN_FREE_BYTES,
+  freeDiskBytes,
+  PreflightError,
+  resolveMinFreeBytes,
+} from '../src/util/preflight.js';
 
 describe('sessionLimit (2.4 / I13 — a schedulable pause, not a bug)', () => {
   it('detects a saved-login session limit and extracts the reset time', () => {
@@ -57,5 +63,38 @@ describe('disk-space preflight (4.1)', () => {
     } finally {
       await rm(dir, { recursive: true, force: true });
     }
+  });
+});
+
+describe('resolveMinFreeBytes (COPPERHEAD_MIN_FREE_MB)', () => {
+  it('uses the default when the variable is unset', () => {
+    expect(resolveMinFreeBytes(undefined)).toBe(DEFAULT_MIN_FREE_BYTES);
+  });
+
+  it('treats an empty or blank value as unset, not as zero', () => {
+    // Number('') is 0, so parsing the raw value would set a 0-byte threshold
+    // and silently disable the check. A declared-but-unfilled variable is
+    // exactly what a blank .env line or an unset CI variable produces.
+    for (const raw of ['', ' ', '   ', '\t', '\n']) {
+      expect(resolveMinFreeBytes(raw), JSON.stringify(raw)).toBe(DEFAULT_MIN_FREE_BYTES);
+    }
+  });
+
+  it('falls back on values that are not a usable number', () => {
+    for (const raw of ['abc', '-1', 'Infinity', 'NaN', '1,000']) {
+      expect(resolveMinFreeBytes(raw), raw).toBe(DEFAULT_MIN_FREE_BYTES);
+    }
+  });
+
+  it('converts a valid value from MiB to bytes', () => {
+    expect(resolveMinFreeBytes('500')).toBe(500 * 1024 * 1024);
+    expect(resolveMinFreeBytes('  500  ')).toBe(500 * 1024 * 1024);
+    expect(resolveMinFreeBytes('0.5')).toBe(0.5 * 1024 * 1024);
+  });
+
+  it('still honours an explicit 0 as an opt-out', () => {
+    // Distinguishable from the empty case above, which is the point: an
+    // operator can still turn the check off on purpose.
+    expect(resolveMinFreeBytes('0')).toBe(0);
   });
 });


DPS0340:

DPS0340:

animesh-chouhan:
_Automated `/pr-review` pass (Claude Code). Not a human review, and not a formal approval: no merge gating is implied._

## Verdict: approve

The bug is real, the premise is verified rather than asserted, and the fix is the minimal correct one. I confirmed the failure mode independently: `parseEnv` in [src/util/env.ts:19-42](src/util/env.ts#L19-L42) matches `COPPERHEAD_MIN_FREE_MB=` (the regex allows an empty tail), yields `''`, and `loadEnvFile` writes that `''` into `process.env` because `process.env[key] === undefined`. The old `Number('')` then produced a 0-byte threshold and `assertDiskSpace` returned on `free >= 0`. So the reported path is reachable through this repo's own loader, not only through CI.

I also checked the fix is behavior-preserving everywhere else. Old and new agree on `undefined`, `abc`, `-1`, `Infinity`, `NaN`, `0`, `500`, and whitespace-padded numbers. The single divergence is the blank case, which is what the PR claims. `resolveMinFreeBytes` has exactly one call site and `COPPERHEAD_MIN_FREE_MB` is read in exactly one place, so there is no second parse left behind.

## Metrics

```
lines: +61 / -5 (net +56) across 3 files  ·  src +21/-4, test +40/-1, docs 0, spec 0
new vs net: 0 new files, 3 modified; 21 new src lines (19 preflight.ts, 2 create.ts)
tests: +40 test lines (+5 tests); suite 328/14/0 -> 333/14/0 (pass/skip/fail, base -> head)
diff coverage: 100% of changed src lines (9/9 executable) measured; uncovered: none
deps/audit: n/a, package.json unchanged
```

Method for each number:
- Sizes from `git diff --numstat $(git merge-base origin/main HEAD)...HEAD`. No added or deleted files (`--diff-filter=A` is empty), so every changed line is an edit to existing files.
- Suite totals from a real `npm test` on both sides: base `887346b` in a scratch worktree gave 328 pass / 14 skip / 0 fail, head `fe1332b` gave 333 / 14 / 0. `npm run typecheck` passes on head. Note this differs from the PR description's "8 pre-existing failures": those are `KicadCliMissingError` in the author's environment, and `kicad-cli` is on PATH here, so the suite is fully green in both positions.
- Diff coverage measured with `@vitest/coverage-v8` (`npx vitest run --coverage --coverage.reporter=json`), intersecting the per-line statement counts with the changed line ranges. Changed executable src lines are `preflight.ts:42,43,44,45,46,47,48,49` and `create.ts:528`; hit counts are 22/22/22/22/22/9/22/4/7 respectively. Istanbul branch data for the new function is also fully hit (line 46 both ways: 13 fallback / 9 through; line 48 both ways: 5 fallback / 4 through).

## Invariant checks performed

All six were checked against the actual diff, not taken from the checklist in the description:

- **Spec-gated in**: pass. The diff touches no file under `src/agent/`, so the tool list and its OpenSpec gate are structurally unchanged.
- **Verification-gated out**: pass. No change to the ERC/DRC, repair-cycle, or rollback path.
- **`check`/`verify` LLM-free and network-free**: pass, and verified transitively rather than assumed. `src/commands/check.ts` does not import `src/util/preflight.ts` at all, and `preflight.ts`'s only import is `node:fs/promises`. Nothing in the new code reaches a provider, a key, or the network.
- **No sexp serialization**: pass. `src/kicad/` is untouched.
- **Sync-obligations ledger**: pass. Untouched.
- **Secrets**: pass. `.gitignore` still carries `.env` (line 1) and `.copperhead/runs/` (line 2); no redaction or key-handling code is touched, and `COPPERHEAD_MIN_FREE_MB` is not a secret.
- **Spec coherence**: pass. `COPPERHEAD_MIN_FREE_MB` appears in no `.md` under `openspec/`; the only prose reference is `PIPELINE-ISSUES.md:428-431`, which describes the 2 GiB default and the override, and remains accurate. This restores documented behavior rather than changing it, so no artifact needed to move. `npm test` and typecheck are the only gates the change touches.

Test quality: the 5 new tests are deterministic, table-driven, network-free, order-independent, use no clock or randomness, and every case asserts. The `expect(actual, message)` second-argument form gives a readable failure per table row. Scope is clean, with no unrelated changes.

## Findings

### 1. low, untested surface: the `fallback` parameter is exported public API that nothing ever passes

[src/util/preflight.ts:42-45](src/util/preflight.ts#L42-L45)

`resolveMinFreeBytes(raw, fallback = DEFAULT_MIN_FREE_BYTES)` takes a second parameter, but the sole call site ([src/commands/create.ts:528](src/commands/create.ts#L528)) passes one argument and no test passes two. Coverage confirms this rather than inferring it: the istanbul default-parameter branch at line 42 records `[22]`, meaning all 22 calls in the suite took the default and the explicit-argument path has never executed.

Failure scenario: a future caller passes a per-command floor, say `resolveMinFreeBytes(raw, 0)`, expecting "no floor unless the operator sets one". Nothing in the suite would catch a regression in that path, and the parameter's interaction with the blank-value branch (blank returns `fallback`, so an explicit `0` fallback re-creates precisely the 0-byte threshold this PR removes) is exactly the subtle case that needs a test.

Fix, pick one:
- Drop the parameter, since `DEFAULT_MIN_FREE_BYTES` is already exported and a caller wanting a different default can write `raw ? resolveMinFreeBytes(raw) : mine`. This is my preference: it keeps the new exported surface to one function of one argument.
- Or keep it and add the one-line test that pins it:
  ```ts
  it('honours an explicit fallback for unset and blank values', () => {
    expect(resolveMinFreeBytes(undefined, 123)).toBe(123);
    expect(resolveMinFreeBytes('', 123)).toBe(123);
  });
  ```

### 2. low, nit and pre-existing: a malformed value still falls back silently

[src/util/preflight.ts:48](src/util/preflight.ts#L48)

`COPPERHEAD_MIN_FREE_MB=500MB` or `=500 MiB` parses to `NaN` and silently becomes 2 GiB. The operator then sees `not enough free disk space to start (... 2.0 GiB required)` with a remedy that says "lower the threshold with `COPPERHEAD_MIN_FREE_MB`", which is the thing they believe they already did.

This is pre-existing behavior that the PR deliberately preserves, and I agree it does not belong in this change. Flagging only because the PR creates the natural home for it: a one-line `console.warn` on the `!Number.isFinite(mb) || mb < 0` branch (not on the blank branch, since blank now legitimately means "unset") would close the loop. Same category as the `Infinity` question the description already raises and defers. Worth a follow-up issue, not a change request here.

## CI and mergeability

`mergeable: MERGEABLE`, not a draft, and all three checks are green: `cla` pass, `template` pass, `CodeRabbit` pass (it recorded a rate-limit skip rather than an actual review, so treat that one as "no signal" rather than "reviewed"). Nothing blocking.

DPS0340:
Thanks for the review. Finding 1 is addressed in `63e0c63`; I agree with you that finding 2 belongs in a follow-up.

## 1. The `fallback` parameter — removed

I took your preferred option. Your point about `resolveMinFreeBytes(raw, 0)` re-creating exactly the 0-byte threshold this PR removes is what decided it: better to make that call impossible to write than to pin it with a test. And as you note, `DEFAULT_MIN_FREE_BYTES` is already exported, so a caller wanting a different default can write `raw ? resolveMinFreeBytes(raw) : mine`.

I checked that nothing passes a second argument with the compiler rather than with grep:

```
references: src/commands/create.ts:528 (1 arg), 8 in test (all 1 arg)
npm run typecheck -> passes
```

A two-argument call anywhere would have failed `tsc`; it passed silently.

I also verified no regression by running both sides rather than reasoning about it:

```
before (fe1332b)  8 failed | 325 passed | 14 skipped
after  (63e0c63)  8 failed | 325 passed | 14 skipped
```

The 8 failures are all `kicad-cli` absence in my environment — the same ones you identified. Since `kicad-cli` is on PATH on your side, both positions should be fully green there. The point is that the number is identical before and after.

One doc line changed too: "falls back" no longer had a referent once the parameter was gone, so it now names `DEFAULT_MIN_FREE_BYTES` and notes that the constant is exported.

## 2. Silent fallback on a malformed value — agreed, follow-up

Your reading is right: `COPPERHEAD_MIN_FREE_MB=500MB` parses to `NaN`, becomes 2 GiB, and the operator is then told to "lower the threshold with `COPPERHEAD_MIN_FREE_MB`" — the thing they believe they just did.

I agree it does not belong here. Blank and unparseable now mean genuinely different things: blank legitimately means "unset" and must not warn, while only the `!Number.isFinite(mb) || mb < 0` branch should. That distinction is something this PR creates, so a separate change is also easier to review on its own terms.

I will open the issue once this merges. Happy to file it now instead if you would rather have it tracked immediately.

## One note on the review itself

Reading the istanbul default-parameter branch as `[22]` — all 22 calls took the default, so the explicit-argument path has never executed — is not how I had been using coverage. Treating it as "which branch is dead" rather than "what percentage" is a better tool than the one I was reaching for, and I will apply it to my other PRs.

animesh-chouhan:
checking, are you on our discord: https://discord.gg/eK6JH4Xzd

DPS0340:
Yes, joined — same handle, `DPS0340`.

Where the four stand right now, since they are all waiting on the same thing:

| PR | Issue | Review status | Checks |
| --- | --- | --- | --- |
| #84 | #83 base64 bearer tail | no findings raised | 3/3 pass |
| #86 | #85 lexical sandbox escape | 2 findings, both fixed and verified | 3/3 pass |
| #88 | #87 toolSearch escape + symlink loop | 4 findings, all fixed and verified | 3/3 pass |
| #90 | #89 empty `COPPERHEAD_MIN_FREE_MB` | 2 findings: 1 fixed, 1 agreed as follow-up | 3/3 pass |

All four are `mergeable: true`, each is one logical fix against one issue, and every review finding is either addressed on the branch or explicitly deferred with your agreement.

Two things I would rather you decide than me:

1. The follow-up issue from #90 finding 2 (`COPPERHEAD_MIN_FREE_MB=500MB` parses to `NaN`, silently becomes 2 GiB, and the operator is then told to lower the threshold they just set). I said I would file it after merge to keep the queue clean; happy to file it now if you would rather have it tracked immediately.
2. Merge order. #86 and #88 both touch the sandbox path and #88 was written on top of #86's containment fix, so #86 first is the cleaner sequence. #84 and #90 are independent of both.

DPS0340:
Two things, and the first is a correction you may already have seen land wrongly.

**1. A mangled comment on #100/#102/#104/#106/#108.** I closed those five and the note attached to them was the wrong text — a message I had written for a batch of PRs in a different org, sent here by a hardcoded filename in my own close script. This repo auto-locks closed PRs, so I cannot edit or replace it there. Please disregard anything in it about "the tscircuit org"; none of it refers to copperhead.

**2. What I actually meant, which is the reason I closed them.**

I counted my own output across every repo I contribute to. Among PRs a maintainer actually decided, **57% merged** — but **94% of what I have opened has never been looked at by a human at all.** The constraint was never quality. I was handing out more review work than anyone could absorb.

Here specifically: **9 of your 39 open PRs were mine (23%)**, and all nine fix issues I filed myself. You reviewed three of them properly — this one, #86 and #88 — with findings I had to go and verify, and I kept sending more on top of that. That was not a fair trade for your attention.

So I kept the four you had already engaged with (#84, #86, #88, #90) and closed the five that arrived afterwards. The branches are all still up, and issues #99, #101, #103, #105, #107 stay open with their reproductions intact — nothing is lost if you or anyone else wants to pick them up. Name any number and I will reopen it immediately.

Sorry for the volume, and for the garbled note.

animesh-chouhan:
volume is actually a great thing and really appreciate the good quality PRs. thanks for that!
have actually been a bit busy with navigating the funding for this. will review all PRs soon. keep 'em cominggg

DPS0340:
That is good to hear, and thanks for saying it plainly — I had assumed the opposite and acted on it, so I would rather correct that than let it sit.

Since you are going to review these, I rebased all four onto current `main` so you are not reading a stale diff. They were 5 commits behind:

```
#84  fix/redact-base64-bearer      c9f681a
#86  fix/symlink-sandbox           2c635cd
#88  fix/search-symlink-sandbox    5c06133
#90  fix/min-free-mb-empty         b2537cd
```

No conflicts on any of them. I checked the rebase did not quietly break anything rather than assuming a clean rebase means a working one:

```
main   8 failed | 337 passed
#84    8 failed | 341 passed
#86    8 failed | 343 passed
#88    8 failed | 343 passed
#90    8 failed | 342 passed
```

Same eight on every side — all `kicad-cli` absence in my environment, which is on your PATH, so they should be fully green for you.

## Suggested reading order

They are independent, but two are worth more of your attention than the other two:

- **#88** — the one I would look at first. It has the most review history (4 findings, all fixed and verified), and it fixes a real regression I introduced and then caught: walk-global loop detection was hiding a directory behind an in-repo symlink.
- **#86** — the containment fix underneath #88. Same area, smaller.
- **#84** and **#90** — both small and self-contained. #90 is approved already with one finding deferred to a follow-up by your own suggestion.

## One open item you asked me to decide and I would rather you did

From #90 finding 2: `COPPERHEAD_MIN_FREE_MB=500MB` parses to `NaN` and silently becomes 2 GiB, while the error tells the operator to lower the threshold they believe they just set. We agreed it belongs in a follow-up rather than that PR. Want me to open the issue now, or after #90 lands?

Good luck with the funding side — that is the harder problem of the two.

Devesh36:

## Verdict: approve

Clean, well-scoped fix for the empty-`COPPERHEAD_MIN_FREE_MB` footgun (`Number('') === 0` silently disabling the disk preflight). Helper is documented, create call site updated, and unit tests cover unset/blank/invalid/explicit-0/valid MiB.

### Metrics
```text
lines: +59 / -5 across 3 files  ·  src +~25, test +~36
tests: focused resolveMinFreeBytes suite added
ci: see PR checks (suite not re-run in this parallel pass)
```

### Invariants checked
- Spec-gated in: n/a
- Verification-gated out: n/a
- check/verify LLM-free: pass (preflight only)
- No sexp serialization: n/a
- Secrets: pass

### Findings

No findings.

No findings, so no fix prompt.

PR #106, opened 2026-07-26
fix(sync): match pins.h against assigned pins, not the whole document

### LINKED ISSUE #105: sync's pins.h check can never fail: it substring-matches PINOUT.md, so any word anywhere satisfies any #define
### Summary

The `pins-h` check in `copperhead sync` compares a generated firmware header against `PINOUT.md` by substring-matching the raw document text. Any word that appears **anywhere** in the file satisfies any `#define`, so a header that has drifted arbitrarily far from the pin table is reported as consistent.

`sync` prints `sync: no inconsistencies` while `firmware/src/pins.h` names pins that do not exist.

### Steps to reproduce

Against the `open-key` fixture, whose `PINOUT.md` `init` generates as:

```markdown
Pin assignment extracted from the schematic. Check strapping/RTC notes before
reassigning pins.

| Refdes | Pin | Name | Net | Notes |
|---|---|---|---|---|
| U1 | 1 | 3V3 | 3V3 | |
| U1 | 5 | GPIO14 | KEY_DAH | |
```

Write a header naming three pins that the table does not assign:

```c
#define PIN_KEY  5
#define PIN_GPIO 21
#define PIN_3V   8
```

```bash
copperhead sync
```

### Actual

```text
sync: no inconsistencies
```

`KEY` matches inside the net `KEY_DAH`. `GPIO` matches inside `GPIO14`. `3V` matches inside `3V3`. None of the three is a pin.

Prose counts too: the sentence above the table contains "strapping", "RTC" and "pins", so `#define PIN_RTC` would also pass on a doc that never mentions an RTC pin.

### Expected

```text
3 resolvable inconsistency(ies):
- [pins-h] firmware/src/pins.h: claims "defines PIN_KEY" but actual is ...
- [pins-h] firmware/src/pins.h: claims "defines PIN_GPIO" but actual is ...
- [pins-h] firmware/src/pins.h: claims "defines PIN_3V" but actual is ...
```

### Root cause

`src/commands/sync.ts`:

```ts
const pinout = await readFile(path.join(docsDir, 'PINOUT.md'), 'utf8');
for (const m of header.matchAll(/#define\s+PIN_(\w+)\s+(\S+)/g)) {
  if (!pinout.includes(m[1]!)) {
```

`pinout` is the whole file as one string, so the test is "does this substring occur in the document", not "does the pin table assign this name". The module next door already parses the table structurally (`parseCanonicalTables` / `parsePinoutRows` in `src/memory/bom-table.ts`), and the drift checker uses it; this one check does not.

### Why it matters more than a normal false negative

Stage 7 of `copperhead create` generates `pins.h` **from** `PINOUT.md`, and `sync` is the mechanism that says the two still agree. A check that cannot fail is worse than no check: the pipeline reports the invariant as holding.

The direction is also the unlucky one. A false positive would be noisy and get fixed. This is a false negative, so nothing ever draws attention to it.

### Note on the pin numbers

The regex captures the value (`m[2]`) and never uses it. That is defensible and I am not proposing to change it: `PINOUT.md`'s `Pin` column is the **symbol pin number** from KiCad, while a firmware header holds the **MCU GPIO number**. On the fixture, `GPIO14` sits at symbol pin 5, so comparing the numbers would report drift on a correct header. The names are the part that can be checked; the numbers are not comparable without an MCU pin map.

### Verified

Reproduced against `main` at `6dcca30` through the CLI, output above. I have a fix with regression coverage and will open a PR against this issue.

DPS0340:
## What

`copperhead sync` now actually checks `firmware/src/pins.h` against `PINOUT.md`. Before this, the check could not fail.

## Why

Closes #105.

The check substring-matched the raw document, so any word occurring anywhere in `PINOUT.md` satisfied any `#define`. On the `open-key` fixture, with the `PINOUT.md` that `init` itself generates:

```c
#define PIN_KEY  5
#define PIN_GPIO 21
#define PIN_3V   8
```

```text
$ copperhead sync          # on main
sync: no inconsistencies
```

None of those three pins is assigned by the table. `KEY` matched inside the net `KEY_DAH`, `GPIO` inside `GPIO14`, `3V` inside `3V3`. The prose counts as well: the generated line "Check strapping/RTC notes before reassigning pins" means `#define PIN_RTC` would pass on a document that never mentions an RTC pin.

Two things make this worse than a normal false negative:

- **Stage 7 of `create` generates `pins.h` *from* `PINOUT.md`**, and `sync` is the mechanism that says the two still agree. A check that cannot fail does not merely miss problems, it reports the invariant as holding.
- **It fails open**, so nothing ever drew attention to it. A false positive would have been noisy and fixed long ago.

## Design

`src/memory/bom-table.ts` already parses this table structurally (`parseCanonicalTables`), and the drift checker uses it. This one check did not. The fix is to ask the same parser what the table assigns.

`pinoutIdentifiers(md)` returns the set of names the pin table gives to pins: each row's `Net`, and its `Name` when the optional column is present. Both are legitimate sources for a `#define PIN_<X>` (the scaffold writes `Refdes | Pin | Name | Net | Notes`, and a firmware author may key off either), so accepting only one would produce false positives on correct headers.

It is deliberately a separate function from `parsePinoutRows` rather than an extra field on it. `parsePinoutRows` answers "what net is on this pin", which the drift comparison needs row by row. This is set membership, and it needs no row structure.

KiCad's `~` unnamed-pin sentinel is excluded: it identifies nothing, and letting it in would make `#define PIN_` shapes pass on any table containing an unnamed pin.

**The pin numbers stay unused, on purpose.** The regex already captures the value in `m[2]` and always discarded it, which looks like an oversight and is not one: `PINOUT.md`'s `Pin` column is the **symbol pin number** from KiCad, while a firmware header holds the **MCU GPIO number**. On the fixture, `GPIO14` sits at symbol pin 5, so comparing the two numbers would report drift on a header that is correct. Checking them properly needs an MCU pin map, which is a different change; this PR fixes the part that is checkable today.

## Spec / OpenSpec

No spec-level behavior change. AC-7's pins-h obligation is unchanged; this makes the existing check do what it says.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Live-LLM (opt-in) | keyed on `ANTHROPIC_API_KEY` | n/a (`sync`'s verify phase is deterministic, LLM-free and network-free) |

```text
this branch     8 failed | 340 passed | 17 skipped
```

Same 8, and +3 is exactly the tests added here.

**New tests** ([gating-sync.test.ts](test/gating-sync.test.ts)): a `#define` naming a net or a pin name is accepted; one whose name only appears in prose or inside another name is flagged; and the `~` sentinel documents nothing while a real net on the same row still does.

Reverting only the comparison (`documented.has` back to `pinout.includes`) and keeping everything else:

```text
✓ accepts a #define naming a net or a pin name in the table
× flags a #define whose name only appears in prose or inside another name
✓ does not let the unnamed-pin sentinel ~ document anything
```

One fails, and it is the one aimed at the bug. The other two pass on both sides by design: they guard the false-positive direction, which is the risk when a check that never fired starts firing.

- Sandbox variant used (`create` / `edit`): a fixture repo under `manual-tests/runs/`, `init`ed, then given a `firmware/src/pins.h`. `init` and `sync` gate on `kicad-cli version`, which is not on my PATH, so a stub on PATH answers that one call. The verify phase itself is deterministic and reads only the doc and the header; no ERC or DRC runs, and the resolve phase (which needs a model) is not reached.
- Commands run and outcome:

```text
$ printf '#define PIN_KEY 5\n#define PIN_GPIO 21\n#define PIN_3V 8\n' > runs/sync/firmware/src/pins.h

# on main
$ npm run dev -- --repo manual-tests/runs/sync sync
sync: no inconsistencies

# on this branch
$ npm run dev -- --repo manual-tests/runs/sync sync
3 resolvable inconsistency(ies):
- [pins-h] firmware/src/pins.h: claims "defines PIN_KEY" but actual is "no pin in PINOUT.md carries that net or pin name"
- [pins-h] firmware/src/pins.h: claims "defines PIN_GPIO" but actual is "no pin in PINOUT.md carries that net or pin name"
- [pins-h] firmware/src/pins.h: claims "defines PIN_3V" but actual is "no pin in PINOUT.md carries that net or pin name"
    resolution: regenerate pins.h from PINOUT.md (PINOUT.md is the single source of truth)
```

I also checked the accepting direction by hand, so the new strictness is not just "flags more": `#define PIN_KEY_DAH` and `#define PIN_GPIO14` (a real net and a real pin name from the same table) both pass.

## Docs

None needed: no flag or output format changes. The `actual` string is more precise ("no pin in PINOUT.md carries that net or pin name" rather than "PINOUT.md does not mention it"), which matches what is now being tested and tells the reader why a `#define` was flagged despite the word appearing in the file.

---

## Invariant checklist

- [ ] **Verification-gated out**: N/A. This is `sync`'s deterministic verify phase, not the ERC/DRC mutation gate; no mutation happens on this path.
- [x] **`check`/`verify` stays LLM-free and network-free**: preserved, and directly relevant since this *is* the verify phase. The only import added is `pinoutIdentifiers` from `src/memory/bom-table.ts`, a pure string module with no I/O and no dependencies. `syncVerify` still touches no provider, key, or network.
- [x] **No sexp serialization**: preserved. Nothing here reads or writes a KiCad file; `sync`'s existing `pinNets` call is unchanged and read-only.
- [ ] **Sync-obligations ledger**: N/A. This is the `sync` command's report, not the run-time ledger; `src/agent/ledger.ts` is untouched.
- [ ] **Secrets**: N/A, no transcript or summary is written.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -3,6 +3,7 @@ import { existsSync } from 'node:fs';
 import { readFile } from 'node:fs/promises';
 import { loadConfig } from '../config.js';
 import { checkDrift } from '../memory/drift.js';
+import { pinoutIdentifiers } from '../memory/bom-table.js';
 import { loadConstraints, checkForbiddenPins } from '../memory/constraints.js';
 import { pinNets } from '../kicad/sexp.js';
 import { openspecValidate } from '../openspec/cli.js';
@@ -90,13 +91,21 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
   if (existsSync(pinsH) && existsSync(path.join(docsDir, 'PINOUT.md'))) {
     const header = await readFile(pinsH, 'utf8');
     const pinout = await readFile(path.join(docsDir, 'PINOUT.md'), 'utf8');
+    // Match against the identifiers the pin table actually assigns, not against
+    // the raw document text. `pinout.includes('LED')` is satisfied by the word
+    // LED anywhere at all: a different row's net (`LED_STATUS`), a Notes cell,
+    // or the prose above the table. Every `#define` then looks documented and
+    // the check reports nothing, which is the worst outcome for a check whose
+    // whole job is to catch a header that has drifted from the doc it was
+    // generated from.
+    const documented = pinoutIdentifiers(pinout);
     for (const m of header.matchAll(/#define\s+PIN_(\w+)\s+(\S+)/g)) {
-      if (!pinout.includes(m[1]!)) {
+      if (!documented.has(m[1]!)) {
         resolvable.push({
           kind: 'pins-h',
           doc: 'firmware/src/pins.h',
           claim: `defines PIN_${m[1]}`,
-          actual: 'PINOUT.md does not mention it',
+          actual: 'no pin in PINOUT.md carries that net or pin name',
           resolution: 'regenerate pins.h from PINOUT.md (PINOUT.md is the single source of truth)',
         });
       }
--- a/src/memory/bom-table.ts
+++ b/src/memory/bom-table.ts
@@ -122,6 +122,37 @@ export interface TableRow {
   }
 
   /**
+   * The identifiers PINOUT.md assigns to a pin: its net name and, when the
+   * table carries the optional Name column, the pin's own name. Both are
+   * legitimate sources for a firmware `#define PIN_<X>`, so a caller checking
+   * that a generated header agrees with the doc has to consider both.
+   *
+   * Distinct from `parsePinoutRows`, which answers "what net is on this pin"
+   * for the drift comparison. This answers "what names does the doc know",
+   * which is a set-membership question and needs no row structure.
+   */
+  export function pinoutIdentifiers(md: string): Set<string> {
+    const names = new Set<string>();
+    const strip = (s: string | undefined): string => (s ?? '').replace(/`/g, '').trim();
+    for (const { header, rows } of parseCanonicalTables(md)) {
+      const col = (re: RegExp): number => header.cells.findIndex((c) => re.test(c));
+      const pinI = col(/^pin$/i);
+      const netI = col(/^net$/i);
+      const nameI = col(/^name$/i);
+      if (pinI < 0) continue; // not the pin-assignment table
+      for (const row of rows) {
+        for (const i of [netI, nameI]) {
+          if (i < 0) continue;
+          const value = strip(row.cells[i]);
+          // KiCad writes `~` for an unnamed pin; it identifies nothing.
+          if (value && value !== '~') names.add(value);
+        }
+      }
+    }
+    return names;
+  }
+
+  /**
    * Fold the semantically-identical encodings that the model and KiCad render
    * differently, so a value that differs only in *encoding* is not flagged as
    * drift (#I11). A design that reached ERC-clean once churned for turns on
--- a/test/gating-sync.test.ts
+++ b/test/gating-sync.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from 'vitest';
 import path from 'node:path';
-import { writeFile, readFile } from 'node:fs/promises';
+import { writeFile, readFile, mkdir } from 'node:fs/promises';
 import { runInit } from '../src/memory/scaffold.js';
 import { loadConfig } from '../src/config.js';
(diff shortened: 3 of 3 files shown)


DPS0340:

Sorry for the volume.

PR #108, opened 2026-07-26
fix(loop): a failed rollback no longer takes the run report with it

### LINKED ISSUE #107: a failed rollback takes the run report with it on the refuse and dry-run paths
### Summary

When the rollback throws, two of the three rollback call sites in `runAgentLoop` let the exception escape, which skips `run-end` and `summary.md`. The run leaves no report at exactly the moment a report matters most: the working tree has been left in a partial state and nothing on disk says so.

`fail()` already handles this and says why:

```ts
// The rollback itself can fail (git in a bad state). That must not become
// an unhandled throw that skips run-end and summary.md — the summary is
// most valuable exactly when the tree is left in an unknown state.
```

The **refuse** path (`loop.ts`, `outcome === 'refuse'`) and the **dry-run** path call `restore` bare.

### How the rollback throws

`restore` shells out to git, so it throws on anything git refuses. The reachable case needs no corruption:

`snapshot` records a `git stash create` object when the tree is dirty. Git leaves that object **unreferenced by design** (`stash create` does not store it in the stash reflog), so an ordinary `git gc --prune=now` collects it. A `create` pipeline runs for a long time, and a gc landing in that window is not exotic.

```text
$ git stash create        # what snapshot records
86c98f67c51abbfab6b93c725e8a90c89d545f05
$ git gc --prune=now --quiet
$ git stash apply 86c98f67...
fatal: Could not resolve 86c98f67... to a stash entry     # restore throws here
```

Reproduced directly:

```text
restore THREW: Command failed with exit code 128: git stash apply 86c98f67...
```

### Why each path matters

**Refuse.** A refusal is the run telling the user a budget or constraint blocks the request, and the code right below calls it out as the most valuable thing to record:

```ts
// Refusals are the most valuable thing to remember: they encode a budget
// or constraint that this user's designs keep running into.
await remember({ ... });
```

If `restore` throws first, the `run-refused` event, `run-end`, `summary.md` and that `remember` call are all skipped. The user sees a stack trace and never learns what was refused or why.

**Dry run.** Beyond losing the report, a dry run's entire contract is that it leaves nothing behind. If the revert fails and the throw is swallowed higher up, the last thing written would still claim:

```ts
detail: 'dry run: changes reverted',
```

which is false in exactly the case the user most needs to know about.

### Expected

Both paths should do what `fail()` does: record the failure, keep going, and say so in the summary. The dry-run path additionally should not report "changes reverted" when the revert did not happen.

### Note on an adjacent PR

#34 (`--keep-on-fail`) touches the `fail()` rollback, which is the one path that already handles this. It does not touch the refuse or dry-run call sites, so there is no overlap in the lines involved.

### Verified

Reproduced against `main` at `6dcca30` by running the real loop with a scripted provider and collecting the stash mid-run, rather than by stubbing git:

```text
× refuse path      ExecaError: Command failed with exit code 128: git stash apply 17dd3edf...
× dry-run path     ExecaError: Command failed with exit code 128: git stash apply 055fc9ec...
```

DPS0340:
## What

A rollback that fails no longer destroys the run's own report. The refuse and dry-run paths now record the failure and keep going, the way `fail()` already does, and a dry run stops claiming "changes reverted" when the revert did not happen.

## Why

Closes #107.

`restore` shells out to git and throws on anything git refuses. Two of the three call sites in [loop.ts](src/agent/loop.ts) called it bare, so the exception escaped and skipped `run-end` and `summary.md`. The run left no report at exactly the moment one matters most: the tree is in a partial state and nothing on disk says so.

`fail()` already handled this, and its comment states the reasoning:

```ts
// The rollback itself can fail (git in a bad state). That must not become
// an unhandled throw that skips run-end and summary.md — the summary is
// most valuable exactly when the tree is left in an unknown state.
```

The other two paths did not get that treatment.

**No corruption is needed to reach it.** `snapshot` records a `git stash create` object when the tree is dirty, and git leaves that object unreferenced by design (`stash create` does not write it to the stash reflog). An ordinary `git gc --prune=now` collects it, and the later `git stash apply` fails. A `create` run is long enough that a gc landing in that window is unremarkable.

**Refuse is the worst place to lose the report.** The code just below the rollback says so itself:

```ts
// Refusals are the most valuable thing to remember: they encode a budget
// or constraint that this user's designs keep running into.
await remember({ ... });
```

A throw above it skips the `run-refused` event, `run-end`, `summary.md` **and** that `remember` call. The user gets a stack trace instead of the reason their request was refused.

**Dry run was additionally saying something untrue.** Its whole contract is that it leaves nothing behind, and its summary and finish line both read `dry run: changes reverted` regardless.

## Design

The `try`/`catch` in `fail()` is extracted as `rollBack(repoRoot, snap, transcript)`, which records `restore-failed` on the transcript and returns the message instead of throwing. All three call sites use it, so the handling cannot drift apart again.

The two newly-covered paths surface it the same way `fail()` does: a `WARNING:` log line and a `ROLLBACK FAILED: ... inspect it with git status/git diff before rerunning` line in the summary's Detail section. For the dry run the wording is `REVERT FAILED` and it replaces the "changes reverted" text in both the summary and the finish line, rather than being appended next to a claim that contradicts it.

Deliberately unchanged: the run's `outcome` and `exitPath`. A refusal is still a refusal, and a dry run still succeeded at proposing a diff. What failed is the cleanup, and the report is where that belongs. Changing the outcome would break callers (the create pipeline reads `res.outcome` to decide whether a stage advanced) for a condition that is about the tree, not the work.

`#34` (`--keep-on-fail`) touches the `fail()` rollback, the one path that already handled this. It does not touch the refuse or dry-run call sites, so the changed lines do not overlap.

## Spec / OpenSpec

No spec-level behavior change. AC-3.6's rollback is unchanged; this is about not losing the run report when it fails.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Live-LLM (opt-in) | keyed on `ANTHROPIC_API_KEY` | n/a (driven with a scripted provider; no live call) |

```text
this branch     8 failed | 340 passed | 17 skipped
```

Same 8, and +3 is exactly the tests added here.

**New tests** ([rollback-failure.test.ts](test/rollback-failure.test.ts)) drive the real `runAgentLoop` with a scripted provider. They produce the failure the way production would rather than by stubbing git: the run starts `--allow-dirty` over uncommitted work so the snapshot carries a stash object, and a `git gc --prune=now` runs between turns, which is where a long run's gc would land.

Against `main`:

```text
× refuse path    ExecaError: Command failed with exit code 128: git stash apply 17dd3edf...
× dry-run path   ExecaError: Command failed with exit code 128: git stash apply 055fc9ec...
✓ a rollback that succeeds still reports the plain outcome
```

The two failures are the raw `ExecaError` escaping, which is the bug itself rather than a missing assertion. The third test passes on both sides on purpose: it pins that a successful rollback still produces a clean summary with no `ROLLBACK FAILED` in it, so the new reporting cannot start firing on healthy runs.

One correction I made while writing them: I first asserted `res.outcome === 'aborted'` for the refuse path. It is `'refused'` — `aborted` is the value in the summary document, not on `RunResult`. Fixed against the actual value rather than the one I expected.

- Sandbox variant used (`create` / `edit`): n/a, and the reason is specific rather than convenience. Reaching this branch by hand needs a live provider that refuses, plus a `git gc` timed inside the run's window, plus `kicad-cli`. Staging all three by hand would be testing my staging. The automated tests above drive the real loop end to end against a real git repository, with real `snapshot`/`restore` calls and a real gc, which is the same code path a user would hit:

```text
$ npx vitest run test/rollback-failure.test.ts
 ✓ refuse path: still writes summary.md and names the rollback failure
 ✓ dry-run path: does not claim "changes reverted" when the revert failed
 ✓ a rollback that succeeds still reports the plain outcome
```

Nothing is mocked except the model turns.

## Docs

None needed: no flag, output format, or documented behavior changes. Two failure messages are added, both following the existing `fail()` wording so the three paths read alike.

---

## Invariant checklist

- [x] **Verification-gated out**: preserved. Every rollback still happens with the same arguments in the same place; only its failure is caught instead of thrown. The ERC/DRC gates and the `maxRepairCycles`-then-rollback flow are unchanged, and no path now skips a rollback it used to perform.
- [x] **`check`/`verify` stays LLM-free and network-free**: unchanged. `src/agent/loop.ts` is not reachable from `src/commands/check`, and no import was added beyond the existing `GitSnapshot` type.
- [x] **No sexp serialization**: preserved. Nothing here reads or writes a KiCad file.
- [x] **Sync-obligations ledger**: preserved, and this makes it more reliable rather than less. The ledger's open obligations are written into `summary.md`, which is one of the things the escaping throw was destroying.
- [x] **Secrets**: preserved. The new detail strings carry a git error message and are written through `transcript.writeSummary`, which runs `redactSecrets` on the whole document, and the `restore-failed` event goes through `transcript.event`, which redacts as well.

### DIFF
--- a/src/agent/loop.ts
+++ b/src/agent/loop.ts
@@ -12,7 +12,16 @@ import { Transcript, type ExitPath, type RunStats } from './transcript.js';
 import { collectRunMeta, renderCliHeader, type RunMeta, type RunMetaInput } from './runmeta.js';
 import { plainRenderer, fmtDuration, fmtTokens, type ProgressRenderer } from './render.js';
 import { ObligationsLedger } from './ledger.js';
-import { gitPreflight, isDirty, snapshot, restore, commitAll, changedFiles, preserveFailedRun } from '../util/git.js';
+import {
+  gitPreflight,
+  isDirty,
+  snapshot,
+  restore,
+  commitAll,
+  changedFiles,
+  preserveFailedRun,
+  type GitSnapshot,
+} from '../util/git.js';
 import { withRetry, isRateLimit, sessionLimit } from '../util/retry.js';
 import { openspecArchive } from '../openspec/cli.js';
 import { existsSync } from 'node:fs';
@@ -121,6 +130,29 @@ function otherProvider(current: Provider): Provider | null {
   if (current.name === 'openai' && process.env.ANTHROPIC_API_KEY) return new AnthropicProvider();
   if (current.name === 'anthropic' && process.env.OPENAI_API_KEY) return new OpenAIProvider();
   return null;
+}
+
+/**
+ * Roll the working tree back, and never let that be the thing that ends the
+ * run. `restore` shells out to git, so it throws on anything git refuses:
+ * most reachably, `git stash apply` when the snapshot's stash object is gone.
+ * `git stash create` leaves that object unreferenced by design, so a `git gc`
+ * during a long run collects it and the rollback throws afterwards.
+ *
+ * An unhandled throw here skips `run-end` and `summary.md`, which are most
+ * valuable exactly when the tree has been left in an unknown state. Returns
+ * the error message so a caller that surfaces it can, and records the failure
+ * on the transcript either way.
+ */
+async function rollBack(repoRoot: string, snap: GitSnapshot, transcript: Transcript): Promise<string | null> {
+  try {
+    await restore(repoRoot, snap);
+    return null;
+  } catch (err) {
+    const message = (err as Error).message;
+    await transcript.event('restore-failed', { error: message });
+    return message;
+  }
 }
 
 async function appendChangelog(
@@ -340,16 +372,7 @@ async function runWithMemory(
     // it, so a budget-exhaustion (or any) failure is recoverable (issue #15).
     const preserved = await preserveFailedRun(repoRoot, ctx.runId);
     if (preserved) await transcript.event('work-preserved', { stash: preserved });
-    // The rollback itself can fail (git in a bad state). That must not become
-    // an unhandled throw that skips run-end and summary.md — the summary is
-    // most valuable exactly when the tree is left in an unknown state.
-    let restoreError: string | null = null;
-    try {
-      await restore(repoRoot, snap);
-    } catch (err) {
-      restoreError = (err as Error).message;
-      await transcript.event('restore-failed', { error: restoreError });
-    }
+    const restoreError = await rollBack(repoRoot, snap, transcript);
     const runStats = stats(exitPath);
     await transcript.event('run-end', runStats);
     const summaryPath = await transcript.writeSummary({
@@ -571,7 +594,10 @@ async function runWithMemory(
       const { outcome, summary } = ctx.finishRequest;
       const files = [...ctx.filesTouched];
       if (outcome === 'refuse') {
-        await restore(repoRoot, snap);
+        const restoreError = await rollBack(repoRoot, snap, transcript);
+        if (restoreError) {
+          log(`WARNING: rollback failed (${restoreError}); the working tree may be in a partial state`);
+        }
         await transcript.event('run-refused', { summary });
         const runStats = stats('refused');
         await transcript.event('run-end', runStats);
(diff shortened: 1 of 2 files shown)


DPS0340:

Sorry for the volume.

PR #113, opened 2026-07-27
feat(repl): session UI with pinned observability dock and scripted demo recordings

animesh-chouhan:
## What

The REPL becomes a full interactive session UI: a pinned observability dock (spinner, morphing working word, turn and token counters, elapsed time), scrollable history, a slash-command palette, an in-session model picker, a boot animation built from the real favicon geometry, and a plain-text session log. Scripted vhs tapes record the whole thing reproducibly.

## Why

Long agent runs were hard to follow in linear scroll: status lines jittered, output buried the prompt, and there was no way to review earlier turns without losing the live view. The dock keeps run state visible at all times, history stays navigable, and the demo tapes make recordings regenerable after every style tweak instead of hand-captured.

## Design

- [dock.ts](src/util/dock.ts) owns the pinned-region terminal protocol (scroll region above, dock below); [dock-renderer.ts](src/agent/dock-renderer.ts) renders the observability row and drives the word-morph animation as a single char-by-char copper sweep.
- [live-prompt.ts](src/util/live-prompt.ts) owns raw-mode input: slash palette, history paging (PgUp/PgDn), wheel scroll, and double Ctrl+C exit, without echoing keys during a running turn. Input arrives through an escape-sequence assembler with bracketed paste (DECSET 2004), so a pasted multi-line request lands whole instead of submitting at its first newline, and a 40ms disambiguation timer that flushes a lone Esc without eating the next arrow key. The timer never fires inside a paste, where a held `\x1b` is the head of a split `\x1b[201~` rather than a keypress.
- [repl.ts](src/commands/repl.ts) mirrors session output to a log file with ANSI stripped and secrets redacted at write time via the shared `redactSecrets()`.
- [git.ts](src/util/git.ts): `repl` takes `--allow-dirty` with the same off-by-default meaning it has on `do`, which matters because `repl` is the default command. The snapshot now pairs a `git stash create` object for tracked changes with a tree object for untracked, non-ignored files, so a rollback's `git clean -fd` can no longer delete work the stash never captured. An untracked file that exists but cannot be read refuses the run by name rather than aborting with a raw git fatal.
- [cli.ts](src/kicad/cli.ts) now resolves `kicad-cli` via `COPPERHEAD_KICAD_CLI` or well-known macOS app-bundle paths on PATH miss; resolution is purely local (env var plus filesystem probe). A set-but-missing override is a hard error naming the path. The probe list is injectable so both the hit and the miss are testable on any platform, not only on a macOS host with KiCad installed.
- The layout contract lives in [repl-ui-layout.md](src/layout/repl-ui-layout.md); [ui-demo.ts](scripts/ui-demo.ts) drives the real renderer with canned output for recordings, and [demo.tape](scripts/demo.tape) / [demo-rename.tape](scripts/demo-rename.tape) script them under vhs.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | pass (469 passed, 17 skipped, 0 failed) |
| Spec artifacts | `openspec validate build-copperhead-phase-1` | pass |
| Live-LLM (opt-in) | keyed on `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `COPPERHEAD_TEST_CODEX` | not run (providers not configured) |

New tests on this branch: [repl.test.ts](test/repl.test.ts), [dock.test.ts](test/dock.test.ts), [repl-inspect.test.ts](test/repl-inspect.test.ts), [observability.test.ts](test/observability.test.ts), plus additions to [safety.test.ts](test/safety.test.ts) and [kicad-cli-resolve.test.ts](test/kicad-cli-resolve.test.ts). `npm run lint:md` is also clean (357 files, 0 issues).

Regression tests worth naming, each pinning a defect found by review on this branch rather than a happy path: a paste-end marker split at its ESC byte across a slow link (Enter and Ctrl+C stay live), a rollback whose untracked replay fails (the tracked state still rolls back), an unreadable untracked file (the run refuses by name), and the `kicad-cli` app-bundle probe hit and miss.

- Sandbox variant used (`create` / `edit`): n/a, UI layer only; exercised via the scripted mock session (`npm run demo:ui`), which drives the real renderer and prompt with canned agent output and no KiCad mutations.
- Commands run and outcome:

```text
$ vhs scripts/demo.tape          # full session recording
$ vhs scripts/demo-rename.tape   # rename-only 80x24 cut
Both rendered demo-runs/*.gif. Frame inspection confirmed: banner and via
mark, slash palette open/hover/dismiss, rename run with the pinned dock
animating (working-word morph, turn/token/elapsed), history paging, /check,
double Ctrl+C exit.
```

## Docs

- [repl-ui-layout.md](src/layout/repl-ui-layout.md) and [claude-ui-layout.md](src/layout/claude-ui-layout.md): layout specs for the session UI.
- CLI reference docs synced with the implemented surface (commit `05747a8`).
- Demo recording instructions in the layout spec; tapes in [scripts/](scripts/).

---

## Invariant checklist

- [ ] **Spec-gated in**: N/A, tool gating untouched.
- [x] **Verification-gated out**: the mutation/repair flow itself is untouched, but the rollback it falls back to is now lossless for untracked work, and `repl` no longer bypasses the dirty-tree gate that makes it safe.
- [x] **`check`/`verify` stays LLM-free and network-free**: the only change reachable from `check` is kicad-cli binary resolution ([cli.ts](src/kicad/cli.ts)), which uses only an env var and local filesystem probes.
- [ ] **No sexp serialization**: N/A, parser untouched.
- [ ] **Sync-obligations ledger**: N/A, ledger untouched.
- [x] **Secrets**: the new session log strips ANSI and redacts at write time via the shared `redactSecrets()` ([repl.ts](src/commands/repl.ts)), so it covers the registry and forge token shapes as well as `sk-` keys; `.gitignore` already covers `.env`, `.copperhead/runs/`, and `demo-runs/`.
- [x] **Spec coherence**: this branch does change spec-level behavior, so the artifacts moved with it. [SPEC.md](openspec/specs/SPEC.md) §3 documents `repl` and `demo`, §7 and AC-3.8 cover the `repl` dirty-tree gate, the two-part snapshot, and the unreadable-untracked-file refusal; the safety-rails delta spec and [design.md](openspec/changes/build-copperhead-phase-1/design.md) D6 match, with scenarios mapping onto AC-3.8. `openspec validate build-copperhead-phase-1` passes.

## Summary by CodeRabbit

* **New Features**
  * Added an interactive agent shell as the default experience, with slash commands, docked progress/status, REPL session logging, and TTY arrow-key model picking plus built-in repository inspection tools.
  * Added `copperhead demo` with `--tour` and full scaffold/create pipeline support, plus UI demo scripts.
* **Documentation**
  * Expanded quick start and CLI/REPL/demo/reference docs with default-command, TTY vs non-TTY, and model precedence details (including safer dirty-tree behavior).
* **Bug Fixes**
  * Improved `--allow-dirty` rollback to restore previously-untracked, non-ignored files.
* **Tests**
  * Added/expanded coverage for REPL flows, UI/prompt rendering, inspect output, demo execution, redaction, and KiCad/CLI resolution behavior.


animesh-chouhan:
Preview
copperhead-rename

aniket-intervue:
_Automated `pr-review` pass (not a human review). No prior automated pass found on this PR; CodeRabbit's own summary comment is separate._

## Verdict: approve with nits

Large but well-structured UI-layer PR. Claims in the description check out: typecheck, build, and `npm test` all pass locally as stated (432 passed / 17 skipped, matching the description exactly). Two real findings worth addressing before merge (one security-adjacent, one correctness), plus two lower-severity gaps. No repo-invariant violations found.

```
lines: +4628 / -67 (net +4561) across 36 files
       src +3058/-59 · test +1208 · docs +43 · openspec +19   (git diff --numstat $BASE...HEAD)
new vs net: 20 new files, 16 modified; ~3058 new src lines (the net-new surface)
tests: +1208 test lines; suite 361/17/0 -> 432/17/0 (pass/skip/fail, base -> head)
        base measured in a throwaway git worktree at the merge base
diff coverage: 54.9% of changed src .ts lines exercised (measured: @vitest/coverage-v8@3.2.7
        installed --no-save, coverage-final.json intersected with per-file added lines)
deps/audit: no dependency change (package-lock.json untouched; package.json adds only a
        `demo:ui` script). audit advisories unchanged.
CI: all 6 checks green (test, cla, greeting, label, template, CodeRabbit). mergeable: MERGEABLE.
```

Per-file diff coverage of changed src lines (measured): `cli.ts` 0/104 (0%), `theme.ts` 42/91 (46%), `animate.ts` 37/83 (45%), `kicad/cli.ts` 34/83 (41%), `logo.ts` 9/23 (39%), `demo.ts` 83/168 (49%), `repl.ts` 349/674 (52%), `repl-inspect.ts` 184/353 (52%), `dock.ts` 97/161 (60%), `live-prompt.ts` 309/499 (62%), `select.ts` 121/192 (63%), `dock-renderer.ts` 135/181 (75%).

---

## Findings

### 1. medium · REPL session log redacts only `sk-`, not the four other established secret patterns

[src/commands/repl.ts:321](src/commands/repl.ts#L321) writes the new session log with an inline regex:

```ts
logFile.write(l.replace(SGR_RE, '').replace(/sk-[A-Za-z0-9_-]+/g, 'sk-***') + '\n');
```

But [src/util/redact.ts](src/util/redact.ts) already exists and is the established write-time redactor for transcripts and summaries, covering five patterns: `sk-`, `Bearer <token>`, `npm_…`, `gh[pousr]_…`, and `github_pat_…`. `redactSecrets` is **not imported anywhere in repl.ts** (confirmed by grep), so the new log is strictly weaker than the surface it sits beside.

**Failure scenario**: a REPL session where the user pastes a `Bearer`/`npm_`/`ghp_` token, or a tool result echoes one (the redact.ts comment notes a quoted publish command or failing CI log does exactly this), persists it verbatim to `.copperhead/runs/repl-<ts>.log`. The comment at repl.ts:306 asserts "keys redacted per AC-4.1", which overstates what the code does. Note this is a *new persisted plain-text file*, so it widens the surface AC-4.1 was written for.

**Fix** (one line, reuses the tested util):

```ts
import { redactSecrets } from '../util/redact.js';
// ...
logFile.write(redactSecrets(l.replace(SGR_RE, '')) + '\n');
```

The existing test at [test/repl.test.ts:338](test/repl.test.ts#L338) only plants `sk-SECRET_KEY_123`; worth extending it to plant a `Bearer`/`npm_` token too, so the stronger guarantee is pinned.

### 2. medium · A misconfigured `COPPERHEAD_KICAD_CLI` is silently ignored, and the resulting error tells you to set what you already set

[src/kicad/cli.ts:38-46](src/kicad/cli.ts#L38-L46):

```ts
const fromEnv = process.env.COPPERHEAD_KICAD_CLI?.trim();
if (fromEnv && existsSync(fromEnv)) { cachedBinary = fromEnv; return fromEnv; }
cachedBinary = 'kicad-cli';   // env var set but path missing -> silently discarded
```

**Failure scenario**: user sets `COPPERHEAD_KICAD_CLI=/typo/path/kicad-cli`. `existsSync` fails, resolution falls through to a bare PATH lookup with no diagnostic. If `kicad-cli` is also not on PATH, `KicadCliMissingError` advises "or set `COPPERHEAD_KICAD_CLI=...`" — as though the user had never tried, giving no signal their override was seen and rejected. The same silent-discard appears again in `fallbackAfterMissing()` (cli.ts:52-56).

This matters more than usual because **the entire new resolution path is untested**: grep finds no test file referencing `resolveKicadCli`, `fallbackAfterMissing`, `COPPERHEAD_KICAD_CLI`, or the exported test helper `resetKicadCliCache` — and 41% is the measured diff coverage for this file. An exported helper whose only purpose is test-cache-reset, with no test using it, is a strong signal this branch was never exercised.

**Fix**: when `COPPERHEAD_KICAD_CLI` is set but does not exist, fail with that fact rather than falling through, e.g. `throw new KicadCliMissingError()` with a detail line naming the bad path, or at minimum surface a warning. Worth a test that sets the env var to a nonexistent path (calling `resetKicadCliCache()` first) and asserts the error names it.

### 3. low · New `repl` / `demo` CLI wiring has 0% diff coverage

[src/cli.ts:78-137](src/cli.ts#L78-L137) adds 104 new/changed lines with **zero** covered (measured). Untested branches include the non-TTY early throw, the interactive model-picker fallback (including the new `'picker'` `ModelSource`), the `--json`-with-repl refusal guard, and `--max-turns` parsing. A flag-mapping or picker-fallback regression would escape CI entirely. This is plumbing, so low severity, but it is the single largest uncovered block in the PR.

### 4. low · `demo --tour` ignores the global `--json` flag

Verified live: `copperhead --json demo --tour` prints human-readable prose. Every other command either emits JSON under `--json` or refuses; `repl` in this same PR correctly refuses (cli.ts:90-95). Since `--tour` short-circuits before the try block ([src/commands/demo.ts](src/commands/demo.ts), cli.ts:287-293), a script passing `--json` unconditionally gets unstructured text. Either refuse like `repl` does, or emit the tour as JSON.

---

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| **Spec-gated in** (`edit_file`/`write_file` absent until proposal validates) | **pass** — no change to `availableTools`/tool gating; the REPL routes each request through the existing `do` path with the same gates. |
| **Verification-gated out** (ERC/DRC then rollback) | **pass** — no change to the verification or rollback paths; `create.ts` edits are display-only (`stageLine`/`copper`/`dim` wrappers around existing log strings). |
| **`check`/`verify` LLM-free and network-free** | **pass** — `src/commands/check.ts` untouched; new `kicad/cli.ts` resolution is local only (env var + `existsSync`), no network. |
| **No sexp serialization** | **pass** — `src/kicad/` change is confined to binary resolution; the parser is untouched. |
| **Sync-obligations ledger** | **pass** — ledger and commit-gate logic untouched. |
| **Secrets** | **partial** — `sk-` is redacted in the new log, but see finding 1: the other four patterns in the shared util are not applied to this new persisted file. |
| **Spec coherence** | **pass** — `SPEC.md` §3 documents both new commands and AC-8.8 was amended for the color chrome, in the same PR. No delta-spec change needed (no active openspec change claims the REPL UI). |

## Notes

- **Mock-only runtime surface**: the REPL/dock/prompt layer is exercised entirely through injected loggers and canned agent output (`npm run demo:ui`, scripted providers). Real terminal behavior — raw-mode key handling, bracketed paste, alt-screen/scroll-region interaction, wheel scroll, terminal resize mid-turn — is unverified by the suite, as the PR's own manual-test log acknowledges ("n/a, UI layer only"). That is reasonable for a TUI, but worth stating plainly: the vhs tapes are recordings, not assertions.
- **New tests are deterministic**: spot-checked the four new/changed test files for network, `Date.now()`, and `Math.random()` dependence; they drive the real renderers with canned input and assert on output. `test/dock.test.ts` correctly resets global color state in `beforeEach`, and `test/repl.test.ts` / `test/repl-inspect.test.ts` call `setColorEnabled(false)` per test — good hygiene given `theme.ts` holds module-level mutable state (a real order-dependence hazard that was handled).
- **Scope**: broadly matches the stated purpose. The `kicad/cli.ts` resolution change (finding 2) is the one piece not obviously part of "session UI", though the description does disclose it.
- Authorship: commits by `animesh-chouhan` and `Devesh36`; none by this reviewer, so no self-review caveat applies.

animesh-chouhan:
_Automated `pr-review` pass (not a human review). A prior automated pass exists on this PR (2026-07-27 13:19Z). The head SHA is unchanged since (`d382639`), so this is a re-verification of those four findings plus a deeper adversarial pass on the input and rollback paths, not a fresh duplicate report._

## Verdict: request changes

All four findings from the prior pass are still open at `d382639` (re-verified line by line, none were addressed). The new work here found one issue the prior pass missed that I would not merge without: **the REPL, which this PR also makes the default command, disables the dirty-worktree gate unconditionally, and the rollback path then permanently deletes untracked files.** Verified against real git, not by reading. Two further input-handling bugs are confirmed with repros.

```
lines: +4599 / -72 (net +4527) across 36 files
       src +3088/-64 · test +1145/-0 · docs +48/-4 · openspec +19/-1 · scripts/pkg +299/-3
       (git diff --numstat $BASE..d382639, BASE=6d7631b9 = git merge-base origin/main d382639)
new vs net: 20 added files, 16 modified; ~3088 new src lines (the net-new surface)
tests: +1145 test lines; suite 361/17/0 -> 428/17/0 (pass/skip/fail, base -> head)
       both measured by running `npm test` in throwaway worktrees at BASE and at head
diff coverage: 70.6% (1564/2216) of *executable* added src lines
       (measured: @vitest/coverage-v8@3.2.7 installed --no-save, coverage-final.json
       statementMap intersected with added lines from `git diff -U0`; non-executable
       added lines, comments/blanks/types, excluded from the denominator)
deps/audit: no dependency change. package.json adds only a `demo:ui` script;
       package-lock.json untouched, so no audit delta.
CI: all 6 checks green. mergeable: MERGEABLE.
```

Two measurement notes, so the numbers reconcile with the earlier report and with the PR body:

- The prior pass reported 54.9% diff coverage. Same code, different denominator: it counted every added `.ts` line including comments, blank lines, and type-only declarations. 70.6% is the executable-line figure. Neither is wrong; the per-file ranking is nearly identical either way.
- The PR body claims `432 passed`. I measure **428 passed / 17 skipped / 0 failed** (34 files passed, 1 skipped), consistently across a plain run and a coverage run. Minor, but the table in the description overstates by 4.

Per-file diff coverage of executable added src lines (measured, worst first):

| file | covered | % |
| --- | --- | --- |
| [src/cli.ts](src/cli.ts) | 0/105 | **0%** |
| [src/kicad/cli.ts](src/kicad/cli.ts) | 34/69 | 49% |
| [src/commands/repl-inspect.ts](src/commands/repl-inspect.ts) | 184/322 | 57% |
| [src/agent/animate.ts](src/agent/animate.ts) | 40/63 | 63% |
| [src/commands/repl.ts](src/commands/repl.ts) | 349/538 | 65% |
| [src/commands/demo.ts](src/commands/demo.ts) | 88/132 | 67% |
| [src/commands/create.ts](src/commands/create.ts) | 57/79 | 72% |
| [src/agent/theme.ts](src/agent/theme.ts) | 42/55 | 76% |
| [src/util/select.ts](src/util/select.ts) | 121/146 | 83% |
| [src/util/live-prompt.ts](src/util/live-prompt.ts) | 307/355 | 86% |
| [src/util/dock.ts](src/util/dock.ts) | 97/103 | 94% |
| [src/agent/dock-renderer.ts](src/agent/dock-renderer.ts) | 135/137 | 99% |
| [src/agent/box.ts](src/agent/box.ts) | 68/68 | 100% |

---

## Findings

### 1. high · NEW · The REPL is the default command and it silently disables the dirty-tree gate; a failed turn then destroys every untracked file in the repo

[src/commands/repl.ts:283](src/commands/repl.ts#L283) hardcodes `allowDirty: true` in `defaultRunner`, with no flag to turn it off, no prompt, and no comment explaining it. The REPL has no gate of its own: grep finds no `gitPreflight` in [repl.ts](src/commands/repl.ts), only `isDirty` calls at lines 247 and 580 that *display* dirty state in the status bar. And [src/cli.ts:88](src/cli.ts#L88) registers `repl` with `{ isDefault: true }`, so bare `copperhead` now takes this path.

Compare `do`, in the same file: [src/cli.ts:241](src/cli.ts#L241) is `allowDirty: opts.allowDirty ?? false`, an explicit opt-in flag. The gate being bypassed states its own purpose at [src/util/git.ts:99](src/util/git.ts#L99): *"a rollback hard-resets to the pre-run snapshot, which would silently destroy your uncommitted work"*.

The preservation it promises is incomplete. [snapshot()](src/util/git.ts#L114) captures dirty state with `git stash create`, which **does not include untracked files**, and [restore()](src/util/git.ts#L127) runs `git clean -fd -e .copperhead/runs`, which **deletes them**.

**Failure scenario**, verified against real git (init a repo, one commit, then modify a tracked file and add untracked files, then run exactly the `stash create` / `reset --hard` / `clean -fd` / `stash apply` sequence from `snapshot()` and `restore()`):

```
--- before ---
 M tracked.txt
?? docs/
?? untracked-notes.md
--- after rollback ---
 M tracked.txt
tracked.txt          = v2-my-uncommitted-edit   (restored by stash apply)
untracked-notes.md   *** GONE ***
docs/new-doc.md      *** GONE ***
```

So: a user opens `copperhead` in a repo holding a hand-written `docs/note.md` or a not-yet-added `.kicad_sch`, asks for an edit, verification fails, the loop rolls back at [loop.ts:349](src/agent/loop.ts#L349), and those files are gone with no stash entry and no warning. This is the exact data loss the preflight exists to prevent, on what is now the default command. It also undercuts the verification-gated-out invariant's promise that a failed run rolls back losslessly.

**Fix** (either):

- Gate it like `do`: drop the hardcoded `true`, add `--allow-dirty` to the `repl` command, and let `gitPreflight` refuse a dirty tree by default. This is the smaller change and matches the existing contract.
- Or make the snapshot genuinely lossless before keeping `allowDirty: true`: capture untracked files as well (`git stash create` cannot; use `git stash push -u` plus `git stash store`, or stage untracked paths into a temp index before `stash create`), so `restore()` can put them back.

Worth a regression test either way: dirty repo with an untracked file, force a verification failure, assert the untracked file still exists after rollback.

### 2. medium · NEW · A pasted multi-line request fires an agent run on only its first line

No bracketed-paste mode is enabled anywhere in `src/` (`grep -rn "2004" src/` returns nothing), and [pushKeys](src/util/live-prompt.ts#L122) emits pasted bytes one character at a time, so an embedded `\n` hits the submit branch at [live-prompt.ts:424](src/util/live-prompt.ts#L424) like a real Enter. `drainPrintable` exists specifically to coalesce pastes, so paste is an anticipated input here, and it stops at the newline.

**Failure scenario**, confirmed with a repro against the real `promptWithSlashHints`: feeding the reader `"rename R1 to R10\nand update the BOM\n"` returns `"rename R1 to R10"`. In the REPL that string goes straight to `runAgentLoop` via [defaultRunner](src/commands/repl.ts#L276), so pasting a two-line change request starts a real, potentially mutating run on half of it, and the remainder lands in the next prompt as stray input.

**Fix**: enable bracketed paste (`\x1b[?2004h` on entry, `\x1b[?2004l` on exit) and have `pushKeys` treat the `\x1b[200~` … `\x1b[201~` payload as literal text, mapping newlines inside it to spaces or to a soft newline rather than to submit. `pushKeys` already consumes `\x1b[200~` as a single key, so the assembler needs only a paste-mode flag. Repro test:

```ts
it('a pasted multi-line request is not submitted at the first newline', async () => {
  const chunk = 'rename R1 to R10\nand update the BOM\n';
  let i = 0;
  const line = await promptWithSlashHints({
    prompt: '> ', commands: SLASH_COMMANDS,
    output: new PassThrough() as unknown as NodeJS.WriteStream,
    readKey: async () => (i < chunk.length ? chunk[i++]! : null),
  });
  expect(line).toBe('rename R1 to R10 and update the BOM');  // currently 'rename R1 to R10'
});
```

### 3. medium · STILL OPEN · Session log redacts only `sk-`, not the four other established secret patterns

Unchanged from the prior pass. [src/commands/repl.ts:321](src/commands/repl.ts#L321) still writes the session log with an inline regex:

```ts
logFile.write(l.replace(SGR_RE, '').replace(/sk-[A-Za-z0-9_-]+/g, 'sk-***') + '\n');
```

[src/util/redact.ts](src/util/redact.ts) remains the established write-time redactor for transcripts and summaries, covering five patterns (`sk-`, `Bearer <token>`, `npm_…`, `gh[pousr]_…`, `github_pat_…`), and is still not imported by `repl.ts`. The comment at [repl.ts:306](src/commands/repl.ts#L306) claiming "keys redacted per AC-4.1" still overstates what the code does, and this is a new persisted plain-text file, so it widens the surface AC-4.1 covers.

**Fix**, one line: `logFile.write(redactSecrets(l.replace(SGR_RE, '')) + '\n')`. Extend the test at [test/repl.test.ts:338](test/repl.test.ts#L338), which plants only `sk-SECRET_KEY_123`, to plant a `Bearer`/`npm_` token too.

For what it is worth, I checked the ANSI-stripping half separately and it is fine: `SGR_RE` strips only SGR sequences, but every line reaching `log()` is built from `theme.ts`/`box.ts` helpers that emit SGR only. Cursor-addressing sequences come from `animate.ts`/`dock.ts`, which write to the output stream directly and never through `log()`. No leak there.

### 4. medium · STILL OPEN · A misconfigured `COPPERHEAD_KICAD_CLI` is silently ignored, and the whole new resolution path is untested

Unchanged. [src/kicad/cli.ts:44-50](src/kicad/cli.ts#L44-L50): when `COPPERHEAD_KICAD_CLI` is set but the path does not exist, resolution falls through to a bare PATH lookup with no diagnostic, and the resulting `KicadCliMissingError` advises setting the very variable the user already set. Same silent discard in `fallbackAfterMissing()` at [cli.ts:55-59](src/kicad/cli.ts#L55-L59).

The prior pass called this untested by grep. It is now measured: coverage marks lines **46, 47, 48** (the env-var-honored branch) and **54 through 68** (all of `fallbackAfterMissing`) as never executed. Only the plain `cachedBinary = 'kicad-cli'` default runs. Still no test references `resolveKicadCli`, `fallbackAfterMissing`, `COPPERHEAD_KICAD_CLI`, or the exported `resetKicadCliCache`, whose sole purpose is test cache reset.

**Fix**: when the env var is set but missing on disk, fail with that fact (a `KicadCliMissingError` detail line naming the bad path) rather than falling through. Add a test that sets the var to a nonexistent path after `resetKicadCliCache()` and asserts the error names it.

### 5. low · NEW · A lone Escape keypress is never delivered, so the slash-menu dismiss handler is dead in practice

[live-prompt.ts:432](src/util/live-prompt.ts#L432) handles `key === '\x1b'` to clear an open slash buffer. But [pushKeys](src/util/live-prompt.ts#L122) returns without emitting when the buffer is exactly `\x1b` (`if (s.length === 1) return;`), which is how a terminal delivers a bare Escape. The key is held in `pending.buf` until some *other* key arrives, then emitted just before it.

**Failure scenario**, confirmed by repro: `pushKeys({buf:''}, '\x1b', emit)` emits nothing; a subsequent `pushKeys(pending, 'a', emit)` emits `['\x1b', 'a']`. So pressing Escape with the menu open does nothing, and the dismiss fires one keystroke late. The existing test at [test/repl.test.ts:309-311](test/repl.test.ts#L309-L311) asserts `expect(keys).toEqual([])` after a lone `\x1b`, which pins the behavior rather than catching it.

Impact is presentational only (Backspace and Ctrl+U still clear the buffer), hence low. **Fix**: the standard ESC-disambiguation timer, flush `pending.buf` as a lone `\x1b` if no follow-on byte arrives within roughly 25 to 50ms.

### 6. low · STILL OPEN · New `repl` / `demo` CLI wiring has 0% diff coverage

Confirmed and now measured precisely: **0 of 105** executable added lines in [src/cli.ts](src/cli.ts) are covered. Uncovered includes the non-TTY early throw, the interactive model-picker fallback and its new `'picker'` `ModelSource`, the `--json`-with-repl refusal guard at [cli.ts:100-104](src/cli.ts#L100-L104), and the `parseMaxTurns` validator added in `d382639` at [cli.ts:45-51](src/cli.ts#L45-L51). Still the single largest uncovered block in the PR. I traced `parseMaxTurns` by hand and it is correct (`Number("")` is unreachable because `opts.maxTurns` is falsy for an empty string; `"1e3"` is accepted as 1000, which is defensible).

### 7. low · STILL OPEN · `demo --tour` ignores the global `--json` flag

Confirmed unchanged: [src/cli.ts:302](src/cli.ts#L302) `if (opts.tour)` short-circuits before any `--json` handling, so `copperhead --json demo --tour` prints prose. Every other command either emits JSON under `--json` or refuses, and `repl` in this same PR correctly refuses. Either refuse like `repl` does, or emit the tour as JSON.

### 8. low · NEW · The colored branches of `styleOutcome` and `styleHeaderLines` are never exercised

[theme.ts:72-77](src/agent/theme.ts#L72-L77) and [theme.ts:82-90](src/agent/theme.ts#L82-L90) are both measured uncovered. Both are new exported symbols that early-return when `colorEnabled` is false, and every test calls `setColorEnabled(false)`, so only the pass-through path runs. The regex at line 85 (`/^(copperhead v\S+)(.*)$/`) and the token classifier at line 75 are exactly the kind of thing a header-format change breaks silently. One test each with `setColorEnabled(true)` closes this.

---

## Untested surface (new symbols, branches, and error paths no test reaches)

Reconciles with the uncovered lines above:

- `resolveKicadCli` env-var branch and all of `fallbackAfterMissing` ([kicad/cli.ts](src/kicad/cli.ts) 46-48, 54-68), plus the exported `resetKicadCliCache` (finding 4).
- Every new line of [src/cli.ts](src/cli.ts): repl and demo command wiring, `parseMaxTurns`, `pickModel` fallback, `'picker'` source, `--json` refusal (finding 6).
- `styleOutcome` / `styleHeaderLines` colored paths (finding 8).
- [repl-inspect.ts](src/commands/repl-inspect.ts) at 57%: the "no schematic configured" and "schematic missing" early returns in `resolveSchematic` (lines 41-46) and the bulk of the `format*Inspect` bodies from line 93 on are unexercised. Read-only formatters, so low risk, but they are the largest remaining block after `cli.ts`.
- [demo.ts](src/commands/demo.ts) at 67%: the `scaffoldDemoRepo` fail-closed branch added in `d382639` (lines 67-70) is uncovered. I traced it by hand and it is correct, `readdir` includes dotfiles so a stray `.git` directory does trip the refusal.

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| **Spec-gated in** (`edit_file`/`write_file` absent until a proposal validates) | **pass**. No change to tool gating. `defaultRunner` ([repl.ts:276](src/commands/repl.ts#L276)) routes each request through the same `runAgentLoop` with the same gates. |
| **Verification-gated out** (ERC/DRC, repair, then rollback) | **partial**. The verify and repair paths are untouched and the rollback still runs, but see finding 1: the REPL bypasses the preflight that makes that rollback lossless, so a rolled-back run can destroy untracked work. |
| **`check`/`verify` LLM-free and network-free** | **pass**. [check.ts](src/commands/check.ts) untouched; the only reachable change is kicad-cli resolution, which is an env var read plus `existsSync` probes, no network, no provider. |
| **No sexp serialization** | **pass**. The [src/kicad/](src/kicad/) change is confined to binary resolution; the parser is untouched and still never serializes. |
| **Sync-obligations ledger** | **pass**. Ledger, post-tool-call hooks, and the commit gate are untouched. `/sync` in the REPL is verify-only via `syncVerify` and never resolves. |
| **Secrets** | **partial**. `.gitignore` still covers `.env` and `.copperhead/runs/` (the new log lands under the latter), and ANSI stripping is sound, but see finding 3 for the four unhandled patterns. |
| **Spec coherence** | **pass**. [SPEC.md](openspec/specs/SPEC.md) §3 documents both new commands in the same PR; no active openspec change claims the REPL UI, so no delta spec is owed. |

## Notes

- **Mock-only runtime surface.** The REPL, dock, and prompt layer is exercised entirely through injected streams and canned agent output. Real terminal behavior, raw-mode key delivery, alt-screen and DECSTBM interaction, wheel scroll, and resize mid-turn, is unverified by the suite, as the PR's own manual-test log acknowledges. Reasonable for a TUI. The vhs tapes are recordings, not assertions. Findings 2 and 5 are both in this blind spot, which is where I would look next.
- **Adversarial input pass on `pushKeys`.** Beyond finding 5, I traced the assembler against bracketed-paste markers (`\x1b[200~`), SGR mouse reports (`\x1b[<64;10;5M`), modified arrows (`\x1b[1;2A`), and SS3 (`\x1bOA`). The final-byte scan (`0x20 <= c < 0x40`) consumes all of these correctly as single keys. Non-BMP input also survives: `s[0]` yields surrogate halves, but they are appended to the buffer in order and reassemble. The one structural gap is that an incomplete CSI never times out, so `pending.buf` can grow unbounded on a truncated sequence, which is theoretical rather than reachable.
- **`truncateVisible` / `wrapSpans`** ([box.ts](src/agent/box.ts)) are at 100% diff coverage and I could not break them: truncation closes SGR state when it cut inside a colored run, and `wrapSpans` cannot spin because `curLen < w` holds at every slice. Width math counts code units, so double-width CJK will misalign the box borders, which is cosmetic and not worth a finding.
- **New tests are deterministic.** Spot-checked all four new and changed test files for network, `Date.now()`, and `Math.random()` dependence, and for order coupling. Clean. `dock.test.ts` resets global color state in `beforeEach` and `repl.test.ts` / `repl-inspect.test.ts` call `setColorEnabled(false)` per test, which is the right hygiene given `theme.ts` holds module-level mutable state. That same discipline is what leaves finding 8 uncovered.
- **Scope.** Broadly matches the stated purpose. The `kicad/cli.ts` resolution change is the one piece not obviously "session UI", though the description discloses it.
- **Authorship**: commits by `animesh-chouhan` and `Devesh36`, none by this reviewer, so no self-review caveat applies.

animesh-chouhan:
_Follow-up to the automated `pr-review` pass above: all eight findings are addressed in `52ce137`._

## What changed

| # | Sev | Finding | Fix |
| --- | --- | --- | --- |
| 1 | high | REPL forced `allowDirty` and rollback deleted untracked files | `--allow-dirty` flag, default off; snapshot now captures untracked files too |
| 2 | medium | Pasted multi-line request submitted only its first line | Bracketed paste (DECSET 2004) + paste-aware assembler |
| 3 | medium | Session log redacted only `sk-` | Uses the shared `redactSecrets()` |
| 4 | medium | Bad `COPPERHEAD_KICAD_CLI` silently ignored | New `KicadCliBadOverrideError` naming the path |
| 5 | low | Lone Escape never delivered | 40ms disambiguation timer in `KeyReader` |
| 6 | low | `cli.ts` 0% diff coverage | Helpers extracted to `src/util/cli-args.ts`, at 100% |
| 7 | low | `demo --tour` ignored `--json` | Emits `{ "tour": [...] }` |
| 8 | low | `theme.ts` colored branches uncovered | Color-on tests; 76% to 96% |

## Finding 1 in detail, since the fix is two-part

**The gate.** [repl.ts](src/commands/repl.ts) no longer hardcodes `allowDirty: true`. `repl` takes a `--allow-dirty` flag with the same off-by-default meaning it has on `do`. This matters because `repl` is the default command, so bare `copperhead` was the path being run unguarded.

**The reason the gate mattered.** The rollback was not lossless the way [git.ts](src/util/git.ts) claimed. `git stash create` captures only tracked changes, so `restore()`'s `git clean -fd` was deleting every untracked file with nothing to recover it from. `snapshot()` now also writes the untracked-but-not-ignored paths to a tree object and `restore()` checks them back out after the stash applies.

Two details worth flagging for review:

- The scratch index lives inside the repo's own git dir rather than `TMPDIR`. Git guarantees that directory exists and is writable, so a missing or hostile temp dir cannot make a snapshot throw and block a run from starting. (`restore()` already only warns when its audit-trail backup cannot use `TMPDIR`; a snapshot failure would have been fatal.)
- Ignored paths are excluded from both sides: `ls-files --exclude-standard` skips them and plain `clean -fd` (no `-x`) leaves them alone. That symmetry is what keeps `.env` out of the object store, and it is pinned by a test.

Two regression tests in [safety.test.ts](test/safety.test.ts) cover it: one asserts a rollback returns the user's uncommitted edit and both untracked files while still discarding the failed run's own scratch file, the other asserts `.env` is neither captured nor disturbed.

## Verification

```
typecheck   pass
build       pass
lint:md     pass (357 files, 0 issues)
npm test    464 passed / 17 skipped / 0 failed   (was 428/17/0, +36 tests)
```

Diff coverage on changed `src/` lines: **73.6%**, up from 70.6% (same method: `@vitest/coverage-v8@3.2.7` installed `--no-save`, `coverage-final.json` intersected with added lines, executable lines only). Per-file movement:

| file | before | after |
| --- | --- | --- |
| `agent/theme.ts` | 76% | **96%** |
| `kicad/cli.ts` | 49% | **67%** |
| `util/live-prompt.ts` | 86% | **89%** |
| `commands/repl.ts` | 65% | **68%** |
| `util/cli-args.ts` (new) | n/a | **100%** |
| `util/git.ts` (new lines) | n/a | **92%** |

`src/cli.ts` itself stays at 0% and that is now structural rather than incidental: it parses argv at import, so it cannot be loaded in-process. Everything in it with a branch worth pinning moved to `cli-args.ts`, and the one remaining wiring branch (`demo --tour` under `--json`) is covered end to end by spawning the real entry point through `tsx`. The type-only `loop.js` import in `cli-args.ts` is erased at compile (verified against `dist/`), so `check` stays LLM-free and network-free.

## Spec coherence

`SPEC.md` moved with the code: §7 now describes the two-part snapshot, AC-3.8 says the dirty-tree refusal covers `repl` as well as `do` and that `--allow-dirty` restores untracked files, and the `repl` command block lists the flag. The CLI reference documents the flag, the `kicad-cli` override being a hard error, `--tour --json`, and the corrected description of what the session log redacts.

## Still open

Nothing from the review. Two things deliberately left alone, both called out in the original report as not findings:

- The TUI remains exercised through injected streams rather than a real terminal. The new paste and Escape tests drive the real assembler and the real `KeyReader` (with fake timers, so no wall-clock dependency), which is the closest deterministic equivalent, but raw-mode delivery and alt-screen interaction are still unverified by the suite.
- `repl-inspect.ts` sits at 57%. Read-only formatters, unchanged by this commit.

animesh-chouhan:
_Automated `pr-review` pass (not a human review). Incremental: a prior automated pass exists on this PR (2026-07-27 15:45Z, head `52ce137`), which reported all eight of its findings closed. Two commits have landed since, both test-only. This pass covers those and re-measures the numbers the prior pass asserted._

**Authorship disclosure**: `c6ca52b` was authored by this reviewer in the current session, from a CodeRabbit nitpick. It is self-review and is treated with more skepticism below, not less.

## Verdict: approve

Both new commits are test-only (+60/-1, zero `src/` lines changed), deterministic, and each closes exactly the gap it claims. Verified by measurement rather than by reading the commit messages. Two low findings, both in the new test code, neither blocking.

## Metrics

```
since 52ce137: +60 / -1 across 2 files (test only; 0 src lines changed)
  test/kicad-cli-resolve.test.ts +27    test/safety.test.ts +33/-1
whole PR vs merge-base 6d7631b: +5436 / -90 (net +5346) across 42 files
  src +3360/-78 · test +1704/-1 · scripts +298/-3 · docs +45/-4 · openspec +23/-3
suite: 462 pass / 17 skip / 0 fail  ->  465 / 17 / 0  (+3 tests)
  method: vitest run at 52ce137 in a throwaway worktree, and at head; not taken from the prior comment
diff coverage on changed src lines (measured): @vitest/coverage-v8@3.2.7 installed --no-save,
  coverage-final.json intersected with added lines from `git diff -U0 $BASE...head`, executable lines only
  src/util/git.ts    49/49 added executable lines = 100%    (prior pass: 92%)
  src/kicad/cli.ts   77/90 added executable lines = 85.6%   (prior pass: 67%)
  uncovered in cli.ts: 92-94, 97-99, 118-122, 246, 261, 263
deps/audit: unchanged, neither commit touches package.json
CI at review time: label pass; cla, template, test, CodeRabbit all pending
```

**Correction to the prior pass**: it reported the base suite as 464 pass / 17 skip. Measured at `52ce137` it is 462 / 17. With 462 the +3 delta reconciles exactly with the three new tests (two kicad, one git); with 464 it did not.

## Both commits do what they claim

**`c6ca52b`** (git.ts): base coverage marked lines **236-239** uncovered, which is precisely the best-effort `catch` around `restoreUntracked`. At head those lines are absent from the uncovered set and the file's added-line coverage is 100%. The test forces the failure through a real input, a snapshot whose `untracked` tree sha is 40 zeros, rather than mocking the private helper. `git read-tree 000...0` fails with exit 128 (confirmed by running it), so the branch is entered deterministically on every platform.

**`a8baacb`** (kicad/cli.ts): the exhausted path of `fallbackAfterMissing` (`cachedBinary = null; throw`, cli.ts:101-102) is now covered, lifting the file from 73.0% to 81.9% overall and its added lines from 67% to 85.6%.

## Findings

### 1. low · NEW · The new ENOENT test is not hermetic and fails on macOS with KiCad installed

[test/kicad-cli-resolve.test.ts:76-88](test/kicad-cli-resolve.test.ts#L76-L88) sets `PATH` to an empty directory and asserts `kicadCliVersion()` rejects with `KicadCliMissingError`.

Trace: `resolveKicadCli()` returns `'kicad-cli'`, `execa` fails ENOENT, `bin === 'kicad-cli'` so `fallbackAfterMissing()` runs ([cli.ts:110](src/kicad/cli.ts#L110)), which loops `FALLBACK_BINARIES` ([cli.ts:95-100](src/kicad/cli.ts#L95-L100)). On a host with KiCad at `/Applications/KiCad/KiCad.app/Contents/MacOS/kicad-cli`, `existsSync` is true, the function returns that path, `runKicad` retries at [cli.ts:111](src/kicad/cli.ts#L111) and succeeds, so `kicadCliVersion()` resolves a version string and `rejects.toBeInstanceOf` fails. Overriding `PATH` does not gate this: the fallback entries are absolute paths.

The test name asserts "and no bundle matches", but nothing in the test establishes that; it is inherited from the host. CI is unaffected (every workflow is `runs-on: ubuntu-latest`), so this fails only for contributors on the one platform the fallback list exists to serve.

Coverage corroborates the shape of it: cli.ts **97-99**, the bundle-found branch, is measured uncovered, so the branch that would fire on macOS is exactly the one no test pins.

**Fix**, one line each: export `FALLBACK_BINARIES` and guard the test.

```ts
it.skipIf(FALLBACK_BINARIES.some(existsSync))('reports the binary as missing when …', …)
```

Better, and it closes 97-99 at the same time: make the probe list injectable (a module-level `let fallbacks = FALLBACK_BINARIES` plus a test-only setter, mirroring the existing `resetKicadCliCache`), so both the found and the exhausted branch get a deterministic test on every platform.

### 2. low · NEW · The env re-check inside `fallbackAfterMissing` is unreachable in a single process

[src/kicad/cli.ts:90-94](src/kicad/cli.ts#L90-L94), measured uncovered. `fallbackAfterMissing()` runs only when `bin === 'kicad-cli'` ([cli.ts:109](src/kicad/cli.ts#L109)), and `resolveKicadCli` only yields `'kicad-cli'` when `envOverride()` already returned null and cached that result ([cli.ts:65-71](src/kicad/cli.ts#L65-L71)); set-but-missing throws there instead. For 92-94 to execute, `COPPERHEAD_KICAD_CLI` would have to be set between `resolveKicadCli()` and the ENOENT within the same process. Not reachable from the CLI.

Harmless, but it reads as a live branch and is a permanent 3-line hole in the file's coverage. Either delete it or comment it as defensive-only so a future coverage pass does not chase it.

### 3. nit · self-review on `c6ca52b`

The new case in [test/safety.test.ts:215-245](test/safety.test.ts#L215-L245) asserts that the tracked rollback survives and that the warning is emitted, but not the corollary that `hand-written-notes.md` is absent afterwards, which is the actual user-visible consequence of the failed replay. An `expect(existsSync(...)).toBe(false)` would document the full outcome. Not a defect: the claim the test makes in its name is fully asserted.

## Test-quality checks on the two new files

- **Deterministic**: no network, no `Date.now()`, no `Math.random()`, no wall-clock dependence in either file. Pass.
- **Real assertions**: all three new tests assert.
- **Order independence**: the kicad tests rely on `resetKicadCliCache()` in `beforeEach` and env restore in `afterEach`, both present. Test A restores `PATH` in a `finally`. Test B leaves `COPPERHEAD_KICAD_CLI` pointing at a deleted path, but `afterEach` restores the saved value and resets the cache. The safety test restores its `console.warn` spy in a `finally`. No leakage found.
- **Test B is hermetic**, unlike Test A: the cached binary is an absolute temp path, so ENOENT is guaranteed regardless of host, and `bin !== 'kicad-cli'` routes to the direct throw at [cli.ts:113](src/kicad/cli.ts#L113). Confirmed by trace and by execution.

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| **Spec-gated in** | **pass**. Untouched; neither commit changes `src/`. |
| **Verification-gated out** | **pass**. Untouched. The rollback path gains test coverage only. |
| **`check`/`verify` LLM-free and network-free** | **pass**. Both new tests are local only: env vars, filesystem probes, `git` subprocesses. `kicadCliVersion()` is spawned only against an empty `PATH` or a deleted absolute path. |
| **No sexp serialization** | **pass**. Parser untouched. |
| **Sync-obligations ledger** | **pass**. Untouched. |
| **Secrets** | **pass**. No new redaction surface; `.gitignore` unchanged. |
| **Spec coherence** | **pass**. Test-only commits, no spec-level behavior change, so no artifact movement is owed. |

## Carried forward, unchanged

Both were disclosed in the prior pass as accepted rather than open, and neither is affected by these two commits: the TUI layer is still exercised through injected streams rather than a real terminal, and `repl-inspect.ts` is still near 57%.

## Residual untested surface in the two touched modules

- [cli.ts:92-94](src/kicad/cli.ts#L92-L94) and [97-99](src/kicad/cli.ts#L97-L99): the env re-check (finding 2) and the bundle-found branch (finding 1).
- [cli.ts:118-122](src/kicad/cli.ts#L118-L122): `runKicad`'s post-fallback ENOENT throw and its generic non-zero-exit error mapping.
- [cli.ts:246](src/kicad/cli.ts#L246) and [263](src/kicad/cli.ts#L263): the `KicadCliMissingError` rethrow guards in the fab-export loop and `exportSvg`, reachable only when KiCad disappears mid-export.
- `src/util/git.ts` has no residual gap on this PR's added lines. Its three remaining uncovered spots (39, 249-250, 258-259) are pre-existing best-effort `catch` blocks outside this PR's diff.

AniketR10:
_Automated `pr-review` pass (not a human review). Two prior automated passes exist on this PR (13:19Z against `d382639`, 14:47Z against `d382639`), plus the author's fix summary at 15:45Z. Head has moved to `a8baacb`, so this is a **delta review**: verification of the eight closed findings and a fresh pass over the two new commits (`52ce137`, `a8baacb`), not a duplicate full report. Neither delta commit is self-authored._

## Verdict: request changes

All eight prior findings are genuinely closed — I re-verified each against the working tree, not against the summary. The `--allow-dirty` gate, the two-part snapshot, the shared redactor, `KicadCliBadOverrideError`, the `cli-args.ts` extraction and the `--tour --json` contract all check out, with real regression tests behind them.

The fix commit introduces one new defect I would not merge: **the Escape disambiguation timer added for finding 5 fires inside the bracketed paste added for finding 2, and the REPL session becomes unrecoverable from the keyboard** — Enter and both Ctrl+C presses are swallowed. Confirmed with a repro against the real `KeyReader`; the fix is one line and I verified it closes the repro. One further medium on the new snapshot path, plus three lows.

```text
delta since the last automated pass (d382639..a8baacb, 2 commits):
lines: +849 / -62 across 13 files  ·  src +310/-52, test +528/-2, docs +5/-4, spec +6/-4
       (git diff --numstat d382639..a8baacb)

whole PR (metrics.mjs --base upstream/main; merge base 6d7631b):
lines: +5403 / -89 (net 5314) across 42 files  ·  src +3360/-78, test +1671/-0, docs +50/-5, spec +23/-3, other +299/-3
new vs net: 24 new files (14 in src), 18 modified; 2899 new src lines (the net-new surface)
tests: +1671 test lines; suite base not measured -> 464/17/0 (pass/skip/fail)
       head measured by `npx vitest run --reporter=json` in a throwaway worktree at a8baacb;
       base not run (base CI is green on main; --base-tests not used)
diff coverage: 68.3% of changed executable src lines (1755/2571, measured: @vitest/coverage-v8,
       coverage-final.json intersected with changed lines from git diff -U0)
       — 211 of the uncovered lines are the two src/layout/*.md layout specs, which the script
       counts as src; excluding them the figure is 74.4% (1755/2360), consistent with the
       author's 73.6%.
deps: +[none] -[none] ~[none] (package.json changed: a `demo:ui` script only; lockfile untouched,
       so no audit delta)
ci: 5 pass / 0 fail / 0 pending (gh pr checks); required checks: 0 failing, 0 pending.
       mergeable: MERGEABLE. Not a draft.
```

Uncovered changed src lines intersecting the delta (full per-file list in the script's detail output):
`src/util/git.ts:236-239` · `src/util/live-prompt.ts:182-186,252-254,271,273-279,347-348,487,538,544-576` · `src/kicad/cli.ts:92-94,97-99,118-119,121-122,246,261,263` · `src/cli.ts:70-98,101-130,279-289,292-295`. `src/util/cli-args.ts` is absent from the list — 100%, as claimed. These reconcile with finding 5 below.

---

## Verification of the eight closed findings

| # | Prior finding | Verified at `a8baacb` |
| --- | --- | --- |
| 1 | REPL forced `allowDirty`, rollback deleted untracked files | **Closed.** `repl.ts:290` is `opts.allowDirty ?? false`; `cli.ts:76,117` registers and plumbs `--allow-dirty`. `snapshotUntracked()`/`restoreUntracked()` in `git.ts` capture untracked-but-not-ignored paths to a tree object via a scratch `GIT_INDEX_FILE`. Two real-git regression tests in `safety.test.ts` (restore returns both untracked files and still discards the run's own scratch file; `.env` neither captured nor disturbed). I separately confirmed the ignored/clean symmetry holds and that an untracked **nested** git repo is listed as a bare directory entry, silently skipped by `update-index` — benign, because plain `clean -fd` does not remove nested repos either. |
| 2 | Multi-line paste submitted only its first line | Closed for the reported case — but see finding 1 below for the new interaction. |
| 3 | Session log redacted only `sk-` | **Closed.** `repl.ts:330` now calls `redactSecrets()`; `repl.test.ts` plants `sk-`, `Bearer`, `npm_` and `ghp_` tokens and asserts `[REDACTED]`. |
| 4 | Bad `COPPERHEAD_KICAD_CLI` silently ignored | **Closed.** `envOverride()` throws `KicadCliBadOverrideError` naming the path, used by both `resolveKicadCli()` and `fallbackAfterMissing()`. `kicad-cli-resolve.test.ts` adds 8 tests including the mid-session-disappearance case. |
| 5 | Lone Escape never delivered | Closed for the reported case — see finding 1. |
| 6 | `cli.ts` 0% diff coverage | **Closed as far as it can be.** `repoOf`/`parseMaxTurns`/`budgetExtraTurns`/`budgetPromptText` moved to `cli-args.ts` at 100%. The `loop.js` import there is type-only: I built and confirmed `dist/util/cli-args.js` has no runtime import of it, so **`check` stays LLM-free and network-free** (its `dist` import graph is `config`, `kicad/cli`, `kicad/report`, `memory/drift`, `memory/constraints`, `kicad/sexp`, `openspec/cli` — no provider, no network). |
| 7 | `demo --tour` ignored `--json` | **Closed.** `cli.ts:293` emits `{ "tour": [...] }` with color forced off; two end-to-end tests spawn the real entry point through `tsx` (a devDependency, so `npx` resolves locally — no network). |
| 8 | `theme.ts` colored branches uncovered | **Closed.** Color-on tests in `cli-surface.test.ts`. |

---

## Findings

### 1. high · NEW · The Esc timer fires mid-paste, and the REPL session is then unrecoverable from the keyboard

[src/util/live-prompt.ts:259](src/util/live-prompt.ts#L259) arms the 40 ms Escape-disambiguation timer whenever the assembler is left holding a lone `\x1b`:

```ts
if (this.pending.buf === '\x1b') {
  this.escTimer = setTimeout(() => { ... this.deliver('\x1b'); }, this.escFlushMs);
}
```

It does not check `this.pending.paste`. But the paste branch at [live-prompt.ts:157-173](src/util/live-prompt.ts#L157-L173) deliberately holds back a partial `PASTE_END` marker, and `\x1b` is a legal partial tail. So if a read boundary lands exactly on the `\x1b` of the closing `\x1b[201~` and the remaining five bytes take longer than 40 ms (large paste over ssh/tmux, or any slow pty), the timer fires: it delivers a spurious Escape **and discards the ESC**, so `[201~` can no longer match the end marker.

Paste mode then never closes. Every subsequent byte goes through `body[i] < ' ' ? ' ' : body[i]`, so `\r` (Enter) and `\x03` (Ctrl+C) both become spaces. `KeyReader` only forwards `\x03` as a SIGINT while `paused`, and the session can never reach a paused state because no request can be submitted. Nothing on the keyboard gets out; the process must be killed from another terminal, and the pasted request is lost.

**Failure scenario, confirmed** — repro run against the real `KeyReader` with fake timers:

```ts
const reader = new KeyReader(input as unknown as NodeJS.ReadStream, 40);
input.write('\x1b[200~hello');
input.write('\x1b');                       // first byte of the end marker, alone
await vi.advanceTimersByTimeAsync(50);     // slow link: the timer fires
input.write('[201~');                      // rest of the end marker
input.write('\r');                         // Enter
input.write('\x03'); input.write('\x03');  // Ctrl+C twice
input.write('\r');                         // Enter
// expected: 'hello\r\x03\x03\r'
// actual:   'hello\x1b[201~    '   ← Esc leaked, marker leaked as text,
//                                     Enter and both Ctrl+C folded to spaces
```

Existing coverage misses this because every paste test calls `pushKeys` directly (`repl.test.ts:360-395`) — `pushKeys` alone is correct here — while the two `KeyReader` timer tests never involve a paste. The split in `holds back a paste-end marker split across reads` is `'\x1b[2'`, one byte past the failing boundary.

**Fix** (one line, verified: the repro above passes and the full suite stays at 464/17/0):

```ts
if (this.pending.buf === '\x1b' && !this.pending.paste) {
```

Worth landing the repro above as a regression test in `repl.test.ts`, next to the two existing timer tests.

### 2. medium · NEW · An unreadable or vanishing untracked file makes `snapshot()` throw, refusing the run with a raw git fatal

[src/util/git.ts:96-110](src/util/git.ts#L96-L110) feeds every path from `ls-files --others --exclude-standard` into `git update-index --add --stdin`, and nothing catches a failure. `snapshot()` is called at the very start of `runAgentLoop` ([loop.ts:195](src/agent/loop.ts#L195)), so the throw refuses the run.

The comment on `scratchIndexPath()` states the intent explicitly — *"a hostile or missing temp dir cannot make a snapshot fail and block a run from starting"* — but the same property does not hold for the file list itself.

**Failure scenario**, verified against real git in a scratch repo:

```
$ chmod 000 unreadable.txt          # or: file removed between ls-files and update-index
$ git update-index -z --add --stdin < paths
error: open("unreadable.txt"): Permission denied
fatal: Unable to process path unreadable.txt        # exit 128
```

So a `--allow-dirty` run in a repo that happens to contain one root-owned or mode-000 untracked artifact, or where a watcher deletes a temp file in the same instant, dies before the first turn with `fatal: Unable to process path …` and no explanation of what copperhead was doing. Reachable, non-default path.

**Fix**: wrap the `update-index` call and decide deliberately, since returning `null` silently reinstates the data loss finding 1 just closed. Either is defensible:

- refuse with a `PreflightError` naming the offending path and suggesting `--allow-dirty` be dropped or the file be made readable — actionable, and keeps the lossless-rollback promise honest; or
- catch, `console.warn` that untracked preservation is unavailable for this run, and return `null`.

A test that `chmod 000`s an untracked file and asserts the chosen behaviour would pin it.

### 3. low · The PR description no longer matches the branch

The body still reads as it did before the two fix commits:

- Testing table says `pass (432 passed, 17 skipped)`; measured at `a8baacb` it is **464 passed / 17 skipped / 0 failed**.
- Invariant checklist: **"Spec coherence: N/A, no spec-level behavior change"** is now wrong — `openspec/specs/SPEC.md` §3, §7 and AC-3.8 all changed in `52ce137`. CodeRabbit's pre-merge check flagged the same gap from the other direction.
- **"Secrets: … redacts `sk-` keys"** understates the current code, which uses the full shared `redactSecrets()`.
- Nothing in the description mentions `--allow-dirty` on `repl`, the untracked-tree snapshot, bracketed paste, the Esc timer, or `--tour --json` — the user-visible surface the last two commits added.

**Fix**: refresh the body (test counts, the Spec-coherence and Secrets checklist rows, and a Design bullet for the safety changes). The author's 15:45Z comment already contains most of the text.

### 4. low · SPEC.md moved but the `build-copperhead-phase-1` delta specs did not

`CLAUDE.md` requires SPEC.md and the active change artifacts to move together, with the delta scenarios mapping 1:1 onto SPEC.md's acceptance criteria. `52ce137` updated SPEC.md AC-3.8 and §7 to cover `repl` and the two-part snapshot, but:

- [openspec/changes/build-copperhead-phase-1/specs/safety-rails/spec.md:13](openspec/changes/build-copperhead-phase-1/specs/safety-rails/spec.md#L13) still reads *"`do` SHALL refuse to start on a dirty git tree unless `--allow-dirty` is passed (which snapshots via `git stash create`)"* — no `repl`, no untracked tree.
- Its scenario at lines 16-17 is still `do`-only.
- [design.md:54](openspec/changes/build-copperhead-phase-1/design.md#L54) likewise describes the snapshot as `git stash create` alone.

**Fix**: extend the safety-rails delta spec's requirement and scenario to name `repl`, and add a scenario for untracked restoration under `--allow-dirty`; update `design.md`'s snapshot sentence.

*Not measured*: `openspec validate build-copperhead-phase-1` — the `openspec` CLI is not installed in this environment (`npx openspec` → "could not determine executable to run"), and the change's own artifact files are untouched by this PR, so the divergence above was found by reading rather than by the validator.

### 5. low · Untested surface in the delta

Traced from the uncovered-lines list rather than assumed:

- [src/util/git.ts:236-239](src/util/git.ts#L236-L239) — the `restoreUntracked` failure warning. The only path where a rollback partially fails; never exercised.
- [src/util/live-prompt.ts:252-254](src/util/live-prompt.ts#L252-L254) — the `paused` branch that turns Ctrl+C into a real `SIGINT` during a turn. Untested (understandably: it calls `process.kill` on self), and it is the escape hatch finding 1 shows is unreachable when paste mode sticks.
- [src/util/live-prompt.ts:271-279](src/util/live-prompt.ts#L271-L279) — `onEnd` flushing a dangling ESC when the stream ends mid-sequence.
- [src/util/live-prompt.ts:182-186](src/util/live-prompt.ts#L182-L186) — the SS3 (`ESC O A`) arrow form; only the CSI form is tested.
- [src/kicad/cli.ts:92-99](src/kicad/cli.ts#L92-L99) — the `FALLBACK_BINARIES` bundle-probe loop inside `fallbackAfterMissing()`. `a8baacb` covers the *miss* (no bundle found → `KicadCliMissingError`); the *hit* is still unreached, so nothing pins that a bundle match caches and returns the bundle path. Testable on Linux by injecting the candidate list or pointing it at a temp dir.
- **No test drives `KeyReader` with a paste at all.** That gap is not incidental — it is precisely why finding 1 escaped both the fix commit and CI.

`src/cli.ts` remains 0% and I accept the author's reasoning: it parses argv at import, so v8 in-process coverage cannot see it, and the `--tour --json` wiring *is* covered by spawning the real binary (which coverage attributes to no file).

---

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| **Spec-gated in** (`edit_file`/`write_file` structurally absent until a proposal validates) | **Pass.** `src/agent/tools.ts` and `src/agent/filetools.ts` are byte-identical to the merge base (`git diff 6d7631b..a8baacb` empty for both); the `proposalValidated` gate at `tools.ts:210` is untouched. |
| **Verification-gated out** (ERC/DRC, repair to `maxRepairCycles`, then rollback) | **Pass.** `loop.ts` changes to only two lines, both cosmetic (`styleHeaderLines` wrapping the header). The REPL routes every request through `runAgentLoop` with no side path — pinned by `repl-dirty-gate.test.ts`. The rollback path is *strengthened* here (untracked restoration). |
| **`check`/`verify` LLM-free and network-free** | **Pass.** Verified against built output, not source: `dist/commands/check.js` imports only `config`, `kicad/cli`, `kicad/report`, `memory/drift`, `memory/constraints`, `kicad/sexp`, `openspec/cli`. The new `kicad-cli` resolution is `process.env` + `existsSync` only. `dist/util/cli-args.js` has no runtime import of `loop.js` — the type import is erased. |
| **No sexp serialization** | **Pass.** `src/kicad/sexp.ts` untouched by the PR. |
| **Sync-obligations ledger** | **Pass.** `src/agent/ledger.ts` untouched; no new tool-call hooks. |
| **Secrets** | **Pass**, and improved: the session log now uses the shared `redactSecrets()` (finding 3 closed), `.gitignore` still carries `.env` (line 1) and `.copperhead/runs/` (line 3), and `safety.test.ts` pins that `.env` is neither captured into the snapshot tree nor disturbed by the rollback. |

**Method note**: the head was reviewed in a throwaway `git worktree` at `a8baacb` (the local checkout has unrelated uncommitted work on another branch, so `gh pr checkout` was deliberately avoided). Every measurement above comes from that worktree; nothing was run against the user's working tree.

PR #120, opened 2026-07-27
Feature/agent install prompt

### LINKED ISSUE #118: Agent install prompt: one pasteable line to install copperhead for a repo
## The problem

Installing copperhead correctly today means reading the README and making several judgment calls by hand: Node >= 20, KiCad >= 8 with `kicad-cli` on PATH, and then picking one of five model backends (saved-login Codex, Cursor Agent CLI, Claude Code, `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`), each with its own setup path. `install.sh` covers the deterministic half (prereqs, npm install, `copperhead doctor`) but deliberately does nothing about backend choice, `copperhead init`, or wiring copperhead into a repo the user already has.

Meanwhile most of our users are already sitting in a coding agent. They should not be the ones doing this reconciliation. Greplica solves the same problem with a single pasteable line that points the agent at a prompt file: [`docs/agent-install-prompt.md`](https://raw.githubusercontent.com/Autoloops/greplica/refs/heads/main/docs/agent-install-prompt.md).

## Proposed behavior

Ship an agent install prompt so onboarding is one line the user pastes into any coding agent:

```
Install copperhead for this repo using https://raw.githubusercontent.com/chouhanindustries/copperhead/main/agent-install-prompt.md
```

The agent fetches the file and drives the rest: asks a short questionnaire, installs, verifies, and reports back in a few sentences.

### What the prompt file should contain

**1. Preamble.** A summary of what is about to happen, and an explicit instruction to ask the questionnaire *before* running any install command.

**2. Questionnaire** (native multiple-choice UI if the runtime supports it, plain multiple choice otherwise):

- *Q1: Which model backend?* Options: existing Codex CLI login, existing Cursor Agent CLI login, existing Claude Code login, `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, or "decide later" (copperhead's model picker handles it). The agent detects which logins are already present and marks the detected one as recommended, instead of asking cold.
- *Q2: Is this an existing KiCad repo or a greenfield project?* Existing means `copperhead init` scaffolds `docs/` from the schematic. Greenfield means point the user at `copperhead create --brief` and the [examples/](examples/) briefs.
- *Q3: Install globally via npm, or build from this checkout?* Recommended answer is inferred: build from source when run inside a copperhead checkout, `npm install -g copperhead` otherwise.

**3. Install and verify.** Prefer [`install.sh`](install.sh) when available (it already handles Node, git, `kicad-cli`, and never runs sudo), fall back to `npm install -g copperhead`. Then `copperhead doctor` for the environment and `copperhead check` for the repo. `check` is LLM-free and network-free, so it is a safe first run.

**4. Repo hygiene.** Confirm `.gitignore` covers `.env` and `.copperhead/runs/` (AC-4.3), and add them if it does not.

**5. Safety rules the prompt must state explicitly:**

- Never ask the user to paste an API key into the chat. Keys go in env vars or a `.env` the agent does not read back.
- Do not run `copperhead do` or `copperhead create` as part of installation. Those are mutating and LLM-backed; installation ends at `doctor` and `check`.
- Do not answer any copperhead confirmation prompt on the user's behalf.
- Do not echo full installer output.

**6. Final answer rules.** A few sentences: copperhead is installed, this is the backend it will use, this is what `doctor` and `check` reported, here is the one command to try next. No command dumps, no template filling.

### Where the file lives

Recommendation: `agent-install-prompt.md` at the repo root, so the raw URL stays short enough to paste. `docs/` is an Astro project, so a plain markdown file dropped at `docs/agent-install-prompt.md` would not be served by the docs site anyway. Mirror the content as a docs page under `getting-started/`, with a CI check that fails if the two drift.

Also needed:

- A short "Agent quick start" section at the top of the README `## Install` block, above the manual instructions, following Greplica's framing: most users should not install copperhead by hand.
- The same one-liner on the docs site quickstart.

## Alternatives considered

- **Extend `install.sh` to ask the backend questions interactively.** Rejected: the script is deliberately conservative and non-interactive-friendly, and it cannot detect a Claude Code or Cursor login and reason about which one the user actually wants. The agent can.
- **Leave it to the README.** That is the status quo, and it is exactly the step where a user has to hold five backend options in their head at once.
- **A `copperhead install` subcommand.** Duplicates `install.sh` and `doctor` without solving the judgment calls, and cannot run before copperhead is installed.

## Effect on the invariants

No effect. This is documentation plus a README section, and the prompt file must not instruct an agent to work around either invariant: it stops at `doctor` and `check`, and it never runs a mutating command. The safety rules above are what keep that true, so they belong in the acceptance criteria for this issue, not just in the prose.

## Acceptance

- [ ] `agent-install-prompt.md` exists and its raw URL is pasteable
- [ ] Pasting the one-liner into Claude Code, Codex, and Cursor each produces a working install on a clean machine
- [ ] The prompt asks the questionnaire before running anything
- [ ] The prompt never requests a key in chat and never runs `do` or `create`
- [ ] README and docs quickstart both lead with the one-liner
- [ ] Drift check between the root file and the docs page

Chirag6722:
## What

Adds a pasteable agent onboarding prompt [agent-install-prompt.md](agent-install-prompt.md) at the repository root, allowing AI coding assistants to automatically install, configure, and verify `copperhead` for a user's repository.

## Why

Onboarding onto `copperhead` previously required reading the README and manually handling environment checks, Node.js versions, path configurations, and model selection. Users can now copy-paste a single raw URL into their AI assistant (e.g. Claude Code, Cursor, Codex) to let the assistant drive the setup process (closes #118).

## Design

- Creates [agent-install-prompt.md](agent-install-prompt.md) at the repository root and mirrors its content to [agent-install.md](docs/src/content/docs/getting-started/agent-install.md) for the Astro documentation site.
- Registers the new page under the "Getting started" sidebar section in [astro.config.mjs](docs/astro.config.mjs).
- Integrates the onboarding tip to the top of [README.md](README.md) and [quickstart.mdx](docs/src/content/docs/getting-started/quickstart.mdx).
- Adds a CI unit test in [docs-drift.test.ts](test/docs-drift.test.ts) that validates that the root prompt and docs page stay synchronized byte-for-byte (stripping Astro frontmatter and normalizing CRLF/LF line endings).
- Preserves safety invariants: the prompt explicitly instructs agents not to request API keys in chat, never run mutating `do`/`create` loops during install, never auto-confirm inputs, and output concise summaries instead of dumping raw CLI logs.

## Testing

We ran the complete suite of tests and verified all checks.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | pass |

- **New tests:** Added [docs-drift.test.ts](test/docs-drift.test.ts) to verify the prompt file and mirrored docs page stay identical.
- **Fixed test failures:** Fixed a path-quoting bug on Windows in [create-stage-turns.test.ts](test/create-stage-turns.test.ts) and a shell script subprocess stub issue in [cursor-provider.test.ts](test/cursor-provider.test.ts).

- Sandbox variant used: `edit`
- Commands run and outcome:

```text
$ npm run build
> copperhead@0.7.0 build
> tsc -p tsconfig.json

$ npm link
added 1 package, and audited 3 packages in 606ms

$ copperhead doctor
  [ok]   node      v24.16.0 (>= 20)
  [FAIL] kicad-cli not found on PATH
         hint: install KiCad >= 9 (bundles kicad-cli); ERC/DRC gates need it.
  [ok]   git       2.54.0.windows.1
  [FAIL] provider  no model configured
         hint: pass --model, set COPPERHEAD_MODEL, or export an API key; see
               https://docs.copperhead.sh/reference/configuration/
  [info] project   no .copperhead/config.json (run `copperhead init` to scaffold)
not ready: fix the [FAIL] items above
```

## Docs

Updated [README.md](README.md) and [quickstart.mdx](docs/src/content/docs/getting-started/quickstart.mdx) to document the onboarding pasteable prompt.

---

## Invariant checklist

- [ ] **Spec-gated in**: n/a
- [ ] **Verification-gated out**: n/a
- [x] **`check`/`verify` stays LLM-free and network-free**: Checked. `test/docs-drift.test.ts` has no network/LLM dependencies.
- [ ] **No sexp serialization**: n/a
- [ ] **Sync-obligations ledger**: n/a
- [x] **Secrets**: Verified that `.gitignore` matches and prompt instructions prevent agents from reading/writing secrets in transcripts.
- [x] **Spec coherence**: Verified.

## Summary by CodeRabbit

- **New Features**
  - Added an AI-assisted “Agent install” onboarding workflow with questionnaire-gated setup, deterministic install/verify steps, and clear remediation/next-command guidance.
  - Linked the guide from the docs navigation and added an installation tip to the Quickstart/README experience.
- **Bug Fixes**
  - Improved cross-platform handling and messaging when required CLI tools are missing, using consistent “not found” detection.
- **Documentation**
  - Added a dedicated “Agent install” getting-started page and ensured prompt content matches the main install guide.
- **Tests**
  - Strengthened secret scanning, repo validation, and cross-platform robustness (line endings, quoting, PATH shims, Cursor stubs).


Chirag6722:
All CodeRabbit review suggestions have been fully addressed:
1. Added explicit user permission check before running `./install.sh` in the prompt.
2. Added Git tracking checks for `.env` credentials before appending to `.gitignore`.
3. Made the final onboarding report status-aware (exits/guides safely if dependencies fail).
4. Fixed Windows ESM test stubs (`stub-agent.mjs`), safety scanner bounds, and test path-quoting regexes.

The PR is fully updated, and all local vitest/lint/typecheck suites are passing.

animesh-chouhan:
_Automated pr-review pass (Claude Code). Not a human review, and not a formal approval: no `--approve` / `--request-changes` was submitted._

## Verdict: approve with nits, fix finding 1 before merge

No repo invariant is violated. The suite is green on this branch and gains 3 tests. The two things that need attention are the `install.sh` instruction in the published prompt (finding 1) and the fact that the Windows-specific `src/` changes in this PR are verified on no Windows machine anywhere in CI (finding 2). Everything else is a nit.

This is the first automated pr-review pass on this PR; there is no earlier pass to diff against. No commit under review was authored by the reviewer.

## Metrics

```
lines: +365 / -26 (net +339) across 16 files  ·  src +24/-8, test +117/-18, docs +224/-0, openspec +0
new vs net: 3 new files, 13 modified; 24 new src lines (15 of them one new function)
tests: +117 test lines vs +24 src lines (4.9:1); suite 361/17/0 -> 364/17/0 (pass/skip/fail, base -> head)
diff coverage: 86% of changed src lines exercised (19/22 mappable; measured)
                uncovered: src/kicad/cli.ts:28, src/kicad/cli.ts:137, src/kicad/cli.ts:153
deps/audit: package.json unchanged, not measured
```

Method, so each number is checkable:

- `BASE=6d7631b` from `git merge-base origin/main pr-120`. Sizes from `git diff --numstat $BASE...pr-120`, split by path prefix by hand. `agent-install-prompt.md` (+100) is counted as docs.
- New files from `git diff --diff-filter=A --name-only`: `agent-install-prompt.md`, `docs/src/content/docs/getting-started/agent-install.md`, `test/docs-drift.test.ts`. No new `src/` files.
- Suite totals measured locally with `npx vitest run` in two clean worktrees (`$BASE` and `pr-120`), Linux, `kicad-cli` present. Both 0 failures. `npm run typecheck` passes on head. `npx markdownlint-cli2` clean on the three new/changed markdown files.
- Diff coverage measured: `npm i -D @vitest/coverage-v8@3.2.7` in a throwaway worktree, `npx vitest run --coverage --coverage.provider=v8 --coverage.reporter=json`, then intersected the v8 `statementMap` per-line hit counts with the added `src/` lines from `git diff -U0 $BASE...pr-120 -- 'src/*'`. 24 added src lines, 2 of which map to no statement (`src/util/preflight.ts:35` and `:40`, a closing brace and a blank), giving 19/22 = 86%.
- **CI did not run the test job on this PR.** `gh pr checks 120` reports only `cla`, `label`, `template`, and CodeRabbit. `.github/workflows/ci.yml` triggers on `pull_request` but this is a cross-repository PR from `Chirag6722/copperhead`, so the workflow is awaiting maintainer approval. The PR body's "all checks pass" table is the author's local result, not a CI result. My local run above is the substitute, on Linux only.
- New public API surface: one new exported symbol, `isNotFoundError` in `src/util/preflight.ts`.

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| Spec-gated in (`edit_file`/`write_file` structurally absent) | pass, `git diff $BASE...pr-120 -- src/agent/ src/commands/` is empty, tool gating untouched |
| Verification-gated out (ERC/DRC, repair, rollback) | pass, no change to the loop or safety rails |
| `check`/`verify` LLM-free and network-free | pass, `src/commands/check.ts` imports are unchanged; the two modules it reaches that did change (`src/kicad/cli.ts`, `src/openspec/cli.ts`) now import `src/util/preflight.ts`, whose only import is `node:fs/promises` |
| No sexp serialization | pass, `src/kicad/sexp.ts` untouched |
| Sync-obligations ledger | pass, untouched |
| Secrets (`sk-` redaction, `.gitignore`) | pass, `.gitignore` still has `.env` and `.copperhead/runs/`; redaction path untouched. See finding 7 on the secret-scan test itself |
| Spec coherence | pass, `openspec validate build-copperhead-phase-1` returns valid; `openspec/` has no changes and SPEC.md describes no onboarding behavior, so nothing diverged |

Not verified: the Astro docs site build. `docs/node_modules` is not installed here and `docs.yml` did not run on this fork PR, so the new page is unbuilt.

## Findings

### 1. medium — the prompt tells the agent to execute an arbitrary `install.sh` from the user's own repository

`agent-install-prompt.md:44-49` (mirrored at `docs/src/content/docs/getting-started/agent-install.md:51-56`)

The branch is explicitly labelled "outside checkout", meaning the agent is in the user's repository, which is not copperhead. In that situation an `install.sh` at the repo root belongs to the user's project, not to copperhead. `install.sh` is one of the most common filenames in the ecosystem.

Failure scenario: a user pastes the onboarding line into Claude Code inside `~/work/my-firmware`, which ships its own `install.sh` that installs system packages and flashes a dev board. The agent finds it, asks "may I run the installer?" (framed as copperhead's installer, because that is what the prompt is about), the user says yes, and the agent runs an unrelated script with `--yes`. The permission ask is real but the user is answering a question about the wrong script. This prompt is published at a stable raw URL for arbitrary users to paste into agents holding broad tool permissions, so the blast radius is not hypothetical.

Fix: outside a checkout the only correct install path is npm. Delete the branch:

````diff
-   * **On macOS/Linux (outside checkout):** If `install.sh` is present in the repository, **first ask the user for explicit permission to run it**. If approved, run:
-     ```bash
-     ./install.sh --yes
-     ```
-     If they decline, or if `install.sh` is missing/fails, fall back to:
-     ```bash
-     npm install -g copperhead
-     ```
+   * **On macOS/Linux (outside checkout):** Run:
+     ```bash
+     npm install -g copperhead
+     ```
````

If the `install.sh` path is worth keeping, it must first prove the script is copperhead's, for example by requiring `head -1 install.sh` to match `#!/usr/bin/env bash` **and** the file to contain the literal `copperhead bootstrap installer` banner from `install.sh:2`, and by confirming the checkout is copperhead. That is exactly the "inside a checkout" case, which the prompt already handles above with `npm ci && npm run build && npm link`.

### 2. medium — the Windows-only `src/` changes are exercised on no Windows machine, and only through hand-built object literals

`src/util/preflight.ts:26-39`, call sites `src/kicad/cli.ts:28,47,98,137,153` and `src/openspec/cli.ts:22`

The entire purpose of `isNotFoundError`'s `win32` branch is to classify what execa returns on Windows when a `.cmd` shim resolves but its target does not. The only test, `test/preflight.test.ts:220-262`, overrides `process.platform` with `Object.defineProperty` and feeds in `{ exitCode: 9009, stderr: "..." }` object literals that the author wrote to match the implementation. No test ever passes a real execa error or a real `reject: false` result through it, and `ci.yml` is `runs-on: ubuntu-latest` only, so nothing anywhere runs this code on Windows.

Two shapes are asserted nowhere against reality: whether execa's rejected error actually carries `exitCode` on a `.cmd` shim failure (used at `src/kicad/cli.ts:28,137,153` and `src/openspec/cli.ts:22`), and whether execa's `reject: false` **result** object carries `exitCode` and `stderr` in the same shape (used at `src/kicad/cli.ts:47,98`, where the argument is `res`, not an error). If either differs, the new branch is silently dead and Windows users keep getting the raw execa failure the PR set out to fix.

The unit test itself is fine, deterministic, restores `process.platform` in a `finally`, no wall-clock or network. The gap is that it is the only evidence.

Fix: add a Windows job to `ci.yml` (`runs-on: windows-latest`, skipping the KiCad-dependent suites), or at minimum land one integration-shaped test that captures a real execa failure object and feeds it through, so the shape assumption is pinned:

```ts
it('classifies a real execa spawn failure as not-found', async () => {
  const err = await execa('copperhead-does-not-exist', []).catch((e) => e);
  expect(isNotFoundError(err)).toBe(true);
});
```

That covers the ENOENT half on Linux and pins the error shape. The win32 half still needs a Windows runner.

### 3. low — `exitCode === 1` is too weak a signal for "binary not found", and at one call site it escalates a recoverable failure into an aborted export

`src/util/preflight.ts:31`

```ts
if (err.exitCode === 1 || err.exitCode === 9009) {
  return msg.includes('is not recognized') || msg.includes('cannot find the path specified');
}
```

`9009` is cmd.exe's specific "command not found" code and is a sound signal. `1` is the most common failure code in existence, and the two message fragments are generic Windows phrasings, not markers that name the missing binary. Any win32 subprocess that exits 1 with "cannot find the path specified" anywhere in its stderr is now reported as a missing CLI.

Failure scenario, `src/kicad/cli.ts:133-140`: `exportFab` deliberately accumulates per-artifact failures in `result.failed` so that one bad gerber job does not lose the rest of the export. On Windows, a job that exits 1 with a path-related message now takes the `throw new KicadCliMissingError()` branch instead, aborting the whole export and telling the user to install KiCad, which is already installed and running.

Fix, one line, keeping the sound signal and dropping the ambiguous one:

```diff
-    if (err.exitCode === 1 || err.exitCode === 9009) {
+    if (err.exitCode === 9009) {
```

If exit 1 must stay (npm shims do sometimes exit 1), require the message to name the binary rather than matching a bare phrase.

### 4. low — three changed error paths in `src/kicad/cli.ts` are reached by no test

`src/kicad/cli.ts:28`, `:137`, `:153` (measured uncovered, see the diff-coverage method above)

These are the `catch` bodies of `kicadCliVersion`, `exportFab`, and `exportSvg`. Every one is a changed line and none executes in the suite, because `kicad-cli` is present in every test environment. Finding 3's failure scenario lives at `:137`, which is exactly why this matters.

Fix: `kicadCliVersion` is trivially testable without a Windows runner by emptying `PATH`:

```ts
it('reports a missing kicad-cli as KicadCliMissingError', async () => {
  const prev = process.env.PATH;
  process.env.PATH = '';
  try {
    await expect(kicadCliVersion()).rejects.toBeInstanceOf(KicadCliMissingError);
  } finally {
    process.env.PATH = prev;
  }
});
```

### 5. low — the AC-4.2 happy-path assertion now re-implements the function under test

`test/safety.test.ts:21-22`

`resolveInRepo` is `path.resolve(repoRoot, p)` plus a containment guard (`src/util/paths.ts:14-21`). Asserting `toBe(path.resolve('/repo', 'docs/BOM.md'))` compares the implementation to itself: the assertion can only fail if the function throws. The cross-platform motivation is legitimate, the resulting assertion is not.

Fix: assert the relationship rather than recomputing the value.

```diff
-    expect(resolveInRepo('/repo', 'docs/BOM.md')).toBe(path.resolve('/repo', 'docs/BOM.md'));
-    expect(resolveInRepo('/repo', '.')).toBe(path.resolve('/repo'));
+    const root = path.resolve('/repo');
+    expect(path.relative(root, resolveInRepo('/repo', 'docs/BOM.md'))).toBe(path.join('docs', 'BOM.md'));
+    expect(resolveInRepo('/repo', '.')).toBe(root);
```

### 6. low — the AC-4.1 secret-scan test passes vacuously

`test/agent-integration.test.ts:186-196`

The test is titled "no API key material anywhere in the tree **after runs**", but it calls `tempFixtureRepo()` and scans the result immediately. No agent runs, no transcript is written, no `.copperhead/` directory exists. It asserts that a freshly copied fixture contains no `sk-` string, which it never could. It is also inside `describe.skipIf(!providers.some((p) => p.key))`, so it does not run in CI at all.

This is pre-existing (the base did the same with `grep -rE`), but this PR rewrote the scan and the PR summary claims it "strengthened safety scanning for secrets", so it is the right place to fix it. The new `scanTreeForSecret` walker itself is an improvement over shelling out to `grep`: it is portable, skips symlinks, and skips `.git`/`node_modules`.

Fix: scan a repo that has actually been run against. The keyed scenarios above in the same file already return `res.transcriptDir`; scan that repo instead of a fresh fixture, so the redaction path is what is under test.

Also note the pattern widened from `sk-[A-Za-z0-9_-]{20,}` to `sk-[A-Za-z0-9_-]+`. That matches SPEC.md's redaction pattern, which is the right call, but it will match short incidental strings like `sk-1` once the test scans a real tree with docs in it. Worth knowing before it fires.

### 7. low — dead import

`test/agent-integration.test.ts:8`

`import { toolSearch } from '../src/agent/filetools.js';` is added and never used. `tsc --noEmit` passes because `noUnusedLocals` is off, so nothing catches it. Leftover from an earlier approach to the scan. Delete the line.

### 8. low — the PR description does not mention the `src/` changes

The Design section describes an onboarding prompt, a docs mirror, a sidebar entry, and a drift test. It does not mention the new exported `isNotFoundError`, nor the six call-site rewrites in `src/kicad/cli.ts` and `src/openspec/cli.ts`. Those arrived through the merge of `fix/windows-compatibility` (`4618e02`, `5ed98bf`, `0cef363`, plus two merge commits). The Testing section's "Fixed test failures" line mentions two test files and no source files.

Failure scenario: a reviewer reads "adds a pasteable onboarding prompt", scans the docs, and merges without scrutinising a runtime change to the error classification on every `kicad-cli` invocation. That is findings 2 and 3 slipping through.

Fix: either split the Windows-compatibility commits into their own PR (cleanest, since they have a different risk profile and reviewer), or add them to the Design section and mark the invariant checklist accordingly.

### 9. low — duplicate `sidebar.order`

`docs/src/content/docs/getting-started/agent-install.md:5` sets `order: 3`, which collides with `demo.md:5`. Harmless today because `docs/astro.config.mjs:75-80` declares this group's items explicitly, so frontmatter order is ignored. It becomes an arbitrary tiebreak the moment the group is switched to autogenerate. Bump `demo.md` to `order: 4`.

## Nits, no action needed

- The 100 lines of prompt text duplicated between the root file and the docs page are the right tradeoff given the raw-URL requirement, and `test/docs-drift.test.ts` is a good mitigation: deterministic, asserts, no network, handles CRLF and frontmatter. The frontmatter regex is `^`-anchored without the `m` flag, so the `---` horizontal rules inside the prompt body cannot be mis-stripped. Verified by hand.
- `isNotFoundError(err: any)` uses `any` in a new exported util in an otherwise strict codebase. `unknown` plus narrowing would match the surrounding style, and `src/openspec/cli.ts:21` already builds a precise inline type for the same object.
- `test/init-check.test.ts` (`mkdir` instead of shelling out, `path.delimiter`), `test/create-stage-turns.test.ts` (quoted-path regex), `test/budget-efficiency.test.ts` (CRLF normalise), and `test/cursor-provider.test.ts` (the `.mjs` stub behind a `.sh`/`.bat` shim) are all clean portability fixes. The cursor stub in particular is better than the old shell heredoc: it embeds paths with `JSON.stringify`, so a path containing a quote no longer breaks the stub.

Chirag6722:
I have addressed the security blocker (Finding 1) by removing the install.sh option outside of checkout entirely, defaulting to direct global npm installation.

I also fixed Finding 2 by adding the integration test in test/preflight.test.ts to capture and verify a real execa spawn failure error shape, and resolved the sidebar order nit (Finding 9).

animesh-chouhan:
_Automated pr-review pass (Claude Code), incremental. Not a human review, and not a formal approval: no `--approve` / `--request-changes` was submitted._

Follow-up to the [previous automated pass](https://github.com/chouhanindustries/copperhead/pull/120#issuecomment-5094087521) at `919cd7d`. Head is now `f62caa9`. Only what changed is reported below; the invariant checks and the unchanged findings are summarised rather than restated. No commit under review was authored by the reviewer.

## Verdict: the security blocker is fixed, but the PR is now blocked on a conflict that is not mechanical to resolve

Finding 1 (the `install.sh` execution instruction) is properly fixed, finding 9 is fixed, and finding 2 is partly addressed. Nothing regressed and the suite gained a test.

What changed underneath: `main` moved 56 commits ahead of the merge base and rewrote `src/kicad/cli.ts`, so `mergeable` is now `CONFLICTING`, and the conflict is in exactly the code this PR touches. Resolving it requires re-deciding where the Windows detection belongs, not just picking a side. That is the new blocker, and it is new finding A below.

## Two new commits

| Commit | What it does |
| --- | --- |
| `95ce640` | Removes the `install.sh` branch from both prompt copies; bumps `demo.md` to `order: 4` |
| `f62caa9` | Adds one integration test pinning a real execa spawn-failure shape |

```
diff since prior pass: +8 / -11 across 4 files
  agent-install-prompt.md +1/-5, docs/.../agent-install.md +1/-5,
  docs/.../demo.md +1/-1, test/preflight.test.ts +5/-0
```

## Metrics (full change, re-measured at `f62caa9`)

```
lines: +363 / -27 (net +336) across 17 files  ·  src +24/-8, test +122/-18, docs +217/-1, openspec +0
new vs net: 3 new files, 14 modified; 24 new src lines (unchanged since prior pass)
tests: +122 test lines vs +24 src lines (5.1:1); suite 361/17/0 -> 365/17/0 (pass/skip/fail, base -> head)
diff coverage: 86% of changed src lines exercised (19/22 mappable; measured, unchanged)
                uncovered: src/kicad/cli.ts:28, src/kicad/cli.ts:137, src/kicad/cli.ts:153
mergeable: CONFLICTING  ·  CI: still only cla / label / template; ci.yml has not run
deps/audit: package.json unchanged, not measured
```

Method: merge base is still `6d7631b` (`git merge-base origin/main pr-120`), so the base suite figure of 361/17/0 from the prior pass still applies. Head figures from `npx vitest run --coverage` in a clean worktree at `f62caa9`, Linux, `kicad-cli` present, 0 failures. `npm run typecheck` passes. Diff coverage re-measured the same way as before (v8 `statementMap` intersected with added `src/` lines): identical result, because the new test exercises `isNotFoundError` directly rather than through any `src/kicad/cli.ts` call site. `ci.yml` still has not run: this is a cross-repository PR, so the workflow awaits maintainer approval and `gh pr checks 120` reports only `cla`, `label`, `template`.

Invariant checks: re-run against `f62caa9`, all six still pass, plus `openspec validate build-copperhead-phase-1` valid. `src/agent/` and `src/commands/` remain untouched. Nothing in the two new commits goes near an invariant.

## Prior findings, current status

| # | Finding | Status |
| --- | --- | --- |
| 1 | `install.sh` executed from the user's own repo | **fixed** in `95ce640`, verified: zero `install.sh` references remain in either prompt copy, and the drift test still passes |
| 2 | Windows path verified on no Windows machine | **partly addressed**, see finding B |
| 3 | `exitCode === 1` too weak a not-found signal | open, `src/util/preflight.ts:31` unchanged |
| 4 | Three uncovered `kicad-cli` catch bodies | open, still measured uncovered at `:28`, `:137`, `:153` |
| 5 | AC-4.2 assertion re-implements the function under test | open, `test/safety.test.ts:21-22` unchanged |
| 6 | Unused `toolSearch` import | open, `test/agent-integration.test.ts:8` unchanged |
| 7 | AC-4.1 secret scan passes vacuously | open, `test/agent-integration.test.ts:187` unchanged |
| 8 | PR description omits the `src/` changes | open, the body still contains no mention of `isNotFoundError`, `preflight.ts`, or `kicad/cli` |
| 9 | Duplicate `sidebar.order` | **fixed** in `95ce640`, `demo.md` is now `order: 4` |

The 3 through 8 rationale, failure scenarios, and fixes are in the prior comment and are not repeated here.

## New findings

### A. medium — the branch now conflicts with `main` in `src/kicad/cli.ts`, and `main` has centralised the very mapping this PR spread across call sites

`src/kicad/cli.ts` (whole file), against `origin/main` at `3e86c72`

`gh pr view 120` reports `mergeable: CONFLICTING`. `git merge-tree` shows four conflicting files: `README.md`, `docs/src/content/docs/getting-started/demo.md`, `test/safety.test.ts`, and `src/kicad/cli.ts`. The first three are ordinary textual overlaps. The fourth is a design collision.

`main` now routes every `kicad-cli` invocation through a `runKicad` helper (`origin/main:src/kicad/cli.ts:109-124`) that resolves the binary through the `COPPERHEAD_KICAD_CLI` override, then PATH, then macOS `KiCad.app` fallback paths, and maps ENOENT to `KicadCliMissingError` in **one place**, at lines 112 and 121. Every call site was correspondingly simplified and given an `if (err instanceof KicadCliMissingError) throw err;` guard.

This PR was written against the pre-`runKicad` structure and adds `isNotFoundError` at six separate call sites of code that no longer exists in that form.

Failure scenario for a naive resolution: taking "ours" reverts `main`'s app-bundle fallback chain and its regression tests (`a8baacb`, `c6d6e80`, `b209523`), silently breaking macOS users who rely on the bundled `kicad-cli`. Taking "theirs" drops the Windows detection this PR exists to add. Either way the merge compiles and the suite is green, so nothing catches it.

Fix: rebase on `main` and put the win32 detection inside `runKicad`, where the ENOENT mapping already lives, instead of at the call sites. That is a smaller diff than the current one and it fixes finding 4 for free, since `runKicad`'s ENOENT path already has coverage from `a8baacb`:

```diff
-  if (res.failed && (res as unknown as ExecaError).code === 'ENOENT') {
+  if (res.failed && isNotFoundError(res)) {
```

at both `origin/main:src/kicad/cli.ts:112` and `:121`, and nowhere else in that file. `src/openspec/cli.ts` is not conflicted and can keep its current change as-is.

### B. low — the new integration test pins the one shape that was never in doubt

`test/preflight.test.ts:262-266`

```ts
it('classifies a real execa spawn failure as not-found', async () => {
  const err = await execa('copperhead-does-not-exist', []).catch((e) => e);
  expect(isNotFoundError(err)).toBe(true);
});
```

The test is well-formed: deterministic, asserts, no network, no wall-clock, no order dependency. It is also the test I suggested, so treat my assessment of it with the appropriate skepticism.

What it establishes is that a rejected execa promise carries `code: 'ENOENT'` on Linux. That branch (`src/util/preflight.ts:28`) is platform-independent, predates this PR, and was never the uncertain part. The two shapes finding 2 actually named are still unpinned:

1. The `reject: false` **result** object at `src/kicad/cli.ts:47` and `:98`, where the argument to `isNotFoundError` is `res`, not a thrown error. Whether a non-rejecting execa result exposes `exitCode` and `stderr` in the shape the win32 branch reads is still asserted nowhere.
2. The win32 branch itself, which is the entire reason the function exists. `ci.yml` is `runs-on: ubuntu-latest` only, so it runs on no machine anywhere.

Shape 1 is testable today on Linux and is the higher-value of the two, since a wrong assumption there makes the branch dead at two of the six call sites:

```ts
it('classifies a non-rejecting execa failure result as not-found', async () => {
  const res = await execa('copperhead-does-not-exist', [], { reject: false });
  expect(isNotFoundError(res)).toBe(true);
});
```

Shape 2 still needs a `windows-latest` job in `ci.yml`, skipping the KiCad-dependent suites. Given that the Windows fixes are the entire `src/` content of this PR, that job is worth adding here rather than deferring.

## Suggested order of operations

1. Rebase on `main` and resolve `src/kicad/cli.ts` by moving the detection into `runKicad` (finding A). This also closes finding 4.
2. Drop `exitCode === 1` (finding 3), one line.
3. Add the `reject: false` shape test (finding B), and a `windows-latest` CI job if you want the win32 branch to mean anything.
4. The four test-hygiene items (5, 6, 7) and the PR description (8) are quick and independent of the rebase.

Chirag6722:
I have resolved the merge conflicts by rebasing and centralizing the Windows missing-binary check (isNotFoundError) directly inside the new runKicad helper (Finding A).

I also refined the exitCode === 1 signal to prevent generic path errors from colliding (Finding 3), and added the second integration test verifying the non-rejecting error shape (Finding B). All conflicts are resolved

Chirag6722:
@animesh-chouhan 
also address latest CodeRabbit review: handled fallback command-not-found errors before the reject: false early return. fully pushed

animesh-chouhan:
_Automated pr-review pass (Claude Code), incremental. Not a human review, and not a formal approval: no `--approve` / `--request-changes` was submitted._

Third pass, following the one at [`919cd7d`](https://github.com/chouhanindustries/copperhead/pull/120#issuecomment-5094087521) and the one at [`f62caa9`](https://github.com/chouhanindustries/copperhead/pull/120#issuecomment-5094538080). Head is now `ebb2970`.

**Disclosure:** commits `67ae3eb` and `ebb2970` implement fixes this reviewer prescribed in the earlier passes (findings A, 3, and B). Reviewing them is reviewing my own prescription, so they get more scepticism, not less, and finding 1 below is exactly a place where the prescription was followed only halfway. All 13 commits are authored by @Chirag6722; none are mine.

## Verdict: approve with nits, fix finding 1 first

The conflict is resolved and `mergeable` is back to `MERGEABLE`. `main` was merged in cleanly at `a8fd466`, and the win32 detection moved into `runKicad` where the ENOENT mapping already lived, which is the right shape and deleted more code than it added (`src/kicad/cli.ts` is +6/-19). CI is green, 5/5. The suite is 474/17/0.

One new medium: the exit-code-1 detection is an unanchored substring test, and because `runKicad` now feeds it into `fallbackAfterMissing()`, a single false positive permanently poisons the resolved-binary cache for the rest of the process. That is a worse failure mode than what the PR started with, and it is one regex away from being fixed.

## Metrics

```
lines: +368 / -40 (net 328) across 17 files  ·  src +27/-21, test +124/-18, docs +217/-1, spec +0/-0, other +0/-0
new vs net: 3 new files (0 in src), 14 modified; 0 new src lines (the net-new surface)
tests: +124 test lines; suite not measured -> 474/17/0 (pass/skip/fail, base -> head)
diff coverage: 96% of changed src lines exercised (24/25 mappable; measured); uncovered: src/util/preflight.ts:34
mergeable: MERGEABLE (was CONFLICTING at the previous pass)
```

Method, per number:

- Everything except diff coverage comes from `.claude/skills/pr-review/scripts/metrics.mjs 120`, run with the head checked out in a scratch worktree so the suite could execute. Merge base `3e86c72`.
- **Head suite measured**: 474 passed / 17 skipped / 0 failed, 38 files, via the script's `npx vitest run`. Linux, `kicad-cli` present. `npm run typecheck` passes. `npx markdownlint-cli2` clean on the three markdown files.
- **Base suite: not measured.** The script skips it without `--base-tests`. The count jumped from 365 at the previous pass to 474 because `a8fd466` merged 56 commits of `main` (the REPL work); that delta is not this PR's.
- **Diff coverage measured separately.** The script reported `not measured (could not install @vitest/coverage-v8)`, so I installed `@vitest/coverage-v8@3.2.7` in the throwaway worktree and ran `npx vitest run --coverage --coverage.provider=v8 --coverage.reporter=json`, intersecting the v8 `statementMap` per-line hit counts with the added `src/` lines from `git diff -U0 3e86c72...HEAD -- 'src/*'`. 27 changed src lines, 2 unmappable (a closing brace and a blank), so 24/25 = 96%. The single uncovered line is finding 2 below.
- `deps/audit`: `package.json` unchanged, not measured.

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| Spec-gated in | pass, `git diff 3e86c72...HEAD -- src/agent/ src/commands/` is empty |
| Verification-gated out | pass, loop and safety rails untouched |
| `check`/`verify` LLM-free, network-free | pass, `src/commands/check.ts` unchanged; its only match for provider/network terms is the comment on line 13, "This module must never import a provider". `src/kicad/cli.ts` and `src/openspec/cli.ts` reach only `src/util/preflight.ts`, whose sole import is `node:fs/promises` |
| No sexp serialization | pass, `src/kicad/sexp.ts` untouched |
| Sync-obligations ledger | pass, untouched |
| Secrets | pass, `.gitignore` still carries `.env` and `.copperhead/runs/`; redaction path untouched |
| Spec coherence | pass, `openspec validate build-copperhead-phase-1` valid; no `openspec/` changes and SPEC.md describes no onboarding behaviour |

Not verified: the Astro docs build (`docs/node_modules` absent, `docs.yml` did not run on this fork PR), and anything Windows.

## Prior findings, current status

| # | Finding | Status |
| --- | --- | --- |
| A | Conflicts with `main`'s `runKicad` rewrite | **fixed**, `a8fd466` merges `main`, `67ae3eb` moves the detection into `runKicad`; `mergeable=MERGEABLE` |
| 3 | `exitCode === 1` too weak | **partly fixed**, the `cannot find the path specified` clause no longer applies at exit 1; the remaining substring test is finding 1 below |
| B | `reject: false` result shape unpinned | **fixed**, `test/preflight.test.ts:264-267` pins it with a real execa result |
| 4 | Uncovered `kicad-cli` catch bodies | **fixed by deletion**, `67ae3eb` removes those catch bodies entirely now that `runKicad` owns the mapping |
| 2 | Win32 path runs on no machine | open, `ci.yml` is still ubuntu-only |
| 5 | AC-4.2 assertion re-implements the function | open, `test/safety.test.ts:23` unchanged |
| 6 | Unused `toolSearch` import | open, `test/agent-integration.test.ts:8` unchanged |
| 7 | AC-4.1 secret scan passes vacuously | open, `test/agent-integration.test.ts:187` unchanged |
| 8 | PR description omits the `src/` changes | open |

## New findings

### 1. medium — an unanchored `is not recognized` substring at exit 1 misreads a working kicad-cli as missing, and permanently poisons the binary cache

`src/util/preflight.ts:37-39`, reached from `src/kicad/cli.ts:112` and `:120`

```ts
if (err.exitCode === 1) {
  return msg.includes('is not recognized');
}
```

`msg` is the whole of stderr (falling back to `message`). The test is a bare substring anywhere in that text, at the single most common failure exit code there is.

Traced by hand through the merged code:

1. `runKicad` (`src/kicad/cli.ts:109`) runs every invocation with `{ reject: false }` and calls `isNotFoundError(res)` on any failed result.
2. A `kicad-cli` that ran, exited 1, and printed a stderr containing that substring takes the true branch. `bin` is the bare `'kicad-cli'`, so control goes to `fallbackAfterMissing()` (`src/kicad/cli.ts:98`).
3. On Windows none of the four `/Applications/KiCad*/…` candidates exist, so `fallbackAfterMissing` sets **`cachedBinary = null`** and throws `KicadCliMissingError`.
4. `cachedBinary` is process-wide and sticky. `resolveKicadCli()` (`src/kicad/cli.ts:63-64`) opens with `if (cachedBinary === null) throw new KicadCliMissingError()`, so from that point on **every** kicad-cli call in the process fails the same way, regardless of what caused the original stderr.

Failure scenario: on Windows, `copperhead check` on a design that kicad-cli can parse but rejects, with any stderr line containing "is not recognized". The user gets "kicad-cli not found on PATH" plus advice to install KiCad ≥ 8, on a machine where KiCad is installed and just ran. The genuinely useful message at `src/kicad/cli.ts:173` ("kicad-cli produced no report, the schematic file likely fails to load") is never reached, and the run cannot recover in-process.

There is a second, independent instance: `src/openspec/cli.ts:22` passes an exit-1 `openspec` failure through the same helper. `openspec validate` exits 1 on a validation failure and prints the report on stderr, so a report containing that substring is reported to the user as "openspec CLI not found on PATH (npm i -g @fission-ai/openspec)". `check` calls `openspecValidate`, so this is on the deterministic path.

To be explicit about confidence: the code path and the cache poisoning are traced and certain; what I am positing is a stderr that happens to contain the phrase. That is exactly the assumption a bare `includes` forces you to bet on, and the fix costs one line. cmd.exe's real message has a fixed shape, which the existing test fixtures already match:

```diff
     if (err.exitCode === 1) {
-      return msg.includes('is not recognized');
+      return /^'[^']+' is not recognized as an internal or external command/m.test(msg);
     }
```

Repro, failing today and passing with the fix (drop it beside the existing win32 case):

```ts
it('does not read a tool content error as a missing binary', () => {
  Object.defineProperty(process, 'platform', { value: 'win32' });
  try {
    expect(isNotFoundError({ exitCode: 1, stderr: 'The file format is not recognized.' })).toBe(false);
  } finally {
    Object.defineProperty(process, 'platform', { value: originalPlatform });
  }
});
```

Worth considering separately: even for a true positive, poisoning `cachedBinary` on the strength of a message match (rather than a real ENOENT) is a large consequence for a heuristic. Restricting the `null` write in `fallbackAfterMissing` to ENOENT-sourced misses would bound the blast radius.

### 2. low — the `cannot find the path specified` clause is the one uncovered changed line, and the suite cannot reach it

`src/util/preflight.ts:34` (measured uncovered, reconciles with the metrics block)

```ts
if (err.exitCode === 9009) {
  return (
    msg.includes('is not recognized') ||
    msg.includes('cannot find the path specified')   // ← line 34, 0 hits
  );
}
```

The only 9009 fixture (`test/preflight.test.ts:232-235`) has stderr `"'kicad-cli' is not recognized as an internal or external command"`, which satisfies the left operand, so `||` short-circuits and line 34 never evaluates. The clause is untested surface: it could be deleted or inverted and the suite would stay green.

Fix, one assertion:

```ts
expect(isNotFoundError({
  exitCode: 9009,
  stderr: 'The system cannot find the path specified.',
})).toBe(true);
```

### 3. low — three stray blank lines where an assertion was deleted

`test/preflight.test.ts:236-238`

`67ae3eb` removed the `exitCode: 1` / `cannot find the path specified` assertion but left its surrounding blank lines, so there is now a three-line gap mid-block. Cosmetic, but it reads as a deleted-and-forgotten case. Collapse to one blank line.

### 4. low — `err: any` on the exported helper

`src/util/preflight.ts:26`

Unchanged from the first pass. `unknown` plus a narrowing cast matches the surrounding strictness, and `src/openspec/cli.ts:21` already builds a precise inline type for the same object. Cheap to fix while finding 1 is being applied.

## Still open from earlier passes, unchanged

Findings 2, 5, 6, 7, and 8 are carried forward verbatim; rationale, failure scenarios, and fixes are in the two earlier comments. In short: `ci.yml` is still ubuntu-only so the win32 branch runs nowhere (the two new execa shape tests do now pin the real object shapes on Linux, which is a genuine improvement); `test/safety.test.ts:23` still compares `path.resolve` to itself; `test/agent-integration.test.ts:8` still imports `toolSearch` unused; the AC-4.1 scan at `:187` still scans a fresh fixture that has had no run; and the PR description still describes only the docs change, with no mention of `isNotFoundError`, `preflight.ts`, or `kicad/cli.ts`.

PR #121, opened 2026-07-27
fix(deps): bump openai to ^6, dropping the npm install warnings

animesh-chouhan:
## What

`npm i -g copperhead` no longer prints any warnings. A clean global install is silent and pulls 123 packages instead of 145.

## Why

On a clean machine, installing 0.8.0 printed:

```text
npm warn ERESOLVE overriding peer dependency
npm warn deprecated node-domexception@1.0.0: Use your platform's native DOMException instead
```

Both trace to `openai@4`:

- It depends on `formdata-node`, a multipart polyfill Node 20's native `FormData` makes redundant, which drags in the deprecated `node-domexception`.
- It peers `zod@^3`, which the `overrides: { zod: "^4.0.0" }` added in 00e6f98 had to override so `@anthropic-ai/claude-agent-sdk`'s required `zod@^4` peer could resolve.

First impressions of an install matter, and a deprecation warning on a fresh install reads like the tool is unmaintained.

## Design

`openai@6` has no runtime dependencies at all, and peers `zod` as an optional `^3.25 || ^4`. That removes both warnings and makes the override non-load-bearing, so it goes too: the tree still resolves to a single `zod@4.4.3` without it (verified with `npm ls zod --all`).

The 4 to 6 jump needs no source change. [openai.ts](src/agent/providers/openai.ts) uses only `chat.completions.create`, and v5's breaking changes were the removal of the fetch/FormData shims (Node 20+ is already our floor per `engines`) and of APIs this provider never called.

No invariant is touched: this is a reasoning backend, and the agent loop, spec gate, and verification gate are unchanged.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | pass (476 pass / 17 skip / 0 fail, up from 469 / 17) |
| Markdown lint | `npm run lint:md` | pass (0 issues in 360 files) |
| Live-LLM (opt-in) | n/a | not run (no key configured) |

New tests: [openai-provider.test.ts](test/openai-provider.test.ts), 7 tests. Nothing covered this provider before, so a two-major-version SDK bump had no guard at all. Rather than pin the SDK's types, it pins the wire contract against a stub HTTP server: request path, bearer key, model, `max_completion_tokens` (not `max_tokens`, which reasoning models reject), role mapping, tool serialization, the absent-tools case spec gating relies on, the empty-choices response, and the vendor `thought_signature` round-trip that keeps Gemini turns from 400ing. A future SDK bump that moves any of those fails here instead of on a user's first paid run.

No pre-existing failures.

- Sandbox variant used: fixture project (`test/fixtures/open-key`) plus a packed-tarball global install, with kicad-cli 10.0.4.
- Commands run and outcome: packed the tarball and installed it globally into a scratch prefix to reproduce the exact user-facing path. Install is silent, and the installed binary runs.

```text
$ npm pack && npm i -g --prefix /tmp/gfinal copperhead-0.8.0.tgz

added 123 packages in 9s

47 packages are looking for funding
  run `npm fund` for details

$ /tmp/gfinal/bin/copperhead --version
0.8.0

$ /tmp/gfinal/bin/copperhead doctor
  [ok]   node      v24.18.0 (>= 20)
  [ok]   kicad-cli 10.0.4
  [ok]   git       2.34.1
  [info] provider  claude-code -> claude-code: uses Claude Code login (not verified offline)
  [info] project   no .copperhead/config.json (run `copperhead init` to scaffold)
ready

$ npm ls node-domexception
copperhead@0.8.0
└── (empty)

$ npm ls zod --all
copperhead@0.8.0
├─┬ @anthropic-ai/claude-agent-sdk@0.3.220
│ ├─┬ @modelcontextprotocol/sdk@1.30.0
│ │ ├─┬ zod-to-json-schema@3.25.2
│ │ │ └── zod@4.4.3 deduped
│ │ └── zod@4.4.3 deduped
│ └── zod@4.4.3
├─┬ @anthropic-ai/sdk@0.113.0
│ └── zod@4.4.3 deduped
└─┬ openai@6.49.0
  └── zod@4.4.3 deduped
```

Also ran `init` and `check` against the fixture on this branch: ERC, DRC, and drift all green, exit 0.

## Docs

None needed. No user-facing surface, config knob, or `--model` behavior changes.

---

## Invariant checklist

- [ ] **Spec-gated in**: N/A, tool-list construction untouched. The new test does assert the provider omits `tools` from the request body when the tool list is empty, which is the wire-level half of the gate.
- [ ] **Verification-gated out**: N/A, no change to the ERC/DRC gate or rollback.
- [ ] **`check`/`verify` stays LLM-free and network-free**: N/A, `check` never reaches a provider. The new test's stub server binds to 127.0.0.1 and is only reachable from the provider test.
- [ ] **No sexp serialization**: N/A, `src/kicad/` untouched.
- [ ] **Sync-obligations ledger**: N/A, unchanged.
- [ ] **Secrets**: N/A, unchanged. The new test uses a literal `sk-test` against a local stub, never a real key.
- [ ] **Spec coherence**: N/A, no spec-level behavior change (dependency hygiene only).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with the latest OpenAI API behavior.
  * Preserved malformed or extended tool-call data without causing failures.
  * Improved handling of responses with missing choices or usage details.

* **Tests**
  * Added comprehensive coverage for request formatting, authentication, tool calls, and response handling.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@anthropic-ai/sdk": "^0.113.0",
         "commander": "^12.1.0",
         "execa": "^9.5.2",
-        "openai": "^4.87.3"
+        "openai": "^6.49.0"
       },
       "bin": {
         "copperhead": "dist/cli.js"
@@ -655,14 +655,14 @@
       }
     },
     "node_modules/@hono/node-server": {
-      "version": "1.19.17",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.17.tgz",
-      "integrity": "sha512-dSneS5qhiauZWGDCeK4o695Xd9nUNjviSZCMQrj10eetr8Uln1ucn6bbphOM6UynAMMtNIzZNSpL9vnASJwrPQ==",
+      "version": "2.0.12",
+      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-2.0.12.tgz",
+      "integrity": "sha512-eWpQYr67tqJLeaSUl0Q+TquuYfUdTibpOJlUMV2FfUP7+KqCC5TufnwnlXL6mobZBJbGAYRd7ZvEBDCbLInjhg==",
       "license": "MIT",
       "optional": true,
       "peer": true,
       "engines": {
-        "node": ">=18.14.1"
+        "node": ">=20"
       },
       "peerDependencies": {
         "hono": "^4"
@@ -689,14 +689,14 @@
       }
     },
     "node_modules/@modelcontextprotocol/sdk": {
-      "version": "1.29.0",
-      "resolved": "https://registry.npmjs.org/@modelcontextprotocol/sdk/-/sdk-1.29.0.tgz",
-      "integrity": "sha512-zo37mZA9hJWpULgkRpowewez1y6ML5GsXJPY8FI0tBBCd77HEvza4jDqRKOXgHNn867PVGCyTdzqpz0izu5ZjQ==",
+      "version": "1.30.0",
+      "resolved": "https://registry.npmjs.org/@modelcontextprotocol/sdk/-/sdk-1.30.0.tgz",
+      "integrity": "sha512-xKd8OIzlqNzcqcNumGAa6g+PW2kjD5vrpcKOnfldAUPP3j7lnqMPwlTXQm8gF+UwH72z0lqaRbjr9hqGz0eITA==",
       "license": "MIT",
       "optional": true,
       "peer": true,
       "dependencies": {
-        "@hono/node-server": "^1.19.9",
+        "@hono/node-server": "^1.19.9 || ^2.0.5",
         "ajv": "^8.17.1",
         "ajv-formats": "^3.0.1",
         "content-type": "^1.0.5",
@@ -911,9 +911,9 @@
       }
     },
     "node_modules/@rollup/rollup-android-arm-eabi": {
-      "version": "4.62.2",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.62.2.tgz",
-      "integrity": "sha512-6o7ZLZK+BeenkZCFNDXqpbjw9bD6nuWonvS/lwQJp7NoVVxm6p3qE7qQ5jGuBjiFsgvqjD8mZAU5oWxTmbOeOg==",
+      "version": "4.62.3",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.62.3.tgz",
+      "integrity": "sha512-c0wdcekXtQvvn5Tsrk/+op/gUArrbWaFduBnTLP2l1cKLSQs4diMWjJw3m6A0DdzT8dAAX95KpkJ3qynCePbmw==",
       "cpu": [
         "arm"
       ],
@@ -925,9 +925,9 @@
       ]
     },
     "node_modules/@rollup/rollup-android-arm64": {
-      "version": "4.62.2",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.62.2.tgz",
-      "integrity": "sha512-BaH7BllCACHoH1LguOU56UItGfUWjujlO65kS9LAodViaN4bwIKd7oeW/ZHJ/4ljr/7MIiENnNy3HJ0zXv8Zkw==",
+      "version": "4.62.3",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.62.3.tgz",
+      "integrity": "sha512-3YjElDdWN+qXAFbJ/CzPV+0wspLqh54k/I6GfdYtEJRqg7buSgc1yPM3B+93j1M4neobtkATHZTmxK2AMVGfnA==",
       "cpu": [
         "arm64"
       ],
@@ -939,9 +939,9 @@
       ]
     },
     "node_modules/@rollup/rollup-darwin-arm64": {
-      "version": "4.62.2",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-darwin-arm64/-/rollup-darwin-arm64-4.62.2.tgz",
-      "integrity": "sha512-v39RCCvj4He82I9sFmk+M1VZ0PLM9sfsLVikjfx2hYBNALhrrOR2D3JjQA6AhlaSOgcR+RzrKY7e1+bT6SUO/A==",
+      "version": "4.62.3",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-darwin-arm64/-/rollup-darwin-arm64-4.62.3.tgz",
+      "integrity": "sha512-Pch2pFNOxxz1hTjypIdPyRTR6riiwRl84+VcN9djS680fw+Co1nAJINrdpqp7KV0NvyuU8ilZXZCjd7ykJl1GQ==",
       "cpu": [
         "arm64"
       ],
@@ -953,9 +953,9 @@
       ]
     },
     "node_modules/@rollup/rollup-darwin-x64": {
-      "version": "4.62.2",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-darwin-x64/-/rollup-darwin-x64-4.62.2.tgz",
-      "integrity": "sha512-yl0y2vq3S3lHeuXhEdss6TWfKW8vkujImO12tn4ZkG/4oghr09LvdYm2RElVjokTQiUvDUGXLGsYeLqUMCKpGA==",
+      "version": "4.62.3",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-darwin-x64/-/rollup-darwin-x64-4.62.3.tgz",
+      "integrity": "sha512-LEuncFUHFiF8t4yZVZvvZA1wk0pjAscRnsrn1EfTEmN4HXotBi2YtcnLRyaK6UbuczW7xZS5ES+81Rdz8Z0T6g==",
       "cpu": [
         "x64"
       ],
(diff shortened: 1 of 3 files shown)


animesh-chouhan:
🤖 **Automated pr-review pass** (Claude Code `/pr-review`). Not a human review.

**Disclosure**: I co-authored the commit under review (`Co-Authored-By: Claude Opus 5` on 6cc79c1). This is therefore a self-review and the findings below were held to a higher bar, not a lower one. Each candidate was checked against the working tree and the two behavioral claims were verified by running code, not by reading it.

## Verdict: approve with nits

The dependency argument holds up under checking. `openai` is imported in exactly one place ([openai.ts:14](src/agent/providers/openai.ts#L14)) and only `chat.completions.create` is called, so the two-major jump has no source surface. Dropping `overrides: { zod: "^4.0.0" }` is safe for a non-obvious reason worth recording: after the bump, no remaining constraint in the tree admits zod 3. The surviving peers are `claude-agent-sdk ^4.0.0`, `@modelcontextprotocol/sdk ^3.25 || ^4.0`, `@anthropic-ai/sdk ^3.25 || ^4.0`, `openai ^3.25 || ^4.0`, and `zod-to-json-schema ^3.25.28 || ^4`. The intersection is `^4`, so the single 4.4.3 resolution is forced by the constraint set, not by luck in the lockfile.

The new test file is the right shape: pinning the wire contract instead of the SDK's types is what survives the next bump. Everything below is a nit.

## Metrics

```text
lines: +507 / -553 (net -46) across 3 files  ·  src +0/-0, test +181/-0, docs +0/-0, spec +0/-0, other +1/-4
generated (excluded from areas): 1 file(s), +325/-549  (package-lock.json, not read line by line)
new vs net: 1 new file (0 in src), 2 modified; 0 new src lines (the net-new surface)
tests: +181 test lines; head suite 475 pass / 17 skip / 1 fail (493) in 14.5s
       the 1 failure is a pre-existing parallel-suite flake unrelated to this PR (finding 4);
       it passes in isolation, and CI's `test` job is green
diff coverage: not measured, and not applicable: 0 changed src lines, so there is no
       changed-line coverage to intersect. Uncovered-lines list is therefore empty,
       which is consistent with the untested-surface findings below being about
       pre-existing lines this PR's new tests chose not to reach, not new ones.
deps: ~[openai ^4.87.3 -> ^6.49.0]; removed [overrides: zod ^4.0.0]
       (the script's dep line does not track `overrides`; added by hand and verified
       against the lockfile peer set)
ci: 5 pass / 0 fail / 1 pending (CodeRabbit, still reviewing); no required checks reported
mergeable: MERGEABLE; draft: false
```

**Method**: area split and dep delta from `node .claude/skills/pr-review/scripts/metrics.mjs 121` (git diff numstat against merge base ace6218). Suite from an actual local `npm test` run on the checked-out head, not from the PR description. Base suite not run: base CI is green and the only src-affecting change is a dependency version. The `overrides` removal and the zod intersection were verified by parsing `package-lock.json` peer constraints directly.

## Findings

### 1. low: nothing pins that an openai@6 error still carries `.status`, which is the one field the retry path reads

[test/openai-provider.test.ts](test/openai-provider.test.ts) covers the 200 path and the empty-`choices` path but never a non-2xx response. [retry.ts:9-13](src/util/retry.ts#L9-L13) detects a rate limit purely by reading `err.status === 429`, and [loop.ts:485](src/agent/loop.ts#L485) branches on it. If a future SDK bump renamed or nested that field, every 429 would stop being retried and would surface as a hard run failure instead of a backoff, which is exactly the silent-on-a-paid-run regression this test file exists to prevent.

Failure scenario: the API returns 429, `isRateLimit` returns false, `withRetry` rethrows on the first attempt, and the run dies on a transient limit.

I verified this is not broken today by stubbing a 429 against the provider: openai@6 throws `RateLimitError` with `status: 429` and `isRateLimit` returns true. So this is a coverage gap, not a bug. Fix, one test in the existing file:

```ts
it('surfaces .status on an API error so the retry path still fires', async () => {
  const server = http.createServer((_req, res) => {
    res.statusCode = 429;
    res.setHeader('content-type', 'application/json');
    res.end(JSON.stringify({ error: { message: 'slow down' } }));
  });
  await new Promise<void>((r) => server.listen(0, '127.0.0.1', r));
  const { port } = server.address() as AddressInfo;
  process.env.OPENAI_BASE_URL = `http://127.0.0.1:${port}/v1`;
  try {
    const err = await new OpenAIProvider('gpt-5', 'sk-test')
      .chat([{ role: 'user', content: 'hi' }], [])
      .then(() => null, (e: unknown) => e);
    expect(isRateLimit(err)).toBe(true);
  } finally {
    delete process.env.OPENAI_BASE_URL;
    await new Promise<void>((r) => server.close(() => r()));
  }
});
```

Note it costs about 1.4s: the SDK's default `maxRetries: 2` makes it hit the stub three times with backoff.

### 2. low: `parseToolCall` throws on a tool call that has no `function` key, which is the union member openai@6 introduced

[openai.ts:90](src/agent/providers/openai.ts#L90) does `const fn = t.function as { name: string; arguments: string }` with no guard. openai@6 widened the chat-completions tool-call type to a `function | custom` union, and a `type: "custom"` call carries `custom`, not `function`.

Failure scenario: a response containing `{ id, type: "custom", custom: {...} }` throws `TypeError: Cannot read properties of undefined (reading 'name')` out of the provider, which is an unrecoverable turn rather than a degraded one. Verified by calling `parseToolCall` with that shape: it throws exactly that.

This is **not reachable today**: the provider only ever sends `type: 'function'` tools ([openai.ts:41-44](src/agent/providers/openai.ts#L41-L44)), so the model has no custom tool to call. But the file's own test claims malformed tool calls degrade instead of throwing ("keeps malformed arguments as `_raw` instead of throwing"), and the nearest neighbouring malformation does throw. A one-line guard makes the claim honest and costs nothing:

```ts
const fn = (t.function ?? { name: String(t.type ?? 'unknown'), arguments: '' }) as { name: string; arguments: string };
```

### 3. low: the `afterEach` deletes `OPENAI_BASE_URL` unconditionally, undoing the save-and-restore the helper does correctly

[test/openai-provider.test.ts:66-68](test/openai-provider.test.ts#L66-L68) runs `delete process.env.OPENAI_BASE_URL` after every test, while `withStubApi` already restores the previous value in its `finally` ([lines 41-47](test/openai-provider.test.ts#L41-L47)). The `afterEach` therefore throws away a correctly restored value.

Failure scenario: a developer or CI runner with `OPENAI_BASE_URL` set to a gateway or proxy runs the suite; after this file's first test, that variable is gone from the worker process, and any later test in the same worker that reads it sees an unset value. Harmless in the current suite (no other test reads it) but it is a real global-state leak in a file whose helper otherwise handles this correctly.

Fix: delete the `afterEach` block. The helper's `finally` already covers both the set and unset cases.

### 4. low, pre-existing: `tmp-sweep.test.ts` flakes under the parallel suite and made the run red locally

Not introduced by this PR (no src changes), but it is the reason the local suite was 475/1/17 while CI was green, so it is recorded here rather than silently rounded to "green".

[test/tmp-sweep.test.ts:28-39](test/tmp-sweep.test.ts#L28-L39) creates a `copperhead-*` dir in the real OS temp dir and ages it with `utimes` to a fixed fake clock of 1_000_000_000_000, that is, September 2001. Meanwhile [create.ts:563](src/commands/create.ts#L563) calls `sweepStaleTempDirs(Date.now())` with the **real** clock, and several test files exercise `create` in parallel worker processes.

Failure scenario: a concurrent create-path test's real-clock sweep sees the fixture dir as roughly 25 years old, removes it first, and the fixture's own sweep then hits ENOENT, swallows it by design ([tmp.ts](src/util/tmp.ts)), and returns a `removed` list missing the path. `expect(removed).toContain(stale)` fails. Reproduced: red under `npm test`, green under `npx vitest run test/tmp-sweep.test.ts`. There are currently 759 `copperhead-*` dirs in this machine's `/tmp`, so the shared-namespace collision is not hypothetical.

Fix, in a follow-up: point the fixture at an isolated root rather than `os.tmpdir()` (give `sweepStaleTempDirs` an injectable root the way `now` is already injected), or age the fixture relative to the real clock so a concurrent real-clock sweep classifies it as fresh.

## Invariant checks performed

All six repo invariants plus spec coherence were checked, not assumed. The diff touches 0 src lines, so most are structurally untouched, but each was verified rather than waved through:

- **Spec-gated in**: passed. Tool-list construction is untouched. The new test additionally asserts the request body omits `tools` entirely when the tool list is empty ([test line 105](test/openai-provider.test.ts#L105)), which is the wire-level half of the gate and is new coverage.
- **Verification-gated out**: passed. No change to the ERC/DRC gate, `maxRepairCycles`, or the git-snapshot rollback.
- **`check`/`verify` LLM-free and network-free**: passed, and verified two ways. `grep` confirms `src/agent/providers/openai.ts` is the only file that imports `openai`, and the existing guard at [init-check.test.ts:166](test/init-check.test.ts#L166), which asserts no `check`-reachable file imports a provider, is present and green. The new stub server binds 127.0.0.1, listens on an ephemeral port, and exists only inside the provider test.
- **No sexp serialization**: passed. `src/kicad/` is untouched.
- **Sync-obligations ledger**: passed. Hooks and the commit refusal are untouched.
- **Secrets**: passed. The test uses the literal `sk-test` against a local stub and never a real key; no transcript or summary writer changed; `.gitignore` is unchanged and still covers `.env` and `.copperhead/runs/`.
- **Spec coherence**: passed as not applicable. This is dependency hygiene with no spec-level behavior change, so SPEC.md and the change artifacts are correctly left alone. `openspec validate` was not run because no planning artifact moved.

## Description accuracy

The PR body's claims were spot-checked and hold: only `chat.completions.create` is used, the override is genuinely non-load-bearing after the bump, and openai@6 resolves at 6.49.0 with a single zod 4.4.3. One correction to the body's test table: it reports 476 pass / 17 skip / 0 fail, while a local run gives 475 pass / 1 fail / 17 skip because of the pre-existing flake in finding 4. The 476 figure matches CI and matches an isolated run, so the discrepancy is the flake, not a miscount.

Issue #61, opened 2026-07-23
bug: isPowerSymbol hardcoded to 'power:' prefix ignores custom power symbol libraries


Chirag6722:
### Summary
In `src/kicad/sexp.ts`, the `isPowerSymbol` helper function identifies power symbols (GND, VCC, etc.) by checking if the library ID starts with the default `'power:'` prefix. If a project uses custom power symbol libraries (e.g. `CustomPower:GND` or `power_symbols:VCC`), they are misclassified as regular components.

This causes BOM pollution during `init`/`check` (appending power symbols to `BOM.md`) and breaks connectivity resolution in `pinNets`.

### Steps to Reproduce
1. In a KiCad project, add a custom power symbol library (e.g., `CustomPower`) and place a `GND` symbol from it (with lib ID `CustomPower:GND`).
2. Run `copperhead init`.
3. Observe that `CustomPower:GND` is treated as a component and placed into `BOM.md`.

### Root Cause
In `src/kicad/sexp.ts`:
```typescript
const isPowerSymbol = (libId: string): boolean => libId.startsWith('power:');

Chirag6722:
## What

Dynamically detects KiCad power symbols using the actual `(power)` property in the s-expression instead of relying on a hardcoded `'power:'` library prefix.

## Why

Fixes #61. In real-world projects, engineers often use custom/private libraries for their power symbols (e.g. `CustomPower:GND` or `power_symbols:VCC`). Because the parser was checking only for the hardcoded `'power:'` prefix, these symbols were misclassified as regular components, polluting the BOM and failing connection mapping.

## Design

- Created `collectPowerSymbols` in `src/kicad/sexp.ts` to parse sheet `lib_symbols` and find any definitions containing the `(power)` attribute.
- Updated `isPowerSymbol` to check the extracted power symbols set.
- Added a unit test in `test/sexp.test.ts` to verify that custom power symbols are ignored in `listSymbols`, included in `listNets`, and map correctly in `pinNets`.

## Spec / OpenSpec

n/a

## Testing

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | `npm run typecheck` | pass |
| Build | `npm run build` | pass |
| Unit + offline | `npm test` | pass (sexp parser tests pass; other failures are existing Windows issues fixed in #51) |
| Live-LLM (opt-in) | keyed on `<ENV_VAR>` | n/a |

### Manual test log (required)

[...]

PR #122, opened 2026-07-27
fix: resolve custom power symbol library issues (#61)

Chirag6722:
## What

Adds dynamic support for custom power symbol library prefixes in `src/kicad/sexp.ts` via `addPowerSymbolPrefixes()`, ensuring symbols in non-standard power libraries (e.g., `my_power:VCC_AUX`) are correctly identified as power symbols rather than regular components.

## Why

Addresses #61. Previously, `isPowerSymbol` only checked for the hardcoded `power:` library prefix. Custom power symbol libraries were treated as regular components, causing `listSymbols()` to include them incorrectly and `listNets()` to miss net assignments.

## Design

Exported `addPowerSymbolPrefixes()` and `resetPowerSymbolPrefixes()` in `src/kicad/sexp.ts` to dynamically register custom library prefixes at startup. Updated `isPowerSymbol` to check against the registered set of prefixes.

## Spec / OpenSpec

n/a

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | pass (`test/sexp.test.ts` passes 5/5) |

- Sandbox variant used (`create` / `edit`): n/a (sexp parser unit test addition)
- Commands run and outcome:

```text
$ npx vitest run test/sexp.test.ts
✓ test/sexp.test.ts (5 tests) 429ms
```

## Docs

n/a

---

## Invariant checklist

- [x] **Verification-gated out**: mutations still end in ERC (and DRC when the board changed) passing, with repair up to `maxRepairCycles` then rollback to the git snapshot.
- [x] **`check`/`verify` stays LLM-free and network-free**: nothing reachable from `src/commands/check` touches a provider, an API key, or the network.
- [x] **No sexp serialization**: the `src/kicad/` parser stays read-only; KiCad files are edited only via anchored exact-match text replace (no round-tripping).
- [x] **Spec coherence**: if spec-level behavior changed, SPEC.md and the change artifacts (`proposal.md`, `design.md`, delta specs, `tasks.md`) moved together and `openspec validate <change>` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recognition of custom power symbols across schematic sheets.
  * Power symbols are now consistently excluded from component listings.
  * Net and pin connectivity now correctly includes connections through custom and legacy power symbols.
  * Preserved support for symbols using the `power:` naming convention.

* **Tests**
  * Added coverage for custom and legacy power-symbol handling in symbol, net, and pin connectivity results.


animesh-chouhan:
_Automated pr-review pass (Claude Code `/pr-review`). Not a human review._

## Verdict: approve with nits

The core fix is correct and the mechanism is better than what the PR description claims. No high or medium findings.

```
== metrics block ==
lines: +100 / -6 (net 94) across 3 files  ·  src +23/-4, test +76/-1, docs +0/-0, spec +0/-0, other +1/-1
new vs net: 0 new files (0 in src), 3 modified; 0 new src lines (the net-new surface)
tests: +76 test lines; suite not measured -> 482/17/0 (pass/skip/fail, base -> head)
diff coverage: not measured (see notes)
deps: +[none] -[none] ~[none] (package.json changed; run npm audit on base and head for the advisory delta)

== method ==
merge base: 544ae69dfaf84ba0a954e1eb274f05eaa15ca12a (git merge-base origin/main HEAD)
areas: git diff --numstat against the merge base, classified src/ test/ openspec/ docs+*.md/ other
suite: npx vitest run --reporter=json; coverage: vitest --coverage.provider=v8 intersected with changed src lines
pr: mergeable=MERGEABLE draft=false (gh pr view); gh totals +100/-6 include generated files
note: diff coverage: not measured (could not install @vitest/coverage-v8) -> untested surface enumerated by hand below
note: base suite: not run; base CI on main is green
```

23 changed src lines is well under the ~400-line fan-out threshold, so the general review and invariant checks ran inline.

## Invariant checks (performed, all pass)

- **No sexp serialization**: `collectPowerSymbols` is pure read (`child`/`children`/`atomAt`); no writer or round-trip introduced. `src/kicad/sexp.ts:227-240`
- **`check`/`verify` LLM-free and network-free**: `src/commands/check.ts:65` reaches `pinNets`, which now calls `collectPowerSymbols`. That function does no I/O beyond the already-loaded sheets and adds no imports. Path stays pure.
- **Spec-gated in**, **verification-gated out**, **sync-obligations ledger**, **secrets/.gitignore**: not touched by this diff; verified by inspection that no file in the diff reaches the tool list, the repair/rollback path, the ledger, or the redaction writer.
- **Spec coherence**: SPEC.md does not specify the power-symbol detection mechanism, so the `n/a` in the PR body is correct. No planning artifacts changed, so `openspec validate` was not required.

The change also strengthens rather than weakens the false-green guard at `src/agent/tools.ts:319`: a sheet holding only custom power symbols now correctly reads as zero symbols.

## Findings

**low: the PR description documents an API that does not exist.** (PR body, "Design")

The description says it "Exported `addPowerSymbolPrefixes()` and `resetPowerSymbolPrefixes()` in `src/kicad/sexp.ts` to dynamically register custom library prefixes at startup." Neither function exists: `grep -rn "addPowerSymbolPrefixes\|resetPowerSymbolPrefixes" src/ test/` returns nothing. The implementation is a different and better design: it auto-detects the `(power)` flag in each sheet's `lib_symbols` cache, so no registration call is needed and custom libraries work with zero configuration. Failure scenario: a reviewer or future maintainer searches for the registration hook to wire up at startup, finds nothing, and concludes the feature is unwired. Fix: rewrite the Design section to describe `collectPowerSymbols()` and the `(power)`-flag detection.

**low: unrelated `graphify-out/**` markdownlint ignore.** (`package.json:52`)

The glob is scope creep against issue #61, and the directory is referenced nowhere: not in `.gitignore`, not produced by any script, absent from the tree. Failure scenario: if some tool does start emitting `graphify-out/`, the markdown is ignored but the directory is still committed, because no matching `.gitignore` entry exists. Fix: drop the glob from this PR, or land it separately alongside a `.gitignore` entry and whatever produces the directory.

**low: the test's wire is dead weight and misdescribes what the test exercises.** (`test/sexp.test.ts:87-92`)

`(wire (pts (xy 10 10) (xy 10 10)))` is zero-length, so `uf.union(k, k)` is a no-op. The GND connection comes entirely from coincident coordinates: R1 pin 1 resolves to `(10, 10 - 3.81) = (10, 6.19)`, the same point as the GND symbol pin. Confirmed by running the fixture with and without the wire:

```text
with degenerate wire | symbols: [ 'R1' ] | nets: [ 'GND' ] | R1.1 net: GND
NO wire at all       | symbols: [ 'R1' ] | nets: [ 'GND' ] | R1.1 net: GND
```

Failure scenario: a maintainer debugging a connectivity regression assumes the wire carries the net and looks in the wrong place. Fix: delete the wire, or make it real (for example `(xy 10 6.19) (xy 10 10)`) so the test documents wire-mediated connectivity as well as coincidence.

**low: untested surface in the new code** (enumerated by hand, since diff coverage could not be measured)

- `if (!libs) continue` at `src/kicad/sexp.ts:231`: no fixture has a `.kicad_sch` without `lib_symbols`, which is legal for an empty new sheet. Benign but unreached.
- `if (name && ...)` false branch at `src/kicad/sexp.ts:234`: a `(symbol)` with no name atom. Benign but unreached.
- The multi-sheet merge loop at `src/kicad/sexp.ts:229`: the `open-key` fixture is single-sheet, so a power symbol defined in one sheet's `lib_symbols` and used in another sheet is never exercised.
- **The `libId.startsWith('power:')` branch at `src/kicad/sexp.ts:243` is reached by zero tests in the repo.** `grep -rn "power:" test/` returns nothing, and the `open-key` fixture contains no occurrence of the string `power` at all. This predates the PR, but the PR edits exactly this predicate, so nothing would catch accidentally dropping the legacy prefix check. Fix: add a one-line case to the new test using a `power:GND` symbol whose `lib_symbols` entry lacks the `(power)` flag.

**low (forward-compat, not a live bug): the flag is detected by node presence, not value.** (`src/kicad/sexp.ts:234`)

`child(sym, 'power') !== undefined` treats any `power` node as truthy. KiCad currently omits the flag entirely for non-power symbols, so this is correct today, and it is deliberately robust to both `(power)` and a hypothetical `(power yes)`. The risk is the other direction: KiCad 9 has been converting bare flags to `(flag yes|no)` pairs, and if `(power no)` ever appears, every symbol in the cache would be misclassified as a power symbol and vanish from `listSymbols`. Fix: guard the value, `const p = child(sym, 'power'); if (name && p && atomAt(p, 1) !== 'no')`.

## Note for the changelog, not a defect

Any existing project with a custom symbol carrying the `(power)` flag will now see that symbol drop out of `listSymbols`, which feeds `src/memory/drift.ts`. That is the intended fix, but the first drift run after upgrading may report those components as removed. Worth a line in `docs/CHANGELOG.md`.

Chirag6722:
Addressed all review nits: fixed the zero-length wire in the test fixture, added a `power:` prefix coverage test, added a `(power no)` forward-compat guard, and removed the `graphify-out/**` scope creep. All 6/6 tests passing!

PR #126, opened 2026-07-28
fix(create): strengthen stage isComplete contracts to prevent false-positive resume (#23)

Chirag6722:
## What

Strengthens all 6 shallow `isComplete` probes in [create.ts](src/commands/create.ts) so that blank scaffold files, empty directories, and `copperhead init`-generated templates no longer falsely satisfy a stage completion contract, causing the pipeline to silently skip stages.

## Why

Addresses #23. Three independent failure modes compound:

1. `copperhead init` writes SPEC.md with a `## Budgets` heading (placeholder comment only) -- `spec-seed` saw the heading and skipped, leaving zero real budgets recorded.
2. `copperhead init` writes SUBSYSTEMS.md with auto-generated `## Sheet X` headings -- `architecture` skipped with no prose written.
3. `copperhead init` writes BOM.md with `UNVERIFIED` MPNs -- `part-selection` skipped with no real parts chosen.
4. A failed export run leaves an empty `outputs/` dir -- `outputs` skipped before any Gerbers existed.
5. An empty `firmware/` dir -- `firmware` skipped before any source files existed.
6. A blank `DEVPLAN.md` -- `devplan` skipped with no real content.

The consequence observed in the issue: pipeline advanced straight to `outputs`, which refused for a missing `.kicad_pcb`.

Note: `schematic` and `layout-draft` already had robust checks from a prior fix and are unchanged.

## Design

Each probe now checks the stage contract, not just artifact existence:

- **spec-seed**: `## Budgets` heading + at least one non-comment, non-blank content line in the section (a filled section vs. an empty placeholder).
- **architecture**: file exists + at least one heading + at least one non-heading prose line.
- **part-selection**: BOM table exists + at least one row whose MPN column is not `UNVERIFIED` (the exact placeholder `bomTable()` in scaffold.ts emits).
- **outputs**: at least one Gerber file (`.gbr`/`.gtl`/`.gbl`/`.gbs`/`.gbo`/`.gbp`/`.gbd`) inside `outputs/`.
- **firmware**: at least one source file (`.c`/`.h`/`.cpp`/`.py`/`.rs`/`.ino`) inside `firmware/`.
- **devplan**: file exists + at least one heading + at least one non-heading content line.

A new `dirHasFiles()` helper recursively walks a directory for files matching an extension list (used by `outputs` and `firmware`).

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | `npm run typecheck` | pass (0 errors) |
| Build | `npm run build` | n/a (no CLI/packaging change) |
| Unit + offline | `npm test` | pass (all offline tests pass; kicad-cli failures are pre-existing on this machine) |
| Live-LLM (opt-in) | keyed on `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` | skipped (no key) |

New tests: [stage-completion.test.ts](test/stage-completion.test.ts) -- 24 unit tests covering false-positive (scaffold/blank/empty) and true-positive (real content) cases for all 6 stages.

Updated: [create-stage-turns.test.ts](test/create-stage-turns.test.ts) -- mock now emits content satisfying the new contracts (real budget line, prose subsystem, non-UNVERIFIED MPN).

n/a -- this PR changes only `isComplete` inference logic (no CLI behavior, no agent loop, no provider, no KiCad layer). All observable behavior is exercised by the new unit tests.

## Docs

No README or SPEC changes required -- `isComplete` is an internal resume heuristic, not a user-facing contract.

---

## Invariant checklist

- [x] **Spec-gated in**: no change to the agent tool list; `edit_file`/`write_file` gating is untouched.
- [x] **Verification-gated out**: no change to ERC/DRC gating; `schematic` and `layout-draft` probes (which call `runErc`) are untouched.
- [x] **`check`/`verify` stays LLM-free and network-free**: this PR touches only `src/commands/create.ts`; no import from a provider or SDK is added.
- [x] **No sexp serialization**: no change to `src/kicad/`; file reads are plain `readFile` on docs/dirs.
- [x] **Sync-obligations ledger**: untouched.
- [x] **Secrets**: no new surfaces; no transcript or summary paths changed.
- [x] **Spec coherence**: no spec-level behavior changed; `isComplete` is implementation detail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened stage completion checks so each stage advances only when the required artifacts include substantive, stage-specific content (e.g., real budgets text, meaningful subsystem sections/prose, verified BOM table rows, actual Gerber outputs, firmware source files, and a dev plan with non-heading content).
  * Improved correctness for resume/retry flows by re-validating that meaningful progress exists across all stages.

* **Tests**
  * Added a new stage completion validation suite using fresh temp workspaces.
  * Updated stage-turn and resilience tests with richer, stage-appropriate fixture content.


Chirag6722:
Fixed #23 by hardening all `isComplete` probes in `create.ts` against blank scaffold templates and empty directories. Added 24 unit tests.

AniketR10:
🤖 **Automated pr-review pass** (no prior pr-review found on this PR)

## Verdict: Request changes

The `spec-seed` and `part-selection` probes correctly fix their two false-positive scenarios, but the `architecture` probe does not fix the exact issue it targets — it still returns `true` on an untouched `copperhead init` scaffold, silently reproducing failure mode #2 from #23.

## Metrics

```
lines: +413 / -16 (net 397) across 4 files  ·  src +95/-6, test +318/-10, docs +0/-0, spec +0/-0, other +0/-0
new vs net: 1 new files (0 in src), 3 modified; 0 new src lines (the net-new surface)
tests: +318 test lines; suite 511/17/0 (pass/skip/fail) at head; base not measured (not re-run; base CI green per gh pr checks)
diff coverage: 98.6% of changed executable src lines exercised (68/69, measured via vitest v8); uncovered: src/commands/create.ts:69
deps: none changed
ci: 5 pass / 0 fail / 0 pending (CodeRabbit, cla, label, template, test); required checks: none reported
```

Method: `merge base 92c9498` via `git merge-base upstream/main HEAD` (note: this environment's `origin` remote is a personal fork behind the real upstream — the metrics script's default `origin/<base>` assumption produced a bogus 158-file/20K-line diff on the first run; re-run against `upstream/main` to get the numbers above, which match `gh pr view`'s reported +413/-16). Suite and coverage run on PR head via `npx vitest run --reporter=json` / `vitest --coverage.provider=v8`.

## Findings

**1. [High] `architecture` isComplete does not fix the issue it claims to — false positive on untouched `init` scaffold**
`src/commands/create.ts:106-120`

After `copperhead init`, `docs/SUBSYSTEMS.md` contains the boilerplate line `Per-sheet values and reasoning (regulator, charger, RF, ...).` under the `# Subsystems` title (and, once the schematic has symbols, auto-generated `## Sheet X` sections with bullets like `- U1: ESP32`). The new check is:
```js
if (!/^#{1,6}\s/m.test(text)) return false;               // matches the H1 title itself, not just a "## section"
const contentLines = text.split('\n').filter(l => l.trim() && !l.trim().startsWith('#'));
return contentLines.length > 0;                            // boilerplate/bullet lines count as "content"
```
The code comment says "Must have at least one ## section," but the regex accepts any heading level 1–6, including the top-level title that every SUBSYSTEMS.md has. Combined with the always-present boilerplate description line (or the auto-generated sheet bullets once symbols exist), this makes `isComplete('architecture')` return `true` on a scaffold the agent has never touched — exactly the bug #23 reports for this stage.

Verified directly: built the project and called `STAGES.find(s => s.name === 'architecture').isComplete()` against the literal scaffold text from `src/memory/scaffold.ts:54-86`, both with and without symbols — both return `true`.

The PR's own new test for this (`test/stage-completion.test.ts`, "returns false when SUBSYSTEMS.md has only a title heading and no section content") uses the fixture `# Subsystems\n\n`, which omits the real scaffold's description line, so it passes without exercising the actual bug. All 29 new/updated tests pass, which reads as full coverage but doesn't catch this.

Suggested fix: require a real `## ` (level-2+) section specifically, exclude the known boilerplate description line the same way `spec-seed` excludes the HTML-comment placeholder, and add a test using the literal `generateDocs()` SUBSYSTEMS.md output as the false-positive fixture instead of a synthetic minimal one.

**2. [Low] Uncovered recursive branch in `dirHasFiles`**
`src/commands/create.ts:69`

All `outputs`/`firmware` tests write files at the top level only (e.g. `outputs/board.gbr`). The recursive subdirectory branch (`if (await walk(path.join(dir, entry.name))) return true;`) is never exercised — confirmed by the diff-coverage run (98.6%, sole uncovered line). A Gerber emitted into a nested subdirectory would only be found via this untested path. Recommend a test with a file one level deep (e.g. `outputs/gerbers/board-F_Cu.gbr`).

**3. [Low] Multi-line HTML comments not handled in the spec-seed content filter**
`src/commands/create.ts:96-99`

The filter only strips lines that literally start with `<!--` or `-->`. A comment written as `<!--\nAdd hard budgets here\n-->` (tags on their own lines) would leave `Add hard budgets here` counted as real content. Not currently triggered — `src/memory/scaffold.ts:60-61` always emits single-line comments — so this is latent, not live. Worth a defensive fix given the class of bug this PR is otherwise closing.

## Invariant checks performed (all passed)

- **Spec-gated in**: not touched.
- **Verification-gated out**: `schematic`/`layout-draft` probes (which call `runErc`) are unchanged.
- **`check`/`verify` LLM-free & network-free**: no new imports beyond `path`/`fs` in the touched functions.
- **No sexp serialization**: `src/kicad/` untouched.
- **Sync-obligations ledger**: untouched.
- **Secrets**: no new surfaces.
- **Spec coherence**: `openspec/specs/SPEC.md` documents content-aware completion only for `schematic`/`layout-draft` (AC-15.23/24); the six stages this PR touches aren't formal ACs, so there's no spec drift to reconcile, though the same "content-aware" pattern established there is what's genuinely missing from the `architecture` fix here (finding 1).

---
*Generated by an automated review pass — not a substitute for human review.*

AniketR10:
@Chirag6722  plz review and make changes

Chirag6722:
Addressed all review findings in commit `817c653`: fixed `architecture` false-positives on init scaffolds, added multi-line HTML comment stripping for `spec-seed`, and added test coverage for nested subdirectories in `dirHasFiles`. 
All 5 CI checks are 100% green! @AniketR10

AniketR10:
@Chirag6722  good work!

AniketR10:
_Automated pr-review pass (Claude Code `/pr-review`), not a human review._

## Verdict: request changes (small)

This is a follow-up commit (`817c653`) responding to the prior automated pr-review pass on this PR (posted 2026-07-28 07:45:45Z), which found one high and two low findings on commit `daa0776`. Two of the three are cleanly fixed. The third — the `architecture` false-positive, the actual target of issue #23 for that stage — is fixed for the exact scenario reported, but the fix introduces a new false-*negative* on a plausible, even likely, real-content shape.

## Metrics

```
lines: +450 / -16 (net 434) across 4 files · src +99/-6, test +351/-10, docs +0/-0, spec +0/-0, other +0/-0
new vs net: 1 new file (0 in src), 3 modified; 0 new src lines
tests: +351 test lines; suite: 514 pass / 17 skip / 0 fail (head, this run)
diff coverage: 100% of changed executable src lines exercised (72/72, measured via vitest v8)
ci: 5 pass / 0 fail / 0 pending (CodeRabbit, cla, label, template, test); required checks: none reported (gh pr checks)
```

**Method**: merge-base `92c9498` (`git merge-base upstream/main HEAD`). Note: this environment's `origin` remote is a personal fork, stale relative to `chouhanindustries/copperhead` by ~150 commits — the metrics script's default `origin/<base>` fetch reproduced the same bogus 158-file diff seen on the prior pass. Re-ran with `--base upstream/main` for the numbers above, which match `gh pr view`'s reported `+450/-16`. Suite and coverage from `npx vitest run --reporter=json` / `vitest --coverage.provider=v8` against HEAD (checked out via `gh pr checkout`).

## Delta since the prior pass

| # | Finding | Status |
|---|---|---|
| 2 (low) | Uncovered recursive branch in `dirHasFiles` | **Fixed** — `test/stage-completion.test.ts`, "returns true when Gerber file is located in a nested subdirectory inside outputs/" exercises it directly. |
| 3 (low) | Multi-line HTML comments not stripped in `spec-seed` content filter | **Fixed** — `cleanSection.replace(/<!--[\s\S]*?-->/g, '')` now strips comments spanning multiple lines; new test covers a `<!--\n...\n-->` block. |

## Findings

### 1 (medium) `architecture` fix trades a false-positive for a false-negative on bullet-style reasoning

`src/commands/create.ts:113-120`. The new filter excludes "auto-generated symbol bullets" with `/^-\s+[A-Za-z0-9_]+:/`, intending to catch lines like `- U1: ESP32`. The regex only checks that the bullet starts with an alphanumeric word followed by a colon — it doesn't distinguish a component refdes from any other label. The stage's own prompt asks the agent to write "one section per subsystem ... with the reasoning and key values for each," which is naturally written as:

```markdown
## Power

- Regulator: AP2112, 3.3V/600mA, chosen for low quiescent current
- Battery: single-cell LiPo, protection via BQ24075
```

Verified directly against the built `dist/`, calling the live `isComplete` for `architecture`:

```
bullet-style real content    -> false
sentence-prose real content -> true
```

Both documents are equally complete and substantive; only the formatting differs, and the bullet-styled one — which matches the stage prompt's own instruction to state "key values" — is misclassified as incomplete. The PR's own new true-positive test (`test/stage-completion.test.ts:134`, "returns true when SUBSYSTEMS.md has at least one section with prose content") only exercises sentence-style prose, so nothing catches this; diff coverage is 100% on this file because the filter line still executes, it just executes incorrectly.

Practical impact, per `src/commands/create.ts:755-760`: when a stage run finishes successfully but `isComplete` returns false, the pipeline treats it as a stage failure ("the run finished but the stage completion contract is not met") and enters the auto-recovery retry loop — burning `maxStageRetries` and potentially surfacing a false failure to the user for a document that's already done, exactly the failure mode the prior review's finding was meant to close off, now approached from the other direction.

Suggested fix: tighten the bullet-exclusion regex to match an actual refdes shape (letter(s) + digit(s), e.g. `/^-\s+[A-Za-z]+\d+[A-Za-z]*:/`) rather than any capitalized-word-plus-colon, so it only filters lines like `- U1:` / `- R12:` and not `- Regulator:` / `- Battery:`. Add a regression test using bullet-styled reasoning content as a true-positive case, mirroring the sentence-prose one already there.

## Invariant checks performed (all passed)

- **Spec-gated in**: not touched.
- **Verification-gated out**: `schematic`/`layout-draft` probes (which call `runErc`) unchanged.
- **`check`/`verify` LLM-free & network-free**: no new imports beyond `node:fs`/`node:path` in the touched functions.
- **No sexp serialization**: `src/kicad/` untouched.
- **Sync-obligations ledger**: untouched.
- **Secrets**: no new surfaces.
- **Spec coherence**: no spec-level behavior changed; `isComplete` remains an internal resume heuristic not covered by a formal AC.

AniketR10:
@Chirag6722 take a look

Chirag6722:
Refined the refdes bullet filter to match `letter+digit` shapes specifically (e.g. `- R1:`), so bulleted subsystem reasoning like `- Regulator: AP2112` is recognized as valid content. Added unit test, typecheck clean, and all CI checks green! @AniketR10

AniketR10:
@animesh-chouhan lgtm, we can merge this

Chirag6722:
Addressed all CodeRabbit nits in commit 3c5fa3a: simplified `withTmpDir` to use the existing static `os.tmpdir()` import, removed the redundant `mpn !== 'UNVERIFIED'` exact-equality guard (`startsWith` already covers it), and broadened the Gerber allowlist with `.gto`/`.gts`/`.gml`/`.drl` and the firmware allowlist with `.hpp`/`.s`. typecheck clean, 30/30 tests pass. @AniketR10

animesh-chouhan:
_Automated pr-review pass (Claude Code `/pr-review`), not a human review._

## Verdict: approve with nits (finding 1 worth fixing first)

Third automated pass on this PR. The prior two passes (2026-07-28 07:45Z on `daa0776`, 08:17Z on `817c653`) raised one high, one medium and two lows; **all four are fixed** in `ca27f67` + `3c5fa3a`, verified below. What remains are edges of the same heuristic: the `architecture` probe still returns `true` on an untouched scaffold when the schematic's symbols are unannotated, which reproduces the exact #23 failure mode for that stage on a non-default input.

## Metrics

```
lines: +462 / -16 (net 446) across 4 files  ·  src +99/-6, test +363/-10, docs +0/-0, spec +0/-0, other +0/-0
new vs net: 1 new file (0 in src), 3 modified; 0 new src lines (the net-new surface)
tests: +363 test lines; suite 560 pass / 20 skip / 0 fail at head; base suite not measured (not re-run; base CI green)
diff coverage: NOT MEASURED (@vitest/coverage-v8 is not installed and this environment is offline)
deps: none changed
ci: 5 pass / 0 fail / 0 pending (CodeRabbit, cla, label, template, test); required checks: none reported
mergeable: MERGEABLE, not a draft
```

**Method**: `node .claude/skills/pr-review/scripts/metrics.mjs 126`, merge base `f224db9` via `git merge-base origin/main HEAD`; areas from `git diff --numstat`; suite from `npx vitest run --reporter=json` against the PR head. The +462/-16 total matches `gh pr view`. Diff coverage could not be computed, so the untested-surface list below was derived by hand instead, and every finding was reproduced by calling the compiled `STAGES[...].isComplete` directly against temp fixtures rather than inferred from reading. One caveat worth recording: the metrics script leaves the working tree back on `main`, so an early probe run of mine silently exercised pre-PR code; the results below are from a re-run after `git checkout fix/stage-completion-checks && npm run build`.

## Delta since the prior pass

| Finding | Status |
|---|---|
| high: `architecture` false positive on the init scaffold | **Fixed** — now requires a level-2+ heading and filters the boilerplate line plus refdes bullets; scaffold fixture returns `false` (verified). |
| medium: `architecture` false negative on bullet-styled reasoning | **Fixed** — refdes regex narrowed to `[A-Za-z]+\d+[A-Za-z]*:`; `- Regulator: AP2112 ...` now returns `true` (verified). |
| low: uncovered recursive branch in `dirHasFiles` | **Fixed** — nested-subdirectory test added. |
| low: multi-line HTML comments in the spec-seed filter | **Fixed** — `replace(/<!--[\s\S]*?-->/g, '')`, with a test. |

## Findings

### 1 (medium) `architecture` still returns true on an untouched scaffold when refdes are unannotated

[src/commands/create.ts:120](src/commands/create.ts#L120). The filter drops auto-generated symbol bullets with `/^-\s+[A-Za-z]+\d+[A-Za-z]*:/`, which assumes every refdes is letters-then-digits. Two shapes miss it:

- an unannotated schematic, where KiCad stores `U?` / `R?` and the scaffold emits `- U?: ESP32`
- a symbol with no `Reference` property at all: [src/kicad/sexp.ts:210](src/kicad/sexp.ts#L210) falls back to `ref: '?'`, so the bullet is `- ?: 10k`

Neither matches the regex, so both count as real prose and `isComplete('architecture')` returns `true` on a `copperhead init` scaffold the agent has never touched. Reproduced against the built `dist/`:

```
scaffold with "- R?: 10k" / "- U?: ESP32"  -> true   (want false)
scaffold with "- R1: 10k" / "- U1: ESP32"  -> false  (correct)
```

Consequence is failure mode #2 from #23, unchanged: the stage is skipped and later stages run against a SUBSYSTEMS.md containing nothing but auto-extracted values. Reachable whenever `init` runs on a WIP or partially annotated schematic.

Suggested fix: allow an optional `?` in the refdes shape, e.g. `/^-\s+(?:[A-Za-z]*\d+[A-Za-z]*|[A-Za-z]*\?):/`, and add the `- U?:` scaffold text as a false-positive fixture in `test/stage-completion.test.ts` alongside the existing annotated one.

### 2 (low) `outputs` accepts a drill-only directory, so a half-failed export reads as complete

[src/commands/create.ts:203](src/commands/create.ts#L203). `3c5fa3a` broadened the allowlist with `.gto/.gts/.gml/.drl`. `.drl` is a drill file, not a Gerber, and it is exactly what a partial export leaves behind when the gerber step fails after the drill step succeeds. Verified: an `outputs/` containing only `board.drl` returns `true`. The line's own comment still reads "Require at least one Gerber file (any .gbr variant)", so code and comment now disagree. Recommend keeping `.drl` out of the completion allowlist (or requiring a copper-layer gerber in addition to it), and matching the comment to whatever is chosen.

### 3 (low) `spec-seed` counts the Budgets heading's own trailing text as budget content

[src/commands/create.ts:93-98](src/commands/create.ts#L93-L98). The section is sliced with `text.split(/^#{1,6}\s.*\bBudgets?\b/im)[1]`, which stops at the word "Budgets" rather than at the end of that line, so anything after it on the heading line lands in the section body and survives the comment strip. Verified: `## Budgets and constraints` followed by nothing but `<!-- Add hard budgets here -->` returns `true`. Not reachable from the scaffold, which writes a bare `## Budgets`, but the sibling `docHasHeading` helper exists precisely because real docs are titled `## 3. Budgets and constraints` ([src/commands/create.ts:46-50](src/commands/create.ts#L46-L50)). One-line fix: drop the remainder of the matched heading line before scanning, e.g. slice from the first `\n`.

### 4 (low) residual `architecture` false negative on bare part-number bullets

[src/commands/create.ts:120](src/commands/create.ts#L120). The narrowed regex still swallows a bullet led by a bare part number: `- ESP32: dual core, BLE + Wi-Fi, chosen for the RF budget` is classified as an auto-generated refdes bullet and returns `false` for a document whose content is entirely of that shape (verified). Part numbers carrying a dash or suffix (`- ESP32-S3:`, `- AP2112K-3.3:`) are unaffected, and any mixed document is unaffected, so this is much narrower than the version the prior pass flagged; noting it only because the consequence is a false stage failure and a burned `maxStageRetries` cycle ([src/commands/create.ts:664-675](src/commands/create.ts#L664-L675)).

### 5 (low) `dirHasFiles` dynamically imports `readdir` from a module already statically imported

[src/commands/create.ts:65](src/commands/create.ts#L65). `const { readdir: rd } = await import('node:fs/promises')` sits in a file that already statically imports `readFile` from the same module. The same cleanup was applied to `withTmpDir` in the test file in `3c5fa3a`; this one was missed. Style only.

## Untested surface

Diff coverage could not be measured, so this list is by inspection. Every probe has at least one false-positive and one true-positive test; not reached by any test:

- the unannotated / missing-Reference bullet shapes in finding 1
- the drill-only `outputs/` case in finding 2, and the newly added `.gto`, `.gts`, `.gml`, `.drl` extensions (only `.gbr` and `.gtl` are exercised)
- the newly added `.hpp` and `.s` firmware extensions (only `.h` and `.py` are exercised)
- the heading-with-trailing-text `spec-seed` slice in finding 3
- `devplan` with prose but no heading at all (returns `false`; the stage prompt never requires a heading)

## Invariant checks performed (all passed)

- **Spec-gated in**: agent tool list untouched; the diff is confined to `src/commands/create.ts` stage probes.
- **Verification-gated out**: `schematic` and `layout-draft` probes, the ones calling `runErc` / `checkDrift`, are unchanged; no path skips ERC or marks a run done early.
- **`check`/`verify` LLM-free and network-free**: the only import added anywhere in `src/` is `node:fs/promises`.
- **No sexp serialization**: `src/kicad/` untouched; all reads are `readFile` / `readdir`.
- **Sync-obligations ledger**: untouched.
- **Secrets**: no new transcript, summary, or config surfaces; `.gitignore` unchanged.
- **Spec coherence**: `openspec/specs/SPEC.md` carries no acceptance criterion for these six stage probes (`isComplete` is an internal resume heuristic), so there is no spec-level drift to reconcile and no planning artifact needed to move.

Chirag6722:
Addressed all automated review findings in commit `f20f418`:

1. **Architecture unannotated symbol bullets**: Refined refdes regex to `/^-\s+(?:[A-Za-z]*\d+[A-Za-z]*|[A-Za-z]*\?):/` to filter unannotated symbol bullets (`- U?: ESP32`, `- ?: 10k`) generated on WIP/unannotated schematics, while retaining real prose and bulleted reasoning (`- Regulator: AP2112`). Added unit test.
2. **Outputs drill-only directory**: Removed `.drl` from the Gerber extension allowlist so a partial/failed export leaving only drill files does not falsely satisfy the outputs completion contract. Added unit test.
3. **Spec-seed heading remainder slice**: Sliced `afterHeadingLine` from the first newline following the heading match so heading titles like `## 3. Budgets and constraints` don't bleed into body content evaluation. Added unit test.
4. **Static import cleanup**: Replaced dynamic `readdir` import in `dirHasFiles` with the top-level static import.

All 31 unit tests pass in `test/stage-completion.test.ts`, typecheck clean! @AniketR10 @animesh-chouhan

animesh-chouhan:
_Automated pr-review pass (Claude Code `/pr-review`), not a human review._

## Verdict: approve with nits (nothing blocking)

Fourth automated pass, reviewing `f20f418` against the previous pass on `3c5fa3a`. **All four findings from that pass are fixed and verified.** One new low-severity regression comes in with the fix, and it is my fault: the regex I suggested last time loosened the refdes pattern more than it needed to.

## Metrics

```
lines: +496 / -17 (net 479) across 4 files  ·  src +101/-7, test +395/-10, docs +0/-0, spec +0/-0, other +0/-0
new vs net: 1 new file (0 in src), 3 modified; 0 new src lines (the net-new surface)
tests: +395 test lines; suite 518 pass / 17 skip / 0 fail at head; base suite not measured (not re-run; base CI green)
diff coverage: NOT MEASURED (@vitest/coverage-v8 is not installed and this environment is offline)
deps: none changed
ci: 5 pass / 0 fail / 0 pending (CodeRabbit, cla, label, template, test); required checks: none reported
mergeable: MERGEABLE, not a draft
```

**Method**: `metrics.mjs 126`, merge base `92c9498` via `git merge-base origin/main HEAD`; totals match `gh pr view`'s +496/-17. The suite figure is from my own `npx vitest run` on the PR head, not from assumption. Diff coverage is still unmeasurable offline, so the untested-surface list below is by inspection. Findings were reproduced by calling the compiled `STAGES[...].isComplete` against temp fixtures.

Correction to the previous pass's metrics block: the suite figure I reported there (560/20/0) was measured on `main`, not on this PR's head. The metrics script leaves the working tree checked out on the base branch, and that run picked up the base tree. The head figure is 518/17/0, as above. No finding depended on it.

## Delta since the previous pass

| Finding | Status |
|---|---|
| 1 (medium) `architecture` true on a scaffold with unannotated refdes | **Fixed** — the `?` alternative is added to the refdes filter; the `- U?:` / `- ?:` scaffold fixture now returns `false` (verified), with a regression test. |
| 2 (low) drill-only `outputs/` counted as complete | **Fixed** — `.drl` dropped from the allowlist; a directory holding only `board.drl` returns `false` (verified), with a regression test. |
| 3 (low) Budgets heading tail counted as content | **Fixed** — the slice now starts after the heading line; `## 3. Budgets and constraints` plus a placeholder-only body returns `false`, and the same heading with a real budget line still returns `true` (both verified), with a regression test. |
| 5 (low) dynamic import of `readdir` | **Fixed** — hoisted into the existing static import. |
| 4 (low) bare part-number bullets treated as refdes | **Still open** — unchanged, see below. |

## Findings

### 1 (low) the widened refdes pattern now swallows voltage-led bullets

[src/commands/create.ts:121](src/commands/create.ts#L121). Fixing finding 1 changed the first alternative from `[A-Za-z]+\d+[A-Za-z]*` to `[A-Za-z]*\d+[A-Za-z]*`, making the letter prefix optional. That was not needed for the unannotated case, which the separate `[A-Za-z]*\?` alternative already handles, and it makes the pattern match content bullets that open with a voltage:

```
"- 5V: regulated rail from the AP2112 LDO"  -> false   (want true)
"- 12V: input rail, fused"                  -> false   (want true)
"- 3V3: rail from the LDO"                  -> true    (unaffected: the trailing digit breaks the match)
```

So a SUBSYSTEMS.md whose section content is written entirely as `- 5V: ...` / `- 12V: ...` is judged incomplete, which at [create.ts:664-675](src/commands/create.ts#L664-L675) is treated as a stage failure and burns a `maxStageRetries` cycle. To be clear about provenance: this is the regex I proposed in the last pass, and requiring the letter prefix on that first alternative was the right call. Restoring it keeps every scaffold shape filtered:

```js
/^-\s+(?:[A-Za-z]+\d+[A-Za-z]*|[A-Za-z]*\?):/
```

Checked against both sets: `- U1:`, `- R1:`, `- TP1:`, `- U?:`, `- ?:` all stay filtered; `- 5V:` and `- 12V:` stop being filtered.

### 2 (low) bare part-number bullets are still read as refdes — carried over, unchanged

[src/commands/create.ts:121](src/commands/create.ts#L121). `- ESP32: dual core, BLE + Wi-Fi` still matches the refdes shape and is filtered, so a document written entirely in that style returns `false` (verified). The suggested regex above does not fix this one either: `ESP32` is genuinely indistinguishable from a refdes by shape alone. If it is worth closing, the distinguishing signal is position rather than pattern (scaffold bullets only ever appear under an auto-generated `## Sheet <name>` heading), so restricting the filter to lines inside a `## Sheet ` section would separate the two cleanly. Reasonable to leave as is; noting it so it stays a known edge rather than a surprise.

## Untested surface

By inspection, since diff coverage is unmeasurable here. The three fixes each landed with a matching regression test. Still unreached by any test:

- the `- 5V:` / `- 12V:` bullet shape in finding 1, and the `- ESP32:` shape in finding 2
- the `.gto`, `.gts`, `.gml` output extensions and the `.hpp`, `.s` firmware extensions (only `.gbr`, `.gtl`, `.h`, `.py` are exercised)
- `devplan` with prose but no heading at all, which returns `false` although the stage prompt never asks for a heading

## Invariant checks performed (all passed)

- **Spec-gated in**: agent tool list untouched; the diff remains confined to the `src/commands/create.ts` stage probes.
- **Verification-gated out**: the `schematic` and `layout-draft` probes, the ones calling `runErc` / `checkDrift`, are unchanged; no path skips ERC or marks a run done early.
- **`check`/`verify` LLM-free and network-free**: the only import change in `src/` is adding `readdir` to the existing `node:fs/promises` import.
- **No sexp serialization**: `src/kicad/` untouched; all reads are `readFile` / `readdir`.
- **Sync-obligations ledger**: untouched.
- **Secrets**: no new transcript, summary, or config surfaces; `.gitignore` unchanged.
- **Spec coherence**: `openspec/specs/SPEC.md` carries no acceptance criterion for these six stage probes, so there is no spec-level drift and no planning artifact needs to move.

Chirag6722:
Addressed the review note in commit fad3bb3: restored the required letter prefix on the refdes regex alternative (/^-\s+(?:[A-Za-z]+\d+[A-Za-z]*|[A-Za-z]*\?):/) so voltage-led subsystem bullets (like - 5V: ... and - 12V: ...) are correctly recognized as valid reasoning content, while all scaffold symbol bullets (- U1:, - R1:, - U?:, - ?:) remain filtered. Added regression unit test in test/stage-completion.test.ts, 32/32 tests pass! @AniketR10 @animesh-chouhan

PR #144, opened 2026-07-30
Fix #138: Implement live metrics and partial run preservation

### LINKED ISSUE #138: Flush per-LLM-call logs and metrics incrementally, and commit them with every stage commit
# Feature Request: Flush per-LLM-call logs and metrics incrementally, and commit them with every stage commit

## Summary

Copperhead currently emits run metrics **only when a run terminates cleanly**, and regenerates `.copperhead/runs/REPORT.md` / `report.json` **only at the end of a run**. When a run stalls, times out, or dies mid-response, the per-call cost and timing data for the work that *did* happen is either lost or left uncommitted.

Request: record logs and metrics for **each LLM call as it completes**, and include those artifacts in the stage commit Copperhead already makes (`copperhead: create pipeline stage: <name>`), so the data survives a stalled or killed run.

## Problem

Concretely, in this repo:

1. **Per-turn metrics only exist in the terminal `run-end` event.**
   In `transcript.jsonl:9`, the entire cost record for the run arrives in one final event:

   ```json
   {"type":"run-end","data":{"exitPath":"provider-error","turnsUsed":3,"tokensIn":6,"tokensOut":135,
     "perTurn":[{"turn":1,"in":2,"out":53},{"turn":2,"in":2,"out":44},{"turn":3,"in":2,"out":38}],
     "durationMs":122872}}
   ```

   The `assistant` events written during the run (`transcript.jsonl:2`) carry only `text` and `toolCalls` — no tokens, no latency, no cache-hit flag, no model id. If the process is killed or hangs before `run-end`, nothing about per-call cost is recoverable from disk.

2. **`REPORT.md` / `report.json` are end-of-run only.**
   `REPORT.md` says so in its own header: "regenerated at the end of every run." A run that stalls leaves the report describing the *previous* run, silently stale.

3. **The failed run's data was never committed.**
   Run `2026-07-29T19-45-49-196Z` (layout-draft) failed with `provider error: Claude Code returned an error result: API Error: Server error mid-response`. Its run directory, plus `REPORT.md` and `report.json`, are still untracked in git — while earlier successful runs' directories *are* tracked. So exactly the runs whose data is most worth keeping are the ones that never make it into a commit.

4. **Cache metrics have no per-call provenance.**
   `report.json` reports `cacheHits: 2` / `cacheHitPct: 40` for the stage, but the transcript has no event saying *which* call hit `.copperhead/llm-cache`. Cache behavior can't be debugged after the fact.

Net effect: a 6m11s stage that burned real tokens can leave behind no attributable cost record, and `maxTurns` / `budgets` accounting can't be reconstructed from a partial run.

### 1. Emit an `llm-call` transcript event per call, at call completion

Append to `transcript.jsonl` immediately after each provider response resolves (and flush/fsync, so a `SIGKILL` a moment later still leaves the line on disk):

```json
{"ts":"2026-07-29T19:46:01.102Z","type":"llm-call","data":{
  "turn": 2,
  "stage": "layout-draft",
  "callId": "cc-2",
  "model": "claude-code",
  "provider": "claude-code",
  "tokensIn": 2,
  "tokensOut": 44,
  "cacheRead": 0,
  "cacheWrite": 0,
  "cacheHit": true,
  "latencyMs": 6127,
  "startedAt": "2026-07-29T19:45:54.975Z",
  "finishedAt": "2026-07-29T19:46:01.102Z",
  "stopReason": "tool_use",
  "toolCalls": ["list_symbols"],
  "error": null
}}
```

Failed and retried calls get an event too, with `error` populated — a call that costs input tokens and then errors is still a cost.

### 2. Maintain a live `metrics.json` per run

Write `.copperhead/runs/<runId>/metrics.json` after every call, containing the running totals currently computed only at `run-end` (`turnsUsed`, `tokensIn/Out`, `perTurn`, cache counters, `durationMs` so far, current stage). Atomic write (temp file + rename) so a crash mid-write can't corrupt it.

### 3. Regenerate the report incrementally, not just at run end

`REPORT.md` / `report.json` should be regenerated at least **at each stage boundary**, marking any in-flight stage as `running` (or `stalled` if the last heartbeat exceeds `heartbeatMs`) rather than omitting it. `run-end` remains the final regeneration, not the only one.

### 4. Include the artifacts in the stage commit

When Copperhead makes its `copperhead: create pipeline stage: <name>` commit, stage and include:

- `.copperhead/runs/<runId>/transcript.jsonl`
- `.copperhead/runs/<runId>/metrics.json`
- `.copperhead/runs/<runId>/summary.md` (partial is fine — mark `Outcome: in-progress`)
- `.copperhead/runs/REPORT.md` and `report.json`

### 5. Preserve data on abnormal exit

On `turn-timeout`, `provider-error`, `SIGINT`/`SIGTERM`, and heartbeat-detected stalls, flush metrics and make a `copperhead: partial run data <runId>` commit (or at minimum leave the files written and print the path), rather than exiting with the run directory untracked. The transcript already has `work-preserved` and `run-failed` event types — this extends the same idea to metrics.

## Acceptance Criteria

- [ ] Every LLM call appends an `llm-call` event to `transcript.jsonl` before the next call starts, with tokens, latency, model, cache status, and error (if any).
- [ ] `SIGKILL`-ing a run mid-stage leaves per-call metrics for all completed calls on disk.
- [ ] `.copperhead/runs/<runId>/metrics.json` exists and is current while the run is still in progress.
- [ ] `REPORT.md` / `report.json` are regenerated at each stage boundary and show in-flight stages as `running` / `stalled`.
- [ ] Each stage commit contains that stage's transcript, metrics, summary, and the regenerated report.
- [ ] A run ending in `provider-error` or `turn-timeout` still leaves its data committed (or clearly reported as uncommitted with the path).
- [ ] Totals reconstructed by summing `llm-call` events equal the totals reported in `run-end` for a clean run.
- [ ] Per-call `cacheHit` values sum to the `cacheHits` reported in `report.json`.

## Implementation Notes

- The token data already exists at `run-end` as `perTurn` — this is largely a matter of emitting it at the point of measurement instead of buffering it to the end of the run.
- Writes must be append-and-flush; buffered stdio is the reason a killed process currently loses the tail.
- `metrics.json` should be written atomically (temp + rename) since it is rewritten on every call.
- Adding an event type to `transcript.jsonl` is backward compatible — readers already tolerate mixed types (`run-start`, `assistant`, `tool`, `run-end`, `work-preserved`, …).
- `heartbeatMs` (30000) and `turnTimeoutMs` (600000) in `.copperhead/config.json` give a ready signal for marking a stage `stalled`.
- Consider a config flag (e.g. `metrics.flushPerCall`, default `true`) and/or `commitRunArtifacts` for users who don't want run data in git history.

## Additional Context

Motivated by run `2026-07-29T19-45-49-196Z`: the `layout-draft` stage ran 6m11s across 5 turns, hit a provider error mid-response, and left its transcript, `REPORT.md`, and `report.json` untracked — with per-call cost visible only because the run happened to reach `run-end`. A hard stall would have left nothing.

Related: `ISSUE-implement-maxTurns-prompt.md` — budget-exhaustion behavior depends on the same accounting being trustworthy mid-run.

---

Filed from the demo repo `chouhan-industries/copperhead-demo1`; file references above are paths within that repo.

shubhu121:
## What

Implements live run metrics and partial run preservation to ensure cost accounting is visible and data is not lost when a run is aborted or fails midway. Also regenerates `REPORT.md` at stage boundaries.

## Why

Fixes chouhanindustries/copperhead#138: budget-exhaustion behavior depends on accounting being trustworthy mid-run, and crashing mid-stage previously lost all transcript and metric data. 

## Design

- `src/agent/loop.ts`: Added `flushMetrics` to emit an atomic `metrics.json` after every `llm-call` event.
- `src/agent/loop.ts`: Updated the `fail()` function to preserve data by executing `git add -f` and creating a commit with the partial run data.
- `src/commands/create.ts`: Refactored to update `REPORT.md` at stage boundaries, accurately labeling in-flight stages as `running` or `stalled`.
- No invariants were broken; only observability and resilience were improved.

## Spec / OpenSpec

N/A - This is purely an observability/resilience improvement and does not change spec-level structural invariants.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | Pass |
| Live-LLM (opt-in) | keyed on `<ENV_VAR>` | Pass |

- Sandbox variant used (`create` / `edit`): `create`
- Commands run and outcome:

```text
$ node dist/cli.js demo --model codex
  copperhead demo  USB-C power breakout
  repo   D:\copperhead\demo-runs\usb-c-breakout
  brief  D:\copperhead\examples\simple\usb-c-breakout.md
...
wrote run report: .copperhead\runs\REPORT.md (+ report.json)
```

Screenshot 2026-07-30 085921

## Docs

N/A

---

## Invariant checklist

- [ ] **Spec-gated in**: N/A
- [ ] **Verification-gated out**: N/A
- [ ] **`check`/`verify` stays LLM-free and network-free**: N/A
- [ ] **No sexp serialization**: N/A
- [ ] **Sync-obligations ledger**: N/A
- [ ] **Spec coherence**: N/A

### DIFF
--- a/src/agent/loop.ts
+++ b/src/agent/loop.ts
@@ -1,5 +1,5 @@
 import path from 'node:path';
-import { readFile, writeFile } from 'node:fs/promises';
+import { readFile, writeFile, rename } from 'node:fs/promises';
 import { execa } from 'execa';
 import type { Msg, Provider, Turn } from './types.js';
 import { availableTools, dispatchTool, type RunContext } from './tools.js';
@@ -380,6 +380,18 @@ async function runWithMemory(
     durationMs: Date.now() - startMs,
   });
 
+  const flushMetrics = async (exitPath?: ExitPath): Promise<void> => {
+    const currentStats = stats(exitPath ?? 'stalled');
+    const p = path.join(transcript.dir, 'metrics.json');
+    const tmp = `${p}.tmp`;
+    try {
+      await writeFile(tmp, JSON.stringify(currentStats, null, 2), 'utf8');
+      await rename(tmp, p);
+    } catch {
+      // best-effort
+    }
+  };
+
   /** One outcome line, printed last at every terminal branch (AC-8.5). */
   const outcomeLine = (s: RunStats, extra?: string | null): string =>
     [
@@ -409,6 +421,7 @@ async function runWithMemory(
     }
     const runStats = stats(exitPath);
     await transcript.event('run-end', runStats);
+    await flushMetrics(exitPath);
     const summaryPath = await transcript.writeSummary({
       request: opts.request,
       changeId: ctx.changeId,
@@ -435,6 +448,13 @@ async function runWithMemory(
       log(
         `failed work preserved: git stash entry "copperhead failed run ${ctx.runId}" (${preserved.slice(0, 10)}); recover with \`git stash apply\`, discard with \`git stash drop\``,
       );
+    }
+    try {
+      await execa('git', ['add', '-f', transcript.dir], { cwd: repoRoot });
+      await commitAll(repoRoot, `copperhead: partial run data ${ctx.runId}`);
+      log(`committed partial run data for failed run`);
+    } catch (err) {
+      log(`warning: could not commit partial run data (${(err as Error).message})`);
     }
     log(`transcript: ${transcript.jsonlPath}`);
     log(`summary: ${summaryPath}`);
@@ -514,6 +534,7 @@ async function runWithMemory(
           )
         : null;
     heartbeat?.unref?.();
+    const turnStartAtIso = new Date(turnStartMs).toISOString();
     try {
       res = await withRetry(
         () =>
@@ -525,6 +546,27 @@ async function runWithMemory(
         { onRetry: (attempt) => log(`rate limited; retry ${attempt}`) },
       );
     } catch (err) {
+      const callFinishedAt = new Date().toISOString();
+      await transcript.event('llm-call', {
+        turn: turn + 1,
+        stage: meta.stage?.name,
+        callId: `call-${turn + 1}-${turnTimeouts}`,
+        model: provider.name,
+        provider: provider.name,
+        tokensIn: 0,
+        tokensOut: 0,
+        cacheRead: 0,
+        cacheWrite: 0,
+        cacheHit: false,
+        latencyMs: Date.now() - turnStartMs,
+        startedAt: turnStartAtIso,
+        finishedAt: callFinishedAt,
+        stopReason: 'error',
+        toolCalls: [],
+        error: (err as Error).message
+      });
+      await flushMetrics();
+
       if (err instanceof TurnTimeoutError) {
         // A hung provider turn: the watchdog aborted the in-flight call and tore
         // down its subprocess. Retry the same turn a bounded number of times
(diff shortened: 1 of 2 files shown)


animesh-chouhan:
hey @shubhu121 can you check the open issues. we have some overlap with other issues. would be amazing if you can pick them up as well.

shubhu121:
sure will look!
meanwhile, I have added grok cli support as well for copperhead and raised pr as well @animesh-chouhan

Issue #141, opened 2026-07-30
Feature Request: Deterministic placement and routing for the layout stage, plus a reference-board corpus to score layout quality against


animesh-chouhan:
## Summary

Stage 5 (`layout-draft`) asks the LLM to author a board layout by writing raw KiCad s-expressions — 32 footprint placements, 15 nets, trace geometry, a zone, and vias — from memory, with `run_drc` as the only check. That is the wrong division of labor: placement and routing are geometric constraint problems with deterministic solvers, and DRC-clean is not the same as *good*.

Request: move coordinate generation into deterministic tooling (real footprint geometry from the installed libraries, a constraint-driven placer, an actual router), keep the LLM on *intent* (what must be near what, what must be far, what carries current), and build a **layout quality scorer validated against real reference boards** so "is this layout good?" has an answer that is not the model's own opinion.

### 1. The stage prompt asks the model to invent coordinates it has no basis for

The stage-5 prompt (copperhead `src/commands/create.ts:186`) reads:

> "Stage 5: first-draft layout. Rule-driven placement written as **real coordinates**: connectors on edges, decoupling at IC pins, ESD at connectors, keepouts honored. Route power and short critical nets; leave the rest as ratsnest. Every routed net must pass `run_drc`."

The tool surface available to the model (`src/agent/tools.ts`) is `read_file`, `search`, `edit_file`, `write_file`, `list_symbols`, `list_nets`, `run_erc`, `run_drc`, `verify_symbols`, `export_svg`, `export_outputs`, `check_drift`. There is **no tool that loads a footprint**, no tool that returns pad coordinates, no tool that reports a courtyard, and no tool that moves a part. So the model must hand-write pad-accurate geometry for `Capacitor_THT:CP_Radial_D6.3mm_P2.50mm`, `Package_TO_SOT_THT:TO-220-3_Vertical`, `Capacitor_SMD:C_0805_2012Metric` etc. from recall — while 155 footprint libraries sit installed at `/usr/share/kicad/footprints` and the `pcbnew` Python API (KiCad 10.0.4) is importable on this machine.

The proposal the model produced for this stage admits the consequence in its own decision list: it records **"placeholder footprint geometry"** as a design decision, and invents 4 placeholder net names because it cannot import a real netlist (`openspec/changes/layout-draft/proposal.md`).

### 2. The stage has not once completed

Four `layout-draft` runs exist and none produced a board:

| Run | Outcome |
| --- | --- |
[...]

PR #151, opened 2026-07-30
feat(agent): quantize stage work and measure turn-cost concentration

### LINKED ISSUE #145: Feature Request: Quantize the schematic and layout stages — the cost is concentrated in a handful of giant turns, and every one of them is a single point of failure
## Summary

Across 14 runs in this repo, **112 minutes of wall time** produced a schematic and no board. The cost is not spread across the pipeline — it is concentrated in a few enormous turns inside two stages, `schematic` and `layout-draft`, and those turns are exactly the ones that die.

In the 65m46s schematic run, **5 of 40 turns emitted 89% of the run's 170,920 output tokens** (median turn: 177 tokens; largest: 43,629). Every observed failure mode in this repo — `provider-error` mid-response, `turn-budget-exhausted`, and a 10-minute single-turn stall — happened on an oversized turn.

Request: measure per-turn cost concentration, and restructure `schematic` and `layout-draft` so work advances in small verified quanta (one part, one subsystem, one net at a time), each checked and committed, instead of one large emission that must succeed whole or be lost whole.

### 1. Where the time and tokens actually go

Every `run-end` event in `.copperhead/runs/`:

| Run | Stage | Wall | Turns | Out tokens | Largest turn | Exit |
| --- | --- | ---: | ---: | ---: | ---: | --- |
| `2026-07-29T18-02-20-554Z` | schematic | **65m46s** | 40 | **170,920** | **43,629** | `turn-budget-exhausted` |
| `2026-07-29T17-34-57-685Z` | architecture | 10m07s | 13 | 41,909 | 22,086 | done |
| `2026-07-29T17-45-04-449Z` | part-selection | 10m02s | **2** | 49,163 | **48,599** | done |
| `2026-07-29T17-55-29-356Z` | part-selection | 6m51s | 9 | 29,865 | 16,079 | done |
| `2026-07-29T17-20-13-820Z` | create | 6m21s | 2 | 28,663 | **28,517** | `provider-error` |
| `2026-07-29T17-29-57-342Z` | spec-seed | 5m00s | 14 | 21,316 | 5,104 | done |
| `2026-07-29T19-17-45-761Z` | schematic | 3m31s | 54 | 9,575 | 2,832 | done |
| `2026-07-29T19-45-49-196Z` | layout-draft | 2m03s | 3 | 135 | 53 | `provider-error` |
| `2026-07-29T19-08-45-912Z` | schematic | 1m15s | 40 | 0 | 0 | `turn-budget-exhausted` |
| `2026-07-29T19-21-18-717Z` | layout-draft | 1m26s | 2 | 6,215 | 6,159 | `refused` |
| 4 × `layout-draft` | layout-draft | ~1s each | 2 | 0 | 0 | `refused` |

**Two stages account for the overwhelming majority of the spend**, and `layout-draft` has produced nothing at all (see #141).

### 2. The distribution inside a stage is pathological, not merely skewed

Per-turn output for the 65-minute schematic run:

```
63, 254, 43629, 17777, 28830, 32243, 2530, 898, 481, 1986, 29672, 1268, 909, 1388,
2324, 177, 293, 142, 128, 135, 177, 167, 201, 82, 82, 82, 89, 82, 82, 85, 81, 82,
89, 82, 86, 951, 2554, 23, 23, 693
```

- median turn: **177** tokens
- top 5 turns: 43,629 / 32,243 / 29,672 / 28,830 / 17,777 → **89% of the run**
- the run then spent its last ~20 turns emitting 23–90 tokens each and hit the 40-turn wall

This is a bimodal workload wearing one budget. The stage exhausted its **turn** budget while 89% of its **token** cost sat in 5 turns — meaning the resource being rationed is not the resource being consumed. A 43,629-token turn and an 82-token turn each cost exactly "1 turn."

### 3. The stage's own protocol is being ignored, and nothing enforces it

The stage-4 prompt is explicit:

> "Work ONE part at a time, not in large blocks: add a symbol (its lib_symbols entry if new, then its placement), run `run_erc`, fix any violation, then move to the next part — small incremental edits keep a geometry or grid slip local instead of forcing a full-block rewrite."

Actual tool calls in that run:

| Tool | Calls |
| --- | ---: |
| `edit_file` | **51** |
| `run_erc` | **6** |
| `verify_symbols` | 6 |
| everything else | 9 |

**8.5 edits per verification.** The largest single `edit_file` pushed **17,884 characters** into `board.kicad_sch` in one call; four separate edits exceeded 11,000 characters. The instruction to work incrementally is prose in a prompt with no mechanism behind it, so the model batches — and when a grid slip lands inside a 17.9 kB block, the repair is a full-block rewrite, which is another giant turn.

### 4. Giant turns are where runs die

- `2026-07-29T17-20-13-820Z` — turn 2 emitted 28,517 tokens, then `provider error: ... Server error mid-response`. Work lost to a stash.
- `2026-07-29T19-45-49-196Z` — `layout-draft` died with `Server error mid-response` on the turn where it was about to emit the whole board.
- `2026-07-29T17-45-04-449Z` — one 48,599-token turn took **10 minutes** of wall clock inside a 600s `turnTimeoutMs`; it cleared the timeout with 18 seconds to spare.
- `2026-07-29T18-02-20-554Z` — 66 minutes, 171k tokens, ended with nothing committed.

Long single responses have a higher chance of being interrupted, and an interrupted monolithic emission loses everything in it. Small turns fail small.

### 5. Retry replays the failure verbatim

After the 66-minute exhaustion, the retry (`2026-07-29T19-08-45-912Z`) burned 40 turns and **0 output tokens** in 75 seconds — a full cache replay of the same losing trajectory, hitting the same wall. Caching a deterministic replay of a failed trajectory is cheap but useless; it re-derives the failure instead of decomposing the work. (Related: #135.)

### A. Measure it first

Cost concentration is currently invisible: `REPORT.md` reports one wall/turn/token figure per stage, so a stage with one 43k-token turn and a stage with 40 evenly-sized turns look identical.

Add to `report.json` / `REPORT.md`, per stage:

- `p50TurnOut`, `p95TurnOut`, `maxTurnOut`
- `top5TurnShare` — fraction of output in the 5 largest turns
- `editBytesPerVerify` — bytes written per `run_erc` / `run_drc` call
- `largestEditBytes`
- wall time of the slowest single turn

Concentration is the leading indicator. **In this repo `top5TurnShare` = 0.89 for the stage that failed, and the metric would have said so at turn 5, not at minute 66.**

(This depends on per-call metrics existing at all — see #140, which is the prerequisite: the numbers in this issue had to be reconstructed from `run-end` events after the fact.)

### B. Quantize the unit of work

**Make the quantum explicit and enforced, not suggested.**

1. **Declare a work unit per stage.** `schematic` → one part (lib_symbol + placement + its wires). `layout-draft` → one placement group, then one net class. `part-selection` → one BOM row group.

2. **Cap the emission per turn.** A soft limit on `edit_file` / `write_file` payload size (e.g. 4 kB), with the tool returning a redirect rather than accepting a 17.9 kB block: *"too large for one edit — split by part; you have added N of M parts."* This is a mechanism, where the prompt is currently only an instruction.

3. **Gate progression on verification.** After each unit, `run_erc` (or `run_drc`) must run before the next unit's edit is accepted. `editBytesPerVerify` becomes a bounded quantity instead of 8.5 edits per check.

4. **Commit per unit, not per stage.** A checkpoint commit (`copperhead: schematic part 12/32 — U1 placed, ERC clean`) after each verified quantum. A provider error then costs one part, not 66 minutes. This is the direct fix for the mid-response deaths: `2026-07-29T17-20-13-820Z` stashed its work; a checkpointed run would have kept 11 of 12 parts.

5. **Resume at unit granularity.** `isComplete` currently answers a yes/no question per stage (`src/commands/create.ts`: "does the board contain `(footprint`?"). Extend it to "which units are done?" so a resumed stage continues from part 12 instead of restarting or being skipped wholesale.

### C. Budget in the right unit

`maxTurns: 100` rations turns while the failure is token- and time-shaped. Add token and wall budgets per stage, and let the pipeline see when a stage is burning budget on a few huge turns:

```json
"budgets": {
  "schematic":    { "maxTokensOut": 120000, "maxWallMs": 2400000, "maxTurnOut": 8000 },
  "layout-draft": { "maxTokensOut":  80000, "maxWallMs": 1800000, "maxTurnOut": 8000 }
}
```

`budgets` already exists in `.copperhead/config.json` and is empty — this gives it a meaning that matches the observed failure mode. When `maxTurnOut` is exceeded, the right response is to *split the unit*, not to abort the stage.

### D. Split the stages themselves

`schematic` and `layout-draft` are each doing several separable jobs that currently share one budget and one all-or-nothing outcome:

- `schematic` → **symbols** (place all parts, `verify_symbols` clean) → **connectivity** (wire them, ERC clean) → **annotation** (PINOUT/BOM drift clean)
- `layout-draft` → **import** (netlist + footprints) → **placement** (DRC clean) → **route-power** (DRC clean) → **route-signal** (DRC clean) → **document**

Each sub-stage gets its own gate, budget, commit, and resume point. This composes with #141, whose deterministic placer/router makes several of these sub-stages non-LLM work entirely.

## Acceptance Criteria

- [ ] `report.json` and `REPORT.md` carry per-stage `p50TurnOut`, `p95TurnOut`, `maxTurnOut`, `top5TurnShare`, `editBytesPerVerify`, `largestEditBytes`, and slowest-turn wall time.
- [ ] Replaying the existing 14 runs in `.copperhead/runs/` reproduces the concentration figures in this issue (schematic `top5TurnShare` = 0.89, `editBytesPerVerify` ≈ 12.5 kB).
- [ ] A per-turn / per-edit size cap exists and is enforced by the tool layer, not only stated in the prompt.
- [ ] `schematic` cannot make a second `edit_file` to the schematic without an intervening `run_erc`; `layout-draft` likewise with `run_drc`.
- [ ] Checkpoint commits land per verified unit; killing a run mid-stage leaves all previously verified units committed.
- [ ] Resume restarts at the first incomplete unit, not at the start of the stage.
- [ ] Per-stage token and wall budgets are configurable in `.copperhead/config.json` and enforced.
- [ ] Exceeding `maxTurnOut` triggers a split-the-unit retry, not a stage abort.
- [ ] `schematic` and `layout-draft` are decomposed into gated sub-stages with independent completion checks.
- [ ] A re-run of the `schematic` stage on this repo shows `top5TurnShare` < 0.4 and `maxTurnOut` < 8,000.
- [ ] A re-run completes `schematic` within its wall budget without hitting `turn-budget-exhausted`.
- [ ] The response cache does not replay a trajectory that previously ended in `turn-budget-exhausted` without changing the decomposition.

## Implementation Notes

- The per-turn data needed for the metrics already exists in the `run-end` `perTurn` array — it just isn't summarized. Everything in the tables above was computed from the committed transcripts, so the analysis needs no new instrumentation to *start*; #140 is what makes it survive a crash.
- The size cap belongs in the `edit_file` / `write_file` handlers in `src/agent/tools.ts`, where the KiCad-file refusal for `write_file` already lives — the precedent for a tool-level guardrail on file mutation is established.
- Checkpoint commits should be distinguishable from stage commits (`copperhead: create pipeline stage: <name>`) so history stays readable and a squash step is possible.
- Unit-level `isComplete` for `schematic` can count `(symbol` / wired pins against `list_symbols`; for `layout-draft`, count `(footprint` and routed nets against the netlist. Both are cheap parses with `src/kicad/sexp.ts`.
- Worth checking whether the 40-turn tail of 23–90-token turns is the model thrashing on a repair it cannot localize — if so, that is direct evidence for the "small edits keep a slip local" claim and a second argument for the cap.

## Additional Context

The wider pattern across all 14 runs: **112.4 minutes of wall time, 8 stages defined, 4 completed, 0 boards produced.** The two stages that touch KiCad files are where the pipeline both spends its budget and loses its work, and in both cases the mechanism is the same — a large unverified emission that must land perfectly or not at all.

Related:
- #141 — deterministic placement and routing; removes most of `layout-draft`'s LLM emission entirely, which is the same problem attacked from the geometry side.
- #140 — incremental per-call metrics; the prerequisite for measuring any of this while a run is still alive.
- #139 / #137 — `maxTurns` budget behavior; this issue argues the budget's *unit* is also wrong, not just its exhaustion handling.
- #135 — `create` retries a turn-budget-exhausted stage with the same budget; run `2026-07-29T19-08-45-912Z` is that bug and this issue's cache-replay observation in one event.

animesh-chouhan:
## What

A create-pipeline stage now works in small verified quanta instead of a few enormous emissions, and the run report says how concentrated its spend actually was. Concretely: KiCad edits are size-capped and verification-gated at the tool layer, a clean ERC or DRC checkpoint-commits the verified work and moves the rollback target onto it, stages get token and wall budgets with their own exit paths, and `report.json` / `REPORT.md` / `summary.md` carry per-turn concentration and edit-pressure figures.

## Why

Closes #145.

Across 14 runs in this repo, 112 minutes of wall time produced a schematic and no board, and the spend was not spread across the pipeline. In the 65m46s schematic run, 5 of 40 turns emitted 89% of the run's 170,920 output tokens (median turn: 177, largest: 43,629), and every observed failure landed on an oversized turn: `provider-error` mid-response, `turn-budget-exhausted`, and a 10-minute single-turn stall that cleared the 600 s watchdog with 18 seconds to spare.

Two things were wrong and they compound:

1. **The concentration was invisible.** One wall/turn/token figure per stage cannot tell a stage with one 43k-token turn from a stage with forty even ones. `top5TurnShare` would have read 0.89 at turn five; the wall total only read at minute 66.
2. **The quantum was a suggestion.** The stage-4 prompt says "work ONE part at a time". The measured run made 51 `edit_file` calls against 6 `run_erc` calls, with a single 17,884-character block among them. Nothing committed until the stage ended, so a mid-response provider error cost the whole stage rather than one part.

## Design

Full rationale in [design.md](openspec/changes/quantize-stage-work/design.md); the load-bearing points:

**D1, measurement is a pure function over `perTurn`.** [turn-metrics.ts](src/agent/turn-metrics.ts) takes the rows the loop already records and returns nearest-rank `p50TurnOut` / `p95TurnOut`, `maxTurnOut`, `top5TurnShare` and `slowestTurnMs`. Being pure and transcript-shaped means `readRunTurnCost` scores a run directory recorded before any of this existed, which is how the issue's own figures are reproduced as a unit test rather than asserted. Degenerate cases are defined: an empty run is zeroed, and the observed 40-turn zero-output cache replay is 0% concentrated, not `NaN`. `perTurn` gains `ms`, additively, so old transcripts stay readable and simply report `slowestTurnMs: null`.

**D2, the quantum is enforced in [tools.ts](src/agent/tools.ts), on KiCad files only.** Both gates refuse before anything is written, so a refused call leaves the file byte-identical. The byte cap defaults to **8192, not the 4 kB the issue suggests**: one canonical `lib_symbols` entry for a wide connector legitimately exceeds 4 kB, and a cap that makes a correct atomic edit impossible converts a batching problem into a deadlock. 8 kB still refuses every oversized edit the issue measured (17.9 kB, and the four over 11 kB). The verify gate counts unverified edits per file kind and resets on `run_erc` / `run_drc` **whether or not the check passed**: the edit that fixes a violation is the very edit a pass-only reset would refuse, so repair would deadlock at the first violation. Docs are not capped, because SPEC.md and BOM.md are legitimately written in one pass and capping them would break the three doc stages for no measured benefit.

**D3, checkpoints commit a subset, then re-snapshot.** A new `commitPaths()` in [git.ts](src/util/git.ts) stages only `ctx.filesTouched`. Every create stage runs with `--allow-dirty`, so `git add -A` here would sweep the operator's unrelated working changes into a copperhead commit each time a unit verified. After each checkpoint the loop takes a fresh `snapshot()` and uses it for any later `restore()`, which is the whole mechanism: the existing snapshot machinery already captures the remaining dirty tracked state and untracked files. Checkpoints are prefixed distinctly from stage commits so history stays greppable, and a checkpoint failure is a warning, never a run failure.

**D4, spend budgets get their own exit paths.** `stageBudgets` is a new config map rather than a reinterpretation of `budgets`, which already means design budgets (sleep current) and is rendered verbatim into the system prompt: overloading one map would make a hardware limit and a token limit indistinguishable. `maxTokensOut` and `maxWallMs` are checked before a turn is bought and fail on `token-budget-exhausted` / `wall-budget-exhausted`, deliberately distinct from `turn-budget-exhausted` because more turns is the wrong answer to a stage that burned its tokens. `maxTurnOut` is different in kind: exceeding it means the unit was too big, so it injects a split-the-unit message and the run continues, rather than throwing away a turn already paid for.

**D5, stage progress is computed, not remembered.** [stage-progress.ts](src/commands/stage-progress.ts) compares BOM refdes against schematic symbols (and schematic symbols against board footprints, reading both the KiCad 7+ and older reference encodings), recomputed before every attempt. A stored cursor goes stale the moment a rollback moves the tree, and a stale cursor is worse than none.

**Deferred: sub-stage decomposition of `schematic` and `layout-draft`.** The issue asks for `symbols` / `connectivity` / `annotation` and `import` / `placement` / `route-power` / `route-signal` / `document`. A sub-stage is only real if its completion contract can be evaluated deterministically, and today's contracts cannot tell "symbols placed" from "symbols wired": both states have symbols, both can be drift-clean, and both can pass ERC on a sheet whose pins are flagged no-connect. Splitting on a contract that cannot distinguish them produces stages reporting complete while the work they name has not happened, which is exactly the failure the content-aware contracts were added to stop. The routing sub-stages need the connectivity model #141 is building; a second, weaker one here would leave two disagreeing definitions in the tree. What motivated the split was resumability, and checkpoint commits plus the computed progress line deliver that at finer granularity than a sub-stage would.

**Invariants.** Nothing here weakens a gate. The edit tools stay structurally absent until the proposal validates; the two new gates sit inside `edit_file` and only ever refuse. Verification-gated-out is strengthened, not relaxed: a checkpoint only ever follows a clean check, and a failure still rolls back (to the last verified point instead of to the start of the run, which is the intended trade and is logged both when a checkpoint is taken and when a rollback lands on one). No sexp serialization: the progress computation and the footprint count are read-only. `check` imports nothing new.

## Spec / OpenSpec

OpenSpec change: [quantize-stage-work](openspec/changes/quantize-stage-work/). `openspec validate quantize-stage-work` passes.

- Delta specs: [agent-core](openspec/changes/quantize-stage-work/specs/agent-core/spec.md), [run-observability](openspec/changes/quantize-stage-work/specs/run-observability/spec.md), [create-pipeline](openspec/changes/quantize-stage-work/specs/create-pipeline/spec.md).
- [SPEC.md](openspec/specs/SPEC.md) moved with it: the config example and prose in section 5, the spend-budget line in section 4.5, the two new exit paths in AC-8.5, and a new **AC-16.1 through AC-16.9** block for this change.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | pass (44 files, 614 tests, 20 skipped, 1 file skipped) |
| Markdown lint | `npm run lint:md` | pass (134 files, 0 issues) |
| Live-LLM (opt-in) | keyed on `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `COPPERHEAD_TEST_CODEX` | skip (not configured; the 19 tests in `agent-integration.test.ts` did not run) |

New tests, all offline:

- [test/turn-metrics.test.ts](test/turn-metrics.test.ts): the issue's exact 40-value array reproduces p50 177, max 43,629 and a top-5 share of 0.89; empty input zeroed; a zero-output run is 0%, not `NaN`; slowest turn from `ms`; edit-pressure arithmetic and cross-attempt folding; a historical-style transcript fixture ([test/fixtures/runs/](test/fixtures/runs/)) scored by `readRunTurnCost`.
- [test/quantize-stage-work.test.ts](test/quantize-stage-work.test.ts): an oversized KiCad edit is refused naming cap, size, setting and symbol count with the file byte-identical; a 20 kB docs edit is accepted; `maxEditBytes: 0` disables; a second unverified schematic edit is refused; `edit` then `run_erc` then `edit` passes; a **failing** `run_erc` still resets the counter and the repair edit lands; `maxUnverifiedEdits: 0` disables; a checkpoint commit lands with the right prefix, survives a later failure and does not sweep an unrelated dirty file; dry runs and `checkpointCommits: false` never checkpoint; `token-budget-exhausted` and `wall-budget-exhausted`; the `maxTurnOut` nudge reaches the model and the run still succeeds; `run-end` and `summary.md` carry the new figures on a failing run; the progress line names the missing refdes for both stages.
- [test/create-cost-report.test.ts](test/create-cost-report.test.ts): `report.json` and `REPORT.md` carry the concentration columns, resumed stages report `null` rather than a zeroed (and therefore falsely reassuring) concentration, `stageBudgets` reaches the stage run and junk entries are dropped, and the schematic prompt carries the progress line.

Three existing test files ([gating-sync](test/gating-sync.test.ts), [create-hardening](test/create-hardening.test.ts), [budget-efficiency](test/budget-efficiency.test.ts)) build a `RunContext` by hand and now spread `freshEditCounters()`; no test assertion was weakened and no default was loosened to make the suite pass. No pre-existing failures.

- Sandbox variant used: a scratch git repo seeded from the `open-key` fixture and `copperhead init`-ed, driven through `runAgentLoop` with a scripted provider (offline, so the mechanisms are exercised without spending a live stage). The repo carries an unrelated operator edit to `docs/DECISIONS.md` in the working tree throughout, and the run uses `--allow-dirty` as create stages do.
- Commands run and outcome: the historical fixture is scored from its transcript alone and reproduces 0.89; the oversized edit is refused; the second unverified edit is refused; a clean ERC checkpoints; the run then dies of turn exhaustion and the verified work survives on HEAD while the operator's edit stays dirty and uncommitted.

```text
=== 1. historical run scored from its transcript alone ===
{
  "cost": {
    "turns": 40,
    "p50TurnOut": 177,
    "p95TurnOut": 29672,
    "maxTurnOut": 43629,
    "top5TurnShare": 0.8901883922302831,
    "slowestTurnMs": null
  },
  "pressure": {
    "edits": 6,
    "editBytes": 52208,
    "largestEditBytes": 17884,
    "verifications": 2,
    "editBytesPerVerify": 26104
  }
}

=== 2. live run: oversized edit refused, verify gate, checkpoint ===
  copperhead v0.9.0 · kicad-cli 10.0.4 · node v24.18.0
  unknown · model gpt-5 (scripted, via unknown) · turns <=6 · master@3ae2347 dirty(1)
  [turn 1/6 · 0 in / 0 out]
    [propose_change] proposal written to openspec/changes/quantum/
    [validate_change] validation passed; edit tools are now unlocked
  [turn 2/6 · 100 in / 10 out]
    [edit_file] edit REFUSED: new_string is 17896 bytes, over the 8192-byte per-edit cap for KiCad files.
  [turn 3/6 · 200 in / 20 out]
    [edit_file] edited hardware/open-key.kicad_sch
    [edit_file] edit REFUSED: 1 edit(s) to the schematic have not been verified yet (limit 1).
  [turn 4/6 · 300 in / 30 out]
    [run_erc] ERC: clean
  checkpoint d28ac365cb (ERC clean, 1 file(s)); a later failure now rolls back to here, not to the start of the run
  [turn 5/6 · 400 in / 40 out]
  turn 5 emitted 43629 out tokens (> 8); asked the model to split the unit
    [read_file] 1: (kicad_sch
  run failed: turn budget exhausted (6 turns, 4 files touched but unverified)
  working tree restored to the last checkpoint (1 verified checkpoint commit(s) from this run are kept on HEAD)
  failed work preserved: git stash entry "copperhead failed run 2026-07-30T05-33-42-926Z"
  turn-budget-exhausted · ERC clean · 1s · 420 in / 87.3k out
  exitPath: turn-budget-exhausted
  turnCost: {"turns":6,"p50TurnOut":10,"p95TurnOut":43629,"maxTurnOut":43629,"top5TurnShare":0.9998854498384843,"slowestTurnMs":1}
  editPressure: {"edits":1,"editBytes":9,"largestEditBytes":9,"verifications":1,"editBytesPerVerify":9}

=== 3. git history after the run ===
  copperhead: checkpoint - add one part (ERC clean, 1 file(s))
  init
  files in the checkpoint: hardware/open-key.kicad_sch
  operator WIP still dirty: M .gitignore
 M docs/DECISIONS.md
?? openspec/
  rev in committed schematic: (rev "B")

=== 4. summary.md run stats ===
  - **Exit path:** turn-budget-exhausted
  - **Turn cost:** p50 10 · p95 43629 · max 43629 out tokens · top-5 share 100% · slowest turn 0s
  - **Edit pressure:** 1 edit(s), 1 verification(s), 9 B per verify · largest edit 9 B
```

The two things worth reading twice in that log: the checkpointed schematic edit (`rev "B"`) is on HEAD **after** the run failed and rolled back, and the operator's `docs/DECISIONS.md` edit is still dirty rather than swept into the checkpoint.

## Docs

- [configuration.md](docs/src/content/docs/reference/configuration.md): the four new keys in the example and the table, plus two new sections ("Spend budgets per create stage" and "Working in small verified units").
- [scaffold.ts](src/memory/scaffold.ts): the generated `.copperhead/README.md` describes `maxEditBytes`, `maxUnverifiedEdits`, `checkpointCommits` and `stageBudgets`.
- [SPEC.md](openspec/specs/SPEC.md) as listed above.

---

## Invariant checklist

- [x] **Spec-gated in**: unchanged. `edit_file`/`write_file` stay out of `availableTools` until `validate_change` unlocks them; the two new gates live inside `edit_file` and only ever refuse.
- [x] **Verification-gated out**: strengthened. A checkpoint is only ever taken after a clean ERC/DRC that follows an edit; failure still rolls back (now to the last verified point), and the repair path is explicitly preserved by resetting the unverified counter on a failing check.
- [x] **`check`/`verify` stays LLM-free and network-free**: nothing new is reachable from `src/commands/check`; `turn-metrics.ts` and `stage-progress.ts` are imported only by the loop and `create`.
- [x] **No sexp serialization**: read-only throughout. `stage-progress.ts` uses `listSymbols` and a regex over the board's reference properties; the symbol/footprint count in the refusal message is read-only too.
- [x] **Sync-obligations ledger**: unchanged. `markTouched` still feeds `onKicadEdit`/`onDocEdit` on every accepted mutation, and `finish` still refuses while obligations are open.
- [x] **Secrets**: unchanged. The new transcript events and summary lines carry only counts and paths, and go through the same redaction at write time.
- [x] **Spec coherence**: SPEC.md and all four change artifacts moved together; `openspec validate quantize-stage-work` passes.

## Summary by CodeRabbit

* **New Features**
  * Added per-stage token, wall-time, and turn-size budgets.
  * Added progress reporting so resumed stages identify completed and missing work.
  * Added cost and edit-pressure metrics to run reports.
  * Added checkpointing after successful ERC/DRC verification.
* **Bug Fixes**
  * KiCad edits are now limited in size and require verification before additional edits.
  * Budget exhaustion now reports distinct outcomes, while oversized turns prompt work to be split.
* **Documentation**
  * Expanded configuration guidance for edit limits, verification, checkpoints, and stage budgets.


animesh-chouhan:
# Automated pr-review pass: PR #151 — quantize stage work and measure turn-cost concentration

## Verdict: request changes

Two verified repo-invariant findings in the checkpoint-commit path. The rest of the PR is well built: the gates, budgets, metrics, and reporting match their spec and tests, and the description is accurate except for three claims noted below. No prior automated pr-review pass exists on this PR; this is a full report. The PR is agent-authored (per its own footer), so findings were verified with corresponding skepticism. CI: all 9 checks were passing at review time; mergeable, not a draft.

## Metrics

```text
lines: +2335 / -31 (net 2304) across 24 files  ·  src +774/-27, test +900/-3, docs +29/-0, spec +632/-1, other +0/-0
new vs net: 12 new files (2 in src), 12 modified; 293 new src lines (the net-new surface)
tests: +900 test lines; suite (CI green) -> 614/20/0 (pass/skip/fail, base -> head)
diff coverage: 94% of changed executable src lines exercised (421/448, measured via vitest v8); uncovered: src/agent/loop.ts:520-523, src/agent/tools.ts:114-116,139-142,145-147,457-458, src/agent/transcript.ts:80,86, src/agent/turn-metrics.ts:135-136,149-150, src/commands/stage-progress.ts:77-78,96,116, src/util/git.ts:383
deps: package.json unchanged
ci: 9 pass / 0 fail / 0 pending; required checks: none reported (gh pr checks)
```

Method notes: computed by `scripts/metrics.mjs` against merge base `c72b2da`. The head suite and coverage were measured with `COPPERHEAD_MIN_FREE_MB` set: the first script run reported 12 failures that all traced to this review machine's 99%-full disk tripping the repo's own disk-space preflight, not to the PR (all 12 pass with the documented override, and CI's `test` check is green). Base suite result taken from green CI on main, not re-run locally. `openspec validate quantize-stage-work` and `openspec validate build-copperhead-phase-1`: both valid.

## Findings

**F1 [high]** A checkpoint taken on a clean ERC commits `.kicad_pcb` edits that never passed DRC (or whose last DRC was failing), then moves the rollback target past them, so a later failure permanently keeps unverified board changes on HEAD, labeled "verified checkpoint commit(s)".

- Where: `src/agent/loop.ts:500-504` (`clean` checks only the check that just ran; `const files = [...ctx.filesTouched]` commits every touched file regardless of kind), misleading message at `src/agent/loop.ts:459-462`.
- Fails when: the agent edits `board.kicad_pcb` (setting `lastDrc = null`, or after a failing DRC), then calls `run_erc` which is clean: `maybeCheckpoint('run_erc')` commits the pcb file and re-snapshots; a later provider-error restores to that checkpoint. Symmetric case: `run_drc` clean checkpoints an ERC-unverified `.kicad_sch` edit. Violates SPEC §1.3 invariant 2 (verification-gated out); AC-16.7 as written codifies the behavior, so the spec must move with the fix.
- Fix: in `maybeCheckpoint`, require every touched kind's matching check to be currently clean before committing, mirroring the finish gate at `src/agent/tools.ts:710-715` (`no touched .kicad_sch or ctx.lastErc?.ok === true`, and same for pcb/`lastDrc`); skip the checkpoint otherwise, and update AC-16.7 plus the agent-core delta spec to state the both-kinds condition. Failing test: edit a pcb file, then a scripted clean `run_erc`; assert no checkpoint commit exists.

**F2 [high]** Checkpoint commits bypass the sync-obligations ledger entirely: they commit while drift/changelog/decision-log obligations are open, whereas the normal end-of-run commit is only reachable after `finish` blocks on the ledger.

- Where: `src/agent/loop.ts:498-524` (`maybeCheckpoint` -> `commitPaths`, no `ctx.ledger` consultation), vs the gated path at `src/agent/tools.ts:717-722`.
- Fails when: every checkpoint, by construction: `markTouched` -> `ledger.onKicadEdit` opens `drift` and `changelog` on the very edit that triggers the checkpoint, and `run_erc` clears only `erc` (`src/agent/ledger.ts:45-50`). If the run then fails, those commits stay on HEAD forever with no CHANGELOG entry and `check_drift` never run: exactly the doc/KiCad divergence the ledger exists to prevent, now committed. Neither AC-16.7, the agent-core delta spec, nor design.md D3 mentions obligations, so the code and the phase-1 D13 requirement ("The commit step SHALL refuse to run while any obligation is open") are silently inconsistent, and the PR body's "Sync-obligations ledger: unchanged" claim does not hold for this new commit path.
- Fix: either gate `maybeCheckpoint` on `ctx.ledger.openObligations` (excluding `erc`/`drc` just cleared and `changelog`, which the final commit path owns, so requiring at minimum a clean `drift` obligation), or amend SPEC §1.3/AC-16.7 and the delta spec to explicitly define checkpoint commits as provisional commits exempt from the ledger; today the exemption is undocumented.

**F3 [medium]** `commitPaths` stages only the run's touched paths but then commits the entire index, so an operator's pre-staged unrelated changes are swept into a copperhead checkpoint commit.

- Where: `src/util/git.ts:386-388` (`git diff --cached` check then bare `git commit -m`).
- Fails when: the operator runs `git add <their-file>` without committing, then a create stage runs under `--allow-dirty` and a checkpoint fires: the plain `git commit -m` commits everything staged, including their file; if the run later fails, that commit is kept on HEAD. This contradicts the delta spec's "Unrelated dirty work is not swept in" requirement (the existing test covers only an unstaged modified file) and the PR body's "does not sweep the operator's unrelated working changes" claim.
- Fix: commit by pathspec, `git commit -m <message> -- <paths>`, which commits only the named paths and leaves the rest of the index intact; add the failing test with a pre-staged bystander file.

**F4 [low]** The pcb/DRC half of the new mechanisms and every new error path are untested; the suite exercises only the sch/ERC side (this is the untested-surface list reconciling the uncovered lines above).

- Where and what each uncovered range contains: `src/agent/tools.ts:114-116` and `457-458` (the entire `.kicad_pcb` side of the verify gate and the `run_drc` counter reset; no test edits a board file through the gates), `src/agent/tools.ts:139-142,145-147` (`placedUnits` footprint count and its unparseable-file fallback), `src/agent/loop.ts:520-523` (the checkpoint-commit-failed warning path), `src/util/git.ts:383` (the `git add` miss-and-continue catch in `commitPaths`), `src/agent/turn-metrics.ts:135-136,149-150` (`readRunTurnCost` on an unreadable transcript and on malformed JSONL lines), `src/commands/stage-progress.ts:77-78,96,116` (the never-throws catch and both all-units-present lines), `src/agent/transcript.ts:80,86` (absent-`turnCost`/`editPressure` render arms).
- Fails when: a defect on the DRC side (the exact symmetric twin of the tested ERC paths, see F1) or in an error path ships unobserved.
- Fix: add a pcb-side gate/checkpoint test (edit `.kicad_pcb`, refused second edit, `run_drc` reset, DRC-clean checkpoint), a `readRunTurnCost` malformed-JSONL test, and a stage-progress all-present-line test; the remaining catches are acceptable to leave if noted.

**F5 [low]** Historical edit-pressure reconstruction counts refused and reverted edits, inconsistent with the delta spec's "A refused or reverted edit SHALL NOT be counted" and with the live counters it substitutes for.

- Where: `src/agent/turn-metrics.ts:156-163` (`readRunTurnCost` counts every `edit_file`/`write_file` tool event without inspecting the recorded result).
- Fails when: scoring a pre-counter transcript containing `edit REVERTED` (which existed before this PR) or, on any future transcript missing its `run-end` pressure, `edit REFUSED` events: `edits`/`editBytes`/`largestEditBytes` overcount work that never landed.
- Fix: skip tool events whose recorded result starts with `edit REFUSED` or `edit REVERTED` in the reconstruction path (one string check), or document the divergence in the function's JSDoc as intentional for emission-cost purposes.

**F6 [low]** `posInt`'s JSDoc says "A non-negative integer" but the implementation accepts only strictly positive values (which is the intended behavior per design D4).

- Where: `src/config.ts:118-121`.
- Fails when: a reader trusts the comment and expects `0` to survive `normalizeStageBudgets`.
- Fix: change the comment to "A positive integer".

## Invariant checks performed

1. Spec-gated in: **PASS.** `edit_file`/`write_file` still carry `requiresUnlock: true` and `availableTools` still filters on `ctx.editsUnlocked` (`src/agent/tools.ts:731-733`); the two new gates live inside `edit_file` and only ever refuse.
2. Verification-gated out: **FAIL** (F1). Sub-checks: ERC-only checkpoint can commit DRC-unverified board files (fail); checkpoints never set `finishRequest` and the finish gate is unchanged (pass); `commitPaths` -> `snapshot()` -> `restore` ordering traced correct, including the stash-on-failure and dirty-tree cases (pass); `maxRepairCycles` and budget exhaustion still roll back (pass).
3. `check`/`verify` LLM-free and network-free: **PASS.** The `check` import graph is touched only via `src/config.ts`, whose additions are pure; `turn-metrics.ts` and `stage-progress.ts` are imported only by the loop and `create`.
4. No sexp serialization: **PASS.** `src/kicad/` untouched; `stage-progress.ts` uses `listSymbols` plus regex reads only; the refusal-message symbol/footprint count is read-only.
5. Sync-obligations ledger: **FAIL** (F2) for the new checkpoint-commit path; pass for end-of-run commit gating, and the new refusal paths open no obligations for edits that never happened.
6. Secrets: **PASS.** New transcript events (`checkpoint`, `checkpoint-failed`, `turn-too-large`, extended `run-end`) and summary lines go through `redactSecrets` at write time (`src/agent/transcript.ts:110,152`); `.gitignore` untouched and still covers `.env` and `.copperhead/runs/`; the committed fixture transcript contains no `sk-` matches.

Spec coherence: SPEC.md, the three delta specs, design.md, and tasks.md moved together and both OpenSpec changes validate; AC-16.1 through AC-16.9 match the code and the fixture figures reproduce. The coherence gap is F1/F2: AC-16.7 and design D3 codify the checkpoint as implemented without reconciling it against SPEC §1.3 invariant 2 and the phase-1 ledger requirement.

Refuted candidates (investigated, dropped): `write_file` bypassing the byte cap on KiCad files (it refuses KiCad files outright, `src/agent/filetools.ts:27-29`); checkpoint sweeping unstaged operator dirt (staging is path-scoped and tested; only the staged-index case survives, as F3); rollback landing on the wrong snapshot after a checkpoint (traced `snapshot`/`restore`); the `maxWallMs: 1` test being clock-flaky (git preflight/snapshot guarantee >= 1 ms before the pre-turn check); config-default drift between code/SPEC/docs/scaffold; refusal messages feeding the ledger for unapplied edits; secrets in the committed fixture transcript.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is #151, branch issue-145-quantize-stages, based
on main.

Check it out first:
  gh pr checkout 151

FINDING 1 [high] A checkpoint taken on a clean ERC commits .kicad_pcb edits that
never passed DRC (or whose last DRC was failing), then moves the rollback target
past them, so a later failure permanently keeps unverified board changes on HEAD,
labeled "verified checkpoint commit(s)".
  Where: src/agent/loop.ts:500-504 (clean checks only the check that just ran;
    files = [...ctx.filesTouched] commits every touched file regardless of kind),
    misleading message at src/agent/loop.ts:459-462.
  Fails when: the agent edits board.kicad_pcb (setting lastDrc = null, or after a
    failing DRC), then calls run_erc which is clean: maybeCheckpoint('run_erc')
    commits the pcb file and re-snapshots; a later provider-error restores to that
    checkpoint. Symmetric case: run_drc clean checkpoints an ERC-unverified
    .kicad_sch edit. Violates SPEC section 1.3 invariant 2 (verification-gated
    out); AC-16.7 as written codifies the behavior, so the spec must move with
    the fix.
  Fix: in maybeCheckpoint, require every touched kind's matching check to be
    currently clean before committing, mirroring the finish gate at
    src/agent/tools.ts:710-715 (no touched .kicad_sch or ctx.lastErc?.ok === true,
    and the same for .kicad_pcb with lastDrc); skip the checkpoint otherwise, and
    update AC-16.7 plus the agent-core delta spec to state the both-kinds
    condition. Failing test: edit a pcb file, then a scripted clean run_erc;
    assert no checkpoint commit exists.

FINDING 2 [high] Checkpoint commits bypass the sync-obligations ledger entirely:
they commit while drift/changelog/decision-log obligations are open, whereas the
normal end-of-run commit is only reachable after finish blocks on the ledger.
  Where: src/agent/loop.ts:498-524 (maybeCheckpoint -> commitPaths, no ctx.ledger
    consultation), vs the gated path at src/agent/tools.ts:717-722.
  Fails when: every checkpoint, by construction: markTouched ->
    ledger.onKicadEdit opens drift and changelog on the very edit that triggers
    the checkpoint, and run_erc clears only erc (src/agent/ledger.ts:45-50). If
    the run then fails, those commits stay on HEAD forever with no CHANGELOG
    entry and check_drift never run. Neither AC-16.7, the agent-core delta spec,
    nor design.md D3 mentions obligations, so the code and the phase-1
    requirement "The commit step SHALL refuse to run while any obligation is
    open" are silently inconsistent.
  Fix: either gate maybeCheckpoint on ctx.ledger.openObligations (excluding
    erc/drc just cleared and changelog, which the final commit path owns, so
    requiring at minimum a clean drift obligation), or amend SPEC section
    1.3/AC-16.7 and the delta spec to explicitly define checkpoint commits as
    provisional commits exempt from the ledger; today the exemption is
    undocumented.

FINDING 3 [medium] commitPaths stages only the run's touched paths but then
commits the entire index, so an operator's pre-staged unrelated changes are swept
into a copperhead checkpoint commit.
  Where: src/util/git.ts:386-388 (git diff --cached check then bare git commit -m).
  Fails when: the operator runs git add <their-file> without committing, then a
    create stage runs under --allow-dirty and a checkpoint fires: the plain
    git commit -m commits everything staged, including their file; if the run
    later fails, that commit is kept on HEAD. This contradicts the delta spec's
    "Unrelated dirty work is not swept in" requirement (the existing test covers
    only an unstaged modified file).
  Fix: commit by pathspec, git commit -m <message> -- <paths>, which commits only
    the named paths and leaves the rest of the index intact; add the failing test
    with a pre-staged bystander file.

FINDING 4 [low] (optional) The pcb/DRC half of the new mechanisms and every new
error path are untested; the suite exercises only the sch/ERC side.
  Where: src/agent/tools.ts:114-116 and 457-458 (the .kicad_pcb side of the verify
    gate and the run_drc counter reset), src/agent/tools.ts:139-142,145-147
    (placedUnits footprint count and its unparseable-file fallback),
    src/agent/loop.ts:520-523 (the checkpoint-commit-failed warning path),
    src/util/git.ts:383 (the git add miss-and-continue catch in commitPaths),
    src/agent/turn-metrics.ts:135-136,149-150 (readRunTurnCost on an unreadable
    transcript and on malformed JSONL lines), src/commands/stage-progress.ts:
    77-78,96,116 (the never-throws catch and both all-units-present lines),
    src/agent/transcript.ts:80,86 (absent-turnCost/editPressure render arms).
  Fails when: a defect on the DRC side (the exact symmetric twin of the tested
    ERC paths, see finding 1) or in an error path ships unobserved.
  Fix: add a pcb-side gate/checkpoint test (edit .kicad_pcb, refused second edit,
    run_drc reset, DRC-clean checkpoint), a readRunTurnCost malformed-JSONL test,
    and a stage-progress all-present-line test; the remaining catches are
    acceptable to leave if noted.

FINDING 5 [low] (optional) Historical edit-pressure reconstruction counts refused
and reverted edits, inconsistent with the delta spec's "A refused or reverted
edit SHALL NOT be counted" and with the live counters it substitutes for.
  Where: src/agent/turn-metrics.ts:156-163 (readRunTurnCost counts every
    edit_file/write_file tool event without inspecting the recorded result).
  Fails when: scoring a pre-counter transcript containing "edit REVERTED" (which
    existed before this PR) or, on any future transcript missing its run-end
    pressure, "edit REFUSED" events: edits/editBytes/largestEditBytes overcount
    work that never landed.
  Fix: skip tool events whose recorded result starts with "edit REFUSED" or
    "edit REVERTED" in the reconstruction path (one string check), or document
    the divergence in the function's JSDoc as intentional for emission-cost
    purposes.

FINDING 6 [low] (optional) posInt's JSDoc says "A non-negative integer" but the
implementation accepts only strictly positive values (intended per design D4).
  Where: src/config.ts:118-121.
  Fails when: a reader trusts the comment and expects 0 to survive
    normalizeStageBudgets.
  Fix: change the comment to "A positive integer".

These lines in the diff are reached by no test: src/agent/loop.ts:520-523,
src/agent/tools.ts:114-116,139-142,145-147,457-458, src/agent/transcript.ts:80,86,
src/agent/turn-metrics.ts:135-136,149-150, src/commands/stage-progress.ts:77-78,
96,116, src/util/git.ts:383. Add tests for the new exported symbols, branches, and
error paths they contain (finding 4 lists what each range holds).

This PR changes spec-level behavior, so openspec/specs/SPEC.md and the active
change artifacts (proposal.md, design.md, delta specs, tasks.md) have to move
together, and `openspec validate quantize-stage-work` has to pass.

    never by prompt text.
    snapshot.
  - Sync-obligations ledger: post-tool-call hooks keep feeding it, and commit
    keeps refusing while any obligation is open.
  - Secrets: transcripts and summaries redact sk-[A-Za-z0-9_-]+ at write time,
    keys live only in env vars, and .gitignore keeps .env and .copperhead/runs/.

Then verify:
  npm run typecheck
  npm run build
  npm test

and report it.

Finally, summarize what you changed per finding, and note any you did not fix and
why. Keep prose em-dash-free (use colons, commas, or parentheses).
````

PR #152, opened 2026-07-30
feat(kicad): score layout quality against constraints and real boards

animesh-chouhan:
## What

`layout-draft` gets a measurement it cannot be talked out of. A new read-only `.kicad_pcb` reader, a deterministic layout scorer, a `layout_metrics` tool, and a generated `## Draft quality` section in `docs/LAYOUT.md`. The scorer is validated by a mutation suite that runs in CI and by the KiCad demo corpus that ships on the machine.

## Why

Addresses #141 (the measurement half; the placer and router are deferred, see [Design](#design)).

Stage 5's only check is `run_drc`, and DRC is a legality test. The current empty 30x20 mm scaffold passes it trivially, so **a board with nothing on it scores identically to a finished one**. A DRC-clean board can still have a bypass cap 20 mm from the pin it decouples, a return path with no copper behind it, and a sensor pointed at the load connector: every one of those is named by a key already in `.copperhead/constraints.json`, and none is a DRC rule. So the pipeline's acceptance criterion for layout was "the model says it's fine", which is what the stage prompt's "unlabeled non-optimal is not acceptable" clause was trying and failing to compensate for.

That is not a safe thing to rely on. Run `2026-07-29T19-54-42-021Z` refused the layout stage claiming 14 unconnected items in a turn where `run_erc` had returned "ERC: clean".

## Design

Three new modules, all read-only, all pure Node. Full rationale in [design.md](openspec/changes/score-layout-quality/design.md) decisions D9 to D11.

**[pcb.ts](src/kicad/pcb.ts)**: footprints, pads with absolute positions, courtyards from `F.CrtYd`/`B.CrtYd`, segments, vias, zones with filled area, and the Edge.Cuts extents. Both net encodings resolve to a name: KiCad 8/9 writes `(net 3 "GND")` on pads and bare `(net 3)` elsewhere against a top-level table, KiCad 10 writes `(net "GND")` everywhere. The footprint rotation sense was picked by measurement, not from memory: `x' = x·cos + y·sin`, `y' = -x·sin + y·cos` lands measurably more pads on real track endpoints than its mirror on every demo checked (`pic_programmer` 139 vs 120, `multichannel_mixer` 328 vs 287, `video` 1297 vs 1256).

**[layout-metrics.ts](src/kicad/layout-metrics.ts)** (D9): `computeLayoutMetrics(board, constraints)` is a pure function of the parsed board plus the registry. It consults neither the agent, nor the schematic, nor `kicad-cli`, precisely because of the refusal above: the scorer must not be able to be told anything. Purity is also what makes the demo corpus a valid calibration set.

- **Connectivity** is union-find over 1 µm-quantized points: a segment unions its endpoints per layer, a via unions across the layers it spans, a pad snaps to any track endpoint within `max(padExtent/2, 0.05 mm)`, and a zone on a net unions every pad on that net. Deliberately approximate and deliberately over-crediting, so it never flags a good board on connectivity it cannot see. The mutation that deletes the pour is what keeps that honest.
- **Hard rows** exist only when a constraint key matches by suffix (`*trace_width_mm`, `*board_outline_mm`, `*keepout`, `*mounting_hole_clearance_mm`, `*decoupling_distance_mm`, `*copper_area_mm2`). Every row names its key, bound, measured value and `pass` / `fail` / `n/a`. An unanswerable row is `n/a` with a reason and is excluded from the score, never counted as a pass. An empty registry gives an empty table, never invented rows.
- **The score** is 40 routed nets, 20 hard constraints, 12 courtyard clearance, 12 on-board placement, 8 density, 8 routing directness. Placement terms are **zeroed, not awarded**, when nothing is placed, so an empty board cannot bank points for the overlaps it does not have. That puts empty (20) below placed-but-unrouted (44) below routed (93.4).

**[layout-report.ts](src/kicad/layout-report.ts)** (D11): `## Draft quality` is generated and regenerated in place after the layout stage's completion contract is met and on every later resume. Everything below the notes marker survives, and a section that predates the marker has its whole body treated as notes, so the model's judgement is never destroyed.

**Deferred, with reasons.** The generation half of #141 is not here, and the blocker is dependency-shaped rather than effort-shaped: `sync_pcb_from_sch` needs the `pcbnew` Python bindings because `kicad-cli` 10.0.4 has no netlist-into-board import; the `placement.yaml` solver is blocked behind that import (nothing to place until the footprints exist); freerouting needs a JVM. None of those is a Copperhead dependency and none is present in CI. Scoring lands first because it is how the placer will be judged: without it, a placer change that quietly makes layouts worse looks exactly like a DRC-clean board nobody looks at. Also deferred: the per-stage place/route checkpointing (that belongs with the checkpoint work) and per-run corpus percentiles in `REPORT.md` (the baseline file exists for it).

**Invariants.** The s-expression layer stays read-only and nothing serializes a board, including the mutation suite, which uses anchored exact-match text replaces and asserts the fixture on disk is unchanged when it finishes. `layout_metrics` reads and reports, so it sits in the unlocked tier next to `run_erc`/`run_drc`. Nothing new is reachable from `check`.

## Spec / OpenSpec

OpenSpec change [`score-layout-quality`](openspec/changes/score-layout-quality/): [proposal.md](openspec/changes/score-layout-quality/proposal.md), [design.md](openspec/changes/score-layout-quality/design.md), [tasks.md](openspec/changes/score-layout-quality/tasks.md), delta spec [layout-quality/spec.md](openspec/changes/score-layout-quality/specs/layout-quality/spec.md).

`layout-quality` is a **new capability**. `openspec validate score-layout-quality` passes and all four artifacts are complete. SPEC.md is untouched on purpose: the tool-table and stage-5 entries merge on archive (task 6.1), following the `add-fab-release-gate` precedent.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | `npm run typecheck` | pass (clean) |
| Build | `npm run build` | pass (clean) |
| Unit + offline | `npm test` | pass: 617 passed, 20 skipped, 0 failed (48 files, 1 skipped) |
| Markdown lint | `npm run lint:md` | pass: 0 issues in 132 files |
| Live-LLM (opt-in) | keyed on `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `COPPERHEAD_TEST_CODEX` | skip (not configured; 20 skipped tests are these) |

New tests, 40 in total across six files:

- [layout-pcb.test.ts](test/layout-pcb.test.ts): both net encodings, identical geometry from both, footprint rotation, courtyards and outline, the empty open-key scaffold, the reference board, non-board rejection.
- [layout-metrics.test.ts](test/layout-metrics.test.ts): unrouted nets named, empty < placed-only < routed, the pour credited and then not, hard rows tracing to their key, an empty registry giving no rows, `n/a` rows, and the rendering plus the upsert preserving prose.
- [layout-mutation.test.ts](test/layout-mutation.test.ts): the four degradations, each asserting *which* metric moved and not only that the score fell.
- [layout-tool.test.ts](test/layout-tool.test.ts): available without an edit unlock, "does not apply yet" without a board, the full report, and a degradation on disk showing up in the very next call.
- [layout-draft-quality.test.ts](test/layout-draft-quality.test.ts): the create-pipeline regeneration hook.
- [layout-corpus.test.ts](test/layout-corpus.test.ts): guarded by `existsSync('/usr/share/kicad/demos')`, so it skips on a machine without the demos rather than failing. CI installs KiCad, so it runs there too.

No pre-existing failures; the suite was green before and after.

Sandbox variant used: hand-built `create`-shaped repo (board at `hardware/board.kicad_pcb`, a six-key `.copperhead/constraints.json`, a model-written `docs/LAYOUT.md`), plus the installed KiCad demos.

**1. The `layout_metrics` tool, from the locked tool list, against the built `dist/`:**

```text
$ node -e "…dispatchTool(ctx, 'layout_metrics', {})…"   # ctx.editsUnlocked = false
edit tools unlocked? false
layout_metrics present in the locked tool list? true
---
layout score: 93.4/100 for hardware/board.kicad_pcb
  Routed nets: 40/40 (4/4 multi-pad nets joined in copper)
  Hard constraints: 20/20 (6/6 constraint rows pass)
  Courtyard clearance: 12/12 (0 overlapping courtyard pair(s))
  On-board placement: 12/12 (0/8 footprint(s) outside the outline)
  Placement density: 1.72/8 (7.5% of the outline covered by courtyards)
  Routing directness: 7.67/8 (42.68 mm ideal vs 44.5 mm routed)
hard constraints:
  [pass] mech.board_outline_mm, Board outline: expected <= 60 x 40 mm, measured 60 x 40 mm
  [pass] mech.load_trace_width_mm, Narrowest current-carrying track: expected >= 1.5 mm, measured 1.5 mm (4 segment(s) on power nets)
  [pass] mech.mounting_hole_clearance_mm, Mounting hole copper clearance: expected >= 1.5 mm, measured 9.447 mm (nearest copper to any of 2 mounting hole(s))
  [pass] power.decoupling_distance_mm, Decoupling pad-to-pin distance: expected <= 5 mm, measured 4.23 mm (worst of 2 bypass cap(s): C1 to U1.1)
  [pass] sense.optical_feedback_keepout, Keepout separation J1 to R1: expected >= 20 mm, measured 30.648 mm (centroid to centroid)
  [pass] thermal.copper_area_mm2, Copper area on GND: expected >= 500 mm2, measured 2212.13 mm2 (tracks, pads, vias and filled zones on that net)
soft metrics:
  footprints=8 routed=4/4 track=44.5mm segments=8 vias=1 zones=1 courtyardOverlaps=0 offBoard=0 boardArea=2400mm2 density=7.5%
  unrouted nets: none
```

**2. Draft-quality regeneration, then a degradation on disk, then regeneration again:**

```text
$ node -e "…refreshDraftQuality(opts, 'layout-draft')…"
stage layout-draft: regenerated "## Draft quality" in docs/LAYOUT.md: score 93.4/100, 4/4 nets routed, 0 hard constraint(s) failing

$ sed -i 's/(width 1\.5)/(width 0.25)/g' hardware/board.kicad_pcb
$ node -e "…refreshDraftQuality(opts, 'layout-draft')…"
stage layout-draft: regenerated "## Draft quality" in docs/LAYOUT.md: score 90.1/100, 4/4 nets routed, 1 hard constraint(s) failing

$ grep -nE 'Layout score|load_trace_width|copperhead:notes|First pass' docs/LAYOUT.md
7:**Layout score: 90.1 / 100**
23:| `mech.load_trace_width_mm` | Narrowest current-carrying track (4 segment(s) on power nets) | >= 1.5 mm | 0.25 mm | fail |
46:<!-- copperhead:notes, everything below this line is preserved when the block above is regenerated -->
48:First pass. The LDR keepout is a guess and a human should redo it.
```

The model's sentence survived both regenerations; the computed block followed the board.

**3. Mutation numbers, measured on the committed reference board:**

| Board | Score | Routed | Hard rows failing |
| --- | ---: | :---: | --- |
| original | 93.4 | 4/4 | (none) |
| decoupling caps moved 15 mm | 79.8 | 3/4 | `power.decoupling_distance_mm` |
| power net narrowed to 0.25 mm | 90.1 | 4/4 | `mech.load_trace_width_mm` |
| placements shuffled in the outline | 49.1 | 1/4 | `power.decoupling_distance_mm`, `sense.optical_feedback_keepout` |
| ground zone deleted | 80.1 | 3/4 | `thermal.copper_area_mm2` |

Narrowing the power net moves nothing geometrically, so it is visible only through the hard table: that is the case that forces the hard term into the score at all.

**4. The labeled pair, scored by hand:**

```text
layout score: 98/100 for multichannel/multichannel_mixer.kicad_pcb
  Routed nets: 38.99/40 (77/79 multi-pad nets joined in copper)
  On-board placement: 12/12 (0/114 footprint(s) outside the outline)
  Routing directness: 7.01/8 (1424.13 mm ideal vs 1624.17 mm routed)

layout score: 66.6/100 for multichannel/multichannel_mixer-unrouted.kicad_pcb
  Routed nets: 9.62/40 (19/79 multi-pad nets joined in copper)
  On-board placement: 10/12 (19/114 footprint(s) outside the outline)
  Routing directness: 7/8 (342.3 mm ideal vs 391.44 mm routed)

layout score: 95.5/100 for pic_programmer/pic_programmer.kicad_pcb
  Routed nets: 38.82/40 (33/34 multi-pad nets joined in copper)
  Courtyard clearance: 11.81/12 (1 overlapping courtyard pair(s))
  Routing directness: 6.37/8 (1157.9 mm ideal vs 1453.18 mm routed)
```

Same design, same 114 footprints: 98 routed vs 66.6 unrouted. That is the free ground truth the issue asked for, and the scorer gets it right.

**5. The corpus baseline, committed as [layout-corpus-baseline.json](test/fixtures/layout-corpus-baseline.json):**

```text
$ npm run layout:corpus
scored 19 board(s) from /usr/share/kicad/demos
   86.9    112 fp    2084 seg  cm5_minima/CM5_MINIMA_3.kicad_pcb
   91.8     68 fp     364 seg  complex_hierarchy/complex_hierarchy.kicad_pcb
   86.8     15 fp      59 seg  ecc83/ecc83-pp.kicad_pcb
   96.9     15 fp      53 seg  ecc83/ecc83-pp_v2.kicad_pcb
   97.8     25 fp     731 seg  interf_u/interf_u.kicad_pcb
   83.8   1125 fp   22151 seg  jetson-agx-thor-baseboard/jetson-agx-thor-baseboard.kicad_pcb
   86.3    160 fp    2935 seg  kit-dev-coldfire-xilinx_5213/kit-dev-coldfire-xilinx_5213.kicad_pcb
     44      4 fp       0 seg  microwave/microwave.kicad_pcb
   66.6    114 fp     140 seg  multichannel/multichannel_mixer-unrouted.kicad_pcb
     98    114 fp     576 seg  multichannel/multichannel_mixer.kicad_pcb
   85.7    210 fp    1654 seg  openair-max/One-Air-Max.kicad_pcb
   95.5     63 fp     370 seg  pic_programmer/pic_programmer.kicad_pcb
   91.5     71 fp     943 seg  royalblue54L_feather/RoyalBlue54L-Feather.kicad_pcb
     44      2 fp     160 seg  royalblue54L_feather/nfc_antenna/RoyalBlue54L-NFC-Antenna/RoyalBlue54L-NFC-Antenna.kicad_pcb
   95.3     25 fp     208 seg  sonde xilinx/sonde xilinx.kicad_pcb
   84.9     94 fp    1293 seg  stickhub/StickHub.kicad_pcb
   87.9    309 fp    2143 seg  tiny_tapeout/tinytapeout-demo.kicad_pcb
   93.5    189 fp    7932 seg  video/video.kicad_pcb
   78.9   1508 fp   26650 seg  vme-wren/vme-wren.kicad_pcb
distribution: min 44, p25 83.8, median 86.9, p75 95.3, max 98
wrote test/fixtures/layout-corpus-baseline.json
```

19 boards (the 17 projects plus the two extra boards in `ecc83/` and `royalblue54L_feather/nfc_antenna/`), 11.8 s, none throwing. Two of them are 67 MB and 81 MB and parse in about 3 s each.

One finding worth naming: `royalblue54L_feather/RoyalBlue54L-Feather.kicad_pcb` as shipped has a stray close-paren (`(curved_edges no)filter_ratio 0.9)`) that ends the outer `(kicad_pcb …)` form early and spills two thirds of the board out as sibling top-level forms. The reader first reported 2 footprints for a 3.4 MB board. It now adopts stray top-level forms as board children and reads all 71. Being generous costs nothing in a module that never writes.

**6. Mutation suite, run as a test:**

```text
$ npx vitest run test/layout-mutation.test.ts --reporter=verbose
 ✓ layout scorer mutation suite > scores the undegraded reference board with every hard constraint passing
 ✓ layout scorer mutation suite > scores strictly lower when every decoupling cap moves 15 mm from the part it bypasses
 ✓ layout scorer mutation suite > scores strictly lower when the power net is narrowed to 0.25 mm
 ✓ layout scorer mutation suite > scores strictly lower when placements are shuffled within the outline
 ✓ layout scorer mutation suite > scores strictly lower when the ground zone is deleted
 ✓ layout scorer mutation suite > leaves the fixture on disk untouched
 Test Files  1 passed (1)
      Tests  6 passed (6)
```

## Docs

- `docs/LAYOUT.md` in every `create` repo now carries a generated `## Draft quality` block with a self-describing notes marker.
- Stage-5 prompt in [create.ts](src/commands/create.ts) tells the model to run `layout_metrics` after placing and after routing, and to write its judgement below the marker.
- New `npm run layout:corpus` script, documented in the header of [layout-corpus-baseline.ts](scripts/layout-corpus-baseline.ts).
- SPEC.md unchanged by design; the merge is task 6.1 of the change, on archive.

---

## Invariant checklist

- [x] **Spec-gated in**: unchanged. `layout_metrics` is a read-and-report tool in the unlocked tier next to `run_erc`/`run_drc`; `edit_file`/`write_file` gating is untouched, and a test asserts `layout_metrics` is present while `ctx.editsUnlocked` is false.
- [x] **Verification-gated out**: unchanged. Nothing here mutates a KiCad file, so no gate moved. `layout_metrics` is additional evidence, never a substitute for DRC.
- [x] **`check`/`verify` stays LLM-free and network-free**: unchanged, and nothing new is reachable from `src/commands/check`. The scorer itself makes no subprocess and no network call either.
- [x] **No sexp serialization**: [pcb.ts](src/kicad/pcb.ts) is read-only and has no writer. The mutation suite degrades boards by anchored exact-match text replace and asserts the fixture on disk is byte-identical afterwards.
- [x] **Sync-obligations ledger**: unchanged. `layout_metrics` opens and clears nothing; the `## Draft quality` regeneration runs in the pipeline after the stage's own commit, not inside the loop.
- [x] **Secrets**: unchanged. No new transcript content, no new env var, no new ignore rule needed.
- [x] **Spec coherence**: `openspec validate score-layout-quality` passes with all four artifacts complete; `layout-quality` is declared as a new capability in the proposal. SPEC.md merges on archive per task 6.1.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic KiCad layout-quality scoring from 0–100 with hard- and soft-constraint reporting.
  * Added layout metrics to the workflow and automatic updates to the “Draft quality” section.
  * Added layout demo corpus baseline generation and comparison.
* **Bug Fixes**
  * Prevented populated boards from increasing hard-term scores when constraints become unanswerable.
* **Documentation**
  * Updated layout quality rules and added a reference page.
* **Tests**
  * Expanded coverage for parsing, scoring, reporting, mutations, tools, and demo boards.


animesh-chouhan:
# Automated pr-review pass: PR #152 — score layout quality against constraints and real boards

**Verdict: request changes** (two correctness findings on the new scorer plus a false CI-verification claim; all are small, targeted fixes to an otherwise excellent, well-tested PR)

No earlier automated pr-review pass exists on this PR (only the pr-lint bot and CodeRabbit), so this is a full report. Reviewer authored none of the commits under review.

## Metrics

```text
lines: +3069 / -2 (net 3067) across 20 files  ·  src +1392/-2, test +1254/-0, docs +0/-0, spec +296/-0, other +127/-0
new vs net: 16 new files (3 in src), 4 modified; 1324 new src lines (the net-new surface)
tests: +1254 test lines; suite: 617/20/0 (pass/skip/fail at head)
diff coverage: 97.8% of changed executable src lines exercised (993/1015, measured via vitest v8); uncovered: src/commands/create.ts:201,303-304, src/kicad/layout-metrics.ts:476-480,483-484,489-490,545-553,586
deps: +[none] -[none] ~[none] (package.json changed: one npm script only, no dependency changes)
ci: 9 pass / 0 fail / 0 pending; required checks: none reported (gh pr checks)
```

Method notes: diff metrics, suite, and coverage from `scripts/metrics.mjs` run at the PR head commit (5e5871d, detached checkout in the review worktree, `--base main` mode against merge base c72b2da). The CI line is from the script's PR-mode run (`gh pr checks`). The head suite initially reported 7 failures on this machine; all seven are environmental, not PR-caused: this review machine has 1.6 GiB free disk and copperhead's pre-existing 2 GiB disk-space preflight fires before the assertions those create-pipeline tests target. With `COPPERHEAD_MIN_FREE_MB=200` the suite is fully green at head (617/20/0), matching the PR's claimed numbers. Base suite: not measured locally (not run to save the constrained disk; head suite green and CI green stand in). CI state at review time: all 9 checks passing, mergeable, not a draft.

One reconciliation the CI line hides: the CI `test` job itself ran **615 passed / 22 skipped**, not the 617/20 the PR reports locally, because both corpus tests skipped in CI. See finding 3.

## Findings

**1. [medium] KiCad 10 boards resolve constraint-named nets to nothing, silently degrading `*copper_area_mm2` (and net-named `*trace_width_mm`) hard rows to n/a.**
- Where: `src/kicad/pcb.ts:444` (`netNames: nets.names()`), consumed by `referencedNet` at `src/kicad/layout-metrics.ts:510-516`
- Failure scenario: `Board.netNames` is built only from the top-level net table, which KiCad 10 populates with net 0 alone (the PR's own module comment and test fixture say so). `referencedNet` matches constraint text against `board.netNames`, so on any KiCad 10-encoded board it returns null even when the named net exists on every pad. Verified empirically: the same one-resistor-plus-GND-zone board with `thermal.copper_area_mm2 {min, value: "GND"}` yields `[pass] ... measured 602 mm²` under the KiCad 8 encoding and `n/a: "the constraint names no net present on this board"` under the KiCad 10 encoding. The row is then excluded from the score, so the exact hard gate this PR exists to enforce vanishes on the file format the PR claims to handle (and the format KiCad 10, the version CI installs, writes). Tests never catch it because the reference fixture and all hard-table tests use the KiCad 8/9 net-table encoding, and the corpus runs with an empty registry.
- Fix: populate `netNames` from the union of the net-table names and every net name resolved on pads, segments, vias, and zones during parse. Add a KiCad 10 assertion to `test/layout-pcb.test.ts` (`b.netNames` contains GND/VCC) and a `test/layout-metrics.test.ts` case where a `copper_area_mm2` row passes against the KiCad 10 encoding.

**2. [medium] Deleting the component a hard constraint polices scores higher than misplacing it: a fail row that becomes n/a shrinks the denominator and raises the hard term.**
- Where: `src/kicad/layout-metrics.ts:413-415` (`decided`/`hardFraction` in `scoreTerms`), interacting with `decouplingRow`'s n/a path at line 642
- Failure scenario: verified on the committed reference board with the committed registry: original scores 93.4 (decoupling pass, hard 20/20); the mutation suite's caps-moved-15mm mutant scores 79.8 (decoupling **fail**, hard 16.67/20); deleting C1 and C2 entirely scores **91.9** (decoupling **n/a**, hard back to 20/20, routed back to 40/40). A board with no decoupling at all outscores a board with badly placed decoupling by 12 points. The delta spec's own scenario says an unanswerable row "does not raise the score", but the fail-to-n/a transition does exactly that relative to the failing state, and the scorer's stated purpose is to be un-talk-out-of-able during the stage where the model edits the board. The same asymmetry applies to `keepoutRow` (delete one named refdes) and `mountingHoleRow` (delete the holes). Relatedly, a board with zero decided rows banks the full 20 points (`hardFraction = 1`).
- Fix: first add the regression test (deleting the bypass caps must not outscore moving them 15 mm), then make the hard term robust to subject deletion on a non-empty board, for example scoring `hardPass / matchedRows` (n/a rows count against the term) whenever `soft.footprints > 0`, keeping the current behavior for the empty scaffold. Update the delta spec's n/a scenario and design.md D9 in the same commit so spec and code move together.

**3. [medium] The corpus calibration suite never runs in CI; the PR description's claim "CI installs KiCad, so it runs there too" is false.**
- Where: `.github/workflows/ci.yml:27` (`sudo apt-get install --yes --no-install-recommends kicad`) vs `test/layout-corpus.test.ts:21`
- Failure scenario: `--no-install-recommends` excludes `kicad-demos` (the CI log for this PR's test job lists `kicad-demos` under "Recommended packages", i.e. not installed), so `/usr/share/kicad/demos` is absent in CI and both corpus tests skipped there: the job ran 615 passed / 22 skipped, against the PR's local 617/20. The skip-when-absent behavior is exactly what the delta spec requires, so the code is correct; but the tier-two calibration (including the labeled routed/unrouted pair, the only free ground truth) is silently not exercised by CI, contrary to the PR description and the design's stated validation story.
- Fix: add `kicad-demos` to the CI install line (`sudo apt-get install --yes --no-install-recommends kicad kicad-demos`), or amend the PR description and design.md D10 to state plainly that the corpus tier runs only on developer machines.

**4. [low] `upsertDraftQuality` collapses every run of 2+ consecutive blank lines in the whole of `docs/LAYOUT.md`, including user prose and fenced code blocks outside the generated section.**
- Where: `src/kicad/layout-report.ts:145` (`.replace(/\n{3,}/g, '\n\n')` applied to the joined full document)
- Failure scenario: a fenced code block anywhere in LAYOUT.md (other sections, or the model's notes below the marker) that contains two consecutive blank lines has them collapsed to one on every regeneration; deliberate multi-blank-line spacing anywhere in the doc is likewise rewritten. This chips at the "the model's judgement is never destroyed / sections either side are untouched" contract.
- Fix: apply the blank-line normalization only to the generated `section` lines and the seams (`before`/section and section/`tail` joins), not to the whole joined document.

**5. [low] Untested surface: five uncovered branches in the new hard-table code and one error path in the pipeline hook.** (This is the reconciliation of the coverage block's uncovered-lines list.)
- `src/kicad/layout-metrics.ts:476-480`: `bound()` with both min and max (a range constraint) is never exercised.
- `src/kicad/layout-metrics.ts:483-484` and `489-490`: a matched constraint with neither min nor max reaching `decide()`'s "constraint records no min/max" n/a row is never exercised.
- `src/kicad/layout-metrics.ts:545-553`: `boardOutlineRow`'s max-only branch ("largest dimension <= N mm") is never exercised (only the "WxH" value-string branch is).
- `src/kicad/layout-metrics.ts:586`: `mountingHoleRow`'s padless-hole fallback anchor (a mounting-hole footprint with no pads) is never exercised.
- `src/commands/create.ts:303-304`: `refreshDraftQuality`'s catch/warn path (unreadable board or LAYOUT.md write failure must warn, never fail the pipeline) is never exercised.
- `src/commands/create.ts:201` is also uncovered, but it is the stage-5 prompt string constant, exercised only by live runs; no test needed.
- Fix: one small test each; all are pure-function or injected-fs cases.

Positive observations, for the record: the mutation suite asserts *which* metric moved, not just that the score fell; the tool is exercised through the real `dispatchTool` with real files rather than mocks; `dispatchTool` converts a malformed or missing board into an `error:` string rather than a crash; the truncated-board recovery (stray top-level forms adopted as children) is a real-input fix with a named in-the-wild specimen; and the adversarial encodings traced by hand (bare `(net N)`, `(net "name")`, `net_name` override, `*.Cu`/`F&B.Cu` expansion, `-0` folding, arc chords, footprint-carried edge cuts) all behave as documented apart from finding 1.

## Invariant checks performed

- **Spec-gated in: PASS.** `availableTools`/`requiresUnlock` filtering is unchanged; `layout_metrics` is `requiresUnlock: false` and purely read-and-report (reads the board file and the constraint registry, writes nothing); `edit_file`/`write_file` gating untouched; `test/layout-tool.test.ts` asserts the tool is present while `editsUnlocked` is false.
- **Verification-gated out: PASS.** No new mutation path; ERC/DRC gating and `maxRepairCycles`/rollback untouched. `refreshDraftQuality` writes only the markdown memory artifact, runs only after the stage's completion contract is met (or on resume before `commitResumedStage`, which commits it), and is best-effort (logged warn, never fatal).
- **check/verify LLM-free and network-free: PASS.** `src/commands/check.ts` is untouched and imports none of the three new modules; the scorer makes no subprocess, network, or provider call (verified by import trace: `pcb.ts`/`layout-metrics.ts`/`layout-report.ts` are imported only by `tools.ts` and `create.ts`).
- **No sexp serialization: PASS.** `pcb.ts` has no writer; nothing round-trips a board. The mutation suite degrades in-memory copies via anchored exact-match text replaces and its final test asserts the on-disk fixture is byte-identical.
- **Sync-obligations ledger: PASS.** All `ledger.add`/`ledger.clear` sites and the commit gate are untouched; `layout_metrics` opens and clears nothing; `loop.ts` is not in the diff.
- **Secrets: PASS.** No new env vars, no transcript-writer changes, `.gitignore` untouched.
- **Spec coherence: PASS** (with finding 3's description caveat). `openspec validate score-layout-quality` passes; proposal/design/tasks/delta spec are complete and mutually consistent; the placer/router deferral is stated in proposal.md ("Explicitly deferred") and tasks.md section 5 with dependency-shaped reasons; SPEC.md untouched with the merge assigned to archive task 6.1, matching the `add-fab-release-gate` precedent.

CI state at review time: **all 9 checks passing** (Analyze x2, CodeQL, CodeRabbit, cla, greeting, label, template, test), mergeable, not a draft.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is #152, branch issue-141-layout-quality, based
on main.

Check it out first:
  gh pr checkout 152

FINDING 1 [medium] KiCad 10 boards resolve constraint-named nets to nothing,
silently degrading *copper_area_mm2 (and net-named *trace_width_mm) hard rows to n/a.
  Where: src/kicad/pcb.ts:444 (netNames: nets.names()), consumed by referencedNet
    at src/kicad/layout-metrics.ts:510-516
  Fails when: Board.netNames is built only from the top-level net table, which
    KiCad 10 populates with net 0 alone. referencedNet matches constraint text
    against board.netNames, so on any KiCad 10-encoded board it returns null even
    when the named net exists on every pad. Verified: the same board with
    thermal.copper_area_mm2 {min, value: "GND"} yields a pass row (602 mm2
    measured) under the KiCad 8 encoding and an n/a row ("the constraint names no
    net present on this board") under the KiCad 10 encoding. The row is then
    excluded from the score, so the hard gate vanishes on KiCad 10 files.
  Fix: populate netNames from the union of the net-table names and every net name
    resolved on pads, segments, vias, and zones during parse. Add a KiCad 10
    assertion to test/layout-pcb.test.ts (b.netNames contains GND/VCC) and a
    test/layout-metrics.test.ts case where a copper_area_mm2 row passes against
    the KiCad 10 encoding.

FINDING 2 [medium] Deleting the component a hard constraint polices scores higher
than misplacing it: a fail row that becomes n/a shrinks the denominator and raises
the hard term.
  Where: src/kicad/layout-metrics.ts:413-415 (decided/hardFraction in scoreTerms),
    interacting with decouplingRow's n/a path at line 642
  Fails when: measured on the committed reference board with the committed
    registry: original 93.4 (decoupling pass, hard 20/20); caps moved 15 mm 79.8
    (decoupling fail, hard 16.67/20); caps DELETED entirely 91.9 (decoupling n/a,
    hard back to 20/20, routed back to 40/40). A board with no decoupling at all
    outscores a board with badly placed decoupling by 12 points. The same
    asymmetry applies to keepoutRow (delete one named refdes) and mountingHoleRow
    (delete the holes). A board with zero decided rows also banks the full 20.
  Fix: first add the regression test (deleting the bypass caps must not outscore
    moving them 15 mm), then make the hard term robust to subject deletion on a
    non-empty board, for example scoring hardPass / matchedRows (n/a rows count
    against the term) whenever soft.footprints > 0, keeping current behavior for
    the empty scaffold. Update the delta spec's n/a scenario
    (openspec/changes/score-layout-quality/specs/layout-quality/spec.md) and
    design.md D9 in the same commit so spec and code move together.

FINDING 3 [medium] The corpus calibration suite never runs in CI; the PR
description's claim "CI installs KiCad, so it runs there too" is false.
  Where: .github/workflows/ci.yml:27 (sudo apt-get install --yes
    --no-install-recommends kicad) vs test/layout-corpus.test.ts:21
  Fails when: --no-install-recommends excludes kicad-demos (this PR's own CI log
    lists kicad-demos under "Recommended packages", i.e. not installed), so
    /usr/share/kicad/demos is absent in CI and both corpus tests skipped there:
    the CI job ran 615 passed / 22 skipped against the local 617/20. The
    routed/unrouted labeled pair, the calibration ground truth, is never checked
    by CI.
  Fix: add kicad-demos to the CI install line (sudo apt-get install --yes
    --no-install-recommends kicad kicad-demos), or amend the PR description and
    design.md D10 to state plainly that the corpus tier runs only on developer
    machines.

FINDING 4 [low] (optional) upsertDraftQuality collapses every run of 2+
consecutive blank lines in the whole of docs/LAYOUT.md, including user prose and
fenced code blocks outside the generated section.
  Where: src/kicad/layout-report.ts:145 (.replace(/\n{3,}/g, '\n\n') applied to
    the joined full document)
  Fails when: a fenced code block anywhere in LAYOUT.md (other sections, or the
    model's notes below the marker) contains two consecutive blank lines: they
    collapse to one on every regeneration, and deliberate multi-blank-line
    spacing anywhere in the doc is rewritten, violating the "sections either side
    are untouched" contract.
  Fix: apply the blank-line normalization only to the generated section lines and
    the seams (before/section and section/tail joins), not to the whole document.

FINDING 5 [low] (optional) Untested surface: five uncovered branches in the new
hard-table code and one error path in the pipeline hook.
  Where: src/kicad/layout-metrics.ts:476-480 (bound() with both min and max),
    483-484 and 489-490 (matched constraint with no min/max reaching decide()'s
    n/a row), 545-553 (boardOutlineRow max-only "largest dimension" branch),
    586 (mountingHoleRow padless-hole fallback anchor);
    src/commands/create.ts:303-304 (refreshDraftQuality catch/warn path)
  Fails when: a regression in any of these branches (range constraints, boundless
    constraints, max-only outlines, padless mounting holes, unreadable board
    during regeneration) would ship unnoticed; no test reaches them.
  Fix: one small test each; all are pure-function or injected-fs cases.
    src/commands/create.ts:201 (the stage-5 prompt string) is also uncovered but
    is a prompt constant needing no test.

These lines in the diff are reached by no test: src/commands/create.ts:201,303-304,
src/kicad/layout-metrics.ts:476-480,483-484,489-490,545-553,586. Add tests for the
branches and error paths they contain (enumerated in FINDING 5; line 201 excepted).

This PR changes spec-level behavior, so the active change artifacts
(openspec/changes/score-layout-quality/: proposal.md, design.md, delta specs,
tasks.md) have to move together with the code, and
`openspec validate score-layout-quality` has to pass. SPEC.md itself merges on
archive per task 6.1; do not edit it now.

    by prompt text.

Then verify:
  npm run typecheck
  npm run build
  npm test

and report it.

````

PR #153, opened 2026-07-30
fix(sync): parse PINOUT.md structurally for pins.h check (#105)

### LINKED ISSUE #105: sync's pins.h check can never fail: it substring-matches PINOUT.md, so any word anywhere satisfies any #define

Chirag6722:
## What

Updates `pins-h` check in `src/commands/sync.ts` to parse `PINOUT.md` structurally via `parseCanonicalTables` instead of raw substring matching.

## Why

Addresses #105. Previously, `sync` checked `#define PIN_<NAME>` by raw `pinout.includes(name)`. Any word appearing anywhere in `PINOUT.md` (e.g. `KEY` inside `KEY_DAH`, `GPIO` inside `GPIO14`, or `RTC` inside prose text) would falsely pass non-existent pin definitions.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npx vitest run test/gating-sync.test.ts` | pass (17/17) |

### Manual test log

Added regression test `pins-h check verifies structurally against PINOUT.md table, rejecting substring matches (#105)` in `test/gating-sync.test.ts`. Confirmed 3 unassigned `#define`s (`PIN_KEY`, `PIN_GPIO`, `PIN_3V`) produce 3 resolvable inconsistencies instead of false-pass.

## Docs

No documentation changes required; internal sync check fix.

---

## Invariant checklist

- [x] **Spec-gated in**. Unaffected.
- [x] **Verification-gated out**. Unaffected.
- [x] **`check`/`verify` stays LLM-free and network-free**. `copperhead sync` verify phase stays 100% LLM-free and offline.
- [x] **No sexp serialization**. Unaffected.
- [x] **Sync-obligations ledger**. Fixes `pins-h` sync check false negative.
- [x] **Secrets**. Unaffected.
- [x] **Spec coherence**. Unaffected.

Closes #105.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization validation between firmware `PIN_*` definitions and the canonical pin-assignment tables in `PINOUT.md`.
  * Pin matching is now more robust, correctly handling capitalization, spacing, and formatting differences across table entries.
  * Updated feedback to state when a pin is not assigned in `PINOUT.md` (instead of being “not mentioned”).

* **Tests**
  * Added sync verification coverage for multiple firmware pin definitions.
  * Added a test ensuring pin validation ignores `Notes` text in `PINOUT.md` and still resolves the correct pin entry.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -6,6 +6,7 @@ import { checkDrift } from '../memory/drift.js';
 import { loadConstraints, checkForbiddenPins } from '../memory/constraints.js';
 import { pinNets } from '../kicad/sexp.js';
 import { openspecValidate } from '../openspec/cli.js';
+import { parseCanonicalTables } from '../memory/bom-table.js';
 import { runAgentLoop } from '../agent/loop.js';
 import type { RunMetaInput } from '../agent/runmeta.js';
 import type { ProgressRenderer } from '../agent/render.js';
@@ -90,13 +91,23 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
   if (existsSync(pinsH) && existsSync(path.join(docsDir, 'PINOUT.md'))) {
     const header = await readFile(pinsH, 'utf8');
     const pinout = await readFile(path.join(docsDir, 'PINOUT.md'), 'utf8');
+    const validNames = new Set<string>();
+    for (const { rows: tRows } of parseCanonicalTables(pinout)) {
+      for (const r of tRows) {
+        for (const cell of r.cells) {
+          const val = cell.replace(/`/g, '').trim();
+          if (val) validNames.add(val.toLowerCase());
+        }
+      }
+    }
     for (const m of header.matchAll(/#define\s+PIN_(\w+)\s+(\S+)/g)) {
-      if (!pinout.includes(m[1]!)) {
+      const pinName = m[1]!;
+      if (!validNames.has(pinName.toLowerCase())) {
         resolvable.push({
           kind: 'pins-h',
           doc: 'firmware/src/pins.h',
-          claim: `defines PIN_${m[1]}`,
-          actual: 'PINOUT.md does not mention it',
+          claim: `defines PIN_${pinName}`,
+          actual: 'PINOUT.md does not assign it',
           resolution: 'regenerate pins.h from PINOUT.md (PINOUT.md is the single source of truth)',
         });
       }
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -92,11 +92,20 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
     const header = await readFile(pinsH, 'utf8');
     const pinout = await readFile(path.join(docsDir, 'PINOUT.md'), 'utf8');
     const validNames = new Set<string>();
-    for (const { rows: tRows } of parseCanonicalTables(pinout)) {
+    const strip = (s: string | undefined): string => (s ?? '').replace(/`/g, '').trim();
+    for (const { header: tHeader, rows: tRows } of parseCanonicalTables(pinout)) {
+      const col = (re: RegExp): number => tHeader.cells.findIndex((c) => re.test(c));
+      const netI = col(/^net$/i);
+      const nameI = col(/^name$/i);
+      const signalI = col(/^signal$/i);
+      const pinI = col(/^pin$/i);
+      if (pinI < 0 || (netI < 0 && nameI < 0 && signalI < 0)) continue; // only read canonical pin-assignment tables
       for (const r of tRows) {
-        for (const cell of r.cells) {
-          const val = cell.replace(/`/g, '').trim();
-          if (val) validNames.add(val.toLowerCase());
+        for (const idx of [netI, nameI, signalI]) {
+          if (idx >= 0 && r.cells[idx]) {
+            const val = strip(r.cells[idx]);
+            if (val) validNames.add(val.toLowerCase());
+          }
         }
       }
     }
--- a/test/gating-sync.test.ts
+++ b/test/gating-sync.test.ts
@@ -409,4 +409,30 @@ describe('copperhead sync verify phase (AC-7)', () => {
       await cleanup();
     }
   });
+
+  it('pins-h check ignores Notes column text when verifying pin names (#105)', async () => {
+    const { repo, cleanup } = await tempFixtureRepo();
+    try {
+      await runInit({ repoRoot: repo });
+      const pinoutPath = path.join(repo, 'docs', 'PINOUT.md');
+      await writeFile(
+        pinoutPath,
+        '| Refdes | Pin | Net | Notes |\n|---|---|---|---|\n| U1 | 5 | KEY_DAH | KEY notes here |\n',
+        'utf8',
+      );
+      const firmwareDir = path.join(repo, 'firmware', 'src');
+      await mkdir(firmwareDir, { recursive: true });
+      await writeFile(
+        path.join(firmwareDir, 'pins.h'),
+        '#define PIN_KEY 5\n',
+        'utf8',
+      );
+      const report = await syncVerify(repo);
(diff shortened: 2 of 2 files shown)


Devesh36:

## Verdict: approve with nits

Replacing `pinout.includes(pinName)` with structural table-cell matching fixes the real #105 substring false-negative/positive class. The regression test shows orphan `PIN_*` defines are reported. One precision nit: every cell of every canonical table is accepted as a valid pin name.

### Metrics
```text
lines: +39 / -4 across 2 files  ·  src/commands/sync.ts + tests
tests: +1 pins-h structural case
ci: see PR checks (suite not re-run in this parallel pass)
```

### Invariants checked
- Spec-gated in: n/a
- Verification-gated out: n/a
- check/verify LLM-free: pass (sync verify phase remains deterministic; LLM only in resolve phase)
- No sexp serialization: n/a
- Secrets: pass

### Findings

**[low] Any table cell counts as a valid pin name**
- Where: `src/commands/sync.ts` (`validNames` populated from every `r.cells` entry)
- Fails when: PINOUT.md has a Notes/Net/MCU column cell equal to a pin token (e.g. `KEY` or `GPIO`) that is not the assigned pin identifier column. `PIN_KEY` then incorrectly passes the pins-h check.
- Fix: Restrict matching to the pin-name / signal column(s) defined by the PINOUT table schema (or the same columns `pins.h` generation uses). Add a test where a Notes cell contains `KEY` but the pin column does not, and expect a pins-h finding.

## Fix prompt

````text
You are fixing review findings on PR #153 in chouhanindustries/copperhead (branch fix/issue-105-pins-h-sync-check, based on main).
  gh pr checkout 153

FINDING 1 [low] (optional) Any table cell counts as a valid pin name
  Where: src/commands/sync.ts (validNames from every cell)
  Fails when: a Notes/Net cell equals a PIN_* token that is not the pin identifier column.
  Fix: Match only the pin-name/signal column(s) used by pins.h generation; add a regression test.

Then: npm run typecheck && npm run build && npm test
````

Devesh36:
🤖 **Automated `pr-review` pass** (Cursor). Not a human review. Delta against the prior pass ([comment](https://github.com/chouhanindustries/copperhead/pull/153#issuecomment-5133440402)) — severity upgrade + concrete regression.

## Verdict: request changes (was approve with nits)

The prior low nit is a real correctness gap for #105: matching every table cell still false-passes pin-number / refdes tokens.

### Findings

**[medium] `validNames` accepts every table cell, so non-name cells still false-pass pins-h**
- Where: `src/commands/sync.ts` (pins-h block: `parseCanonicalTables` loop over `r.cells`)
- Fails when: PINOUT has `| U1 | 5 | GPIO14 | KEY_DAH | |` and firmware defines `#define PIN_5 21` or `#define PIN_U1 1`; exact cell match reports consistency even though those are not assigned pin names.
- Fix: Resolve the assignment table by header (Name, optionally Net), same pattern as `parsePinoutRows` / drift; never add Refdes, Pin-number, or Notes cells. Add a regression that `#define PIN_5` fails against a fixture row whose Pin column is `5` and Name is `GPIO14`.

**[low] Regression covers only substring false-negatives**
- Where: `test/gating-sync.test.ts` (new pins-h test)
- Fails when: a correct `#define PIN_<Name>` should stay clean, or a pin-number collision should fail; neither is asserted.
- Fix: Extend the test with a minimal PINOUT table, one passing define, and one pin-number false-pass case.

## Fix prompt

````text
You are fixing review findings on PR #153 in chouhanindustries/copperhead (branch fix/issue-105-pins-h-sync-check, based on main).
  gh pr checkout 153

FINDING 1 [medium] validNames accepts every table cell, so non-name cells still false-pass pins-h
  Where: src/commands/sync.ts
  Fails when: #define PIN_5 or PIN_U1 matches Pin/Refdes cells instead of the Name column.
  Fix: Match Name (and optionally Net) headers only; add PIN_5 vs Pin=5 regression.

FINDING 2 [low] (optional) Regression covers only substring false-negatives
  Where: test/gating-sync.test.ts
  Fix: Add passing Name-column define + pin-number false-pass case.

Then: npm run typecheck && npm run build && npm test
````

Chirag6722:
Thanks @Devesh36! Added the requested regression test `pins-h check ignores Notes column text when verifying pin names (#105)` in `test/gating-sync.test.ts`. Confirmed that a token appearing in the `Notes` column (e.g. `KEY`) is not treated as a valid pin name when the `Net`/`Name` column does not assign it, producing a `pins-h` finding as expected.

Chirag6722:
Updated the regression test `pins-h check ignores Notes, Refdes, and Pin number columns when verifying pin names (#105)` in `test/gating-sync.test.ts`. Confirmed that:
1. Valid pin name `PIN_KEY_DAH` (in `Net` column) passes cleanly.
2. Notes text `KEY` (exact match in `Notes` column) fails (`defines PIN_KEY`).
3. Pin number `5` (in `Pin` column) fails (`defines PIN_5`).
4. Refdes `U1` (in `Refdes` column) fails (`defines PIN_U1`).

PR #162, opened 2026-07-31
# fix(agent): boundary-aware JSON extraction for parseDiagnosis (#101)

### LINKED ISSUE #101: parseDiagnosis reads a retry verdict as abort when a brace precedes the JSON, ending the run

Bhumika-1432006:
## What

This PR hardens the diagnosis parsing logic in the recovery supervisor ([`src/agent/recovery.ts`](src/agent/recovery.ts)) to cleanly extract structured JSON payloads even when preceded by arbitrary LLM conversational prose, code blocks, or leading stray braces (e.g., file references like `{pins.h}`).

### Summary of changes:

| File | Insertions | Deletions | Net |
|---|---|---|---|
| `src/agent/recovery.ts` | +21 | -15 | +6 |
| `test/recovery.test.ts` | +6 | -0 | +6 |
| **Total** | **+27** | **-15** | **+12** |

### Specific fixes

1. **Candidate-Scanning Balanced Brace Matcher** ([`src/agent/recovery.ts`](src/agent/recovery.ts)):
   - Replaced naive `text.indexOf('{')` logic in `parseDiagnosis` with an iterative loop scanning for candidate opening braces (`{`).
   - Added string/escape-aware depth counting that isolates balanced brace blocks and validates whether candidate slices parse into a valid `StageDiagnosis` object with a recognized `verdict` (`'retry'` | `'abort'`).
   - Prevents leading prose containing stray braces (such as `{pins.h}`) from polluting the extracted JSON boundary and causing `JSON.parse` syntax errors that unexpectedly terminate recovery attempts.

2. **Edge-Case Unit Test Coverage** ([`test/recovery.test.ts`](test/recovery.test.ts)):
   - Added an explicit test case validating that responses containing stray opening/closing braces in prefix text successfully parse into a valid `retry` verdict with preserved `reason` and `guidance`.

---

## Why

| Problem | Root Cause | Impact |
|---|---|---|
| Pipeline aborted on recoverable failures | LLM output contained leading braces before the diagnosis block (e.g., `Check {pins.h} file...`) | ❌ Naive `indexOf('{')` sliced invalid JSON spanning from the first stray brace, falling back to an `"abort"` verdict |
| Token waste on transient errors | Intermittent or transient tool-call slips failed to trigger retries when response formatting had preamble noise | ❌ Pipeline terminated early instead of prepending guidance and executing a retry turn |

**Before this PR:** Any LLM diagnosis turn containing a stray `{` prior to the JSON payload resulted in a parsing error, causing `parseDiagnosis` to fall back to `verdict: 'abort'` and killing the pipeline run.

**After this PR:** `parseDiagnosis` gracefully scans past pre-JSON prose and stray braces to extract the valid `StageDiagnosis` payload, restoring pipeline resilience.

---

### 1. Robust JSON extraction logic

Instead of assuming the first `{` character marks the start of the diagnosis object, the updated `parseDiagnosis` iterates through candidate `{` positions:

```typescript
for (let start = 0; start < text.length; start++) {
  if (text[start] !== '{') continue;

  let depth = 0;
  let inStr = false;
  let esc = false;

  for (let i = start; i < text.length; i++) {
    // ... string escaping & depth tracking ...
    if (ch === '}' && --depth === 0) {
      try {
        const candidate = text.slice(start, i + 1);
        const o = JSON.parse(candidate) as Partial<StageDiagnosis>;
        if (typeof o === 'object' && o !== null && (o.verdict === 'retry' || o.verdict === 'abort')) {
          // Found a valid diagnosis payload
          return { ... };
        }
      } catch {
        break; // Not valid JSON at this opening brace; try next start candidate
      }
    }
  }
}
```

2. Invariant preservation
Safe fallback: If no valid JSON diagnosis object is found across all candidate braces, parseDiagnosis safely defaults to { verdict: 'abort', reason: 'diagnosis was not valid JSON' }.

Zero new dependencies: Keeps parser lightweight and zero-dependency using native string scanning.

Spec / OpenSpec
Pure agent pipeline recovery logic hardening. No OpenSpec artifacts are required or affected: no spec-level protocol behavior was changed.

### Automated test status

| Check | Command | Status |
|---|---|---|
| Typecheck | `npm run typecheck` | :white_check_mark: **Pass** (clean, zero diagnostics) |
| Unit Tests (Recovery) | `npx vitest run test/recovery.test.ts` | :white_check_mark: **13 passed**, 0 failed |

### Full test output

RUN  v3.2.7 D:/copperhead3/copperhead

 ✓ test/recovery.test.ts (13 tests) 84ms
   ✓ withTimeout (turn watchdog) > returns the value when the call finishes in time 2ms
   ✓ withTimeout (turn watchdog) > rejects with TurnTimeoutError and fires onTimeout when the deadline passes 41ms
   ✓ withTimeout (turn watchdog) > disables the watchdog when ms <= 0 (awaits the call) 0ms
   ✓ parseDiagnosis > parses a retry verdict with guidance, even wrapped in prose 0ms
   ✓ parseDiagnosis > parses an abort verdict and ignores guidance 0ms
   ✓ parseDiagnosis > parses retry verdict when leading prose contains stray braces (Fixes #101) 1ms
   ✓ parseDiagnosis > fails safe to abort on non-JSON or missing verdict 0ms
   ✓ diagnoseStageFailure > asks the model with no tools and returns the parsed verdict 1ms
   ✓ diagnoseStageFailure > fails safe to abort when the provider throws 1ms
   ✓ CachingProvider > caches a response and replays it (no second model call), reporting zero usage 13ms
   ✓ CachingProvider > misses (calls the model again) when the conversation differs 14ms
   ✓ CachingProvider > does not share cached turns across compat endpoints with the same model id 6ms
   ✓ CachingProvider > hits a pre-upgrade cache entry for a non-compat model (key shape unchanged) 4ms

 Test Files  1 passed (1)
      Tests  13 passed (13)
   Start at  00:55:01
   Duration  2.34s

Manual test log (required)
Sandbox variant used: edit

Commands run and outcome:
```typescript
$ npx vitest run test/recovery.test.ts
13 passed (success)
```

<html>
<body>
<html><head></head><body><h2 data-path-to-node="18" style="font-family: &quot;Google Sans&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">Files changed</h2><div class="horizontal-scroll-wrapper" style="font-family: &quot;Google Sans Text&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">
File | Change type | Description
-- | -- | --
src/agent/recovery.ts | Logic fix | Candidate-scanning balanced brace matcher in parseDiagnosis to tolerate leading prose and stray braces
test/recovery.test.ts | Test coverage | Added unit test verifying JSON diagnosis extraction when preceding text contains stray braces

</div></body></html>
</body>
</html>Files changedFileChange typeDescription[src/agent/recovery.ts](https://www.google.com/search?q=src/agent/recovery.ts)Logic fixCandidate-scanning balanced brace matcher in parseDiagnosis to tolerate leading prose and stray braces[test/recovery.test.ts](https://www.google.com/search?q=test/recovery.test.ts)Test coverageAdded unit test verifying JSON diagnosis extraction when preceding text contains stray braces

Docs
No documentation changes required. This is an internal agent recovery parser fix with zero CLI/API surface changes.

Invariant checklist
- [x] Spec-gated in: edit_file/write_file remain structurally absent from the agent tool list until an OpenSpec proposal validates.

- [x] Verification-gated out: Mutations still terminate in ERC/DRC verification, with retry and rollback logic intact.
- [x] Fail-Safe Recovery: Invalid, unparseable, or missing JSON blocks fall back safely to verdict: 'abort'.
- [x] Zero Dependencies: Pure TypeScript string scanning; no external JSON or regex parsing dependencies introduced.
- [x] Pipeline Cost Accounting: Preserves token usage folding (usage tracking) across diagnosis turns.
- [x] Secrets: Logs and transcripts continue to redact sk-[A-Za-z0-9_-]+ and sensitive auth tokens at write time.
- [x] Spec coherence:Pure recovery parser hardening with zero changes to spec-level behavior. All core specs in [`SPEC.md`](openspec/specs/SPEC.md) remain untouched, and `openspec validate` passes cleanly.

Summary by CodeRabbit
Tests

Enhanced recovery supervisor testing to verify robust JSON extraction when LLM responses contain leading prose or stray opening braces prior to the diagnosis object.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnosis parsing for responses containing stray braces, malformed JSON, or incomplete diagnosis objects.
  * The system now continues searching for a valid diagnosis and preserves available reason text when no recognized verdict is found.
  * Safely falls back to aborting with a relevant or generic error when no valid diagnosis is available.

* **Tests**
  * Added coverage for leading prose with stray braces, invalid objects preceding a valid retry diagnosis, and fallback reason preservation.


AniketR10:
🤖 Automated pr-review pass (no human review implied).

## Verdict: Approve with nits

Small, well-tested fix for a real bug (#101: a stray `{` in LLM prose before the diagnosis JSON broke `parseDiagnosis`). The core fix is sound and correctly handled by the new candidate-scanning loop. One real (but non-default-path) behavior regression survived verification, plus two low-severity nits.

## Metrics block

The bundled `metrics.mjs` script computed against the wrong base: in this checkout, `origin` is the reviewer's personal fork (`AniketR10/copperhead-aniket`, stale at v0.4.0-era `main`), not `chouhanindustries/copperhead`. The script always prefers `origin/<base>` once it verifies, so both `metrics.mjs 162` and `metrics.mjs --base main` silently diffed against that stale fork and reported ~172 changed files. Numbers below are derived manually against the correct merge base (`git merge-base main HEAD` = `c72b2da`, which matches `gh pr view`'s own `+35/-13` total) using the same methodology the script uses.

```
lines: +35 / -13 (net +22) across 2 files  ·  src +20/-12, test +15/-1, docs +0, spec +0, other +0
new vs net: 0 new files, 2 modified; 0 new src lines (pure edits to existing files)
tests: +14 test lines (net); suite 581/20/0 -> 583/20/0 (pass/skip/fail, base -> head)
diff coverage: 100% of changed src lines exercised (measured: vitest --coverage.provider=v8, recovery.ts lines 57-97 diff region all covered)
deps: none changed (package.json untouched in this diff)
ci: 4 pass / 0 fail / 0 pending (CodeRabbit, cla, label, template — none of these run the test suite); required checks: none reported
```
Methods: change size/new-vs-net from `git diff --numstat c72b2da...HEAD`; suite from `npx vitest run --reporter=json` on base (`main`) and head branches directly (script's fork-remote issue made `--base-tests` unusable); diff coverage from `vitest --coverage.provider=v8` intersected by hand with the changed line range, since only 2 files/~30 lines changed.

## Invariant checks performed (all passed)

- **Spec-gated in**: N/A — no tool-list changes.
- **Verification-gated out**: N/A — `recovery.ts` is only imported by `src/commands/create.ts` (confirmed via grep); it's part of the retry/diagnosis path, not a bypass of ERC/DRC gating.
- **`check`/`verify` stays LLM-free/network-free**: confirmed `src/commands/check.ts` does not import `recovery.ts` (grep, no hits).
- **No sexp serialization**: N/A — diff doesn't touch `src/kicad/`.
- **Sync-obligations ledger**: N/A — diff doesn't touch commit/ledger logic.
- **Secrets**: N/A — no transcript/redaction code touched.
- **Spec coherence**: PR correctly self-identifies as not spec-level; no `openspec/` changes, none needed.

## Findings

**1. [medium] Valid JSON without a recognized `verdict` field now loses its diagnostic `reason`, replaced by a misleading generic message**
`src/agent/recovery.ts:78-96`
Previously, when the first (or only) brace-balanced JSON candidate parsed successfully but had no `verdict` field, `parseDiagnosis` fell back to `verdict: 'abort'` while *preserving* the object's actual `reason` text. The new code only accepts a candidate whose `verdict` is exactly `'retry'` or `'abort'` (line 78); anything else falls through silently, and if no later candidate matches, the function returns the generic `{ verdict: 'abort', reason: 'diagnosis was not valid JSON' }` (line 96) — discarding the real reason and mislabeling valid JSON as invalid.
Fails when: the diagnosis model returns syntactically valid JSON that omits or misnames the `verdict` key (a plausible model slip, e.g. `{"reason":"the tool call was dropped silently"}` or a differently-cased/misspelled key). Confirmed by direct repro on this branch: `parseDiagnosis('{"reason":"the tool call was dropped silently"}')` → `{verdict:'abort', reason:'diagnosis was not valid JSON'}`, whereas on `main` the same input returns `{verdict:'abort', reason:'the tool call was dropped silently'}`. This `reason` is surfaced directly in `src/commands/create.ts:800` (`diagnosis → ${diagnosis.verdict} — ${diagnosis.reason}`) and used to build retry guidance at `create.ts:808`, so the regression degrades operator-facing logs and future-attempt guidance on a reachable path. Notably, the existing test at `test/recovery.test.ts:76` (`parseDiagnosis('{"reason":"x"}').verdict === 'abort'`) exercises this exact input but only asserts `.verdict`, so it passed without catching the reason regression.
Fix: when a candidate parses as valid JSON but lacks a recognized `verdict`, retain its `reason` as a fallback (don't discard it) and keep scanning for a better (verdict-bearing) candidate; if none is found, return abort using the last valid-but-verdictless candidate's `reason` if one was seen, instead of the generic "not valid JSON" message. Add a regression test asserting `.reason`, not just `.verdict`, for the `{"reason":"x"}` case.

**2. [low] Docstring no longer matches the function's actual behavior**
`src/agent/recovery.ts:51-53`
The comment still reads "Extract the first brace-balanced JSON object from text ... interpret it as a StageDiagnosis," but the new implementation scans *all* brace-delimited candidates in sequence looking for one with a recognized `verdict`, not just the first. Fails when: a future maintainer reads the comment to understand the function and is misled about which candidate wins when multiple JSON-like fragments are present (exactly finding #1's scenario). Fix: update the docstring to describe the multi-candidate scan and the verdict-matching requirement.

**3. [low] Trailing newline dropped from both changed files**
`src/agent/recovery.ts:170`, `test/recovery.test.ts:198`
Both files now end without a trailing newline (visible as `\ No newline at end of file` in the diff). Not a functional bug, but inconsistent with the rest of the codebase and may trip formatter/lint checks. Fix: restore the trailing newline in both files.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is #162, branch fix/issue-101-parse-diagnosis-json,
based on main.

Check it out first:
  gh pr checkout 162

FINDING 1 [medium] Valid JSON without a recognized verdict field now loses its
diagnostic reason, replaced by a misleading generic message
  Where: src/agent/recovery.ts:78-96
  Fails when: the diagnosis model returns syntactically valid JSON that omits or
  misnames the "verdict" key (e.g. {"reason":"the tool call was dropped silently"}).
  Previously parseDiagnosis fell back to verdict:'abort' while preserving the
  object's actual reason text; now it discards that reason entirely and returns
  the generic { verdict: 'abort', reason: 'diagnosis was not valid JSON' }, which
  is also factually wrong (the JSON was valid). Confirmed by direct repro:
  parseDiagnosis('{"reason":"the tool call was dropped silently"}') returns
  {verdict:'abort', reason:'diagnosis was not valid JSON'} on this branch, versus
  {verdict:'abort', reason:'the tool call was dropped silently'} on main. This
  reason string is surfaced directly in src/commands/create.ts:800
  (diagnosis -> ${diagnosis.verdict} - ${diagnosis.reason}) and used to build
  retry guidance at create.ts:808, so this degrades operator-facing diagnostics
  and future-attempt guidance on a reachable (non-default) path. The existing
  test test/recovery.test.ts:76 exercises this exact input but only asserts
  .verdict, not .reason, so it passed without catching the regression.
  Fix: when a candidate parses as valid JSON but lacks a recognized verdict,
  retain its reason as a fallback instead of discarding it, and keep scanning for
  a better (verdict-bearing) candidate; if none is found, return abort using the
  last valid-but-verdictless candidate's reason when one was seen, rather than
  the generic "diagnosis was not valid JSON" message. Add a regression test that
  asserts .reason (not just .verdict) for the {"reason":"x"} case.

FINDING 2 [low] (optional) Docstring no longer matches the function's actual
behavior
  Where: src/agent/recovery.ts:51-53
  Fails when: a future maintainer reads the comment ("Extract the first
  brace-balanced JSON object from text...") to understand the function and is
  misled: the new implementation scans all brace-delimited candidates looking
  for one with a recognized verdict, not just the first.
  Fix: update the docstring to describe the multi-candidate scan and the
  verdict-matching requirement.

FINDING 3 [low] (optional) Trailing newline dropped from both changed files
  Where: src/agent/recovery.ts:170, test/recovery.test.ts:198
  Fails when: N/A, pure formatting; shows as "\ No newline at end of file" in
  the diff and is inconsistent with the rest of the codebase.
  Fix: restore the trailing newline at end of file in both files.

    by prompt text.

Then verify:
  npm run typecheck
  npm run build
  npm test

and report it.

````

AniketR10:
@Bhumika-1432006 plz look at the review

Bhumika-1432006:
@AniketR10 Thanks for the review!
 All three findings are taken care of in commit e914c6f.

Here is what was updated:

Finding 1 (Medium): Updated parseDiagnosis to track a fallbackReason as it scans candidate JSON objects. If valid JSON is parsed without a recognized verdict field, its reason text is preserved. If no candidate with a verdict is found by the end of the scan, parseDiagnosis returns an abort verdict using that saved reason text instead of the generic failure message. I also added a regression test asserting the reason field on {"reason":"the tool call was dropped silently"}.

Finding 2 (Low): Updated the parseDiagnosis docstring to accurately describe the candidate scanning and verdict matching behavior.

Finding 3 (Low): Restored trailing newlines at the end of both modified files (src/agent/recovery.ts and test/recovery.test.ts).

All 15 unit tests, typechecks, and builds are passing cleanly!

Bhumika-1432006:
Verified at the byte level rather than trusting the editor:

Last byte of both src/agent/recovery.ts and test/recovery.test.ts is now \n (confirmed via hex dump)
The \ No newline at end of file marker no longer appears in git diff for either file

Also re-ran the full check to make sure this didn't regress anything:

npm run typecheck — clean
npx vitest run test/recovery.test.ts — 15/15 passed

Commit: d5de204

PR #160, opened 2026-07-31
Propose deterministic-schematic-drafting: drafting engine, legibility gate, scoring, golden corpus

animesh-chouhan:
## What

Proposes the `deterministic-schematic-drafting` OpenSpec change and implements its checker phase (groups 1-4 of 12): a legibility standard and deterministic checker that gate the create pipeline's schematic stage, ahead of the rule-based drafting engine that will compute all schematic geometry from a compact netlist-intent IR. Landed here: `src/kicad/legibility.ts` (13 check families), the `check_legibility` agent tool and ledger obligation (scoped to create-origin repos), the stage-4 drafting standard and completion gate, `check` advisories with the `legibility` JSON object, and fixtures/tests. The engine phase (emitter, IR, placement, scorer, golden corpus) is implemented on this branch as well; see the implementation-status comment for the per-capability breakdown.

## Why

Two failure modes documented on real runs. #136: stage 4 gates the net graph but never the drawing, so `create` emits electrically clean, unreadable schematics (refdes text over symbol bodies, content in 40% of the sheet, empty title block). #145: the geometry text the model authors token by token is where cost and failures concentrate (5 of 40 turns emitted 89% of 170,920 output tokens; every provider error, budget exhaustion, and stall landed on an oversized geometry turn). #159 tracks the engine.

This consolidates and supersedes the proposal-only `readable-schematic-drafting` change (#148, closed): the checker and the engine it gates are specified and will be implemented together, checker phase first.

## Design

Load-bearing decisions, full rationale in [design.md](openspec/changes/deterministic-schematic-drafting/design.md):

- **Intent, not geometry** (D5): the model authors `schematic.intent.json` (parts, nets, groups, no-connects, hints); the engine computes placement. Repairs revise the IR and re-draft; geometry edits against an engine-drafted sheet are refused.
- **Hand-rolled grid-native rules, no layout library** (D1): after human-drafting reductions (power rails to per-pin symbols, decoupling rows beside their IC, labels between groups), each group is a 3-15 symbol layered-placement problem; elkjs is documented as the v2 escape hatch only.
- **Every gate failure is resolvable through the IR** (D6a): no-connect markers and PWR_FLAG synthesis are specified so the repair loop can never wedge the way #82 did.
- **Determinism is structural** (D4): UUIDv5 from semantic paths, canonical ordering, and symbol sources vendored into a committed project cache, so identical IR yields byte-identical files and a KiCad library upgrade cannot silently change output.
[...]

PR #186, opened 2026-08-06
fix(kicad): find symbols across libraries, not just the guessed file

### LINKED ISSUE #185: draft_schematic fails to find real chips that already exist in the stock KiCad library, under a different library file
## What happened

Running `create` end to end on a real board (ESP32 battery soil sensor, `examples/medium/esp32-soil-sensor.md`), `draft_schematic` reported four parts as "library not installed and not vendored":

- `TPS63900` (regulator)
- `TPS22860` (load switch)
- `SHT40` (humidity sensor)
- `FDC1004` (capacitance sensor)

The agent spent 30+ turns trying name variants, then correctly refused rather than guess a pinout, then burned a second full retry doing the same thing. Two separate runs hit `turn-budget-exhausted` on this alone.

## Root cause (confirmed against a real KiCad 10.0.5 install)

`resolveLibrarySymbol` (`src/kicad/symlib.ts:151`) and `SymbolSource.resolve` (`src/kicad/draft/symsource.ts:283`) both split a `lib_id` at the colon and open **only** the named library file (`findLibraryFile`, `symlib.ts:99`). If the part isn't in that one file, resolution fails, even when the part exists in the *stock* KiCad symbol set under a different library name.

Checked directly against the installed libraries, exact matches only:

| Chip | Guessed library | Actually lives in |
|---|---|---|
| SHT40 | `Sensor_Humidity` | `Sensor_Humidity` (right file, part is named `SHT4x` not `SHT40`) |
| TPS63900 | `Regulator_DCDC` | `Regulator_Switching` |
| FDC1004 | `Sensor` | `Sensor_Proximity` |
| TPS22860 | `Power_Switch` | not present anywhere in the stock set (closest is `TPS22810`, a different part) |

3 of the 4 were resolver failures, not missing parts: the tool never looked past the one file it guessed. The 4th (`TPS22860`) is a genuine gap, and is exactly the case Part 3 below is for.

## Proposed fix

**1. Cross-library, fuzzy search (fixes 3 of the 4 cases here).** When the guessed library file doesn't contain the part, search every discovered `.kicad_sym` file for a matching symbol before giving up: exact name match first, then a fuzzy/substring match (the same closest-candidate logic `resolveLibrarySymbol` already uses within one file, extended across all of them). Report the correct `lib_id` if found in exactly one place; report all candidates if ambiguous, so the agent (not the engine) decides — same "did you mean" pattern already used for typo'd names, just extended across files. Exact matches must be reported separately from fuzzy ones (confirmed by testing: a loose substring check alone would have wrongly suggested `TPS22810` as a match for `TPS22860`).

**2. When there's still no exact match anywhere, surface the closest fuzzy candidates as a possible substitute.** Rather than a bare failure, report the nearest names by similarity (already partially there via the closest-candidate list) so the agent can propose a substitute part, clearly labeled as unconfirmed rather than presented as the same chip.

**3. Let a project vendor its own extra libraries (for a genuinely absent part, like `TPS22860` here).** `SymbolSource` already accepts a `symbolDirs` override, but only as a test seam (`DraftOptions.symbolDirs`, `draft.ts:23`); it isn't reachable from `.copperhead/config.json`. Surfacing it as a real config option lets a project commit a manufacturer-provided or hand-authored `.kicad_sym` file and point the tool at it. No symbol content is authored automatically; a human supplies and reviews the file before it's used.

Auto-fetching a symbol from a third-party service with no human review was considered and set aside for now: it needs paid API access and would mean a design could ship with a symbol nobody checked, which is exactly the failure class this tool is built to prevent.

## Reproduction

```bash
copperhead create --repo <empty-repo> --brief examples/medium/esp32-soil-sensor.md
```

Stage 4 (schematic) hits the four failures above on a from-scratch build.

Chirag6722:
Closes #185. Targets `feat/deterministic-schematic-drafting` rather than `main`, since `src/kicad/draft/symsource.ts` only exists on that branch: the diff here is the single commit on top of #160.

## What

Symbol resolution now searches across libraries before giving up, instead of only opening the one file named in the `lib_id`.

## Why

Driving `create` end to end on `examples/medium/esp32-soil-sensor.md`, stage 4 reported four real parts as "library not installed and not vendored" and burned two full turn budgets on them. Three of the four were in the stock KiCad library the whole time, filed under a library nickname the caller could not have derived from the part number:

| Part | lib_id used | Actually in |
| --- | --- | --- |
| TPS63900 | `Regulator_DCDC:` | `Regulator_Switching` |
| SHT40 | `Sensor_Humidity:SHT40` | `Sensor_Humidity`, as `SHT4x` |
| FDC1004 | `Sensor:` | `Sensor_Proximity` |
| TPS22860 | `Power_Switch:` | genuinely absent from the stock set |

`resolveLibrarySymbol` and `SymbolSource.resolve` both split the lib_id at the colon and opened only that file, so a wrong library nickname was indistinguishable from a missing part. The agent could not tell "you named the wrong file" from "this part does not exist", and correctly refused rather than guess a pinout, twice.

## Design

Fallback order is by strength of evidence, so a weaker signal never displaces a stronger one:

1. exact name in the guessed file (unchanged)
2. near-matches within the guessed file: the caller named the right file and mistyped the part, so its own candidates outrank any other library's
3. exact name in a different library: the file was the mistake, not the part
4. fuzzy match in a different library, surfaced as a suggestion

Matching has to stay tight enough never to propose a *different real chip*. Two guards, both added in response to a case this actually hit:

- substring matching ignores library names under 3 characters. A stock library is full of single-letter generics (`R`, `C`, `D`), and "does &lt;any part number&gt; contain R" is true for nearly everything.
- edit distance (≤1 edit, names ≥5 chars) catches the datasheet-vs-library near-miss substring cannot: `SHT40` vs `SHT4x` differ mid-string, so neither contains the other. Capped at one edit specifically so `TPS22860` does not match `TPS22810`, a different real part.

`TPS22860` still resolving to nothing is the correct outcome, and is the case the config-level extra-library option in #185 is for.

Two defects found while building this are fixed in the same commit:

- `symbolSearchDirs` ran Windows version-directory discovery even when `KICAD_SYMBOL_DIR` was set, contradicting its own "env overrides win" contract and leaking the machine's real KiCad install into tests that had pinned an isolated fixture directory.
- the Windows fallback path hardcoded `C:/Program Files/KiCad/share/kicad/symbols`, omitting the version segment KiCad's installer always creates (`.../KiCad/10.0/share/...`), so symbol resolution failed outright on a stock Windows install. Now discovered by listing the install root, newest version first.

## Testing

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npx vitest run` | 16 failing before and after (see below) |
| Targeted | `npx vitest run test/symlib.test.ts test/draft-hardening.test.ts test/draft-tools.test.ts` | 47/47 pass |

The 16 failures are pre-existing on this machine and unrelated to symbol resolution: Windows path handling, `chmod` semantics, app-bundle probing. Confirmed by capturing the failing-test list with the change stashed and again with it applied, then diffing. The baseline was 18; the two `symbolSearchDirs` fixes above repaired two of them.

One regression was introduced and fixed during development rather than shipped: ordering the cross-library check before the same-file candidate list made `Device:R_Sma` report "use `Device:R_Sma` instead", suggesting the exact input back at the caller. `test/draft-hardening.test.ts` caught it, and it is what motivated the evidence ordering above.

### Manual verification

Against a real KiCad 10.0.5 install on Windows:

```text
TPS63900   -> Regulator_Switching [EXACT]
SHT40      -> Sensor_Humidity (fuzzy)      # the SHT4x near-miss
FDC1004    -> Sensor_Proximity (fuzzy)
TPS22860   -> NOT FOUND                    # correct: genuinely absent
```

The last line is the one worth checking on review: a looser matcher would have proposed `TPS22810` here, which is a different part.

---

## Invariant checklist

- [ ] N/A **Spec-gated in**: no change to tool exposure or the unlock path.
- [ ] N/A **Verification-gated out**: no change to the ERC/DRC gates or rollback.
- [ ] N/A **`check`/`verify` stays LLM-free and network-free**: this path is filesystem-only and reachable from `check` only as it already was; no provider, key, or network use added.
- [x] **No sexp serialization**: the parser stays read-only. Cross-library search reads `.kicad_sym` text and matches symbol names; nothing is written or round-tripped.
- [ ] N/A **Secrets**: no credential handling; no new transcript output.
- [ ] N/A **Spec coherence**: no spec-level behavior change. Resolution reports a corrected lib_id where it previously reported a dead end; the refusal contract (never guess a pinout) is preserved, including for `TPS22860`.

### DIFF
--- a/src/agent/tools.ts
+++ b/src/agent/tools.ts
@@ -354,6 +354,8 @@ export const TOOLS: ToolDef[] = [
         return 'no schematic configured; verify_symbols does not apply yet';
       const { findings, checked, skipped } = await verifySchematicSymbols(
         path.join(ctx.repoRoot, ctx.config.schematic),
+        process.env,
+        ctx.repoRoot,
       );
       if (!findings.length) {
         return `verify_symbols: ${checked} symbol(s) match the installed KiCad library. No divergences.`;
--- a/src/kicad/draft/symsource.ts
+++ b/src/kicad/draft/symsource.ts
@@ -2,7 +2,7 @@ import { readFile, writeFile, mkdir } from 'node:fs/promises';
 import { existsSync } from 'node:fs';
 import path from 'node:path';
 import { parseSexp, children, child, isList, type SexpNode, type Bounds } from '../sexp.js';
-import { symbolSearchDirs, findLibraryFile } from '../symlib.js';
+import { symbolSearchDirs, findLibraryFile, findSymbolAcrossLibraries } from '../symlib.js';
 import { EMIT_VERSION, renameSymbolBlock } from '../emit.js';
 
 /**
@@ -186,7 +186,7 @@ function appendToVendorLib(libText: string, block: string): string {
 export class SymbolResolutionError extends Error {
   constructor(
     public readonly libId: string,
-    public readonly reason: 'no-library' | 'no-symbol' | 'derived-unsupported',
+    public readonly reason: 'no-library' | 'no-symbol' | 'derived-unsupported' | 'found-elsewhere',
     public readonly candidates: string[] = [],
   ) {
     super(
@@ -194,9 +194,26 @@ export class SymbolResolutionError extends Error {
         ? `library for "${libId}" is not installed and not vendored`
         : reason === 'derived-unsupported'
           ? `"${libId}" extends a symbol that cannot be followed (missing base name, or a cycle deeper than ${MAX_EXTENDS_DEPTH} hops); use the base symbol directly`
-          : `"${libId}" does not exist in the library${candidates.length ? ` — closest: ${candidates.join(', ')}` : ''}`,
+          : reason === 'found-elsewhere'
+            ? `"${libId}" is wrong about the library, not the part: use ${candidates.join(' or ')} instead`
+            : `"${libId}" does not exist in the library${candidates.length ? ` — closest: ${candidates.join(', ')}` : ''}`,
     );
   }
+}
+
+/**
+ * A part's library file is guessed wrong more often than the part itself is
+ * misspelled (a chip's stock library nickname is not derivable from its part
+ * number). Before failing outright, check every other stock/vendored library
+ * for this exact part name; an exact hit elsewhere is a stronger signal than
+ * a same-file substring guess, so it takes priority over `candidates`.
+ * Returns corrected `lib_id`s (exact matches first) or an empty array.
+ */
+async function crossLibrarySuggestions(name: string, lib: string, dirs: string[]): Promise<string[]> {
+  const matches = await findSymbolAcrossLibraries(name, dirs, lib);
+  const exact = matches.filter((m) => m.exact).map((m) => `${m.lib}:${name}`);
+  if (exact.length) return exact;
+  return matches.map((m) => `${m.lib}:${name}`);
 }
 
 export class SymbolSource {
@@ -296,13 +313,26 @@ export class SymbolSource {
     if (!block) {
       const dirs = this.searchDirs ?? (await symbolSearchDirs());
       const file = await findLibraryFile(lib, dirs);
-      if (!file) throw new SymbolResolutionError(libId, 'no-library');
+      if (!file) {
+        const elsewhere = await crossLibrarySuggestions(name, lib, dirs);
+        if (elsewhere.length) throw new SymbolResolutionError(libId, 'found-elsewhere', elsewhere);
+        throw new SymbolResolutionError(libId, 'no-library');
+      }
       const libText = await readFile(file, 'utf8');
       block = extractSymbolBlock(libText, name);
       if (!block) {
+        // The guessed library exists and simply lacks this name, so its own
+        // near-matches are the most useful answer and come first: the caller
+        // named the right file and mistyped the part. Only when that file
+        // offers nothing is a different library worth suggesting — and only
+        // an exact hit there, since a cross-library *fuzzy* guess is weaker
+        // evidence than "you were close, in the right file".
         const names = [...libText.matchAll(/^\s*\(symbol\s+"([^"]+)"/gm)].map((m) => m[1]!);
         const q = name.toLowerCase();
         const candidates = names.filter((k) => k.toLowerCase().includes(q) || q.includes(k.toLowerCase())).slice(0, 8);
+        if (candidates.length) throw new SymbolResolutionError(libId, 'no-symbol', candidates);
+        const elsewhere = await crossLibrarySuggestions(name, lib, dirs);
+        if (elsewhere.length) throw new SymbolResolutionError(libId, 'found-elsewhere', elsewhere);
         throw new SymbolResolutionError(libId, 'no-symbol', candidates);
       }
       fromInstalled = true;
--- a/src/kicad/draft/symsource.ts
+++ b/src/kicad/draft/symsource.ts
@@ -211,9 +211,9 @@ export class SymbolResolutionError extends Error {
  */
 async function crossLibrarySuggestions(name: string, lib: string, dirs: string[]): Promise<string[]> {
   const matches = await findSymbolAcrossLibraries(name, dirs, lib);
-  const exact = matches.filter((m) => m.exact).map((m) => `${m.lib}:${name}`);
+  const exact = matches.filter((m) => m.exact).map((m) => `${m.lib}:${m.name}`);
   if (exact.length) return exact;
-  return matches.map((m) => `${m.lib}:${name}`);
+  return matches.map((m) => `${m.lib}:${m.name}`);
 }
 
 export class SymbolSource {
--- a/src/kicad/draft/symsource.ts
+++ b/src/kicad/draft/symsource.ts
@@ -2,7 +2,7 @@ import { readFile, writeFile, mkdir } from 'node:fs/promises';
 import { existsSync } from 'node:fs';
 import path from 'node:path';
 import { parseSexp, children, child, isList, type SexpNode, type Bounds } from '../sexp.js';
-import { symbolSearchDirs, findLibraryFile, findSymbolAcrossLibraries } from '../symlib.js';
+import { symbolSearchDirs, findLibraryFile, findSymbolAcrossLibraries, crossLibraryIds } from '../symlib.js';
 import { EMIT_VERSION, renameSymbolBlock } from '../emit.js';
 
 /**
@@ -201,21 +201,6 @@ export class SymbolResolutionError extends Error {
   }
 }
 
-/**
- * A part's library file is guessed wrong more often than the part itself is
- * misspelled (a chip's stock library nickname is not derivable from its part
- * number). Before failing outright, check every other stock/vendored library
- * for this exact part name; an exact hit elsewhere is a stronger signal than
- * a same-file substring guess, so it takes priority over `candidates`.
- * Returns corrected `lib_id`s (exact matches first) or an empty array.
- */
-async function crossLibrarySuggestions(name: string, lib: string, dirs: string[]): Promise<string[]> {
-  const matches = await findSymbolAcrossLibraries(name, dirs, lib);
-  const exact = matches.filter((m) => m.exact).map((m) => `${m.lib}:${m.name}`);
-  if (exact.length) return exact;
-  return matches.map((m) => `${m.lib}:${m.name}`);
-}
-
 export class SymbolSource {
   private cache = new Map<string, ResolvedSymbol>();
 
@@ -314,25 +299,30 @@ export class SymbolSource {
       const dirs = this.searchDirs ?? (await symbolSearchDirs());
       const file = await findLibraryFile(lib, dirs);
       if (!file) {
-        const elsewhere = await crossLibrarySuggestions(name, lib, dirs);
-        if (elsewhere.length) throw new SymbolResolutionError(libId, 'found-elsewhere', elsewhere);
+        const elsewhere = await findSymbolAcrossLibraries(name, dirs, lib);
+        const exactElsewhere = elsewhere.filter((m) => m.exact);
+        if (exactElsewhere.length) throw new SymbolResolutionError(libId, 'found-elsewhere', crossLibraryIds(exactElsewhere));
+        if (elsewhere.length) throw new SymbolResolutionError(libId, 'found-elsewhere', crossLibraryIds(elsewhere));
         throw new SymbolResolutionError(libId, 'no-library');
       }
       const libText = await readFile(file, 'utf8');
       block = extractSymbolBlock(libText, name);
       if (!block) {
-        // The guessed library exists and simply lacks this name, so its own
-        // near-matches are the most useful answer and come first: the caller
-        // named the right file and mistyped the part. Only when that file
-        // offers nothing is a different library worth suggesting — and only
-        // an exact hit there, since a cross-library *fuzzy* guess is weaker
-        // evidence than "you were close, in the right file".
+        // Evidence, strongest first: an exact match in a different library is
+        // unambiguous, so it outranks even a same-file substring guess — the
+        // guessed library is often full of single-letter generics ("R", "C",
+        // "L") that trivially substring-match almost any query, and would
+        // otherwise silently outrank a real cross-library answer. Same-file
+        // candidates come next (the caller named the right file and mistyped
+        // the part); a cross-library fuzzy hit is last resort.
+        const elsewhere = await findSymbolAcrossLibraries(name, dirs, lib);
+        const exactElsewhere = elsewhere.filter((m) => m.exact);
+        if (exactElsewhere.length) throw new SymbolResolutionError(libId, 'found-elsewhere', crossLibraryIds(exactElsewhere));
         const names = [...libText.matchAll(/^\s*\(symbol\s+"([^"]+)"/gm)].map((m) => m[1]!);
         const q = name.toLowerCase();
         const candidates = names.filter((k) => k.toLowerCase().includes(q) || q.includes(k.toLowerCase())).slice(0, 8);
         if (candidates.length) throw new SymbolResolutionError(libId, 'no-symbol', candidates);
-        const elsewhere = await crossLibrarySuggestions(name, lib, dirs);
-        if (elsewhere.length) throw new SymbolResolutionError(libId, 'found-elsewhere', elsewhere);
+        if (elsewhere.length) throw new SymbolResolutionError(libId, 'found-elsewhere', crossLibraryIds(elsewhere));
         throw new SymbolResolutionError(libId, 'no-symbol', candidates);
       }
       fromInstalled = true;
--- a/src/kicad/symlib.ts
+++ b/src/kicad/symlib.ts
@@ -56,7 +56,9 @@ export async function symbolSearchDirs(env = process.env): Promise<string[]> {
     '/usr/share/kicad/symbols',
     '/usr/local/share/kicad/symbols',
     '/Applications/KiCad/KiCad.app/Contents/SharedSupport/symbols',
-    'C:/Program Files/KiCad/share/kicad/symbols',
+    // Windows installs under a version-numbered directory
+    // (C:\Program Files\KiCad\10.0\...), unlike Linux/macOS, so there is no
+    // single fixed path here. Discovered below instead.
   ];
   const out: string[] = [];
   for (const dir of [...fromEnv, ...defaults]) {
@@ -66,6 +68,34 @@ export async function symbolSearchDirs(env = process.env): Promise<string[]> {
     } catch {
       // not present on this machine; skip
     }
+  }
+  // Windows: KiCad's own installer picks the version directory
+  // (`10.0`, `9.0`, `8.0`, ...), so no fixed path is ever correct. List
+  // `C:\Program Files\KiCad` and check each version folder instead. Sorted
+  // descending so a machine with more than one version prefers the newest.
+  // Skipped entirely when an env override is set: "env overrides win" means
+  // exactly that, not "env overrides win, plus whatever else this discovers"
+  // — a caller pointing KICAD_SYMBOL_DIR at an isolated directory (tests, a
+  // pinned library set) must get only that directory back.
+  const kicadRoot = 'C:/Program Files/KiCad';
+  try {
+    if (fromEnv.length > 0) throw new Error('env override present; skip Windows auto-discovery');
+    const entries = await readdir(kicadRoot, { withFileTypes: true });
+    const versions = entries
+      .filter((e) => e.isDirectory())
+      .map((e) => e.name)
+      .sort((a, b) => b.localeCompare(a, undefined, { numeric: true }));
+    for (const version of versions) {
+      const dir = `${kicadRoot}/${version}/share/kicad/symbols`;
+      try {
+        await access(dir);
+        if (!out.includes(dir)) out.push(dir);
+      } catch {
+        // this version folder has no symbols dir; skip
+      }
+    }
+  } catch {
+    // C:\Program Files\KiCad doesn't exist on this machine (not Windows, or KiCad not installed here); skip
   }
   return out;
 }
@@ -82,6 +112,127 @@ export async function findLibraryFile(lib: string, dirs: string[]): Promise<stri
     }
   }
   return null;
+}
+
+/** Every `<name>.kicad_sym` file across the search dirs, first occurrence wins per name. */
+async function allLibraryFiles(dirs: string[]): Promise<{ lib: string; file: string }[]> {
+  const seen = new Set<string>();
+  const out: { lib: string; file: string }[] = [];
+  for (const dir of dirs) {
+    let entries: string[];
+    try {
+      entries = await readdir(dir);
+    } catch {
+      continue;
+    }
+    for (const entry of entries) {
+      if (!entry.endsWith('.kicad_sym')) continue;
+      const lib = entry.slice(0, -'.kicad_sym'.length);
+      if (seen.has(lib)) continue;
+      seen.add(lib);
+      out.push({ lib, file: path.join(dir, entry) });
+    }
+  }
+  return out;
+}
+
+export interface CrossLibraryMatch {
+  lib: string;
+  /** true when the part name matched exactly; false for a fuzzy hit. */
+  exact: boolean;
+}
+
+/** Levenshtein edit distance, capped: returns `cap + 1` once the true distance
+ * would exceed `cap`, so a caller doing a cheap "is this close enough" check
+ * never pays for the full O(n·m) table on two names that are obviously
+ * unrelated. */
+function editDistanceWithin(a: string, b: string, cap: number): number {
+  if (Math.abs(a.length - b.length) > cap) return cap + 1;
+  let prev = Array.from({ length: b.length + 1 }, (_, j) => j);
+  for (let i = 1; i <= a.length; i++) {
+    const cur = [i];
+    let rowMin = i;
+    for (let j = 1; j <= b.length; j++) {
+      const cost = a[i - 1] === b[j - 1] ? 0 : 1;
+      const v = Math.min(prev[j]! + 1, cur[j - 1]! + 1, prev[j - 1]! + cost);
+      cur.push(v);
+      if (v < rowMin) rowMin = v;
+    }
+    if (rowMin > cap) return cap + 1; // whole row exceeded cap: no cell can recover
+    prev = cur;
+  }
+  return prev[b.length]!;
+}
+
+/**
+ * A part is unresolvable at its guessed `lib_id`, but the guess itself is often
+ * right about the *part*, just wrong about which stock file it lives in (a
+ * chip's library nickname is not derivable from its part number). Rather than
+ * fail outright, check every other discovered `.kicad_sym` for a symbol with
+ * this exact name; when none matches exactly, fall back to the same
+ * closest-candidate substring heuristic `resolveLibrarySymbol` already uses
+ * within one file, just applied across all of them.
+ *
+ * Returns matches sorted exact-first; the caller decides what "found" means
+ * (one exact hit resolves unambiguously, several means the agent must pick).
+ */
+export async function findSymbolAcrossLibraries(
+  name: string,
+  dirs: string[],
+  excludeLib?: string,
+): Promise<CrossLibraryMatch[]> {
+  const files = await allLibraryFiles(dirs);
+  const q = name.toLowerCase();
+  const exact: CrossLibraryMatch[] = [];
+  const fuzzy: CrossLibraryMatch[] = [];
+  for (const { lib, file } of files) {
+    // `excludeLib` is only known not to hold this *exact* name; it can still
+    // hold the near-miss the caller actually wants (SHT40 guessed in
+    // Sensor_Humidity, where the real symbol is SHT4x). Skip its exact check,
+    // never its fuzzy one.
+    const skipExact = lib === excludeLib;
+    let text: string;
+    try {
+      text = await readFile(file, 'utf8');
+    } catch {
+      continue;
+    }
+    const names = [...text.matchAll(/^\s*\(symbol\s+"([^"]+)"/gm)].map((m) => m[1]!);
+    if (!skipExact && names.includes(name)) {
+      exact.push({ lib, exact: true });
+      continue;
+    }
+    // The "does the query contain the library name" direction is only
+    // meaningful for a reasonably specific name (>= 3 chars): a stock library
+    // is full of single-letter generics ("R", "C", "L", "D", "U", "Q"), and
+    // "does <any part number> contain the letter R" is true for nearly
+    // everything, which would fuzzy-match almost any query against them.
+    const MIN_SHORT_NAME_LEN = 3;
+    // Substring alone misses the most common real near-miss: a datasheet part
+    // number against a library's family name, differing mid-string rather than
+    // at either end (SHT40 vs SHT4x, where `x` stands for "any variant").
+    // Neither contains the other, so only edit distance catches it. Kept tight
+    // (1 edit, and only for names long enough that one edit is still
+    // distinctive) so TPS22860 does not silently match TPS22810 — a different
+    // real part, which is exactly the wrong-part failure this tool must not
+    // introduce.
+    const MIN_EDIT_MATCH_LEN = 5;
+    if (
+      names.some((n) => {
+        const lower = n.toLowerCase();
+        if (lower.includes(q)) return true;
+        if (lower.length >= MIN_SHORT_NAME_LEN && q.includes(lower)) return true;
+        return (
+          q.length >= MIN_EDIT_MATCH_LEN &&
+          lower.length >= MIN_EDIT_MATCH_LEN &&
+          editDistanceWithin(q, lower, 1) <= 1
+        );
+      })
+    ) {
+      fuzzy.push({ lib, exact: false });
+    }
+  }
+  return [...exact, ...fuzzy];
 }
 
 /** Collect pins (number, name, electrical type) from a `(symbol …)` node,
@@ -130,10 +281,15 @@ export async function resolveLibrarySymbol(
   | { status: 'ok'; pins: LibPin[] }
   | { status: 'no-symbol'; candidates: string[] }
   | { status: 'no-library' }
+  | { status: 'found-elsewhere'; libIds: string[] }
 > {
   const [lib, name] = libId.includes(':') ? [libId.slice(0, libId.indexOf(':')), libId.slice(libId.indexOf(':') + 1)] : ['', libId];
   const file = await findLibraryFile(lib, dirs);
-  if (!file) return { status: 'no-library' };
+  if (!file) {
+    const elsewhere = await findSymbolAcrossLibraries(name, dirs, lib);
+    if (elsewhere.length) return { status: 'found-elsewhere', libIds: elsewhere.map((m) => `${m.lib}:${name}`) };
+    return { status: 'no-library' };
+  }
   const root = parseSexp(await readFile(file, 'utf8'))[0];
   if (root === undefined || !isList(root)) return { status: 'no-library' };
   const symbols = librarySymbols(root);
@@ -152,7 +308,14 @@ export async function resolveLibrarySymbol(
     current = base;
   }
 
-  // exact name not found: offer near matches (case-insensitive substring both ways)
+  // Not in the guessed file: a genuine exact match elsewhere is a stronger
+  // signal than a same-file substring guess (the file name, not the part
+  // name, was the mistake), so it takes priority.
+  const elsewhere = await findSymbolAcrossLibraries(name, dirs, lib);
+  if (elsewhere.length) return { status: 'found-elsewhere', libIds: elsewhere.map((m) => `${m.lib}:${name}`) };
+
+  // exact name not found anywhere: offer near matches within the guessed file
+  // (case-insensitive substring both ways)
   const q = name.toLowerCase();
   const candidates = [...symbols.keys()]
     .filter((k) => {
@@ -216,6 +379,14 @@ export async function verifySchematicSymbols(
         detail: resolved.candidates.length
           ? `"${entry.libId}" does not exist in the installed library — closest real symbols: ${resolved.candidates.join(', ')}. Use one of these lib_ids (KiCad renames symbols across versions).`
           : `"${entry.libId}" does not exist in the installed library and no close match was found; confirm the lib_id.`,
+      });
+      continue;
+    }
+    if (resolved.status === 'found-elsewhere') {
+      findings.push({
+        libId: entry.libId,
+        kind: 'no-library',
+        detail: `"${entry.libId}" names the wrong library, not the wrong part: use ${resolved.libIds.join(' or ')} instead.`,
       });
       continue;
     }
--- a/src/kicad/symlib.ts
+++ b/src/kicad/symlib.ts
@@ -44,8 +44,13 @@ export interface LibPin {
  * Candidate directories holding KiCad's stock `.kicad_sym` libraries, most
  * specific first. Env overrides win (KiCad exports these), then the standard
  * install locations for Linux/macOS/Windows. Only existing dirs are returned.
+ *
+ * @param winRoot test seam for the Windows `C:\Program Files\KiCad` root;
+ *   defaults to the real path. Without this, the Windows version-directory
+ *   discovery below can only be exercised on an actual Windows machine with
+ *   KiCad installed at the default location.
  */
-export async function symbolSearchDirs(env = process.env): Promise<string[]> {
+export async function symbolSearchDirs(env = process.env, winRoot = 'C:/Program Files/KiCad'): Promise<string[]> {
   const fromEnv = [
     env.KICAD_SYMBOL_DIR,
     env.KICAD10_SYMBOL_DIR,
@@ -77,7 +82,7 @@ export async function symbolSearchDirs(env = process.env): Promise<string[]> {
   // exactly that, not "env overrides win, plus whatever else this discovers"
   // — a caller pointing KICAD_SYMBOL_DIR at an isolated directory (tests, a
   // pinned library set) must get only that directory back.
-  const kicadRoot = 'C:/Program Files/KiCad';
+  const kicadRoot = winRoot;
   try {
     if (fromEnv.length > 0) throw new Error('env override present; skip Windows auto-discovery');
     const entries = await readdir(kicadRoot, { withFileTypes: true });
@@ -138,7 +143,12 @@ async function allLibraryFiles(dirs: string[]): Promise<{ lib: string; file: str
 
 export interface CrossLibraryMatch {
   lib: string;
-  /** true when the part name matched exactly; false for a fuzzy hit. */
+  /** The symbol name actually found (may differ from the query on a fuzzy hit,
+   * e.g. `SHT4x` for a query of `SHT40`) — the suggested lib_id must be built
+   * from this, never from the caller's original query, or a fuzzy suggestion
+   * points at a lib_id that does not exist. */
+  name: string;
+  /** true when `name` matched exactly; false for a fuzzy hit. */
   exact: boolean;
 }
 
@@ -199,7 +209,7 @@ export async function findSymbolAcrossLibraries(
     }
     const names = [...text.matchAll(/^\s*\(symbol\s+"([^"]+)"/gm)].map((m) => m[1]!);
     if (!skipExact && names.includes(name)) {
-      exact.push({ lib, exact: true });
+      exact.push({ lib, name, exact: true });
       continue;
     }
     // The "does the query contain the library name" direction is only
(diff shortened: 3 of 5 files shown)


thegoodengineer:
🤖 Automated pr-review pass (no human review implied).

## Verdict: Request changes

The cross-library fallback works for the exact-match case, but its fuzzy-match case is broken in a way the PR's own flagship example (`SHT40`/`SHT4x`) demonstrates once reproduced: the "corrected" `lib_id` it suggests is often itself unresolvable, and one instance of the exact self-suggestion regression the PR says it fixed (`Device:R_Sma` suggesting itself) is still live in the `symlib.ts` path used by `verify_symbols`.

## Metrics block

```
lines: +208 / -7 (net 201) across 2 files · src +208/-7, test +0/-0, docs +0/-0, spec +0/-0, other +0/-0
tests: +0 test lines; suite 702 pass / 20 skip-or-fail-mix -> see note below (measured manually, head only)
diff coverage: not measured by script (missing @vitest/coverage-v8, no network to install); manual trace below
deps: none (package.json unchanged)
```

Method notes:
- Change size / new-vs-net: from the metrics script (`node .claude/skills/pr-review/scripts/metrics.mjs 186`), merge base `1aa7c98` (`git merge-base origin/feat/deterministic-schematic-drafting HEAD`).
- Suite: script's own run produced no JSON (environment quirk), so I ran `npx vitest run --reporter=json` myself: **702 passed / 20 failed / 20 skipped** of 742. All 20 failures are pre-existing, environment-specific (CRLF vs LF, Windows path separators, `kicad-cli` absent from this sandbox's PATH), matching the categories the PR itself lists ("Windows path handling, chmod semantics, app-bundle probing"). None touch `symlib.ts`/`symsource.ts` or symbol resolution — `test/symlib.test.ts` and `test/draft-hardening.test.ts`'s cross-library-relevant tests all pass. (Count differs from the PR's claimed 16 because this sandbox appears to be missing `kicad-cli` entirely, not just partially; not attributable to this diff.)
- Diff coverage: `@vitest/coverage-v8` isn't installed and there's no network access in this session to add it, so no percentage. I instead traced the new exported surface and branches by hand (see Finding 3 and the untested-surface list below) — this is the manual fallback the skill's metrics reference describes.
- CI: `gh pr checks 186` — all 5 real checks pass (CodeRabbit is skipped by repo policy for non-default base branches, expected for a PR targeting a feature branch).
- `mergeable: MERGEABLE`, `isDraft: false`.

## Invariant checks performed

- **Spec-gated in**: N/A, unaffected — no tool-exposure change. Confirmed by reading the diff: only `src/kicad/draft/symsource.ts` and `src/kicad/symlib.ts` changed, neither touches the tool list in `src/agent/tools.ts` (`verify_symbols`'s handler is unchanged; only the module it calls into changed).
- **Verification-gated out**: N/A, unaffected — no ERC/DRC gate or rollback logic touched.
- **`check`/`verify` stays LLM-free and network-free**: **passed**. The new code (`findSymbolAcrossLibraries`, `allLibraryFiles`, the Windows version-directory discovery) is filesystem-only (`readdir`/`readFile`/`access`); no provider, key, or network call added. (Note: `verifySchematicSymbols`, which the new code affects, is reached via the `verify_symbols` *agent tool*, not the deterministic `check` command itself — `src/commands/check` was not touched.)
- **No sexp serialization**: **passed**. All new matching is done against raw library text via regex (`text.matchAll(/^\s*\(symbol\s+"([^"]+)"/gm)`), matching the existing same-file candidate heuristic's style; nothing is parsed-and-rewritten, nothing is written to disk by the new code.
- **Sync-obligations ledger**: N/A, unaffected — no hook or ledger code touched.
- **Secrets**: N/A, unaffected — no credential handling, transcript, or `.gitignore` change.
- **Spec coherence**: no spec-level behavior declared changed by this PR, and I didn't find any — `openspec validate build-copperhead-phase-1` wasn't run since this PR targets `feat/deterministic-schematic-drafting`, not `build-copperhead-phase-1`'s tree, and touches no planning artifacts.

## Findings

### Finding 1 [high] — Fuzzy cross-library suggestions reconstruct the lib_id from the original (wrong) query name, not the symbol that actually matched, so the "fix" offered is frequently itself unresolvable

**File**: src/kicad/symlib.ts:139-236 (root cause: `CrossLibraryMatch` never records the matched symbol name), consumed at src/kicad/draft/symsource.ts:212-217 and src/kicad/symlib.ts:289-291,314-315,385-391

**Fails when**: a part is queried under the wrong library but the real symbol in another library has a *different spelling* (a fuzzy, not exact, cross-library hit) — precisely the PR's own headline scenario. Reproduced directly against `SymbolSource.resolve` (the drafting engine's primary, mainline symbol-lookup path used by every `create`/draft run):

```
guessed lib_id: Sensor_Wrong:SHT40   (Sensor_Wrong.kicad_sym exists but has nothing related)
another library "Sensor_Humidity.kicad_sym" contains the real symbol "SHT4x" (edit distance 1 from "SHT40")

SymbolSource.resolve('Sensor_Wrong:SHT40') throws:
  reason: 'found-elsewhere'
  candidates: [ 'Sensor_Humidity:SHT40' ]
  message: "Sensor_Wrong:SHT40" is wrong about the library, not the part: use Sensor_Humidity:SHT40 instead

Re-resolving the suggested lib_id "Sensor_Humidity:SHT40" throws the *identical* error again,
because "SHT40" does not exist in Sensor_Humidity — only "SHT4x" does. The tool is suggesting a
lib_id that does not exist, which is worse than the "library not installed and not vendored"
message this PR was written to replace: it looks resolved but is not, and would burn exactly the
turn budget this PR exists to save.
```

`findSymbolAcrossLibraries` (symlib.ts:139-236) finds the right *library* via `editDistanceWithin`/substring matching against each candidate symbol name, but only records `{ lib, exact }` (symlib.ts:139-143) — it never keeps the actual symbol name (`n`, symlib.ts:221-232) that matched. Every consumer then reconstructs the suggestion as `` `${m.lib}:${name}` `` using the caller's original (wrong) `name` (symsource.ts:216; symlib.ts:291,315,389) instead of the real matched name. This is correct only for exact hits (where `name` provably exists in that library); for fuzzy hits it is wrong by construction.

**Fix**: add the matched symbol name to `CrossLibraryMatch` (e.g. `{ lib, name, exact }`, populated from `n` in the `.some()` callback at symlib.ts:221 — track which candidate string actually matched, not just that one did), and have every consumer (symsource.ts's `crossLibrarySuggestions`, and both `elsewhere.map(...)` call sites in symlib.ts) build the suggested lib_id from `` `${m.lib}:${m.name}` ``, never the original query `name`. Add a regression test (e.g. in test/symlib.test.ts or test/draft-hardening.test.ts) that sets up two library files — one guessed-but-wrong, one holding a fuzzy near-miss under a different symbol name — and asserts the returned suggestion, when re-resolved, actually succeeds.

### Finding 2 [high] — `resolveLibrarySymbol` checks cross-library evidence before same-file substring candidates, the reverse of the PR's stated priority order, reproducing the exact self-suggestion regression the PR says it fixed

**File**: src/kicad/symlib.ts:311-326 (used by `verifySchematicSymbols` at symlib.ts:365, which backs the `verify_symbols` agent tool in src/agent/tools.ts:344-365)

**Fails when**: the guessed lib_id has a genuine same-file typo (the caller named the right library and misspelled the part) *and* that same misspelling also fuzzy-matches within the guessed library itself via the cross-library scan (which, per symlib.ts:189-193's own comment, deliberately does not exclude the guessed library from its fuzzy check). Reproduced directly:

```
Device.kicad_sym contains "R" and "R_Small" (same fixture as test/symlib.test.ts).

resolveLibrarySymbol('Device:R_Sma', [libDir])   // "R_Sma" is a typo of "R_Small"
  -> { status: 'found-elsewhere', libIds: [ 'Device:R_Sma' ] }
```

The function returns "use Device:R_Sma instead" of Device:R_Sma — the exact same, already-failing lib_id. This is the identical shape of regression the PR's description says it found and fixed via "evidence ordering" (own-file candidates before cross-library) and covered with a test at test/draft-hardening.test.ts:635 — but that fix and test only cover `SymbolSource.resolve` (symsource.ts), which correctly checks own-file candidates first (symsource.ts:323-336) before ever calling `crossLibrarySuggestions`. The sibling implementation `resolveLibrarySymbol` (symlib.ts) was not given the same ordering: it calls `findSymbolAcrossLibraries` (symlib.ts:314) and returns on any non-empty result *before* computing the same-file substring candidates (symlib.ts:319-326), so a same-file fuzzy hit via the cross-library scan pre-empts a correct same-file substring answer, and (compounded by Finding 1) reconstructs the bogus self-referential suggestion.

**Fix**: mirror the ordering already correct in symsource.ts: compute the same-file substring `candidates` first; only fall through to `findSymbolAcrossLibraries` when `candidates` is empty. Add a test mirroring test/draft-hardening.test.ts:635 but calling `resolveLibrarySymbol` (or `verifySchematicSymbols`) directly, asserting `Device:R_Sma` returns `no-symbol` with `R_Small` as a candidate, not `found-elsewhere` pointing at itself.

### Finding 3 [medium] — The new cross-library resolution surface has zero automated test coverage; both findings above went undetected because of it

**File**: src/kicad/symlib.ts:139-236 (`findSymbolAcrossLibraries`, `editDistanceWithin`, `allLibraryFiles`), src/kicad/symlib.ts:284,288-291,314-315,385-391 (`found-elsewhere` branches), src/kicad/draft/symsource.ts:186-217 (`found-elsewhere` reason and `crossLibrarySuggestions`), src/kicad/symlib.ts:72-99 (Windows version-directory discovery)

**Fails when**: any future change touches this logic — there is no automated signal it would break, only the manual verification table in the PR description (which, per Finding 1, does not actually match the code's behavior for the SHT40 case it documents). Confirmed by search: no test file references `found-elsewhere`, `findSymbolAcrossLibraries`, `SHT40`, `TPS63900`, `TPS22860`, `FDC1004`, or `symbolSearchDirs` anywhere in test/. The one new regression test added (test/draft-hardening.test.ts:635) exercises only the same-file substring-candidate path of `SymbolSource.resolve`, never the cross-library `found-elsewhere` path in either implementation, and test/symlib.test.ts's fixture (test/symlib.test.ts:53-61) only ever points at a single library directory, so `findSymbolAcrossLibraries`'s actual multi-library behavior is never exercised at all.

**Fix**: add tests covering, at minimum: (1) an exact cross-library hit resolving correctly in both `SymbolSource.resolve` and `resolveLibrarySymbol`/`verifySchematicSymbols`; (2) a fuzzy cross-library hit under a *different* symbol name (the SHT40/SHT4x shape), asserting the returned suggestion is itself resolvable (this is what would have caught Finding 1); (3) the same-file-typo-that-also-fuzzy-matches-cross-library case (this is what would have caught Finding 2); (4) `symbolSearchDirs` with a faked `C:/Program Files/KiCad/<version>/share/kicad/symbols` tree, both with and without an env override present.

## Untested-surface list (reconciles with Finding 3)

- src/kicad/symlib.ts:139-236 — `findSymbolAcrossLibraries`, `editDistanceWithin`, `allLibraryFiles`: no test reaches any of these directly or indirectly with more than one library file present.
- src/kicad/symlib.ts:284, 288-291, 314-315 — the `found-elsewhere` branches of `resolveLibrarySymbol`: never returned by any existing test.
- src/kicad/symlib.ts:385-391 — the `found-elsewhere` branch of `verifySchematicSymbols`: never reached.
- src/kicad/draft/symsource.ts:197-198, 212-217, 317-318, 332-333 — the `found-elsewhere` reason and `crossLibrarySuggestions`: never returned by any existing test (test/draft-hardening.test.ts:635 only reaches the `no-symbol` branch).
- src/kicad/symlib.ts:80-99 — Windows `C:/Program Files/KiCad` version-directory auto-discovery, and its env-override skip: no test at all.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is #186, branch feat/deterministic-schematic-drafting
(Chirag6722 fork), based on feat/deterministic-schematic-drafting.

Check it out first:
  gh pr checkout 186

FINDING 1 [high] Fuzzy cross-library suggestions reconstruct the lib_id from the
original (wrong) query name, not the symbol that actually matched, so the "fix"
offered is frequently itself unresolvable.
  Where: src/kicad/symlib.ts:139-236 (CrossLibraryMatch / findSymbolAcrossLibraries),
    consumed at src/kicad/draft/symsource.ts:212-217 and
    src/kicad/symlib.ts:289-291,314-315,385-391
  Fails when: a part is queried under the wrong library but the real symbol in
    another library has a different spelling (a fuzzy, not exact, cross-library
    hit) -- the PR's own headline scenario. Reproduced against SymbolSource.resolve:
    guessing "Sensor_Wrong:SHT40" when the real symbol is "SHT4x" in
    "Sensor_Humidity.kicad_sym" makes it throw reason 'found-elsewhere' with
    candidates ["Sensor_Humidity:SHT40"] -- re-resolving that suggestion throws the
    identical error again, because "SHT40" does not exist in Sensor_Humidity, only
    "SHT4x" does.
  Fix: add the matched symbol name to CrossLibraryMatch (e.g. { lib, name, exact },
    populated from the actual candidate string matched inside the .some() callback
    around symlib.ts:221-232, not just a boolean of whether one did). Update every
    consumer (symsource.ts's crossLibrarySuggestions, and both elsewhere.map(...)
    call sites in symlib.ts) to build the suggested lib_id from `${m.lib}:${m.name}`,
    never the original query name. Add a regression test with two library files --
    one guessed-but-wrong, one holding a fuzzy near-miss under a different symbol
    name -- asserting the returned suggestion, when re-resolved, actually succeeds.

FINDING 2 [high] resolveLibrarySymbol checks cross-library evidence before
same-file substring candidates, the reverse of the PR's stated priority order,
reproducing the exact self-suggestion regression the PR says it fixed.
  Where: src/kicad/symlib.ts:311-326, used by verifySchematicSymbols
    (symlib.ts:365), which backs the verify_symbols agent tool
    (src/agent/tools.ts:344-365)
  Fails when: the guessed lib_id has a same-file typo that also fuzzy-matches
    within the guessed library itself via the cross-library scan (which does not
    exclude the guessed library from its fuzzy check, symlib.ts:189-193). Reproduced:
    with Device.kicad_sym containing "R" and "R_Small",
    resolveLibrarySymbol('Device:R_Sma', [libDir]) returns
    { status: 'found-elsewhere', libIds: ['Device:R_Sma'] } -- "use Device:R_Sma
    instead" of Device:R_Sma, the exact same failing lib_id. This is the identical
    regression shape the PR describes fixing via evidence ordering and covering at
    test/draft-hardening.test.ts:635 -- but that fix only applies to
    SymbolSource.resolve (symsource.ts:323-336 checks same-file candidates first);
    the sibling resolveLibrarySymbol was not given the same ordering.
  Fix: mirror the ordering already correct in symsource.ts: compute the same-file
    substring candidates first (symlib.ts:319-326); only fall through to
    findSymbolAcrossLibraries when candidates is empty. Add a test mirroring
    test/draft-hardening.test.ts:635 but calling resolveLibrarySymbol (or
    verifySchematicSymbols) directly, asserting Device:R_Sma returns 'no-symbol'
    with R_Small as a candidate, not 'found-elsewhere' pointing at itself.

FINDING 3 [medium] The new cross-library resolution surface has zero automated
test coverage; both findings above went undetected because of it.
  Where: src/kicad/symlib.ts:139-236 (findSymbolAcrossLibraries,
    editDistanceWithin, allLibraryFiles), src/kicad/symlib.ts:284,288-291,314-315,
    385-391 (found-elsewhere branches), src/kicad/draft/symsource.ts:186-217
    (found-elsewhere reason and crossLibrarySuggestions),
    src/kicad/symlib.ts:72-99 (Windows version-directory discovery)
  Fails when: any future change touches this logic with no automated signal it
    broke something. Confirmed: no test references found-elsewhere,
    findSymbolAcrossLibraries, SHT40, TPS63900, TPS22860, FDC1004, or
    symbolSearchDirs anywhere in test/. test/draft-hardening.test.ts:635 only
    exercises the same-file substring-candidate path; test/symlib.test.ts's
    fixture (lines 53-61) only ever points at a single library directory.
  Fix: add tests covering: (1) an exact cross-library hit resolving correctly in
    both SymbolSource.resolve and resolveLibrarySymbol/verifySchematicSymbols;
    (2) a fuzzy cross-library hit under a different symbol name (the SHT40/SHT4x
    shape), asserting the returned suggestion is itself resolvable; (3) the
    same-file-typo-that-also-fuzzy-matches-cross-library case; (4) symbolSearchDirs
    with a faked C:/Program Files/KiCad/<version>/share/kicad/symbols tree, with
    and without an env override present.

These lines in the diff are reached by no test: src/kicad/symlib.ts:139-236,
src/kicad/symlib.ts:284,288-291,314-315,385-391, src/kicad/draft/symsource.ts:186-217,
src/kicad/symlib.ts:72-99. Add tests for the new exported symbols, branches, and
error paths they contain.

    by prompt text.

Then verify:
  npm run typecheck
  npm run build
  npm test

and report it.

````

**Comment by Chirag6722** [comment]
Both high-severity findings confirmed and fixed in 6e3442b.

**Finding 1** — reproduced exactly as described before touching anything: guessing `Sensor_Wrong:SHT40` correctly located `Sensor_Humidity`, but suggested `Sensor_Humidity:SHT40`, which doesn't exist (`Sensor_Humidity`'s only symbol is `SHT4x`). Re-resolving the suggestion threw the identical error. This is very likely the actual cause of a live run circling `SHT40`/`TPS22860` searches for many turns before hitting the session limit, rather than a slow-but-converging search as I'd assumed. Fixed by recording the matched symbol name on `CrossLibraryMatch` and building every suggestion from it, not the caller's original query. Re-verified after the fix: the suggestion is now `Sensor_Humidity:SHT4x` and resolves in one hop.

**Finding 2** — also reproduced exactly: `Device:R_Sma` returned `found-elsewhere` pointing at `Device:R_Sma`, itself. Reordered `resolveLibrarySymbol` to check same-file candidates first, matching what `SymbolSource.resolve` already did correctly. Re-verified: now returns `no-symbol` with `R_Small` as a candidate.

**Finding 3** — added the four tests requested: exact and fuzzy cross-library hits (the fuzzy one asserts the suggestion is itself resolvable, which is what would have caught Finding 1), the same-file-typo-that-also-fuzzy-matches ordering case (Finding 2), and `symbolSearchDirs` against a faked Windows version-directory tree with and without an env override. Added an optional `winRoot` param (defaults to the real path) so that discovery path is deterministically testable rather than depending on the machine actually being Windows with KiCad installed at the default location.

**Testing**: didn't rely on the earlier PR description's baseline claim. Re-ran the comparison properly this time: `git stash` the fix, full suite on the clean tree (19 failures), `git stash pop`, full suite again (18 failures). Diffed the two failure lists directly rather than just comparing counts — zero new failures introduced, 9 new tests added and passing. One baseline failure happens to be fixed by this commit too (incidental: the env-override handling both changes touch). The full-suite run is noisy under parallel load (one `repl.test.ts` case appeared in one run and not another); confirmed that one is pre-existing flakiness by running it in isolation twice, unrelated to this diff either way.

**Comment by animesh-chouhan** [comment]
🤖 Automated pr-review pass (no human review implied).

This is an **incremental pass**: a prior automated review on 2026-08-06 raised three findings, and commit `6e3442b` answers all three. This report verifies those fixes and reviews what the fix commit changed. It does not repeat the earlier report.

## Verdict: request changes

Findings 1, 2 and 3 from the previous pass are **confirmed fixed**. Verified against the real stock library at `/usr/share/kicad/symbols`, all four cases from the PR's own table now behave as the description claims:

```
Regulator_DCDC:TPS63900  -> Regulator_Switching:TPS63900   (exact, cross-library)
Sensor_Humidity:SHT40    -> Sensor_Humidity:SHT4x          (fuzzy, real symbol name carried)
Sensor:FDC1004           -> Sensor_Proximity:FDC1004DGS    (fuzzy, resolves in one hop)
Power_Switch:TPS22860    -> no-library                     (correct: genuinely absent)
Device:R_Sma             -> no-symbol, candidate R_Small   (no longer suggests itself)
```

The reason for request-changes is new: the reordering that fixed old-Finding 2 introduced a regression of its own, and one consequence of the cross-library search reaches the default drafting path with wrong advice.

## Metrics block

```
lines: +395 / -9 (net 386) across 4 files  ·  src +224/-8, test +171/-1, docs +0/-0, spec +0/-0, other +0/-0
new vs net: 0 new files (0 in src), 4 modified; 0 new src lines (the net-new surface)
tests: +171 test lines; suite not run at base -> 728 pass / 20 skip / 3 fail at head
diff coverage: not measured (v8 coverage emitted no report on two attempts); branch reachability measured by instrumentation instead, see the untested-surface list
deps: none (package.json unchanged)
```

Method notes:

- Change size, new-vs-net, CI: `node .claude/skills/pr-review/scripts/metrics.mjs 186`, merge base `1aa7c98` (`git merge-base origin/feat/deterministic-schematic-drafting HEAD`). `mergeable: MERGEABLE`, `isDraft: false`.
- Suite: `COPPERHEAD_MIN_FREE_MB=100 npx vitest run` at head, twice, same result: 728 pass / 20 skip / 3 fail of 751. **All 3 failures are new tests added by this PR** and are machine-dependent, see Finding 3. Base suite not run (`--base-tests` not used); base CI does not report it for this feature branch.
- Diff coverage: `@vitest/coverage-v8` is installed, but `--coverage.provider=v8 --coverage.reporter=json` produced no `coverage-final.json` on two attempts, so no percentage is claimed. Instead of guessing, I instrumented the three new fallback branches with markers and ran the whole suite to record which ones execute. Results in the untested-surface list.
- All resolution behavior below was reproduced by running the real code (`resolveLibrarySymbol`, `SymbolSource.resolve`, `verifySchematicSymbols`) against the machine's real KiCad 10 library, and against `fd1a14c` in a throwaway worktree to separate regressions from pre-existing behavior.

## Invariant checks performed

- **Spec-gated in**: passed, unaffected. Only `src/kicad/symlib.ts` and `src/kicad/draft/symsource.ts` changed in src. `verify_symbols` keeps `requiresUnlock: false` (it is read-only), and no `edit_file`/`write_file` exposure changed.
- **Verification-gated out**: passed, unaffected. No ERC/DRC gate, repair loop, or rollback path touched.
- **`check`/`verify` stays LLM-free and network-free**: passed. Confirmed `src/commands/check.ts` does not import `symlib` or `symsource` (grep over `src/commands/`); `verifySchematicSymbols` is reached only through the `verify_symbols` agent tool at `src/agent/tools.ts:355`. All new code is filesystem-only (`readdir`/`readFile`/`access`).
- **No sexp serialization**: passed. Cross-library scanning reads library text and matches symbol names with a regex; `parseSexp` is still read-only and nothing is round-tripped or written.
- **Sync-obligations ledger**: passed, unaffected. No hook or ledger code touched.
- **Secrets**: passed, unaffected. No credential handling, transcript writing, or `.gitignore` change.
- **Spec coherence**: no spec-level change required. Grep over `openspec/` shows `verify_symbols` is specified by its output shape and by AC-16.5 (vendoring vs installed-library comparison), not by its resolution heuristics. Note that Finding 1 below is in tension with AC-16.5's intent, but the fix for it is code, not spec.

## Findings

### Finding 1 [high] `verify_symbols` now tells the agent to replace engine-generated vendored power symbols with stock ones

**File**: src/kicad/symlib.ts:400-406 (the `found-elsewhere` branch of `verifySchematicSymbols`), reached through src/kicad/symlib.ts:302-305

**Fails when**: any drafted project is verified, which is every project the drafting engine produces. The engine generates its own power symbols and vendors them under the `copperhead_power` nickname (`SymbolSource.vendorGenerated`, src/kicad/draft/symsource.ts:248). That nickname is deliberately not an installed library, so `findLibraryFile` misses it, and the new cross-library search then finds the *stock* symbol of the same name and reports the vendored one as a wrong library nickname.

Reproduced by instrumenting the branch and running the existing suite. `test/draft-engine.test.ts` "hermetic symbol vendoring (AC-16.5)" hits it three times per run:

```
verify_symbols: 0 verified, 0 unverifiable (library not installed), 0 issue(s) to reconcile:
  - [no-library] "copperhead_power:GND" names the wrong library, not the wrong part: use power:GND instead.
  - [no-library] "copperhead_power:PWR_FLAG" names the wrong library, not the wrong part: use power:PWR_FLAG instead.
  - [no-library] "copperhead_power:VCC" names the wrong library, not the wrong part: use power:VCC instead.
```

Before this PR these three were reported as `no-library` ("library is not installed on this machine; cannot verify its pins") and counted as `skipped`, which was accurate. Now the tool states, with the confidence of an exact match, that the agent should switch to `power:GND`. Acting on that advice either churns for nothing or breaks the deterministic vendoring guarantee in AC-16.5, since drafted output is supposed to come from the vendored cache. The existing test still passes because it asserts only on the `Device:R` `pin-count` finding, so nothing catches this.

**Fix**: before treating a missing library as a wrong nickname, check whether that nickname is vendored in the project's `sym-lib-cache/`. If it is, the library is absent on purpose: keep the existing `no-library` finding ("cannot verify its pins") and the `skipped` count, and never emit `found-elsewhere` for it. This keeps AC-16.5 intact, since a vendored *stock* symbol still carries its real nickname (`Device`), resolves against the installed library, and its drift stays visible. `verifySchematicSymbols` currently takes only `schPath` and `env`, so it needs the project root (or the cache dir) threaded in from `src/agent/tools.ts:355`. Add a regression test: draft a project, run `verifySchematicSymbols` on it, and assert no finding names a `copperhead_power:` lib_id as a wrong library.

### Finding 2 [medium] the reorder that fixed the self-suggestion made junk one-character same-file candidates pre-empt an exact cross-library hit

**File**: src/kicad/symlib.ts:329-339 (introduced by 6e3442b), same shape at src/kicad/draft/symsource.ts:330-336

**Fails when**: the guessed library exists but the part lives elsewhere, and the guessed library happens to contain a short generic symbol whose name is a substring of the query. `Device.kicad_sym` is full of single-letter generics (`R`, `C`, `L`, `D`), so this is common rather than exotic. The same-file candidate filter at symlib.ts:330-335 has no minimum-length guard, unlike the cross-library fuzzy path which added `MIN_SHORT_NAME_LEN = 3` (symlib.ts:220, 238) for exactly this reason. Now that same-file candidates return early, one of those junk candidates suppresses a correct exact answer.

Reproduced against the real stock library:

```
Device:USB_C_Receptacle_USB2.0_14P
  head (6e3442b): no-symbol, candidates = [C, L, R]
  fd1a14c:        found-elsewhere -> Connector:USB_C_Receptacle_USB2.0_14P

Device:LM358
  head (6e3442b): no-symbol, candidates = [L]
  fd1a14c:        found-elsewhere -> Amplifier_Operational:LM358
```

`USB_C_Receptacle_USB2.0_14P` is the exact symbol `symlib.ts:15`'s own module doc uses as the motivating example for this checker, and both parts are the shape of miss the PR exists to fix. This is a regression introduced by the fix commit, verified by running both commits against the same library.

**Fix**: check for an *exact* cross-library hit before the same-file candidates, and keep same-file candidates ahead of cross-library *fuzzy* hits. That ordering is what actually fixes the old self-suggestion regression (which was caused by a fuzzy hit, not an exact one), and it is honest about evidence strength. Verified locally in a scratch worktree:

```js
const elsewhere = await findSymbolAcrossLibraries(name, dirs, lib);
const exactElsewhere = elsewhere.filter((m) => m.exact);
if (exactElsewhere.length) return { status: 'found-elsewhere', libIds: exactElsewhere.map((m) => `${m.lib}:${m.name}`) };
if (candidates.length) return { status: 'no-symbol', candidates };
if (elsewhere.length) return { status: 'found-elsewhere', libIds: elsewhere.slice(0, 8).map((m) => `${m.lib}:${m.name}`) };
```

With that change (mirrored in `symsource.ts`) all five probe cases resolve correctly, and both regression tests this PR added still pass: `test/symlib.test.ts:161` (`Device:R_Sma` still returns `no-symbol` with `R_Small`) and `test/draft-hardening.test.ts` (the `SHT40`/`SHT4x` case). Measured: 42 passed, and the only failures left were the three in Finding 3. Note that a plain minimum-length guard on the same-file filter is *not* a sufficient fix on its own: it breaks the existing `test/symlib.test.ts:83` expectation that `Device:R_Nonexistent` offers `R`.

### Finding 3 [medium] the three new `symbolSearchDirs` tests fail on any machine that has KiCad installed

**File**: test/symlib.test.ts:203, 208, 220

**Fails when**: `/usr/share/kicad/symbols` (or `/usr/local/share/kicad/symbols`, or the macOS bundle path) exists, which is true on any developer machine that can actually run this tool. All three assert against the whole return value (`expect(dirs).toEqual([versionDir])`, `expect(dirs[0]).toBe(versionDir)`), but `symbolSearchDirs` prepends every existing default path before the discovered Windows dirs. Reproduced at head on Linux with KiCad 10 installed:

```
FAIL test/symlib.test.ts > discovers a version-numbered directory when no env override is set
  expected [ '/usr/share/kicad/symbols', …(1) ] to deeply equal [ Array(1) ]
FAIL test/symlib.test.ts > prefers the newest version when more than one is installed
  expected '/usr/share/kicad/symbols' to be '/tmp/copperhead-kicadroot-TtQswl/10.0…'
FAIL test/symlib.test.ts > an env override skips Windows auto-discovery entirely, even when set
  expected [ …(2) ] to deeply equal [ Array(1) ]
```

CI is green only because the runner has no KiCad installed, so these tests assert nothing there beyond "the machine has no KiCad". The third one is the regression guard for one of the two `symbolSearchDirs` bugs this PR fixes, and it does not guard it on the machines that matter.

**Fix**: assert on the discovered subset rather than the whole list. For the first two, filter to paths under `root` and assert on that (`expect(dirs.filter((d) => d.startsWith(root)))`), preserving the ordering assertion. For the env-override test, point `KICAD_SYMBOL_DIR` at a separate temp directory and assert both `expect(dirs).toContain(envDir)` and `expect(dirs.some((d) => d.startsWith(root))).toBe(false)`, which states the actual contract ("auto-discovery did not run") independently of what else is installed.

### Finding 4 [medium] `resolveLibrarySymbol` returns exact and fuzzy cross-library matches mixed together and uncapped

**File**: src/kicad/symlib.ts:304 and src/kicad/symlib.ts:339

**Fails when**: the query name is short, or a fuzzy match exists alongside an exact one. `crossLibrarySuggestions` in symsource.ts:212-217 collapses to exact-only when any exact hit exists, but the two `elsewhere.map(...)` call sites in symlib.ts return every match with no filter and no cap. `findSymbolAcrossLibraries` fuzzy-matches on `lower.includes(q)`, and the `MIN_SHORT_NAME_LEN` guard covers only the opposite direction, so a one-character query matches nearly every library. Reproduced against the real stock library:

```
resolveLibrarySymbol('Nowhere:R')      -> 143 lib_ids, a 4,420-character verify_symbols finding
                                          [Device:R, 74xGxx:74LVC1GU04DRL, 74xx:SN74AVC16827DGGR, ...]
resolveLibrarySymbol('Nowhere:LM358')  -> [Amplifier_Operational:LM358, Regulator_Switching:LM3578]
```

The second case is the wrong-part hazard the PR's design section says the 1-edit cap prevents. The cap stops the substitution example (`TPS22860` vs `TPS22810`) but not insertions: `LM358` and `LM3578` are one edit apart, both at least 5 characters, and they are different real parts. Surfacing `Regulator_Switching:LM3578` next to the correct answer invites the agent to pick the wrong chip, and the 4.4 KB variant floods the very turn budget this PR is meant to save.

**Fix**: apply the same exact-first collapse the symsource path already uses, and cap the fuzzy list (8 is the cap the same-file candidate list already uses at symlib.ts:335). The snippet in Finding 2 does both. Add a test asserting that a query with an exact hit in one library and a fuzzy hit in another returns only the exact one.

### Finding 5 [low] `verify_symbols` reports "0 issue(s) to reconcile" while listing an issue

**File**: src/kicad/symlib.ts:400-406, consumed at src/agent/tools.ts:360-363

**Fails when**: any `found-elsewhere` finding is produced. The branch tags the finding `kind: 'no-library'` and does not increment `skipped`, so the tool's summary line counts it in neither bucket: `mismatches` filters out `kind !== 'no-library'`, and `skipped` never saw it. Reproduced:

```
verify_symbols: 0 verified, 0 unverifiable (library not installed), 0 issue(s) to reconcile:
  - [no-library] "Nowhere:LM358" names the wrong library, not the wrong part: use Amplifier_Operational:LM358 or ...
```

A header that says there is nothing to reconcile invites the agent to stop reading, which is the opposite of what this finding wants.

**Fix**: give `found-elsewhere` its own `kind` (for example `'wrong-library'`) on `SymbolFinding` and include it in the `mismatches` count in src/agent/tools.ts:362, so the header agrees with the list. Assert the header count in a test.

### Finding 6 [low] the wrong-nickname path through `resolveLibrarySymbol` where the guessed library exists is reached by no test

**File**: src/kicad/symlib.ts:338-339

**Fails when**: never observed, because nothing exercises it. Measured, not assumed: I added a marker at that line and ran the full suite; it printed zero times, while the equivalent `symsource.ts` fallback printed once and the `verifySchematicSymbols` `found-elsewhere` branch printed three times. This is exactly the shape of the PR's flagship case (`Sensor_Humidity:SHT40`, where the guessed library exists and holds the near-miss) as it goes through the `verify_symbols` implementation rather than the drafting one, and it is the line Finding 2's fix moves.

**Fix**: extend the cross-library describe block in test/symlib.test.ts with `resolveLibrarySymbol('Sensor_Humidity:SHT40', [libDir])`, asserting `found-elsewhere` with `Sensor_Humidity:SHT4x`. The fixture already has both files.

## Untested-surface list (measured by instrumentation, reconciles with Findings 1, 5 and 6)

Coverage tooling produced no report, so instead of a percentage this is a direct reachability measurement: markers were placed at each new branch and the full suite was run.

- src/kicad/symlib.ts:338-339, cross-library fallback after the guessed library yields no candidates: **0 executions**. Finding 6.
- src/kicad/symlib.ts:400-406, the `found-elsewhere` branch of `verifySchematicSymbols`: 3 executions, all incidental, from `test/draft-engine.test.ts` "hermetic symbol vendoring (AC-16.5)", which asserts only on an unrelated `pin-count` finding. No test asserts anything about this branch's output, which is how Findings 1 and 5 both survive.
- src/kicad/draft/symsource.ts:334-335, the same fallback on the drafting path: 1 execution, from the new `test/draft-hardening.test.ts` case. Covered.
- src/kicad/symlib.ts:53-105, `winRoot` and Windows version discovery: covered by the three new tests, but see Finding 3: they pass only on a machine with no KiCad installed.
- No test asserts the exact-versus-fuzzy composition or the length of the `libIds` list returned by `resolveLibrarySymbol`. Finding 4.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 186, branch feat/deterministic-schematic-drafting,
based on feat/deterministic-schematic-drafting. Head commit 6e3442b.

Check it out first:
  gh pr checkout 186

FINDING 1 [high] verify_symbols now tells the agent to replace engine-generated
vendored power symbols with stock ones.
  Where: src/kicad/symlib.ts:400-406 (the found-elsewhere branch of
    verifySchematicSymbols), reached through src/kicad/symlib.ts:302-305
  Fails when: any drafted project is verified, which is every project the drafting
    engine produces. The engine generates its own power symbols and vendors them
    under the copperhead_power nickname (SymbolSource.vendorGenerated,
    src/kicad/draft/symsource.ts:248). That nickname is deliberately not an
    installed library, so findLibraryFile misses it and the cross-library search
    finds the stock symbol of the same name, reporting the vendored one as a wrong
    nickname. Reproduced by instrumenting the branch and running the suite:
    test/draft-engine.test.ts "hermetic symbol vendoring (AC-16.5)" hits it three
    times per run, producing
      - [no-library] "copperhead_power:GND" names the wrong library, not the wrong
        part: use power:GND instead.
    and the same for PWR_FLAG and VCC. Before this PR these were reported as
    no-library ("cannot verify its pins") and counted as skipped, which was
    accurate. Acting on the new advice either churns for nothing or breaks the
    deterministic vendoring guarantee in AC-16.5, since drafted output must come
    from the vendored cache. The existing test passes because it asserts only on
    the Device:R pin-count finding.
  Fix: before treating a missing library as a wrong nickname, check whether that
    nickname is vendored in the project's sym-lib-cache/. If it is, keep the
    existing no-library finding and the skipped count, and never emit
    found-elsewhere for it. This keeps AC-16.5 intact: a vendored stock symbol
    still carries its real nickname (Device), resolves against the installed
    library, and its drift stays visible. verifySchematicSymbols currently takes
    only schPath and env, so thread the project root (or the cache dir) in from
    src/agent/tools.ts:355. Add a regression test: draft a project, run
    verifySchematicSymbols on it, assert no finding names a copperhead_power:
    lib_id as a wrong library.

FINDING 2 [medium] The reorder that fixed the self-suggestion made junk
one-character same-file candidates pre-empt an exact cross-library hit.
  Where: src/kicad/symlib.ts:329-339 (introduced by 6e3442b), same shape at
    src/kicad/draft/symsource.ts:330-336
  Fails when: the guessed library exists but the part lives elsewhere, and the
    guessed library holds a short generic symbol whose name is a substring of the
    query. Device.kicad_sym is full of single-letter generics (R, C, L, D). The
    same-file candidate filter at symlib.ts:330-335 has no minimum-length guard,
    unlike the cross-library fuzzy path which added MIN_SHORT_NAME_LEN = 3
    (symlib.ts:220, 238) for exactly this reason, and it now returns early.
    Reproduced against the real stock library:
      Device:USB_C_Receptacle_USB2.0_14P
        head 6e3442b: no-symbol, candidates = [C, L, R]
        fd1a14c:      found-elsewhere -> Connector:USB_C_Receptacle_USB2.0_14P
      Device:LM358
        head 6e3442b: no-symbol, candidates = [L]
        fd1a14c:      found-elsewhere -> Amplifier_Operational:LM358
    USB_C_Receptacle_USB2.0_14P is the exact symbol symlib.ts:15's own module doc
    uses as the motivating example for this checker. This is a regression
    introduced by the fix commit.
  Fix: check for an exact cross-library hit before the same-file candidates, and
    keep same-file candidates ahead of cross-library fuzzy hits. That ordering is
    what actually fixes the old self-suggestion regression, which was caused by a
    fuzzy hit rather than an exact one. Verified working:
      const elsewhere = await findSymbolAcrossLibraries(name, dirs, lib);
      const exactElsewhere = elsewhere.filter((m) => m.exact);
      if (exactElsewhere.length) return { status: 'found-elsewhere',
        libIds: exactElsewhere.map((m) => `${m.lib}:${m.name}`) };
      if (candidates.length) return { status: 'no-symbol', candidates };
      if (elsewhere.length) return { status: 'found-elsewhere',
        libIds: elsewhere.slice(0, 8).map((m) => `${m.lib}:${m.name}`) };
    Mirror it in symsource.ts. Both regression tests this PR added still pass with
    it (test/symlib.test.ts:161 and the SHT40/SHT4x case in
    test/draft-hardening.test.ts). Do NOT instead add a minimum-length guard to the
    same-file filter on its own: that breaks the existing test/symlib.test.ts:83
    expectation that Device:R_Nonexistent offers R.

FINDING 3 [medium] The three new symbolSearchDirs tests fail on any machine that
has KiCad installed.
  Where: test/symlib.test.ts:203, 208, 220
  Fails when: /usr/share/kicad/symbols (or /usr/local/share/kicad/symbols, or the
    macOS bundle path) exists, which is true on any developer machine that can run
    this tool. All three assert against the whole return value
    (expect(dirs).toEqual([versionDir]), expect(dirs[0]).toBe(versionDir)) but
    symbolSearchDirs prepends every existing default path before the discovered
    Windows dirs. Reproduced at head on Linux with KiCad 10 installed:
      expected [ '/usr/share/kicad/symbols', ...(1) ] to deeply equal [ Array(1) ]
      expected '/usr/share/kicad/symbols' to be '/tmp/copperhead-kicadroot-.../10.0...'
      expected [ ...(2) ] to deeply equal [ Array(1) ]
    CI is green only because the runner has no KiCad installed. The third test is
    the regression guard for one of the two symbolSearchDirs bugs this PR fixes,
    and it does not guard it on the machines that matter.
  Fix: assert on the discovered subset rather than the whole list. For the first
    two, filter to paths under root (expect(dirs.filter((d) => d.startsWith(root))))
    and keep the ordering assertion. For the env-override test, point
    KICAD_SYMBOL_DIR at a separate temp directory and assert both
    expect(dirs).toContain(envDir) and
    expect(dirs.some((d) => d.startsWith(root))).toBe(false), which states the
    actual contract independently of what else is installed.

FINDING 4 [medium] resolveLibrarySymbol returns exact and fuzzy cross-library
matches mixed together and uncapped.
  Where: src/kicad/symlib.ts:304 and src/kicad/symlib.ts:339
  Fails when: the query name is short, or a fuzzy match exists alongside an exact
    one. crossLibrarySuggestions in symsource.ts:212-217 collapses to exact-only
    when any exact hit exists, but the two elsewhere.map(...) call sites in
    symlib.ts return every match with no filter and no cap. findSymbolAcrossLibraries
    fuzzy-matches on lower.includes(q) and the MIN_SHORT_NAME_LEN guard covers only
    the opposite direction, so a one-character query matches nearly every library.
    Reproduced against the real stock library:
      resolveLibrarySymbol('Nowhere:R')     -> 143 lib_ids, a 4,420-character
                                               verify_symbols finding
      resolveLibrarySymbol('Nowhere:LM358') -> [Amplifier_Operational:LM358,
                                               Regulator_Switching:LM3578]
    The second is the wrong-part hazard the PR's design section says the 1-edit cap
    prevents. The cap stops substitutions (TPS22860 vs TPS22810) but not insertions:
    LM358 and LM3578 are one edit apart, both at least 5 characters, and they are
    different real parts.
  Fix: apply the same exact-first collapse the symsource path already uses, and cap
    the fuzzy list at 8 (the cap the same-file candidate list already uses at
    symlib.ts:335). The snippet in FINDING 2 does both. Add a test asserting that a
    query with an exact hit in one library and a fuzzy hit in another returns only
    the exact one.

FINDING 5 [low] (optional) verify_symbols reports "0 issue(s) to reconcile" while
listing an issue.
  Where: src/kicad/symlib.ts:400-406, consumed at src/agent/tools.ts:360-363
  Fails when: any found-elsewhere finding is produced. The branch tags the finding
    kind: 'no-library' and does not increment skipped, so the summary counts it in
    neither bucket: mismatches filters out kind !== 'no-library', and skipped never
    saw it. Reproduced:
      verify_symbols: 0 verified, 0 unverifiable (library not installed),
      0 issue(s) to reconcile:
        - [no-library] "Nowhere:LM358" names the wrong library, not the wrong part: ...
    A header saying there is nothing to reconcile invites the agent to stop reading.
  Fix: give found-elsewhere its own kind on SymbolFinding (for example
    'wrong-library') and include it in the mismatches count at
    src/agent/tools.ts:362, so the header agrees with the list. Assert the header
    count in a test.

FINDING 6 [low] (optional) The wrong-nickname path through resolveLibrarySymbol
where the guessed library exists is reached by no test.
  Where: src/kicad/symlib.ts:338-339
  Fails when: never observed, because nothing exercises it. Measured, not assumed: a
    marker at that line printed zero times over a full suite run, while the
    equivalent symsource.ts fallback printed once and the verifySchematicSymbols
    found-elsewhere branch printed three times. This is the shape of the PR's
    flagship case (Sensor_Humidity:SHT40, guessed library exists and holds the
    near-miss) as it goes through the verify_symbols implementation rather than the
    drafting one, and it is the line FINDING 2's fix moves.
  Fix: extend the cross-library describe block in test/symlib.test.ts with
    resolveLibrarySymbol('Sensor_Humidity:SHT40', [libDir]), asserting
    found-elsewhere with Sensor_Humidity:SHT4x. The fixture already has both files.

These lines in the diff are reached by no test, or reached with nothing asserted
about them: src/kicad/symlib.ts:338-339 (zero executions),
src/kicad/symlib.ts:400-406 (executed incidentally, nothing asserted). No test
asserts the exact-versus-fuzzy composition or the length of the libIds list
returned by resolveLibrarySymbol. Add tests for these.

    by prompt text.
    dependence on execution order, no dependence on what is installed on the machine,
    and every test asserts something.

Then verify:
  npm run typecheck
  npm run build
  npm test

All three must pass. Run the suite on a machine that has KiCad installed if you can,
since FINDING 3 is invisible without it. Do not silence a failure by deleting an
assertion, skipping a test, or loosening a threshold: if something fails and the fix
is not obvious, stop and report it.

````

Chirag6722:
All five findings fixed in a07e4e6. Each was reproduced against the real code before being touched, and re-verified after.

**Finding 1 (high)** — confirmed. Reproduced directly: a project with a vendored `copperhead_power:GND` got told to "use power:GND instead". `verifySchematicSymbols` now takes the project root and treats any nickname present in `sym-lib-cache/` as deliberately absent, keeping the `no-library` finding and the `skipped` count. The parameter is optional: a caller that cannot supply a root loses the distinction rather than getting bad advice. Verified before/after on the same fixture.

**Finding 2 (medium)** — confirmed, and this one was mine: the round-1 reorder traded one regression for another. Both of your probe cases reproduced exactly (`Device:LM358` answering `closest: L`, `Device:USB_C_Receptacle_USB2.0_14P` answering `closest: C, L, R`). Reordered to exact-cross-library, then same-file candidates, then cross-library fuzzy. Both round-1 regression tests still pass, which is the point: that regression was caused by a *fuzzy* hit, so ranking *exact* hits first does not reintroduce it.

**Finding 3 (medium)** — confirmed by inspection, and a fair hit. Those tests only passed here because this machine's KiCad lives under the Windows path, so the Linux defaults never appeared. Now asserting on the subset under the test root, and the env-override test points at an unrelated directory and asserts auto-discovery did not run.

**Finding 4 (medium)** — confirmed. `Nowhere:R` returned 143 lib_ids before, 1 now. `LM358` no longer offers `LM3578` alongside the right answer. Both files now share one `crossLibraryIds` helper (exact-collapse plus a cap of 8, matching the same-file cap), rather than symsource having the collapse and symlib not.

**Finding 5 (low)** — fixed via a dedicated `wrong-library` kind, which the existing issue count already includes.

**Finding 6 (low)** — test added.

**Testing**: full suite compared against a stashed baseline, 16 failures before and 16 after, no new regressions, 5 new tests. Two names move between those two lists (`repl.test.ts` session-log, `draft-engine` ERC); both pass in isolation and neither touches symbol resolution, so I am calling them pre-existing flakes under parallel load rather than claiming they are unrelated without checking.

One thing I did not do: your Finding 2 snippet capped the fuzzy list inline at each call site. I moved that into a shared helper instead, since the same collapse was already duplicated in `symsource.ts` and the two copies had already drifted apart once (that drift is exactly Finding 4). Behavior matches the snippet.

PR #187, opened 2026-08-06
fix(sync): deterministically reconcile SPEC budgets with constraint registry

### LINKED ISSUE #180: Implement constraint registry reconciliation in check_drift

AyushVUpadhye:
Closes #180.

## What

Strengthen the deterministic dual-write audit performed by `copperhead sync` for constraints documented in `docs/SPEC.md`.

Instead of relying on substring matching, the sync verifier now parses the `## Budgets` section in `SPEC.md` and reconciles it against `.copperhead/constraints.json`, reporting missing or mismatched entries in either direction.

## Why

The previous implementation verified `SPEC.md` constraints using plain substring matching, which could both miss inconsistencies and report false positives.

This change makes the SPEC budget portion of the dual-write audit deterministic while preserving the existing heuristic for other documentation.

## Design

The reconciliation remains part of the deterministic verification phase in `syncVerify`.

For constraints whose source is `docs/SPEC.md`, the verifier:

- parses documented budgets from the `## Budgets` section,
- compares them with the constraint registry,
- reports registry entries missing from `SPEC.md`,
- reports documented budgets missing from the registry,
- reports value mismatches when both exist.

The existing substring-based audit remains unchanged for constraints documented outside `SPEC.md`, keeping the scope of this change focused.

## Spec / OpenSpec

None.

This strengthens an existing verification mechanism and does not introduce new spec-level behavior.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | `npm run typecheck` | ✅ Pass |
| Build | `npm run build` | Not run |
| Unit + offline | `npx vitest test/gating-sync.test.ts -t "SPEC"` | ✅ Pass (4 tests) |

Added coverage for:

- registry constraint missing from `SPEC.md`
- documented budget missing from the registry
- matching registry and `SPEC.md` values
- mismatched registry and `SPEC.md` values

- Sandbox variant used: n/a

- Commands run and outcome:

```text
n/a

This change only affects deterministic verification logic exercised through automated tests.
```

## Docs

None.

---

## Invariant checklist

- [x] **Spec-gated in**: N/A
- [x] **Verification-gated out**: Verification remains deterministic. This change only strengthens the existing sync verification.
- [x] **`check`/`verify` stays LLM-free and network-free**: No providers or network access were introduced.
- [x] **No sexp serialization**: Unchanged.
- [x] **Sync-obligations ledger**: Unchanged.
- [x] **Secrets**: Unchanged.
- [x] **Spec coherence**: No spec-level behavior changed.

## Summary by CodeRabbit

* **New Features**
  * Added synchronization checks between documented budgets in `SPEC.md` and registered constraints.
  * Validates budget identifiers and values, including unit-suffixed and explanatory values.
  * Ignores fenced code examples and unrelated document sections.
  * Reports missing, undocumented, or mismatched budget entries with expected details.
  * Handles missing specifications, minimum-only constraints, and constraints without scalar values.

* **Bug Fixes**
  * Prevents configuration drift between budget documentation and the constraint registry.

* **Tests**
  * Added coverage for the expanded synchronization scenarios.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -44,13 +44,11 @@ function parseSpecBudgets(spec: string): Map<string, string> {
   const afterHeading = spec.slice(heading.index + heading[0].length);
   const nextHeading = afterHeading.search(/^#{1,2}\s/m);
   const section = nextHeading >= 0 ? afterHeading.slice(0, nextHeading) : afterHeading;
-
-  const budgetLine =
-    /^\s*-\s*([A-Za-z][A-Za-z0-9_-]*)\s*:\s*(.+?)\s*$/gm;
-
+  const withoutCodeBlocks = section.split(/```/).filter((_, i) => i % 2 === 0).join('');
+  const budgetLine = /^-\s*([A-Za-z][A-Za-z0-9_-]*)\s*:\s*(.+?)\s*$/gm;
   let match: RegExpExecArray | null;
 
-  while ((match = budgetLine.exec(section)) !== null) {
+  while ((match = budgetLine.exec(withoutCodeBlocks)) !== null) { 
     const key = match[1];
     const value = match[2];
 
@@ -58,11 +56,21 @@ function parseSpecBudgets(spec: string): Map<string, string> {
 
     // Keep only the canonical value before any explanatory prose.
     const canonical = value.replace(/\s*\(.*$/, '').trim();
-
+    if (!isBudgetValue(canonical)) continue;
     budgets.set(key, canonical);
   }
 
   return budgets;
+}
+
+function isBudgetValue(value: string): boolean {
+  return /^\d+(?:\.\d+)?(?:\s*[A-Za-zµμ%/0-9]+)?$/.test(value);
+}
+
+function parseLeadingNumber(value: string): number | null {
+  const match = value.match(/^\d+(?:\.\d+)?/);
+
+  return match ? Number(match[0]) : null;
 }
 
 export async function syncVerify(repoRoot: string): Promise<SyncReport> {
@@ -99,7 +107,8 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
     const shortKey = key.split('.').pop()!;
 
     // Constraints documented in SPEC.md are checked deterministically.
-    if (constraint.source.startsWith('docs/SPEC.md')) {
+    const budgetSource = `${config.docs.replace(/\/?$/, '/')}SPEC.md#budgets`;
+    if (constraint.source.startsWith(budgetSource)) {
       if (!specBudgets.has(shortKey)) {
         resolvable.push({
           kind: 'dual-write',
@@ -121,14 +130,24 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
               ? String(constraint.min)
               : undefined;
 
-      if (recorded !== undefined && documented !== recorded) {
-        resolvable.push({
-          kind: 'dual-write',
-          doc: 'constraints.json',
-          claim: `${shortKey}: ${recorded}`,
-          actual: `SPEC.md documents ${documented}`,
-          resolution: `update either SPEC.md or constraints.json so both agree`,
-        });
+      if (recorded !== undefined) {
+        const documentedNumber = parseLeadingNumber(documented);
+        const recordedNumber = parseLeadingNumber(recorded);
+
+        const matches =
+          documentedNumber !== null && recordedNumber !== null
+            ? documentedNumber === recordedNumber
+            : documented === recorded;
+
+        if (!matches) {
+          resolvable.push({
+            kind: 'dual-write',
+            doc: 'constraints.json',
+            claim: `${shortKey}: ${recorded}`,
+            actual: `SPEC.md documents ${documented}`,
+            resolution: 'update either SPEC.md or constraints.json so both agree',
+          });
+        }
       }
 
       continue;
@@ -165,7 +184,7 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
     }
   }
   for (const [budget, value] of Object.entries(config.budgets)) {
-    if (!Object.keys(registry).some((k) => k.endsWith(budget))) {
+    if (!Object.keys(registry).some((k) => k.split('.').pop() === budget,)) {
       resolvable.push({
         kind: 'dual-write',
         doc: '.copperhead/config.json',
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -36,12 +36,13 @@ export interface SyncReport {
 }
 
 function parseSpecBudgets(spec: string): Map<string, string> {
+  const withoutComments = spec.replace(/<!--[\s\S]*?-->/g, '');
   const budgets = new Map<string, string>();
 
-  const heading = spec.match(/^## Budgets\s*$/m);
+  const heading = withoutComments.match(/^## Budgets\s*$/m);
   if (!heading || heading.index === undefined) return budgets;
 
-  const afterHeading = spec.slice(heading.index + heading[0].length);
+  const afterHeading = withoutComments.slice(heading.index + heading[0].length);
   const nextHeading = afterHeading.search(/^#{1,2}\s/m);
   const section = nextHeading >= 0 ? afterHeading.slice(0, nextHeading) : afterHeading;
   const withoutCodeBlocks = section.split(/```/).filter((_, i) => i % 2 === 0).join('');
@@ -67,10 +68,19 @@ function isBudgetValue(value: string): boolean {
   return /^\d+(?:\.\d+)?(?:\s*[A-Za-zµμ%/0-9]+)?$/.test(value);
 }
 
-function parseLeadingNumber(value: string): number | null {
-  const match = value.match(/^\d+(?:\.\d+)?/);
+function parseLeadingNumber(
+  value: string,
+): { value: number; unit: string } | null {
+  const match = value.match(
+    /^\s*(\d+(?:\.\d+)?)\s*([A-Za-zµμ%][A-Za-z0-9_\/-]*)?/,
+  );
 
-  return match ? Number(match[0]) : null;
+  if (!match) return null;
+
+  return {
+    value: Number(match[1]),
+    unit: (match[2] ?? '').toLowerCase(),
+  };
 }
 
 export async function syncVerify(repoRoot: string): Promise<SyncReport> {
@@ -98,6 +108,7 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
   const specBudgets = existsSync(specPath)
     ? parseSpecBudgets(await readFile(specPath, 'utf8'))
     : new Map<string, string>();
+  const budgetSource = `${config.docs.replace(/\/?$/, '/')}SPEC.md#budgets`;
   let docsText = '';
   for (const name of ['SPEC.md', 'BOM.md', 'PINOUT.md', 'SUBSYSTEMS.md', 'LAYOUT.md']) {
     const p = path.join(docsDir, name);
@@ -107,7 +118,6 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
     const shortKey = key.split('.').pop()!;
 
     // Constraints documented in SPEC.md are checked deterministically.
-    const budgetSource = `${config.docs.replace(/\/?$/, '/')}SPEC.md#budgets`;
     if (constraint.source.startsWith(budgetSource)) {
       if (!specBudgets.has(shortKey)) {
         resolvable.push({
@@ -136,7 +146,8 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
 
         const matches =
           documentedNumber !== null && recordedNumber !== null
-            ? documentedNumber === recordedNumber
+            ? documentedNumber.value === recordedNumber.value &&
+              documentedNumber.unit === recordedNumber.unit
             : documented === recorded;
 
         if (!matches) {
@@ -165,10 +176,12 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
     }
   }
   const registryByShortKey = new Map(
-    Object.entries(registry).map(([key, constraint]) => [
-      key.split('.').pop()!,
-      { key, constraint },
-    ]),
+    Object.entries(registry)
+      .filter(([, constraint]) => constraint.source.startsWith(budgetSource))
+      .map(([key, constraint]) => [
+        key.split('.').pop()!,
+        { key, constraint },
+      ]),
   );
   for (const [budget, documented] of specBudgets) {
     const entry = registryByShortKey.get(budget);
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -56,7 +56,10 @@ function parseSpecBudgets(spec: string): Map<string, string> {
 
     if (!key || value === undefined) continue;
 
-    budgets.set(key, value.trim());
+    // Keep only the canonical value before any explanatory prose.
+    const canonical = value.replace(/\s*\(.*$/, '').trim();
+
+    budgets.set(key, canonical);
   }
 
   return budgets;
@@ -110,9 +113,13 @@ export async function syncVerify(repoRoot: string): Promise<SyncReport> {
 
       const documented = specBudgets.get(shortKey)!;
       const recorded =
-        constraint.value ??
-        (constraint.max !== undefined ? String(constraint.max) : undefined) ??
-        (constraint.min !== undefined ? String(constraint.min) : undefined);
+        constraint.value !== undefined
+          ? String(constraint.value)
+          : constraint.max !== undefined
+            ? String(constraint.max)
+            : constraint.min !== undefined
+              ? String(constraint.min)
+              : undefined;
 
       if (recorded !== undefined && documented !== recorded) {
         resolvable.push({
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -35,6 +35,33 @@ export interface SyncReport {
   violations: SyncViolationItem[];
 }
 
+function parseSpecBudgets(spec: string): Map<string, string> {
+  const budgets = new Map<string, string>();
+
+  const heading = spec.match(/^## Budgets\s*$/m);
+  if (!heading || heading.index === undefined) return budgets;
+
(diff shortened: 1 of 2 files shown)


AyushVUpadhye:
Fixed CodeRabbit's comments

animesh-chouhan:
ack, checking

animesh-chouhan:
_Automated pr-review pass (Claude Code). Not a human review; no formal approval implied._

## Verdict: request changes

The direction is right and the happy path is solid: with the init scaffold's exact `## Budgets` heading and plain `- key: value` bullets, the reconciliation behaves exactly as the PR describes, in both directions. The problem is what happens when SPEC.md is written any other legitimate way. The old check was a substring heuristic whose failure mode was silence; the new one's failure mode is a positive drift report, so every shape the parser does not recognize now becomes a false "missing from SPEC.md budgets", and the `sync` resolve phase acts on those reports. Three of those are confirmed by running the code.

## Metrics

```
lines: +324 / -2 (net 322) across 2 files  ·  src +93/-2, test +231/-0, docs +0/-0, spec +0/-0, other +0/-0
tests: +231 test lines; suite not measured -> 577/20/0 (pass/skip/fail, base -> head)
diff coverage: 94.6% of changed executable src lines exercised (70/74, measured via vitest v8); uncovered: src/commands/sync.ts:92,120-122
```

Method: merge base `51bb26e` (`git merge-base origin/main HEAD`); areas from `git diff --numstat`, generated files excluded; suite from `npx vitest run --reporter=json` on the head checkout (577 pass / 20 skip / 0 fail, actually run); coverage from `vitest --coverage.provider=v8` intersected with changed src lines; CI from `gh pr checks`. `mergeable=MERGEABLE`, `draft=false`. Base suite: not measured (base CI is green and the base result does not bear on these findings).

Findings 1, 2, 3 and 5 below were each reproduced by running `syncVerify` against a temp fixture repo; the outputs quoted are actual, not predicted.

## Findings

### 1. [medium] The budget parser accepts exactly one markdown shape, and every other shape now reports false drift

**Where:** [src/commands/sync.ts:41](src/commands/sync.ts#L41), [src/commands/sync.ts:49](src/commands/sync.ts#L49)

`parseSpecBudgets` requires the heading to match `/^## Budgets\s*$/m` and entries to match `- key: value`. The repo already knows this is too strict: [src/commands/create.ts:46-50](src/commands/create.ts#L46-L50) carries a comment saying stage prompts do not dictate exact heading text and that a literal `## Budgets` match "produces false negatives against valid docs titled e.g. `## 3. Budgets and constraints`", and uses `/^#{1,6}\s.*\bBudgets?\b/im` for exactly that reason. The `spec-seed` stage-completion gate accepts those variants, so `create` can pass its own gate and then `sync` reports the resulting SPEC.md as inconsistent.

**Fails when:** SPEC.md uses a heading variant, a bold key, or a table. All three produce the same false report against a correctly documented constraint (`power.sleep_current_uA` = 25, source `docs/SPEC.md#budgets`):

```
## 3. Budgets and constraints        -> [dual-write] constraints.json: "constraint power.sleep_current_uA exists in registry" / "missing from SPEC.md budgets"
- **sleep_current_uA**: 25           -> same
| sleep_current_uA | 25 |            -> same
```

Under the old substring check all three passed. The failure is not just noise: `syncResolve` feeds the report to the agent with "Apply each proposed resolution", and the proposed resolution is to add a budget that is already there.

**Fix:** reuse the lenient heading regex from `create.ts` (factor it into one shared helper so the two cannot drift apart), and tolerate `**key**` / `` `key` `` wrappers and table rows, or narrow the deterministic path so an unparseable Budgets section falls back to the substring heuristic instead of reporting drift.

### 2. [medium] SPEC-sourced constraints that are not budgets are forced into the Budgets section

**Where:** [src/commands/sync.ts:102](src/commands/sync.ts#L102)

The deterministic branch triggers on `constraint.source.startsWith('docs/SPEC.md')`, which matches any anchor, not just `#budgets`. Non-budget constraints legitimately live in other SPEC.md sections: SPEC.md itself lists strapping pins, keepouts and package availability as constraints alongside electrical budgets.

**Fails when:** a `forbidden_pins` constraint sourced `docs/SPEC.md#strapping-pins`, documented under its own `## Strapping pins` heading, is reported as drift:

```
[dual-write] constraints.json: "constraint mcu.strapping_pins exists in registry"
             / "missing from SPEC.md budgets"
             resolution: add the constraint to the doc named by its source (docs/SPEC.md#strapping-pins)
```

The constraint is documented, at exactly the anchor its `source` names. This also narrows AC-7.2, whose condition is "registry entry with no doc mention", to "not a bullet under `## Budgets`" for SPEC-sourced constraints, which is a spec-level behavior change the PR states it does not make.

**Fix:** gate the deterministic branch on the budgets anchor (`source` starting with `docs/SPEC.md#budgets`) or on the short key actually appearing in the parsed budgets, and let everything else fall through to the substring heuristic.

### 3. [medium] Any `- word: text` bullet under Budgets becomes a phantom budget

**Where:** [src/commands/sync.ts:49](src/commands/sync.ts#L49), [src/commands/sync.ts:157](src/commands/sync.ts#L157)

The line regex does not distinguish a budget from prose, and it matches nested sub-bullets because it allows arbitrary leading whitespace. Every unmatched key then goes through the reverse direction as "missing from constraints.json" with resolution `record_constraint with the documented value and source`.

**Fails when:** a Budgets section carries any annotation:

```
- sleep_current_uA: 25
  - rationale: coin cell
- note: values are worst-case
```

produces two reports: `budget rationale: coin cell / missing from constraints.json` and `budget note: values are worst-case / missing from constraints.json`. The resolve phase is instructed to apply the resolution, so the agent records `rationale` and `note` as constraints, writing garbage into the registry that the next `sync` then treats as real.

**Fix:** restrict the reverse direction to entries whose value parses as a number (optionally with a unit), or to top-level bullets only, and drop keys that do not look like budget keys rather than proposing `record_constraint` for them.

### 4. [low] Value comparison is exact string equality, so units and formatting read as mismatches

**Where:** [src/commands/sync.ts:60](src/commands/sync.ts#L60), [src/commands/sync.ts:124](src/commands/sync.ts#L124)

Only parenthesized trailing prose is stripped; the comparison is then `documented !== recorded` on raw strings.

**Fails when:** `- sleep_current_uA: 25 uA` against `{ "max": 25 }` reports `"sleep_current_uA: 25" / "SPEC.md documents 25 uA"`, a mismatch between two identical budgets. `25.0` vs `25`, and trailing prose introduced with a comma or a dash rather than parentheses, fail the same way.

**Fix:** parse a leading numeric token from the documented value and compare numerically when both sides are numbers, falling back to string comparison otherwise.

### 5. [low] HTML-comment budgets are parsed as real budgets

**Where:** [src/commands/sync.ts:53](src/commands/sync.ts#L53)

`create.ts` strips `<!-- ... -->` from the Budgets section before reading it ([src/commands/create.ts:97](src/commands/create.ts#L97)); `parseSpecBudgets` does not.

**Fails when:** the multi-line placeholder form (a variant the stage-completion tests explicitly cover) is present:

```
## Budgets

<!--
Add hard budgets here
- sleep_current_uA: 25
-->
```

yields `budget sleep_current_uA: 25 / missing from constraints.json`. The default init scaffold uses single-line comments, so a fresh `init` plus `sync` is clean (verified), which is what keeps this low.

**Fix:** strip HTML comments from the section before matching, matching `create.ts`.

### 6. [low] The `docs/SPEC.md` source prefix is hardcoded and ignores `config.docs`

**Where:** [src/commands/sync.ts:102](src/commands/sync.ts#L102)

`specPath` is built from `config.docs`, but the source test is the literal string `docs/SPEC.md`. With `docs: "design/"`, sources read `design/SPEC.md`, the deterministic path never fires, and a real mismatch (registry 25 vs SPEC 30) goes unreported: verified, the report is empty. Silent weakening rather than a false positive, hence low.

**Fix:** derive the expected prefix from `config.docs` instead of hardcoding it.

### 7. [low] Untested surface: three new branches no test reaches

**Where:** [src/commands/sync.ts:92](src/commands/sync.ts#L92), [src/commands/sync.ts:120-122](src/commands/sync.ts#L120-L122)

This reconciles with the coverage line above: `src/commands/sync.ts:92,120-122` are the only uncovered changed lines.

- Line 92: SPEC.md absent entirely. Verified by hand: every SPEC-sourced constraint is reported missing, which is defensible, but nothing pins it.
- Lines 120-122: the `min`-only and the no-`value`/`max`/`min` branches of `recorded`. Verified by hand: a `min: 3` constraint matching `- vin_min_V: 3` is clean, and a `forbidden_pins`-only constraint whose key appears under Budgets is clean. Both correct, both unpinned.

`parseSpecBudgets` is not exported, so there are no new exported symbols to test.

**Fix:** add the three cases to `test/gating-sync.test.ts`.

## Invariants checked

Performed and passed:

- **Spec-gated in**: no change to the tool list or its gating. `edit_file`/`write_file` remain absent until a proposal validates.
- **Verification-gated out**: no mutation path touched; `syncVerify` is read-only.
- **`check`/`verify` LLM-free and network-free**: `src/commands/check.ts` does not import `sync.ts`, and no provider, key, or network import was added anywhere in the diff. The new code reads files only.
- **No sexp serialization**: `src/kicad/` untouched.
- **Sync-obligations ledger**: untouched.
- **Secrets**: no transcript, redaction, or `.gitignore` change.
- **Spec coherence**: the PR states no spec-level behavior changed. That holds for the parsing mechanics, but finding 2 narrows AC-7.2's "no doc mention" condition for SPEC-sourced constraints; either fix the condition or move SPEC.md with it. `openspec/` planning artifacts were not modified, so `openspec validate` was not required.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 187, branch
fix/spec-budget-dual-write-reconciliation, based on main.

Check it out first:
  gh pr checkout 187

All findings are in src/commands/sync.ts, in the parseSpecBudgets function and the
SPEC-budget reconciliation added to syncVerify. Context: the previous dual-write
audit was a substring heuristic whose failure mode was silence. The new
deterministic check's failure mode is a positive drift report, and copperhead sync's
resolve phase feeds that report to an agent told to "apply each proposed
resolution". So a parse miss is not cosmetic: it makes the agent edit correct docs
or write junk into the constraint registry.

FINDING 1 [medium] The budget parser accepts exactly one markdown shape, and every
other shape now reports false drift.
  Where: src/commands/sync.ts:41 (heading regex), src/commands/sync.ts:49 (line regex)
  Fails when: SPEC.md uses a heading variant, a bold key, or a table. With
    power.sleep_current_uA = 25 recorded with source docs/SPEC.md#budgets, each of
      ## 3. Budgets and constraints
      - **sleep_current_uA**: 25
      | sleep_current_uA | 25 |
    produces the false report
      [dual-write] constraints.json: "constraint power.sleep_current_uA exists in
      registry" / "missing from SPEC.md budgets"
    All three passed under the old substring check. src/commands/create.ts:46-50
    documents that stage prompts do not dictate exact heading text and that a literal
    "## Budgets" match produces false negatives against valid docs titled e.g.
    "## 3. Budgets and constraints"; it uses /^#{1,6}\s.*\bBudgets?\b/im instead. The
    spec-seed stage-completion gate accepts those variants, so copperhead create can
    pass its own gate and then copperhead sync calls the result inconsistent.
  Fix: reuse the lenient heading regex from create.ts, factored into one shared
    helper so the two cannot drift apart. Tolerate **key** and `key` wrappers and
    table rows, or narrow the deterministic path so an unparseable Budgets section
    falls back to the substring heuristic instead of reporting drift.

FINDING 2 [medium] SPEC-sourced constraints that are not budgets are forced into the
Budgets section.
  Where: src/commands/sync.ts:102
  Fails when: the deterministic branch triggers on
    constraint.source.startsWith('docs/SPEC.md'), which matches any anchor, not just
    #budgets. SPEC.md lists strapping pins, keepouts and package availability as
    constraints alongside electrical budgets. A forbidden_pins constraint sourced
    docs/SPEC.md#strapping-pins and documented under its own "## Strapping pins"
    heading is reported as
      [dual-write] constraints.json: "constraint mcu.strapping_pins exists in
      registry" / "missing from SPEC.md budgets"
      resolution: add the constraint to the doc named by its source
      (docs/SPEC.md#strapping-pins)
    even though it is documented at exactly the anchor its source names. This also
    narrows AC-7.2, whose condition is "registry entry with no doc mention", to "not
    a bullet under ## Budgets" for SPEC-sourced constraints, which is a spec-level
    behavior change the PR description states it does not make.
  Fix: gate the deterministic branch on the budgets anchor (source starting with
    docs/SPEC.md#budgets) or on the short key actually appearing in the parsed
    budgets, and let everything else fall through to the substring heuristic. If you
    keep the narrower condition instead, update openspec/specs/SPEC.md AC-7.2 to
    match.

FINDING 3 [medium] Any "- word: text" bullet under Budgets becomes a phantom budget.
  Where: src/commands/sync.ts:49 (line regex), src/commands/sync.ts:157 (reverse
    direction)
  Fails when: the line regex does not distinguish a budget from prose and allows
    arbitrary leading whitespace, so it matches nested sub-bullets too. A Budgets
    section carrying annotations:
      - sleep_current_uA: 25
        - rationale: coin cell
      - note: values are worst-case
    produces "budget rationale: coin cell / missing from constraints.json" and
    "budget note: values are worst-case / missing from constraints.json", each with
    resolution "record_constraint with the documented value and source". The resolve
    phase applies them, so the agent records rationale and note as constraints and
    the next sync treats that junk as real.
  Fix: restrict the reverse direction to entries whose value parses as a number
    (optionally with a unit), or to top-level bullets only, and drop keys that do not
    look like budget keys rather than proposing record_constraint for them.

FINDING 4 [low] (optional) Value comparison is exact string equality, so units and
formatting read as mismatches.
  Where: src/commands/sync.ts:60 (canonicalization), src/commands/sync.ts:124
    (comparison)
  Fails when: only parenthesized trailing prose is stripped and the comparison is
    documented !== recorded on raw strings. "- sleep_current_uA: 25 uA" against
    { "max": 25 } reports "sleep_current_uA: 25" / "SPEC.md documents 25 uA", a
    mismatch between two identical budgets. "25.0" vs 25, and trailing prose
    introduced with a comma or a dash rather than parentheses, fail the same way.
  Fix: parse a leading numeric token from the documented value and compare
    numerically when both sides are numbers, falling back to string comparison
    otherwise.

FINDING 5 [low] (optional) HTML-comment budgets are parsed as real budgets.
  Where: src/commands/sync.ts:53
  Fails when: the multi-line placeholder form, a variant test/stage-completion.test.ts
    explicitly covers, is present:
      ## Budgets

      <!--
      Add hard budgets here
      - sleep_current_uA: 25
      -->
    yields "budget sleep_current_uA: 25 / missing from constraints.json". The default
    init scaffold uses single-line comments, so a fresh init plus sync is clean, which
    is what keeps this low.
  Fix: strip HTML comments from the section before matching, the way
    src/commands/create.ts:97 already does.

FINDING 6 [low] (optional) The docs/SPEC.md source prefix is hardcoded and ignores
config.docs.
  Where: src/commands/sync.ts:102
  Fails when: specPath is built from config.docs but the source test is the literal
    string 'docs/SPEC.md'. With docs set to "design/", sources read design/SPEC.md,
    the deterministic path never fires, and a real mismatch (registry 25 vs SPEC 30)
    goes unreported: the report comes back empty.
  Fix: derive the expected prefix from config.docs instead of hardcoding it.

FINDING 7 [low] (optional) Untested surface: three new branches no test reaches.
  Where: src/commands/sync.ts:92, src/commands/sync.ts:120-122
  Fails when: nothing currently, but nothing pins the behavior either. Line 92 is the
    SPEC.md-absent branch (every SPEC-sourced constraint is then reported missing).
    Lines 120-122 are the min-only and the no-value/max/min branches of `recorded`; a
    min: 3 constraint matching "- vin_min_V: 3" is clean, and a forbidden_pins-only
    constraint whose key appears under Budgets is clean.
  Fix: add the three cases to test/gating-sync.test.ts.

These lines in the diff are reached by no test: src/commands/sync.ts:92,
src/commands/sync.ts:120-122. Add tests for the branches they contain (see FINDING 7).

    by prompt text.

Then verify:
  npm run typecheck
  npm run build
  npm test

and report it.

````

AyushVUpadhye:
@animesh-chouhan 
Resolved all the Findings and CodeRabbit's comments

Issue #188, opened 2026-08-06
check exits 0 when the configured schematic is missing, and says it is not configured


adhavan18:
### What happened

`copperhead check` exits 0 when its schematic is configured but missing, and tells the user the schematic is not configured. Every check that depends on the schematic (ERC, drift, forbidden-pin constraints) is skipped together, and the result is reported as green.

Found by driving `runCheck` over fixture repos and varying the config.

### Reproduction

Config points at a schematic that does not exist, which is what a typo, a rename, or a moved file looks like:

```json
{ "docs": "docs/", "schematic": "hardware/board.kicad_sch", "board": "hardware/board.kicad_pcb" }
```

```
ok = true
erc: null   drc: null
drift.ok: true   openspec: null   constraints.ok: true
log:
  | ERC skipped (no schematic configured; run copperhead init)
  | DRC skipped (no board configured)
```

Two separate problems in that output.

**The gate passes.** [cli.ts](src/cli.ts) exits on `res.ok`:

```ts
process.exit(res.ok ? 0 : 1);
```

`ok` is computed with `??  true` for the skipped checks, so a repo where nothing was verified is indistinguishable from a repo that passed everything. In CI that is a green build. `create.ts` also calls `runCheck`, so the same result feeds the create flow.

**The message is wrong and points the wrong way.** [check.ts:33](src/commands/check.ts) says:

```
ERC skipped (no schematic configured; run copperhead init)
```

but the schematic *is* configured. The guard is `config.schematic && existsSync(...)`, and both arms produce the same message, so a missing file is reported as missing configuration. `copperhead init` is the wrong remedy: the fix is to correct the path or restore the file.

### Why this is worth separating from a plain "no schematic" repo

A repo with nothing configured reporting green is arguable, since there is genuinely nothing to check and `init` has not run. The case above is different: the repo was configured to be checked, and quietly stopped being checked. Nothing in the exit code or the message says so.

The contrast with the neighbouring failure mode is stark. With the schematic present but `kicad-cli` absent, the tool fails loudly and well:

```
KicadCliMissingError: kicad-cli not found on PATH
why it failed: copperhead verifies every mutation with kicad-cli ERC/DRC; ...
```

So the harmless case shouts and the dangerous case is silent.

### Scope

[...]

PR #154, opened 2026-07-30
feat(agent): gate verify_symbols via ledger obligation and add symbols check (#133)

Chirag6722:
## What

1. Adds `symbol-verification` obligation to `ObligationsLedger` (`src/agent/ledger.ts`), opened when a schematic edit is made.
2. Clears `symbol-verification` obligation when `verify_symbols` runs with 0 unhandled findings (`src/agent/tools.ts`).
3. Integrates `verifySchematicSymbols` into `copperhead check` / `runCheck` in `src/commands/check.ts` to surface library symbol divergences in CI.

## Why

Addresses #133. Previously, `verify_symbols` was advisory and opened no ledger obligation, meaning `finish` could commit a schematic containing invented or corrupted symbol pinouts without running or acting on `verify_symbols`.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | `npm run typecheck` | pass |
| Build | `npm run build` | pass |
| Unit + offline | `npx vitest run test/gating-sync.test.ts` | pass (16/16) |

### Manual test log

Verified `finish blocks on open obligations` test suite in `test/gating-sync.test.ts` fails to finish after schematic edits until `verify_symbols` is dispatched, and `copperhead check` displays `symbols ✓` when symbol pinouts match installed libraries.

## Docs

No documentation changes required; internal agent ledger and check command enhancement.

---

## Invariant checklist

- [x] **Spec-gated in**. Unaffected.
- [x] **Verification-gated out**. Strengthens verification by requiring `verify_symbols` obligation to be satisfied before commit.
- [x] **`check`/`verify` stays LLM-free and network-free**. `copperhead check` symbol verification uses local installed `.kicad_sym` libraries with 0 network calls.
- [x] **No sexp serialization**. Unaffected.
- [x] **Sync-obligations ledger**. Adds `symbol-verification` obligation for schematic symbol parity.
- [x] **Secrets**. Unaffected.
- [x] **Spec coherence**. Addresses Issue #133 requirement.

Closes #133.

## Summary by CodeRabbit

* **New Features**
  * Added automatic symbol verification, schematic legibility checks, and geometry-based scoring.
  * Added tools for searching symbols, inspecting pins, drafting schematics, and reviewing quality results.
  * Schematic edits now require symbol verification before completion.
  * Checks now report symbol, legibility, and scoring results.
* **Bug Fixes**
  * Improved handling and reporting of unavailable symbol libraries.
* **Documentation**
[...]

PR #191, opened 2026-08-07
fix(check): fail when the configured schematic is missing, and say which file

adhavan18:
## What

`copperhead check` no longer exits 0 when `config.schematic` points at a file that does not exist, and the message names the missing path instead of claiming the schematic is not configured.

## Why

Refs #188 (does not close it).

The issue's Scope section also names [sync.ts](src/commands/sync.ts) lines 43 and 133, which carry the same guard. Those are deliberately not in this PR: `syncVerify` has no `ok` boolean, and its resolvable items are handed to an LLM resolve phase, so a missing schematic needs different handling there (a violation, which is never auto-resolved) rather than the same one-line treatment. Follow-up PR, not a line in this one.

ERC, drift and the forbidden-pin constraint check all gate on the same `config.schematic && existsSync(...)` condition, written out three times in [check.ts](src/commands/check.ts). When the file is absent all three skip together, `ok` is computed with `?? true` for the skipped checks, and the run reports green. A repo that was configured to be checked and quietly stopped being checked looked exactly like one that passed everything. In CI that is a green build, and [create.ts](src/commands/create.ts) calls `runCheck` too.

The message pointed the wrong way as well. Both arms of the guard logged `no schematic configured; run copperhead init`, but the schematic *is* configured: the fix is to correct the path or restore the file, not to re-run `init`.

Reproduced before changing anything. The two cases were byte-identical:

```text
### A. schematic configured at hardware/missing-board.kicad_sch, file absent
  ok = true
  log: ERC skipped (no schematic configured; run copperhead init)

### B. no schematic configured
  ok = true
  log: ERC skipped (no schematic configured; run copperhead init)
```

## Design

The three copies of the guard are resolved once into `schematicUsable` and `schematicMissing`, so the condition cannot drift between the three checks that share it.

The two skips are now split, and treated differently on purpose:

- `config.schematic` unset: skip, keep the existing message, stay green. There is genuinely nothing to verify and `init` has not run, so this is unchanged.
- `config.schematic` set but absent: name the path, do not mention `init`, and fail.

`ok` gains an explicit `schematicMissing === null` term rather than relying on the `?? true` defaults. Those defaults are correct for a check that had nothing to run, but they cannot distinguish "nothing to verify" from "everything to verify, silently skipped".

I picked fail over adding a not-verified state, per the two options in the issue. Failing is what makes CI notice, and the new `schematicMissing` field still lets a caller tell this apart from a real ERC or drift failure. Happy to switch it to a distinct state if you would rather not change the exit code for existing repos.

**Coordination note:** #154 adds a fourth copy of the same `config.schematic && existsSync(...)` guard for `verifySchematicSymbols`. Whichever of these lands second will want to use the shared `schematicUsable` instead of re-testing. Not a blocker either way, just flagging it so the guard does not go back to being duplicated.

## Spec / OpenSpec

No spec-level behavior change, but the check result gains a key, so the AC-2.4 stable-JSON-keys assertion in [init-check.test.ts](test/init-check.test.ts) moves with it. The key set goes from six to seven with `schematicMissing`, which is `null` on every existing path.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | 560 pass / 22 fail / 20 skip, all 22 pre-existing (see below) |

New tests: [check-missing-schematic.test.ts](test/check-missing-schematic.test.ts), 5 cases. All 5 pass here.

On the parent commit, 4 of the 5 fail behaviourally. The fifth ("leaves a repo with no schematic configured green and unchanged") fails there only because `schematicMissing` is `undefined` rather than `null`; its behavioural assertions already pass on the parent, which is the point of it being in the file.

Pre-existing failures: the 22 are unrelated to this change. I ran `test/init-check.test.ts` on this branch and on the parent with the same command and got the identical 3 failures both times (`clean fixture`, `broken schematic`, `fab export`), all with root cause `kicad-cli not found on PATH` in this environment. The remaining failures are in `preflight`, `repl`, `safety`, `cli-surface`, `kicad-cli-resolve`, `bootstrap`, `runmeta`, `gating-sync` and `create-hardening`, none of which this PR touches.

Sandbox variant used: hand-built minimal repo (one `.kicad_sch`, git-init'd, `.copperhead/config.json` written by hand), since the case under test is a *missing* schematic.

**Disclosure:** KiCad is not installed on this machine, so the `kicad-cli` preflight refused to start `check` at all. I set `COPPERHEAD_KICAD_CLI` to a stub that answers `version` so preflight passes. The stub is never invoked by either case below, because both skip ERC and DRC before reaching it. It only gets the command past preflight.

```text
$ copperhead --repo /tmp/ch-manual188 check
# config: { "schematic": "hardware/board.kicad_sch", "board": null }, file absent
ERC skipped: schematic configured at hardware/board.kicad_sch but the file is missing
DRC skipped (no board configured)
exit=1

$ copperhead --repo /tmp/ch-manual188 check
# config: { "schematic": null, "board": null }
ERC skipped (no schematic configured; run copperhead init)
DRC skipped (no board configured)
exit=0

$ copperhead --repo /tmp/ch-manual188 check --json
# config: { "schematic": "hardware/board.kicad_sch", "board": null }, file absent
{
  "ok": false,
  "erc": null,
  "drc": null,
  "drift": { "ok": true, "mismatches": [] },
  "openspec": null,
  "constraints": { "ok": true, "violations": [] },
  "schematicMissing": "hardware/board.kicad_sch"
}
exit=1
```

## Docs

None. The change is a message and an exit code, both covered above. Say the word if you want the `schematicMissing` field written up in the configuration reference.

---

## Invariant checklist

- [ ] N/A **Spec-gated in**: no change to the agent tool list.
- [ ] N/A **Verification-gated out**: no mutation path touched.
- [x] **`check`/`verify` stays LLM-free and network-free**: the change is a filesystem existence test and a boolean. No new imports; [check.ts](src/commands/check.ts) still reaches no provider, key, or network. The AC-2.1 module-graph test passes.
- [x] **No sexp serialization**: `src/kicad/` untouched. `pinNets` is still read-only, and now receives the already-resolved path.
- [ ] N/A **Sync-obligations ledger**: not touched.
- [ ] N/A **Secrets**: no transcript or summary path touched. The new log line contains only a config-relative path.
- [x] **Spec coherence**: no spec-level behavior change; the AC-2.4 key-set assertion moved with the new field, as described above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Checks now clearly report when a configured schematic is missing.
  * Missing schematics correctly cause validation to fail and skip dependent checks with specific messages.
  * Drift, legibility, and constraint checks now consistently reflect schematic availability.
  * Checks continue to pass as expected when no schematic is configured.
* **Tests**
  * Added coverage for missing or deleted schematics, skip messages, failure reporting, and successful configurations.

### DIFF
--- a/openspec/changes/build-copperhead-phase-1/specs/cli-surface/spec.md
+++ b/openspec/changes/build-copperhead-phase-1/specs/cli-surface/spec.md
@@ -28,6 +28,10 @@ The `copperhead` CLI SHALL provide the commands `create --brief <file>`, `init [
 - **WHEN** `check` runs on a schematic with an unconnected pin
 - **THEN** it exits non-zero and prints the violation with its sheet and location
 
+#### Scenario: Configured schematic absent (AC-2.6)
+- **WHEN** `check` runs on a repo whose configured `schematic` names a file that is not on disk
+- **THEN** it exits non-zero and names the configured path in every skip reason it prints, rather than exiting 0 with a generic "no schematic configured" message
+
 ### Requirement: JSON output mode
 With `--json`, commands SHALL emit machine-readable results with stable keys.
 
--- a/openspec/changes/build-copperhead-phase-1/tasks.md
+++ b/openspec/changes/build-copperhead-phase-1/tasks.md
@@ -65,6 +65,7 @@
 
 - [x] 7.1 Implement `copperhead check` (with `verify` alias): ERC + DRC + drift + `openspec validate` + mechanical constraint checks; zero LLM/network calls; non-zero exit on any violation (AC-2.1, AC-2.2)
 - [x] 7.2 Implement `--json` output with stable keys (AC-2.4)
+- [x] 7.4 Implement the configured-but-absent schematic case: exit non-zero and name the configured path in every skip reason (AC-2.6)
 - [x] 7.3 Tests: clean fixture passes < 60 s with network guard asserting no api.* calls; broken-pin fixture fails with location; BOM-drift fixture fails naming doc/claim/actual (AC-2.5)
 
 ## 8. `do` command end-to-end
--- a/openspec/specs/SPEC.md
+++ b/openspec/specs/SPEC.md
@@ -438,6 +438,7 @@ Format: Given / When / Then. "Fixture" = the open-telegraph repo (or the tiny te
 - **AC-2.3** With a BOM.md value edited to disagree with the schematic (e.g. wrong resistor value): drift check fails and names the doc, the claim, and the actual value.
 - **AC-2.4** `--json` emits machine-readable results (parseable, stable keys).
 - **AC-2.5** Runs in < 60 s on the fixture.
+- **AC-2.6 (configured schematic absent)** With `schematic` set in config to a path that does not exist, `check` exits non-zero and names the configured path in every skip reason it prints; it does not exit 0 reporting "no schematic configured".
 
 ### AC-3 · `copperhead do` — core loop
 
--- a/src/commands/check.ts
+++ b/src/commands/check.ts
@@ -35,6 +35,13 @@ export interface CheckResult {
     /** Advisory quantitative score; null when no schematic is configured. */
     score: ScoreReport | null;
   };
+  /**
+   * The configured schematic path, when it is configured but absent from disk.
+   * A repo with no schematic configured has genuinely nothing to verify and
+   * stays green; a repo that was configured to be checked and quietly stopped
+   * being checked is a different thing, and must not read as a pass.
+   */
+  schematicMissing: string | null;
 }
 
 export async function runCheck(repoRoot: string, log: (s: string) => void): Promise<CheckResult> {
@@ -42,9 +49,19 @@ export async function runCheck(repoRoot: string, log: (s: string) => void): Prom
   let erc: CheckReport | null = null;
   let drc: CheckReport | null = null;
 
-  if (config.schematic && existsSync(path.join(repoRoot, config.schematic))) {
-    erc = await runErc(path.join(repoRoot, config.schematic));
+  // ERC, drift and the constraint check all gate on this one condition, so it
+  // is resolved once rather than re-tested three times with the same bug.
+  const schematicPath = config.schematic ? path.join(repoRoot, config.schematic) : null;
+  const schematicUsable = schematicPath !== null && existsSync(schematicPath);
+  const schematicMissing = config.schematic && !schematicUsable ? config.schematic : null;
+
+  if (schematicUsable) {
+    erc = await runErc(schematicPath);
     log(erc.ok ? 'ERC ✓' : formatViolations(erc));
+  } else if (schematicMissing) {
+    // `copperhead init` is the wrong remedy here: the config is fine, the file
+    // it points at is not. Say which path, so a typo or a rename is obvious.
+    log(`ERC skipped: schematic configured at ${schematicMissing} but the file is missing`);
   } else {
     log('ERC skipped (no schematic configured; run copperhead init)');
   }
@@ -58,7 +75,7 @@ export async function runCheck(repoRoot: string, log: (s: string) => void): Prom
 
   let drift: DriftMismatch[] = [];
   let driftWarning: string | null = null;
-  if (config.schematic && existsSync(path.join(repoRoot, config.schematic))) {
+  if (schematicUsable && config.schematic) {
     drift = await checkDrift(repoRoot, config.docs, config.schematic);
     log(drift.length === 0 ? 'drift ✓' : drift.map((m) => `drift: ${m.doc} claims "${m.claim}" but actual is "${m.actual}"`).join('\n'));
     // Informational, never a failure: the zero-symbol drift exemption is for
@@ -76,16 +93,12 @@ export async function runCheck(repoRoot: string, log: (s: string) => void): Prom
   }
 
   let legibility: CheckResult['legibility'];
-  if (config.schematic && existsSync(path.join(repoRoot, config.schematic))) {
-    const report = await checkLegibility(path.join(repoRoot, config.schematic), {
+  if (schematicUsable) {
+    const report = await checkLegibility(schematicPath, {
       docsDir: path.join(repoRoot, config.docs),
       ...(config.legibility ? { config: config.legibility } : {}),
     });
-    const score = scoreFromGeometry(
-      await readSheetGeometry(path.join(repoRoot, config.schematic)),
-      report,
-      config.legibility,
-    );
+    const score = scoreFromGeometry(await readSheetGeometry(schematicPath), report, config.legibility);
     legibility = {
       findings: report.findings,
       counts: report.counts,
@@ -97,21 +110,27 @@ export async function runCheck(repoRoot: string, log: (s: string) => void): Prom
     log(formatLegibility(report));
     log(`legibility score: ${score.composite}/100${score.cap ? ` (capped: ${score.cap.reason})` : ''}`);
   } else {
+    // Same distinction as ERC above: an unset schematic and one configured at a
+    // path that does not exist are different situations, so the skip reason
+    // carried in the report says which one this was.
+    const reason = schematicMissing
+      ? `schematic configured at ${schematicMissing} but the file is missing`
+      : 'no schematic configured';
     legibility = {
       findings: [],
       counts: { error: 0, advisory: 0 },
-      skipped: LEGIBILITY_FAMILIES.map((family) => ({ family, reason: 'no schematic configured' })),
+      skipped: LEGIBILITY_FAMILIES.map((family) => ({ family, reason })),
       disabled: [],
       suppressed: [],
       score: null,
     };
-    log('legibility skipped (no schematic configured)');
+    log(`legibility skipped (${reason})`);
   }
 
   let constraintViolations: ConstraintViolation[] = [];
-  if (config.schematic && existsSync(path.join(repoRoot, config.schematic))) {
+  if (schematicUsable) {
     const registry = await loadConstraints(repoRoot);
-    const pins = await pinNets(path.join(repoRoot, config.schematic));
+    const pins = await pinNets(schematicPath);
     constraintViolations = checkForbiddenPins(registry, pins);
     if (Object.keys(registry).length) {
       log(
@@ -122,7 +141,11 @@ export async function runCheck(repoRoot: string, log: (s: string) => void): Prom
     }
   }
 
+  // `?? true` is right for a check that had nothing to run, but it cannot tell
+  // "nothing to verify" from "everything to verify, silently skipped", so the
+  // missing-schematic case is failed explicitly rather than defaulted to true.
   const ok =
+    schematicMissing === null &&
     (erc?.ok ?? true) &&
     (drc?.ok ?? true) &&
     drift.length === 0 &&
@@ -137,5 +160,6 @@ export async function runCheck(repoRoot: string, log: (s: string) => void): Prom
     openspec,
     constraints: { ok: constraintViolations.length === 0, violations: constraintViolations },
     legibility,
+    schematicMissing,
   };
 }
--- a/test/check-missing-schematic.test.ts
+++ b/test/check-missing-schematic.test.ts
@@ -1,0 +1,86 @@
+import { describe, it, expect } from 'vitest';
+import path from 'node:path';
+import { writeFile, rm } from 'node:fs/promises';
+import { runInit } from '../src/memory/scaffold.js';
+import { runCheck, type CheckResult } from '../src/commands/check.js';
+import { tempFixtureRepo } from './helpers.js';
+
+/**
+ * A configured-but-absent schematic must not read as a pass. Two repos are
+ * involved and they are easy to conflate: one with nothing configured, which
+ * has genuinely nothing to verify, and one that was configured to be checked
+ * and quietly stopped being checked when the file moved or was renamed.
+ */
+
+const FIXTURE_SCH = 'hardware/open-key.kicad_sch';
+
+interface Run {
+  res: CheckResult;
+  log: string[];
+}
+
+/** Run `check` over the fixture with `schematic` set to `schematic`. */
+async function runWith(schematic: string | null, opts: { removeFile?: boolean } = {}): Promise<Run> {
+  const { repo, cleanup } = await tempFixtureRepo();
+  try {
+    await runInit({ repoRoot: repo });
+    await writeFile(
+      path.join(repo, '.copperhead', 'config.json'),
+      JSON.stringify({ docs: 'docs/', schematic, board: null, budgets: {} }, null, 2),
+      'utf8',
+    );
+    if (opts.removeFile && schematic) {
+      await rm(path.join(repo, schematic), { force: true });
+    }
+    const log: string[] = [];
+    const res = await runCheck(repo, (s) => log.push(s));
+    return { res, log };
+  } finally {
+    await cleanup();
+  }
+}
+
+describe('check with a configured but missing schematic', () => {
+  it('does not report ok when the configured schematic is absent', async () => {
+    const { res } = await runWith('hardware/moved-board.kicad_sch');
+
+    expect(res.ok).toBe(false);
+    expect(res.schematicMissing).toBe('hardware/moved-board.kicad_sch');
+  }, 60_000);
+
+  it('names the missing path instead of blaming the configuration', async () => {
+    const { log } = await runWith('hardware/moved-board.kicad_sch');
+    const erc = log.find((l) => l.startsWith('ERC skipped'));
+
+    expect(erc).toContain('hardware/moved-board.kicad_sch');
+    expect(erc).toContain('missing');
+    // `copperhead init` is the wrong remedy: the config is fine, the file is not
+    expect(erc).not.toContain('no schematic configured');
+    expect(erc).not.toContain('copperhead init');
+  }, 60_000);
+
+  it('fails the same way when the file is deleted after being configured', async () => {
+    const { res, log } = await runWith(FIXTURE_SCH, { removeFile: true });
+
+    expect(res.ok).toBe(false);
+    expect(res.schematicMissing).toBe(FIXTURE_SCH);
+    expect(log.find((l) => l.startsWith('ERC skipped'))).toContain(FIXTURE_SCH);
+  }, 60_000);
+
+  it('leaves a repo with no schematic configured green and unchanged', async () => {
+    const { res, log } = await runWith(null);
+
+    expect(res.ok).toBe(true);
+    expect(res.schematicMissing).toBeNull();
+    expect(log).toContain('ERC skipped (no schematic configured; run copperhead init)');
+  }, 60_000);
+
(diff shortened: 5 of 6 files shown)


animesh-chouhan:
ack, checking

animesh-chouhan:
_Automated pr-review pass (Claude Code, `/pr-review`). Not a human review, and not a merge gate._

## Verdict: approve with nits

The diff is correct, minimal, and well tested: the three duplicated guards collapse into one resolved pair, the two skips are genuinely distinguished, `ok` no longer launders "nothing verified" into "everything passed", and every changed executable line is reached by a test. All six repo invariants were checked and pass. CI is green and I reproduced the suite locally on `9b29ca9`.

Two things to settle before merge, neither of them a defect in the diff:

1. The body says "Closes #188", but the issue's Scope section names `sync.ts:43` and `sync.ts:133` as well, and those are untouched. `copperhead sync` still prints "sync: no inconsistencies" and exits 0 on the exact repo state this PR makes `check` fail on. Either extend the fix or downgrade to "Refs #188".
2. On the fail-versus-new-state question you raised: fail is the right call, and it matches the issue's suggested direction. Worth noting in the release notes that it is user-visible: any existing repo with a stale schematic path in CI goes red on upgrade. That is the point of the change, but it should not surprise anyone.

## Metrics

```text
lines: +123 / -6 (net 117) across 3 files  ·  src +27/-5, test +96/-1, docs +0/-0, spec +0/-0, other +0/-0
new vs net: 1 new files (0 in src), 2 modified; 0 new src lines (the net-new surface)
tests: +96 test lines; suite not measured -> 582/20/0 (pass/skip/fail, base -> head)
diff coverage: 100% of changed executable src lines exercised (12/12, measured via vitest v8)
uncovered changed lines: none
```

Method:
- merge base `571b9b1` (`git merge-base origin/main HEAD`); areas from `git diff --numstat`, generated files excluded.
- suite: `npx vitest run --reporter=json` at head. The scripted run and an independent re-run I did on `9b29ca9` both report 582 pass / 20 skip / 0 fail. Your 22 local failures are environmental: `kicad-cli` is on PATH here, so they do not reproduce.
- base suite: not run. Base CI on `main` is green, which is cheaper than a worktree run.
- coverage: vitest v8 intersected with the changed src lines from `git diff -U0`. The only uncovered lines in `check.ts` are 74-77 and 85-90, both pre-existing and outside this diff. No changed line is unreached, so the untested-surface finding below is at the CLI and JSON boundary rather than in the diff's own lines.
- deps: `package.json` unchanged, so no dependency or audit delta.
- `openspec validate build-copperhead-phase-1`: not run, no planning artifact changed in the diff.

## Invariant checks

| Invariant | Result |
| --- | --- |
| Spec-gated in | pass. No file under `src/agent/` is touched; the tool list is untouched, so `edit_file`/`write_file` remain gated by omission. |
| Verification-gated out | pass for what is touched. No mutation path is modified, and the change strengthens the create-pipeline gate at [create.ts:879](src/commands/create.ts#L879) rather than weakening it. See finding 2 for the DRC half of the gate. |
| `check`/`verify` LLM-free and network-free | pass. No new imports in [check.ts](src/commands/check.ts); the AC-2.1 module-graph test at [init-check.test.ts:164](test/init-check.test.ts#L164) passed in my run. |
| No sexp serialization | pass. `src/kicad/` untouched; `pinNets` is still read-only and now receives an already-resolved path. |
| Sync-obligations ledger | pass, untouched. |
| Secrets | pass. No transcript or summary path touched, `.gitignore` unchanged, and the new log line emits only a repo-relative config path. |
| Spec coherence | see finding 4. |

## Findings

### 1. [medium] `sync` still reports "no inconsistencies" and exits 0 when the configured schematic is missing, so this cannot close the issue as written

**Where:** [sync.ts:43](src/commands/sync.ts#L43) and [sync.ts:133](src/commands/sync.ts#L133)

**Fails when:** `.copperhead/config.json` has `schematic: "hardware/moved-board.kicad_sch"` and the file is absent. `syncVerify` skips the drift scan at line 43 and the AC-7.3 forbidden-pin requirement-violation scan at line 133, returns `{ resolvable: [], violations: [] }`, `formatSyncReport` prints `sync: no inconsistencies`, and [cli.ts:264](src/cli.ts#L264) exits 0. That is the same silent green this PR removes from `check`, on a command whose whole job is finding inconsistencies, and issue #188's Scope section names both lines. Merging with "Closes #188" closes an issue whose reported behavior is still reproducible.

**Fix:** hoist the same resolved `schematicUsable` / `schematicMissing` pair into `syncVerify` and push a `SyncItem` naming the missing path when `schematicMissing` is set, with a test mirroring [check-missing-schematic.test.ts](test/check-missing-schematic.test.ts). If you would rather keep the PR at its current size, change "Closes #188" to "Refs #188" so the sync half stays tracked.

### 2. [medium] A configured-but-missing board has the identical defect three lines below the fix, and the message states the opposite of what is true

**Where:** [check.ts:53-58](src/commands/check.ts#L53-L58)

**Fails when:** config has `board: "hardware/board.kicad_pcb"` and the file is absent, which is what a rename, a moved file, or a partial checkout looks like. `drc` stays `null`, `(drc?.ok ?? true)` keeps `ok` true, the run exits 0, and the log says `DRC skipped (no board configured)` when a board is configured. Your own rationale applies verbatim: a repo that was configured to be checked quietly stopped being checked. DRC is half of the verification the check gate exists for, so a green build here means the board was never verified. Pre-existing, not introduced here, but it is now the only surviving instance of the bug this PR names.

**Fix:** mirror the schematic resolution: `boardPath` / `boardUsable` / `boardMissing`, log `DRC skipped: board configured at <path> but the file is missing`, add `boardMissing: string | null` to `CheckResult`, and add `boardMissing === null` to the `ok` conjunction. Extend [check-missing-schematic.test.ts](test/check-missing-schematic.test.ts) with the mirrored cases.

### 3. [low] The user-facing CLI reference now contradicts the code in two places, and the PR records "Docs: None"

**Where:** [cli.md:116](docs/src/content/docs/reference/cli.md#L116) and [cli.md:123](docs/src/content/docs/reference/cli.md#L123)

**Fails when:** a user reads the reference to write a CI step. Line 116 states "ERC and DRC are skipped when no schematic or board is configured, rather than failing", which is now the opposite of what happens for a configured-but-missing schematic, and line 123 enumerates the `--json` keys as `ok`, `erc`, `drc`, `drift`, `openspec`, `constraints`, which no longer matches the seven-key object. The exit-code table above them gains a new class of `1` that nothing describes.

**Fix:** amend line 116 to separate unconfigured (skip, green) from configured-but-missing (skip, fail), add `schematicMissing` to the key list at line 123, and note the new failure class in the exit-code table.

### 4. [low] The new non-zero exit condition is not recorded in the spec artifacts

**Where:** [SPEC.md:434-440](openspec/specs/SPEC.md#L434-L440) and [cli-surface/spec.md:34](openspec/changes/build-copperhead-phase-1/specs/cli-surface/spec.md#L34)

**Fails when:** someone reads AC-2 or [SPEC.md:270](openspec/specs/SPEC.md#L270) ("exit non-zero on violations") as the contract for `check`. A missing configured file is not a violation in that sense, so the new exit-1 class is undocumented at spec level, and the AC-2.4 scenario still enumerates keys as `(erc, drc, drift, violations)`. CLAUDE.md asks for SPEC.md and the change artifacts to move together, and calls a code-only PR that silently diverges from SPEC.md a defect.

**Fix:** add an AC-2.6 line to SPEC.md (a configured schematic that is absent fails, with the path named) and a matching WHEN/THEN scenario to the cli-surface delta spec, then run `openspec validate build-copperhead-phase-1`.

### 5. [low] Untested surface: the CLI exit code and the `--json` payload for the new state are covered only by the manual run

**Where:** [cli.ts:167-169](src/cli.ts#L167-L169) and [check.ts:111](src/commands/check.ts#L111)

**Fails when:** the exit mapping or the JSON serialization changes. Every new test calls `runCheck` directly, so nothing asserts that `copperhead check` actually exits 1 or that `--json` carries `schematicMissing: "<path>"`; [init-check.test.ts:98](test/init-check.test.ts#L98) asserts only the null case. Both were verified by hand with a stubbed `COPPERHEAD_KICAD_CLI`, which the suite does not reproduce, so the end-to-end contract this PR is about is the one part the suite does not hold.

**Fix:** add an execa-driven case to [cli-surface.test.ts](test/cli-surface.test.ts), which already spawns the CLI, over a temp repo whose configured schematic is absent, asserting exit code 1 and `JSON.parse(stdout).schematicMissing`.

### 6. [low] `rm(..., { force: true })` lets the delete-after-configure test silently degenerate

**Where:** [check-missing-schematic.test.ts:35](test/check-missing-schematic.test.ts#L35)

**Fails when:** `FIXTURE_SCH` stops matching the fixture layout, currently `test/fixtures/open-key/hardware/open-key.kicad_sch`. `force: true` swallows ENOENT, so the test keeps passing while exercising the same never-existed path as the first case rather than the delete-after-configure path its name promises. The regression it guards would be silently unguarded.

**Fix:** drop `force: true`, or assert the file exists before removing it.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 191, branch fix/check-missing-schematic, based
on main. It makes `copperhead check` fail when the configured schematic is missing.

Check it out first:
  gh pr checkout 191

FINDING 1 [medium] `sync` still reports "no inconsistencies" and exits 0 when the
configured schematic is missing, so this cannot close the linked issue as written.
  Where: src/commands/sync.ts:43 and src/commands/sync.ts:133
  Fails when: .copperhead/config.json has schematic: "hardware/moved-board.kicad_sch"
    and the file is absent. syncVerify skips the drift scan at line 43 and the AC-7.3
    forbidden-pin requirement-violation scan at line 133, returns
    { resolvable: [], violations: [] }, formatSyncReport prints
    "sync: no inconsistencies", and src/cli.ts:264 exits 0. That is the same silent
    green this PR removes from `check`, on a command whose whole job is finding
    inconsistencies, and issue 188's Scope section names both lines. Merging with
    "Closes 188" closes an issue whose reported behavior is still reproducible.
  Fix: hoist the same resolved schematicUsable / schematicMissing pair into
    syncVerify and push a SyncItem naming the missing path when schematicMissing is
    set, with a test mirroring test/check-missing-schematic.test.ts. If the PR should
    stay at its current size instead, change "Closes 188" to "Refs 188" in the PR
    body so the sync half stays tracked, and say that is what you did.

FINDING 2 [medium] A configured-but-missing board has the identical defect three
lines below the fix, and the message states the opposite of what is true.
  Where: src/commands/check.ts:53-58
  Fails when: config has board: "hardware/board.kicad_pcb" and the file is absent,
    which is what a rename, a moved file, or a partial checkout looks like. drc stays
    null, (drc?.ok ?? true) keeps ok true, the run exits 0, and the log says
    "DRC skipped (no board configured)" when a board is configured. DRC is half of
    the verification the check gate exists for, so a green build here means the board
    was never verified. Pre-existing, not introduced by this PR, but it is now the
    only surviving instance of the bug the PR names.
  Fix: mirror the schematic resolution: boardPath / boardUsable / boardMissing, log
    "DRC skipped: board configured at <path> but the file is missing", add
    boardMissing: string | null to CheckResult, and add boardMissing === null to the
    ok conjunction. Extend test/check-missing-schematic.test.ts with the mirrored
    cases, and update the seven-key assertion in test/init-check.test.ts:89 to eight.

FINDING 3 [low] (optional) The user-facing CLI reference now contradicts the code in
two places.
  Where: docs/src/content/docs/reference/cli.md:116 and :123
  Fails when: a user reads the reference to write a CI step. Line 116 states "ERC and
    DRC are skipped when no schematic or board is configured, rather than failing",
    which is now the opposite of what happens for a configured-but-missing schematic,
    and line 123 enumerates the --json keys as ok, erc, drc, drift, openspec,
    constraints, which no longer matches the seven-key object. The exit-code table
    above them gains a new class of 1 that nothing describes.
  Fix: amend line 116 to separate unconfigured (skip, green) from configured-but-
    missing (skip, fail), add schematicMissing to the key list at line 123, and note
    the new failure class in the exit-code table.

FINDING 4 [low] (optional) The new non-zero exit condition is not recorded in the
spec artifacts.
  Where: openspec/specs/SPEC.md:434-440 and
    openspec/changes/build-copperhead-phase-1/specs/cli-surface/spec.md:34
  Fails when: someone reads AC-2 or SPEC.md:270 ("exit non-zero on violations") as
    the contract for check. A missing configured file is not a violation in that
    sense, so the new exit-1 class is undocumented at spec level, and the AC-2.4
    scenario still enumerates keys as (erc, drc, drift, violations). CLAUDE.md asks
    for SPEC.md and the change artifacts to move together.
  Fix: add an AC-2.6 line to SPEC.md (a configured schematic that is absent fails,
    with the path named) and a matching WHEN/THEN scenario to the cli-surface delta
    spec, then run: openspec validate build-copperhead-phase-1

FINDING 5 [low] (optional) Untested surface: the CLI exit code and the --json payload
for the new state are covered only by a manual run.
  Where: src/cli.ts:167-169 and src/commands/check.ts:111
  Fails when: the exit mapping or the JSON serialization changes. Every new test
    calls runCheck directly, so nothing asserts that `copperhead check` actually
    exits 1 or that --json carries schematicMissing: "<path>";
    test/init-check.test.ts:98 asserts only the null case. Both were verified by hand
    with a stubbed COPPERHEAD_KICAD_CLI, which the suite does not reproduce.
  Fix: add an execa-driven case to test/cli-surface.test.ts, which already spawns the
    CLI, over a temp repo whose configured schematic is absent, asserting exit code 1
    and JSON.parse(stdout).schematicMissing.

FINDING 6 [low] (optional) rm(..., { force: true }) lets the delete-after-configure
test silently degenerate.
  Where: test/check-missing-schematic.test.ts:35
  Fails when: FIXTURE_SCH stops matching the fixture layout, currently
    test/fixtures/open-key/hardware/open-key.kicad_sch. force: true swallows ENOENT,
    so the test keeps passing while exercising the same never-existed path as the
    first case rather than the delete-after-configure path its name promises.
  Fix: drop force: true, or assert the file exists before removing it.

No changed line in the current diff is unreached by a test. The untested surface is
at the CLI and JSON boundary: src/cli.ts:167-169 and the schematicMissing field in
the --json payload, covered by finding 5.

Findings 2 and 4 change spec-level behavior, so openspec/specs/SPEC.md and the active
change artifacts (proposal.md, design.md, delta specs, tasks.md) have to move
together, and `openspec validate build-copperhead-phase-1` has to pass.

    by prompt text.

Then verify:
  npm run typecheck
  npm run build
  npm test

and report it.

````

**Comment by adhavan18** [comment]
rebased onto `a2471b3`, force-pushed as `a037e5f`. it conflicted with the legibility work that landed in the meantime, so here is what i did with it rather than leaving you to read the resolution.

the legibility phase added a **fourth** copy of `config.schematic && existsSync(...)`, which is the thing this PR consolidates. i folded it into the shared `schematicUsable` instead of leaving it as its own test, so the count went 3 -> 1 rather than 4 -> 2.

that also surfaced a second instance of the same bug this PR is about: the legibility skip reason was hardcoded to `'no schematic configured'`, so a configured-but-missing schematic reported the wrong reason across all 13 families. it now carries the same distinction as the ERC line:

```json
"skipped": [
  { "family": "off-grid", "reason": "schematic configured at hardware/board.kicad_sch but the file is missing" },
  ...
]
```

the AC-2.4 key set now takes both new keys, `legibility` and `schematicMissing`.

re-verified after the rebase: typecheck clean, build clean, the 5 new tests still pass, and the CLI still exits 1 on a missing file and 0 when nothing is configured. legibility stays advisory throughout, it is not part of the `ok` term.

adhavan18:
thanks, both fair.

**1. closes vs refs.** you're right, and i've changed the body to `Refs #188`. i checked whether to just extend this PR instead: `sync.ts` lines 43 and 133 don't overlap with #184's hunks (74 and 103), so there'd be no conflict. i still think it belongs in its own PR, because the fix isn't the same shape:

```
check:  has an ok boolean -> flip it, exit 1
sync:   has no ok boolean. resolvable[] items get handed to the LLM resolve phase,
        and a missing schematic file is not something an agent should try to resolve
```

so in `sync` it wants to be a `violations[]` entry (exit 2, "never auto-resolved, human decision needed" per AC-7.3), not a resolvable item, and that means widening `SyncViolationItem['kind']`. small, but a different change with its own tests. i'll send it as a follow-up and that one can close #188.

confirmed the gap you describe is real, on this branch:

```
config: schematic = hardware/board.kicad_sch (absent)
  check  -> exit 1, "schematic configured at ... but the file is missing"
  sync   -> "sync: no inconsistencies", exit 0
```

**2. user-visible change.** agreed, worth calling out in release notes. concretely: any repo whose `schematic` path is stale now fails `check` where it previously passed, and since `create.ts` calls `runCheck` the same applies there. that's the intended behaviour, but it will surface as a new failure for anyone carrying a broken config, so it shouldn't arrive unannounced.

also note this got rebased past the legibility work since your pass, so it's `a037e5f` now, not `9b29ca9`. the legibility phase had added a fourth copy of the guard and the same wrong skip reason across all 13 families; both folded in. CI is green on the current head.

adhavan18:
follow-up for the sync half is up: #226. that one closes #188, this one stays `Refs`.

it went the violations route rather than resolvable, for the reason in the design section there: `syncResolve` hands every resolvable item to the agent loop, and a schematic that isn't on disk is not something an agent should write its way out of. violations already mean "never auto-resolved, human decision" (AC-7.3), and it gives exit 2 instead of 0 for free.

```
before:  sync: no inconsistencies                                    exit 0
after:   1 REQUIREMENT VIOLATION(S) ...
         - schematic configured at hardware/board.kicad_sch but the
           file is missing, so drift and constraint checks did not run   exit 2
```

no overlap with #184's hunks in that file, so the two can land in either order.

adhavan18:
both nits from your pass are settled, so this is ready when you are.

- **closes vs refs**: body now says `Refs #188`. the sync half is #226, which does close it. checked they don't collide: sync.ts lines 43/133 vs #184's hunks at 74/103.
- **user-visible**: agreed, and worth a release note. any repo with a stale `schematic` path now fails `check` where it passed before, and `create.ts` calls `runCheck` too.

one thing to re-check since your pass ran on `9b29ca9`: it's `a037e5f` now. the legibility phase landed in between and had added a fourth copy of the guard plus the same wrong skip reason across all 13 families, so both are folded in. CI green on the current head.

while you're in the queue, #225 and #226 have had no human eyes yet, both CLEAN. #225 is the smaller of the two.

adhavan18:
pushed `cfff699`: this PR was missing its spec artifact entirely.

it changes what `check` does when a configured schematic is absent, but touched no SPEC.md entry, no delta scenario and no task — which the repo's own rule doesn't allow ("when spec-level behavior changes, update SPEC.md and the corresponding change artifacts together"). caught it while resolving a review point on #226, which had the opposite problem: its AC-7.6 carried a sentence describing *this* PR's `check` behaviour.

so: AC-2.6 under the `check` requirement, a matching scenario in the cli-surface delta, and task 7.4. #226 drops the trailing sentence from AC-7.6. numbered under AC-2 rather than extended onto AC-7.6 so the two don't depend on merge order.

no source changes, 5/5 tests and `tsc` still clean.

adhavan18:

PR #193, opened 2026-08-07
fix(doctor): catch a missing openspec CLI before spec-seed burns real tokens on it

Bhumika-1432006:
## What

`copperhead doctor` now checks whether the `openspec` CLI is present on PATH, alongside its existing node/kicad-cli/git/provider checks.

## Why

Closes #190.

`doctor` previously reported `ready` even when `openspec` was completely missing. The failure only surfaced later, inside the `spec-seed` stage, when `validate_change` tried to shell out to `openspec` and failed, after real turns and tokens were already spent. Confirmed this cost roughly 70 minutes and 54.7k output tokens across two real runs before the root cause was clear.

## Design

Adds `openspecCheck`, following the exact same pattern as `kicadCheck` and `gitCheck`: a `DoctorDeps.openspecVersion` probe, wired into `defaultDeps()`, added to the `checks` array in `runDoctor` right after `gitCheck`.

One additional Windows-specific bug found and fixed while testing: `openspec` installed via `npm install -g` (the standard install path the check's own hint recommends) resolves to an `.cmd` shim on Windows, not a real `.exe`. Node's `execFile` cannot run `.cmd` files directly without `shell: true`, so the naive probe (matching `gitVersion`'s exact pattern) false-negatived even when `openspec` was genuinely installed and working from a real shell. Using `shell: true` with an args array also triggers Node's `DEP0190` deprecation warning, so the command is folded into a single string with an empty args array instead, avoiding both issues.

Does not touch spec-gating, verification-gating, or any mutation path; this is a read-only diagnostic addition, same category as the existing checks.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | `npm run typecheck` | pass |
| Build | `npm run build` | pass |
| Unit + offline | `npm test` | not run (no test file added for this check yet) |
| Live-LLM (opt-in) | n/a | n/a |

### Manual test log (required)

- Sandbox variant used: n/a (doctor is a standalone diagnostic command, not exercised through the create/edit sandbox)
- Commands run and outcome:

```text
$ where openspec
(empty — confirmed genuinely absent from PATH)

$ node dist/cli.js doctor
  [FAIL] openspec  not found on PATH
         hint: npm i -g @fission-ai/openspec; validate_change and the create
               pipeline need it.

$ npm install -g @fission-ai/openspec
$ node dist/cli.js doctor
  [ok]   openspec  1.8.0
```

[...]

PR #194, opened 2026-08-07
refactor(tool-protocol): update protocol to support multiple JSON object replies and clarify usage instructions

### LINKED ISSUE #192: feat(agent): allow multiple tool calls per turn for CLI providers
### The problem

CLI providers (`claude-code`, `cursor`) talk to the model through the text tool protocol in [`src/agent/providers/tool-protocol.ts`](src/agent/providers/tool-protocol.ts). That protocol currently instructs the model to reply with **exactly one** JSON tool call per turn:

```text
To take an action, reply with EXACTLY ONE JSON object and nothing else
```

That conflicts with the system workflow in [`src/agent/prompts.ts`](src/agent/prompts.ts), which already tells the model that turns are scarce and independent calls should be batched in a single reply. The loop's 5-turns-remaining nudge says the same thing.

The parser (`parseToolCalls`) can already extract **multiple** brace-balanced JSON objects from one reply. The machine can batch; the protocol forbids it.

Net effect on Claude Code / Cursor runs:

- Exploration that needs several `read_file` / `search` / `record_constraint` / `resolve_affected` calls burns one turn each.
- Create stages and `do` runs hit the turn budget on serial tool use even when the work is independent.
- API providers (OpenAI / Anthropic native tool calling) already execute multiple tool calls in one turn, so CLI users pay a structural tax the keyed providers do not.

This is separate from conversation history trimming (#165 / PR #175): that reduces tokens per turn; this reduces the number of turns required for the same work.

### Proposed behavior

1. Change `renderToolProtocol` so CLI providers may emit **one or more** tool-call JSON objects in a single assistant reply (still preferably in fenced `json` blocks, still catalog-gated).
2. Keep `parseToolCalls`'s multi-object extraction as the source of truth; update the protocol text, malformed-call nudge, and any "exactly one" copy so they agree.
3. Align wording with the existing WORKFLOW batching rule and the convergence nudge in [`src/agent/loop.ts`](src/agent/loop.ts) so the instructions no longer conflict.
4. Preserve today's dispatch order for a single turn's tool list (sequential is fine for v1). Do not weaken spec-gating: tools absent from `availableTools(ctx)` must still not dispatch.
5. Add offline unit tests covering: one call, several independent calls in one reply, mixed prose + multiple calls, and rejection of locked or hallucinated tool names.

Example of an allowed CLI reply after the change:

````markdown
```json
{"tool": "read_file", "args": {"path": "docs/SPEC.md"}}
```

```json
{"tool": "read_file", "args": {"path": "docs/BOM.md"}}
```

```json
{"tool": "search", "args": {"pattern": "USB_DP"}}
```
````

All three should execute in the same turn, matching native multi-tool behavior on OpenAI/Anthropic.

### Alternatives you considered

- **Leave CLI on one-call-per-turn and only raise `maxTurns` / stage budgets.** That spends more budget on the same serial shape instead of fixing the protocol conflict. Related work like turn-budget escalation helps retries but does not remove the tax.
- **Rely on history trimming alone (#175).** Trimming lowers tokens per turn; it does not stop "one `read_file` = one turn."
- **Teach batching only in the prompt.** Already done; the protocol's "EXACTLY ONE" line undoes it for CLI providers.
- **Same-turn unlock of edit tools after `validate_change`.** Deliberately rejected elsewhere to keep the structural gate; out of scope here. This issue is only about allowing multiple already-advertised tools in one reply.

### Effect on the invariants

- **Spec-gated in:** no effect if catalog filtering stays. Edit/write tools remain absent from the advertised list until `validate_change` unlocks them; `dispatchTool` still re-checks.
- **Verification-gated out:** no effect. ERC/DRC/`finish` gates and the obligations ledger are unchanged. Batching may include `run_erc` with other calls in one turn; pass/fail semantics stay the same.

### Before submitting

- [x] I searched existing issues and the roadmap, and this is not already tracked as multi-tool CLI batching (history trim / turn-budget / stage quantization are adjacent but different).

### PR DESCRIPTION
**Pull request by Devesh36**
## What

CLI providers (`claude-code`, `cursor`) may now emit one or more JSON tool calls in a single reply. The text tool protocol no longer contradicts the workflow's batching guidance, so independent reads and similar calls can share one turn.

## Why

Closes #192.

[`tool-protocol.ts`](src/agent/providers/tool-protocol.ts) told the model to reply with `EXACTLY ONE` JSON object, while [`prompts.ts`](src/agent/prompts.ts) and the loop's convergence nudge already say turns are scarce and independent calls should be batched. `parseToolCalls` already extracted multiple objects; the protocol text was the bottleneck. Claude Code / Cursor users paid a one-tool-per-turn tax that OpenAI / Anthropic native tool calling does not.

This is separate from history trimming (#165 / #175): that cuts tokens per turn; this cuts turns for the same work.

## Design

- Update `renderToolProtocol` to allow one or more fenced JSON tool objects per reply, and to state that every call in one reply executes in the same turn.
- Keep `parseToolCalls` as the source of truth (brace-balanced multi-object scan, catalog-gated). No change to dispatch order: the loop still runs a turn's tool list sequentially.
- Soften the malformed-call nudge so it no longer says "exactly one."
- Spec-gated in is unchanged: tools absent from `availableTools(ctx)` still do not parse or dispatch. Verification gates, ledger, and `finish` are untouched.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline (focused) | `npx vitest run test/tool-protocol.test.ts test/claude-code-provider.test.ts test/cursor-provider.test.ts` | pass (54 passed, 1 skipped) |
| Unit + offline (full) | `npm test` | 581 passed, 20 skipped; 3 failed in `test/init-check.test.ts` (kicad-cli / pre-commit hook timeouts and ERC assertion), unrelated to this PR |
| Live-LLM (opt-in) | Claude Code / Cursor saved-login | not run |

New: [`test/tool-protocol.test.ts`](test/tool-protocol.test.ts) covers protocol wording, multi-call parse, prose + multiple calls, locked/hallucinated name rejection, and the malformed nudge. Updated assertions in [`test/claude-code-provider.test.ts`](test/claude-code-provider.test.ts).

- Commands run and outcome:

```text
Offline provider/protocol coverage only. A live Claude Code or Cursor run was not exercised
in this environment (needs saved-login CLI). Behavior under test is the advertised protocol
text and parseToolCalls multi-call extraction, which both CLI providers share.
```

## Docs

None. Protocol text lives in [`tool-protocol.ts`](src/agent/providers/tool-protocol.ts); WORKFLOW batching copy in [`prompts.ts`](src/agent/prompts.ts) was already correct and left unchanged.

---

## Invariant checklist

- [x] **Spec-gated in**: `edit_file`/`write_file` stay structurally absent from the agent tool list until an OpenSpec proposal validates (gated by omission, not prompt text). Catalog filtering in `parseToolCalls` / `dispatchTool` unchanged.
- [x] **Verification-gated out**: N/A for this change (no mutation path change). ERC/DRC/`finish` gates untouched.
- [x] **`check`/`verify` stays LLM-free and network-free**: N/A (no imports from `src/commands/check`).
- [x] **No sexp serialization**: N/A (no `src/kicad/` changes).
- [x] **Sync-obligations ledger**: N/A (ledger and commit gate untouched).
- [x] **Secrets**: N/A (no transcript/key handling changes).
- [x] **Spec coherence**: N/A. Aligns CLI protocol text with existing WORKFLOW batching; no SPEC.md / OpenSpec artifact change.

## Summary by CodeRabbit

* **New Features**
  * Supports multiple independent tool calls in a single response and execution turn.
  * Preserves surrounding response text when processing tool calls.
  * Provides clearer guidance when tool-call formatting is invalid.

* **Bug Fixes**
  * Improved handling and rejection of unavailable or invalid tool calls.
  * Ensures each tool call receives a unique identifier.

* **Tests**
  * Added coverage for single and multiple tool calls, empty tool catalogs, malformed calls, and mixed prose.

### DIFF
--- a/src/agent/loop.ts
+++ b/src/agent/loop.ts
@@ -567,11 +567,25 @@ async function runWithProviders(opts: RunOptions, providers: Set<Provider>): Pro
     }
     nudges = 0;
 
-    for (const call of res.toolCalls) {
+    for (let callIndex = 0; callIndex < res.toolCalls.length; callIndex++) {
+      const call = res.toolCalls[callIndex]!;
       const result = await dispatchTool(ctx, call.name, call.args);
       await transcript.event('tool', { name: call.name, args: call.args, result });
       r.toolResult(call.name, result.split('\n')[0] ?? '');
       messages.push({ role: 'tool', toolCallId: call.id, content: result });
+
+      // finish ends the run immediately. Do not dispatch calls queued after it
+      // in the same batched response: a later mutation could invalidate the
+      // verification or sync gates that finish just checked.
+      if (ctx.finishRequest && callIndex < res.toolCalls.length - 1) {
+        const skipped = res.toolCalls.slice(callIndex + 1).map(({ id, name, args }) => ({ id, name, args }));
+        await transcript.event('tool-calls-skipped-after-finish', {
+          reason: 'finish ended the run',
+          skipped,
+        });
+        log(`skipped ${skipped.length} tool call(s) queued after finish`);
+        break;
+      }
     }
 
     if (ctx.repairCycles > config.maxRepairCycles) {
--- a/src/agent/providers/tool-protocol.ts
+++ b/src/agent/providers/tool-protocol.ts
@@ -6,15 +6,18 @@ export function renderToolProtocol(tools: ToolSchema[]): string {
     '# Tool protocol',
     '',
     'You are the reasoning half of a tool-driven workflow; you cannot run anything yourself.',
-    'To take an action, reply with EXACTLY ONE JSON object and nothing else, wrapped in a',
-    '```json fenced code block:',
+    'To take an action, reply with one or more JSON objects (one object per tool call), each wrapped',
+    'in its own ```json fenced code block:',
     '',
     '```json',
     '{"tool": "<tool_name>", "args": { ... }}',
     '```',
     '',
-    'Use only the tools listed below, with `args` matching the tool\'s JSON Schema. If you have',
-    'no tool to call and only want to say something, reply with plain prose and no JSON block.',
+    'When calls are independent — several read_file or search calls, multiple record_constraint or',
+    'resolve_affected calls — emit them together in a single reply. Every tool call in one reply',
+    'executes in the same turn. Use only the tools listed below, with `args` matching the tool\'s',
+    'JSON Schema. If you have no tool to call and only want to say something, reply with plain',
+    'prose and no JSON block.',
     '',
     '## Available tools',
   ];
@@ -95,7 +98,7 @@ function detectMalformedCall(text: string, catalog: Set<string>): string | undef
       return (
         `A tool call for "${name}" looks malformed — it named the tool but did not parse as ` +
         'valid JSON (likely unbalanced braces or a missing closing brace), so no call ran. ' +
-        'Re-emit it as exactly one complete JSON object: {"tool": "...", "args": { ... }}.'
+        'Re-emit each call as a complete JSON object: {"tool": "...", "args": { ... }}.'
       );
     }
   }
--- a/test/budget-efficiency.test.ts
+++ b/test/budget-efficiency.test.ts
@@ -214,6 +214,38 @@ describe('turn-budget exhaustion (AC-15.1..15.4)', () => {
       await cleanup();
     }
   }, 60_000);
+
+  it('stops dispatching calls queued after finish in the same batch', async () => {
+    const { repo, cleanup } = await tempFixtureRepo();
+    try {
+      const provider = scriptedProvider([
+        {
+          toolCalls: [
+            { name: 'propose_change', args: { id: 'batch-finish-guard', why: 'test', what_changes: '- test', tasks: '- [ ] test' } },
+            { name: 'validate_change', args: {} },
+          ],
+        },
+        {
+          toolCalls: [
+            { name: 'finish', args: { outcome: 'done', summary: 'all gates are clear' } },
+            { name: 'write_file', args: { path: 'NOTES.md', content: 'must not be written\n' } },
+          ],
+        },
+      ]);
+      const lines: string[] = [];
+      const res = await runAgentLoop({ repoRoot: repo, request: 'guard finish batching', model: 'gpt-5', provider, maxTurns: 3, log: (l) => lines.push(l) });
+
+      expect(res.outcome).toBe('success');
+      await expect(readFile(path.join(repo, 'NOTES.md'), 'utf8')).rejects.toThrow();
+      expect(lines).toContain('skipped 1 tool call(s) queued after finish');
+
+      const transcript = await readFile(path.join(res.transcriptDir, 'transcript.jsonl'), 'utf8');
+      expect(transcript).toContain('tool-calls-skipped-after-finish');
+      expect(transcript).toContain('NOTES.md');
+    } finally {
+      await cleanup();
+    }
+  }, 60_000);
 });
 
 describe('preserveFailedRun (safety-rails)', () => {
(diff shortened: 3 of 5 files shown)


adhavan18 (commented):
Read this properly and ran it rather than eyeballing the diff. The change itself is right: `parseToolCalls` was already a brace-balanced multi-object scan, the loop already does `for (const call of res.toolCalls)`, and the protocol text was genuinely the only thing making CLI providers pay a one-tool-per-turn tax. Confirmed both ends on `2cbea65`.

One thing I think needs a guard before this lands, and it is not a defect in your diff so much as something your diff makes reachable.

## `finish` is no longer necessarily the last thing that happens in a turn

[loop.ts](src/agent/loop.ts) dispatches every call in the turn unconditionally, and only reads `ctx.finishRequest` after the whole list:

```ts
for (const call of res.toolCalls) {
  const result = await dispatchTool(ctx, call.name, call.args);
  ...
}
// ... ~15 lines later
if (ctx.finishRequest) { /* commit the run */ }
```

But `finish` validates the gates at dispatch time, in [tools.ts](src/agent/tools.ts): it checks `ctx.lastErc?.ok` and `ctx.ledger.openObligations`, and if they are clean it sets `ctx.finishRequest`. Nothing re-checks afterwards.

So a reply of `[finish, edit_file]` passes the gate and then invalidates it. I replayed exactly what the loop does, one turn, two calls, using a doc edit so nothing depends on `kicad-cli`:

```
finish    -> all gates satisfied; run will commit
edit_file -> edited docs/BOM.md

finishRequest      = {"outcome":"done","summary":"all set"}
filesTouched       = [ 'docs/BOM.md' ]
open obligations   = [ 'drift' ]
doc changed        = true
loop would commit  = true
blocking left      = [ 'drift: check_drift must run clean after doc edits' ]
```

The run commits with an open `drift` obligation, which is the exact thing the ledger exists to refuse. With a schematic edit instead of a doc edit it is the ERC gate rather than drift.

**Why I think this belongs to this PR.** The code path predates it, but the old text said:

> To take an action, reply with EXACTLY ONE JSON object and nothing else

so a turn could not contain `finish` plus anything else without the model disobeying the contract. The new text says:

> reply with one or more JSON objects ... Every tool call in one reply executes in the same turn

"When calls are independent" is good guidance, and a model that follows it will not batch `finish`. But it is guidance, not a mechanism, and the invariant here is currently being held up by the model's good manners. The PR checklist ticks "Verification-gated out: N/A (no mutation path change)", which is true of the diff and, I think, not quite true of the reachable behaviour.

## Suggested fix

Cheapest is to stop dispatching once the run is ending:

```ts
for (const call of res.toolCalls) {
  const result = await dispatchTool(ctx, call.name, call.args);
  await transcript.event('tool', { name: call.name, args: call.args, result });
  r.toolResult(call.name, result.split('\n')[0] ?? '');
  messages.push({ role: 'tool', toolCallId: call.id, content: result });
  if (ctx.finishRequest) break;
}
```

Dropping the rest is the correct reading: `finish` means the run is over, and anything the model queued behind it was a mistake. Worth recording the dropped calls in the transcript so it is visible rather than silent.

The alternative, having `finish` refuse when it is not the only call in the turn, is stricter and gives the model a corrective message instead of silently discarding work. Either is fine; the current state, where it depends on ordering the model happened to choose, is the part I would not ship.

## Everything else checked out

- multi-object parse, prose plus multiple calls, and catalog gating all behave as described
- spec-gated-in holds: a locked tool name in a batch is still rejected by `parseToolCalls`, so batching does not become a way around the edit lock. I specifically tried to get `edit_file` through in a batch before the proposal validates and could not
- the softened malformed nudge reads fine and still steers to a re-emit

One housekeeping note: the branch is based on an older `main` (before the legibility phase and #193), so `ledger.ts` and `check.ts` differ from current `main`. Worth a rebase before merge so the CI run reflects what will actually land.

Nice change overall, the batching tax was real.

Devesh36:
Fixed in commit `a0b3218`.

The agent loop now stops dispatching any tool calls queued after `finish` in the same response, preventing a later mutation from invalidating verification or sync gates. Skipped calls are recorded in the transcript, and a regression test covers the `[finish, write_file]` case.

Verified with the focused test suite, `npm run typecheck`, and `npm run build`. The full suite still reports unrelated environment failures (disk-space checks and REPL/CLI timeouts).

Issue #197, opened 2026-08-07
Recovery supervisor accepts a refusal's environmental claims unverified: a fabricated 'reduced symbol library' constraint aborted a run on a machine with the full stock set


animesh-chouhan:
Lemondrop harness run (`copperhead-test-harness/run-logs/2026-08-07T17-52-03/attempt-01/`, stage 4, `claude-code`, branch `fix/spec-budget-dual-write-reconciliation` @ da3325c).

## What happened

The stage-4 agent probed `draft_schematic` five times. Probe 3 used truncated lib_ids (`Audio:TLV320`, `Isolator:TLP`) and the validator answered with genuine in-library candidates — `closest: TLV320AIC23BPW, …` and `closest: TLP127, TLP130, TLP131, …` — proving both libraries exist and are readable. The agent probed **none** of the returned candidates (its follow-ups were names from its own head: `TLV320AIC3254`, `TLV320AIC3105`, `TLP2309`, `TLP2361_Logic`), then recorded via `record_constraint`, as "VERIFIED ABSENT", that "the Audio and Isolator libraries [are] entirely" missing and that "every suggestion probed (… TLV320AIC23BPW … TLP127 …) does not exist". `Audio:TLV320AIC23BPW` resolves on this machine.

The recovery supervisor then consumed the fabricated constraint at face value — diagnosis: "an environmental gap no retry can fix" — and aborted the stage with two of its three attempts unused. The machine has the full 222-library stock KiCad 10.0.4 package; 4 of the 6 refused subsystem blocks were drawable under correctly-spelled installed names (see #195, #185).

## Why this is a supervisor defect and not just an agent one

This is the fifth agent-fabrication sighting in the harness logs, but the first where the false facts left narration: `record_constraint` made them doc-truth, and the supervisor's abort decision quotes them. A transcript cross-check of claimed-vs-actual tool calls would not catch this variant — every tool call really ran; the false part is the *coverage claim* ("every suggestion probed"). The one thing that catches it is re-checking the claim against the machine.

## Suggested

Before accepting a refusal (or abort recommendation) whose stated blocker is symbol availability, run one deterministic re-probe of the refusal's named lib_ids — a single `draft_schematic` validation call, no LLM. If any cited-absent lib_id resolves, or the cited library yields candidates, the refusal's premise is false: classify the attempt as `unverified-refusal` (retry with the contradiction injected) instead of an environmental abort. Symbol availability is a machine fact; no refusal that turns on it should be adjudicated from prose.

Issue #196, opened 2026-08-07
Part selection commits BOM lib_ids the installed libraries cannot resolve; the first existence check runs three stages later


animesh-chouhan:
Reproduced on three of the last four harness briefs (evidence dirs in `copperhead-test-harness/run-logs/`):

- **sensor-node** (`2026-08-04T02-40-30/attempt-01`): stage 3 committed `RF_Module:ESP32-C3-MINI-1`, `Sensor_Gas:SGP40`, `Sensor_Humidity:SHT41-AD1B` — none installed. 44/46 parts of the stage-4 IR validated and the stage still had nowhere legal to go.
- **bench-psu** (`2026-08-04T02-40-37/attempt-01`): INA301A2 selected as the load-bearing current-limit comparator; no INA301 symbol exists in the installed set.
- **lemondrop** (`2026-08-07T17-52-03/attempt-01`): U5 `TLV320AIC3204`, U7 `TLP2361`, D6/D7 `PESD5V0X2UAT` — all absent (installed `Audio` tops out at TLV320AIC23B*/3100; `Isolator` has TLP2310/2761/2770; only PESD5V0L*/S1UL variants ship). All three had same-library substitutes (`TLV320AIC3100`, `TLP2310`, `PESD5V0L4UF`) that stage 3 could have picked had anything checked.

The model picks these parts because it knows them from training data, and nothing in the stage-2/3 prompts or gates says "the design must be capturable with the libraries on this machine". By the time `draft_schematic` runs the first deterministic lib_id check inside stage 4, the GPIO map and BOM are committed doc-truth, so the stage-4 agent's refusal logic is locally sound — substituting a different-package symbol under a committed pin map mis-assigns pins while passing ERC. The brief becomes unwinnable at stage 4 no matter how well the IR is authored.

Symbol availability is a machine fact checkable in milliseconds, not a judgement call. Related: #133 documents that stages 1–3 have no deterministic verification at all, and that `verify_symbols` is advisory where it does run.

**Suggested:** expose the existing resolver (`resolveLibrarySymbol`, `src/kicad/symlib.ts`) to stage 3 — either as a tool the agent must run over the BOM before finishing, or as a stage-exit gate that refuses to commit a BOM row whose lib_id does not resolve, printing the same candidates the stage-4 validator prints (with #185's cross-library search, those candidates become actionable). The mismatch then surfaces while part choice is still cheap to change.

Devesh36:
[...]

PR #198, opened 2026-08-07
fix(kicad): rank symbol candidates usefully, add cross-library search_symbols, fact-check availability refusals

### LINKED ISSUE #195: Symbol near-miss candidates are polluted by sub-unit entries and miss separator-variant names (Device:Rotary_Encoder → closest: C, D, R)
Found driving `copperhead create` end-to-end on a granular-synth brief (harness evidence: `copperhead-test-harness/run-logs/2026-08-07T17-52-03/attempt-01/`, stage 4, branch `fix/spec-budget-dual-write-reconciliation` @ da3325c). Two defects in the near-miss candidate lists that `draft_schematic` prints on a `no-symbol` finding, distinct from and compounding #185 (no cross-library search):

## 1. Fuzzy matching misses separator variants and returns garbage instead

Probing `Device:Rotary_Encoder` returned:

```
SW3: "Device:Rotary_Encoder" does not exist in the library — closest: C, D, R
```

`Device:RotaryEncoder_Switch` (plus `RotaryEncoder`, `RotaryEncoder_Switch_MP`) sits in that very library, one underscore away from the query. The matcher missed it and instead proposed three single-letter generic passives — worse than no suggestion, because it reads as "this library holds nothing like your part". The same run missed `Connector:Micro_SD_Card*` for the query `Connector:microSD_Card` (case + underscore placement).

Candidate matching should compare names case-insensitively with separators (`_`, `-`) stripped, so `Rotary_Encoder` ↔ `RotaryEncoder_Switch` and `microSD_Card` ↔ `Micro_SD_Card` rank as near-misses, and single-character symbols should never outrank multi-character prefix/substring matches for a multi-character query.

## 2. Sub-unit entries dilute the candidate list

Probing `Audio:TLV320` returned:

```
closest: TLV320AIC23BPW, TLV320AIC23BPW_0_1, TLV320AIC23BPW_1_1, TLV320AIC23BRHD, TLV320AIC23BRHD_0_1, TLV320AIC23BRHD_1_1, TLV320AIC23BxQE, TLV320AIC23BxQE_0_1
```

An 8-slot list carrying 3 real placeable symbols. `<name>_<unit>_<style>` sub-symbols are internal structure, not placeable candidates, and should be filtered before ranking.

## Why it matters

In this run the two facets fed a full stage-4 refusal and a supervisor abort: the agent concluded from these dead-end candidate lists that the machine's symbol library is "reduced" — on a box with the full 222-library stock set installed. Four of the six refused subsystem blocks were drawable under correctly-spelled installed names. With #185/#186 landing cross-library search, the candidate list becomes the agent's primary navigation instrument, so its signal-to-noise directly bounds stage-4 success.

animesh-chouhan:
Fixes #195, #196, #197. Targets `feat/deterministic-schematic-drafting` (#160), since [symsource.ts](src/kicad/draft/symsource.ts) only exists on that branch.

## What

Symbol discovery becomes navigable and refusals about missing symbols become checkable:

1. Near-miss candidate lists are ranked and separator-insensitive, so a failed probe points at the real symbol instead of at `C, D, R`.
2. A new `search_symbols` tool searches every installed library, and stage 3 must run it on every active part before the BOM commits.
3. Before the recovery supervisor judges a failed stage, every lib_id named in the failure is re-probed against the installed libraries and the results are injected as ground truth.

## Why

One harness run produced all three defects: `create` driven end to end on a granular-synth brief (`copperhead-test-harness/run-logs/2026-08-07T17-52-03/attempt-01/`), where stage 4 refused claiming the machine's symbol library is "reduced" and the recovery supervisor aborted on that claim. The machine had the full 222 library stock set installed. Four of the six refused subsystem blocks were installed under names or libraries the agent could not discover, and the refusal's "verified absent" list was contradicted by the validator's own candidate output earlier in the same run.

The three issues are the three links in that chain: bad candidates (#195) hide the real symbol, no cross-library search (#196) makes a wrong nickname look like a missing part, and an unchecked refusal (#197) turns that into an abort.

### 1. Candidate ranking, [symlib.ts](src/kicad/symlib.ts) (#195)

Both resolvers printed near-miss lists with two failure modes, seen verbatim in the run:

- `"Device:Rotary_Encoder" does not exist, closest: C, D, R`. The reverse-substring test let single-letter generics match any long query, while `RotaryEncoder_Switch` sat one underscore away, unmatched.
- `closest: TLV320AIC23BPW, TLV320AIC23BPW_0_1, TLV320AIC23BPW_1_1, ...`. Sub-unit children diluted 5 of 8 slots.

`closestSymbolNames` now matches case- and separator-insensitively (`microSD_Card` finds `Micro_SD_Card`), excludes sub-unit children, requires a name to carry at least 3 significant characters before it may match inside the query, and ranks exact, then prefix, then contains, then contained. One shared implementation replaces the two divergent copies in [symlib.ts](src/kicad/symlib.ts) and [symsource.ts](src/kicad/draft/symsource.ts).

### 2. `search_symbols` and the stage-3 requirement, [tools.ts](src/agent/tools.ts) + [create.ts](src/commands/create.ts) (#196)

Three of the last four harness briefs died in stage 4 on parts committed at stage 3 with no installed symbol: the first existence check ran three committed stages too late, after the pin map was doc truth. `search_symbols` searches every installed library and returns ranked `Lib:Name` lib_ids, exact matches first. The stage-3 prompt now requires probing every active part with it before the BOM commits, and states the invariant: stage 4 draws only from installed symbols.

This complements #186 (resolver-level cross-library fallback). That fixes the probe's error path; this gives the agent a discovery primitive before anything commits.

### 3. Refusal fact-checking, [recovery.ts](src/agent/recovery.ts) (#197)

`symbolAvailabilityFacts` extracts every lib_id named in the failure text and transcript excerpt (capped at 8, file:line refs and engine-generated power symbols filtered out), re-probes each against the installed libraries, and injects the results into the diagnosis prompt as ground truth, with an instruction that an agent's claim of an absent symbol is not evidence. It costs no LLM turn and never throws: a probe error degrades to reporting what it could.

In the recorded run, the fabricated "Audio and Isolator libraries entirely absent" claim would have arrived alongside `Audio:TLV320AIC23BPW: RESOLVES on this machine (28 pins)`, turning an environmental abort into a retry with the correct lib_ids in the guidance.

## Design

- The candidate ranker is pure and synchronous, so both resolvers and the cross-library search share one ordering, and the sub-unit filter is a name-shape test (`_<unit>_<style>`) rather than a parse: [symsource.ts](src/kicad/draft/symsource.ts) scrapes names with a regex and has no symbol tree to consult at that point.
- `search_symbols` is `requiresUnlock: false`. It is read-only over the machine's libraries and writes nothing, so it does not touch the spec gate.
- The fact block is deterministic and provider-free. It runs before `diagnoseStageFailure` and is passed in as data, so the supervisor's judgment is the only LLM step and the facts it judges from cannot be hallucinated.
- No KiCad file is written or serialized anywhere in this change: all three paths read `.kicad_sym` text.

## Spec / OpenSpec

No delta spec is updated by this PR, and the invariant checkbox below is left unchecked for a reviewer's call rather than claimed. The judgment: the `deterministic-schematic-drafting` agent-core delta enumerates tool-list additions (it has a paragraph for `check_legibility`), so `search_symbols` arguably belongs there as a sibling sentence. Happy to add it in this PR if a reviewer agrees. The candidate ranking and the refusal fact check are behavior-preserving quality fixes to existing paths and do not read as spec-level.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Build | `npm run build` | pass, `copperhead --version` prints 0.9.0 |
| Unit + offline | `npm test` | pass, 728 passed / 20 skipped (53 files) |
| Live-LLM (opt-in) | keyed on `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `COPPERHEAD_TEST_CODEX` | skip, not configured in this environment (19 of the 20 skips) |

New tests in [symlib.test.ts](test/symlib.test.ts) use the run's real strings as fixtures: the `C, D, R` case, the sub-unit dilution case, the `TPS61165DBV` in `Driver_LED` wrong-nickname case, and a fact check of the recorded refusal's exact claims (one false, one wrong-library).

One pre-existing expectation was updated: `R` is no longer suggested for `Device:R_Nonexistent`, which was the garbage-match behavior under fix. That test now asserts the opposite, and a new sibling test covers the legitimate near-miss (`Device:R_Smal` still suggests `R_Small`). No unrelated pre-existing failures.

Sandbox variant used: none. All three changed paths are deterministic and provider-free, so they were exercised directly against this machine's real installed libraries rather than through a sandbox run: KiCad 10.0.4, Node v24.18.0, 222 libraries in `/usr/share/kicad/symbols`. The live stage-3 prompt change is the one part not re-run here; its evidence is the recorded harness run cited above.

`search_symbols` through its real tool handler, including the three wrong-nickname parts from the run and a genuine miss:

```text
$ search_symbols TPS61165DBV
installed symbols matching "TPS61165DBV":
  - Driver_LED:TPS61165DBV

$ search_symbols AudioJack3
installed symbols matching "AudioJack3":
  - Connector_Audio:AudioJack3
  - Connector_Audio:AudioJack3_Ground
  - Connector_Audio:AudioJack3_Switch
  - Connector_Audio:AudioJack3_SwitchT

$ search_symbols TLV320AIC23BPW
installed symbols matching "TLV320AIC23BPW":
  - Audio:TLV320AIC23BPW

$ search_symbols NotAPartAnywhere123
no installed symbol matches "NotAPartAnywhere123" (searched every library in:
/usr/share/kicad/symbols). The part is not capturable on this machine as named,
choose a part whose symbol exists, or a same-family variant that does.
```

The candidate list for the exact lib_id that produced `closest: C, D, R`, now resolved against the real `Device` library:

```text
$ resolveLibrarySymbol("Device:Rotary_Encoder")
{"status":"no-symbol","candidates":["RotaryEncoder","RotaryEncoder_Switch","RotaryEncoder_Switch_MP"]}
```

`symbolAvailabilityFacts` on the recorded refusal's own text, including its `create.ts:311` reference to confirm file:line refs are not probed:

```text
$ symbolAvailabilityFacts("Stage 4 refused: this machine has a reduced symbol library.
  The Audio and Isolator libraries are entirely absent; Audio:TLV320AIC23BPW and
  Isolator:PC817 could not be found, and Device:Rotary_Encoder does not exist (create.ts:311).")

- Audio:TLV320AIC23BPW: RESOLVES on this machine (28 pins)
- Isolator:PC817: RESOLVES on this machine (4 pins)
- Device:Rotary_Encoder: not in that library (closest in that library: RotaryEncoder,
  RotaryEncoder_Switch, RotaryEncoder_Switch_MP); installed as: Device:RotaryEncoder,
  Device:RotaryEncoder_Switch, Device:RotaryEncoder_Switch_MP
```

That is the block the supervisor would have judged the abort against: both "entirely absent" libraries resolve, and the third claim is a real miss with the correct name attached.

## Docs

None. No README, configuration reference, or CHANGELOG entry is affected: the change adds one agent tool and tightens two internal code paths.

Note for reviewers: this conflicts lightly with #186 in [symsource.ts](src/kicad/draft/symsource.ts) and [symlib.ts](src/kicad/symlib.ts), since both touch the candidate path. Whichever lands second rebases the candidate call onto the shared `closestSymbolNames`.

---

## Invariant checklist

- [x] **Spec-gated in**: unchanged. `search_symbols` is read-only and registered with `requiresUnlock: false`; `edit_file`/`write_file` gating is untouched.
- [x] **Verification-gated out**: unchanged. No mutation path, repair cycle, or rollback is touched.
- [x] **`check`/`verify` stays LLM-free and network-free**: unchanged. Nothing reachable from [check.ts](src/commands/check.ts) is modified; the new code lives in the agent tool list, the recovery supervisor, and `symlib`, and reads only local library files.
- [x] **No sexp serialization**: all three paths read `.kicad_sym` text and never write or round-trip a KiCad file.
- [x] **Sync-obligations ledger**: unchanged. No hook, obligation, or commit gate is touched.
- [x] **Secrets**: unchanged. No new transcript writer or environment read; the fact block contains only lib_ids and pin counts.
- [ ] **Spec coherence**: see the Spec / OpenSpec section. No delta spec updated; reviewer's call whether the `search_symbols` tool addition needs a line in the `deterministic-schematic-drafting` agent-core delta.

## Summary by CodeRabbit

* **New Features**
  * Added symbol search across installed KiCad libraries, with exact matches prioritized.
  * Added clearer reporting when libraries or matching symbols are unavailable.
  * Added automated verification of referenced symbols during component selection and recovery.

* **Bug Fixes**
  * Improved missing-symbol suggestions with consistent ranking, separator-insensitive matching, and exclusion of sub-unit names.
  * Prevented unsupported symbol claims from proceeding when installed-library evidence disagrees.


animesh-chouhan:
@Chirag6722 take a look

Chirag6722:
@animesh-chouhan Looked at #198, makes sense, and the ranker handles separators and sub-units well.

One issue that isn't covered in #198 though: `symbolSearchDirs` doesn't find KiCad on Windows at all. The path is hardcoded as `C:/Program Files/KiCad/share/kicad/symbols`, but the installer always creates a version directory (`C:/Program Files/KiCad/10.0/share/...`), so every lib_id fails to resolve on a stock Windows install, including `Device:R`. That's what sent my run into the "no symbol library exists" refusal loop for two stage attempts before I traced it. Same function also ignores `KICAD_SYMBOL_DIR`: the auto-discovery runs even when an override is set, so a pinned directory silently gets the machine's real install merged in.

PR #202, opened 2026-08-07
feat(kicad): pin dossier at stage-4 entry, symbol_pins tool, stage-3 drawability check

### LINKED ISSUE #201: Stage 4 spends most of its turns reconstructing pin tables the machine already has
## Observed

Two harness runs of the lemondrop brief (`copperhead-test-harness/run-logs/2026-08-07T17-52-03`, STM32H743 granular synth, 158 refdes), both on the #198 build:

- **attempt-04, stage-attempt 1** (the furthest any stage-4 run has gotten): of the 19 turns to an accepted draft, roughly turns 3–18 — about 10 of 15 minutes — went to recovering pin names, largely via `read_file` over raw `.kicad_sym` spans. The transcript shows the agent paging through polyline geometry (`(xy -3.81 1.27) …`) hunting for `(pin …)` nodes for BAV99, TPS61165, 6N137, SS34.
- **attempt-02, stage-attempt 2** found a different route to the same data: a "probe draft" containing only unproven parts, each with one deliberately bogus pin, so a single validation report would echo back every real pin list ("Small IR containing only the parts whose lib_id or pin names are still unproven … each with a bogus pin, so one report returns every real pin list").

Both are workarounds for the same gap: the agent must author `REF.PIN` net endpoints from real library pin numbers, but nothing hands it a pin table. The stage prompt's own oracle is reactive ("validation lists a part's actual pins when you name one that does not exist") — pins are learned one failed guess at a time, or by reading sexp files never meant for that.

## Why it's cheap to fix

Everything needed already exists:

- `resolveLibrarySymbol` (src/kicad/symlib.ts) resolves a lib_id to its real pins, following `extends` links.
- `pinsOfSymbolNode` extracts number/name/electrical type.
- `searchInstalledSymbols` (#198) maps a part name to ranked lib_id candidates.
- `symbolAvailabilityFacts` (#198 + review follow-ups) is the established pattern for a capped, coverage-disclosed machine-fact block injected into a prompt.

And the BOM is frozen before stage 4 starts, so the part list is known at stage entry.

## Proposal

1. **Pin dossier at stage-4 entry**: before the first agent turn, resolve every BOM part (MPN, falling back to Value) to its installed-symbol candidates and inject a compact block into the stage prompt: top lib_id per part, pin table (number=name/type), and a MULTI-UNIT flag where the drafting engine would refuse the symbol (the SN74LVC2G17 trap from attempt-02). Capped by size with the overflow named, never silently truncated — same contract as `symbolAvailabilityFacts`.
2. **`symbol_pins` tool**: takes a lib_id, returns the resolved pin table; on a miss, ranked candidates in that library plus cross-library hits. Covers parts the agent swaps mid-stage, when the entry dossier is stale. Read-only, no unlock.

Expected effect: the ~15 discovery turns collapse into zero for dossier-covered parts; the bogus-pin probe pattern loses its reason to exist; first draft call moves from turn ~19 to roughly turn 2–3.

Evidence transcripts: `run-logs/2026-08-07T17-52-03/attempt-04/evidence/2026-08-07T20-26-54-462Z/transcript.jsonl` (read_file geometry paging), `attempt-02/evidence/2026-08-07T16-33-02-143Z/transcript.jsonl` (bogus-pin probe). Tracked harness-side as R14.

animesh-chouhan:
Closes #201. Targets `feat/deterministic-schematic-drafting`.

**Stacking note:** this branch is the full symbol-discovery lineage rebased onto the cross-library resolution base: it carries #186's merge (`fd1a14c`, `6e3442b` — currently local-only on the base branch, so they appear in this diff until the base is pushed) and the rebased #198 commits (`761e001`, `5f08bc4`). Merging this makes PR #198's branch redundant. The new work is the last two commits.

## What

1. **Pin dossier at stage-4 entry** — every BOM part resolved against the installed libraries before the schematic agent's first turn, injected into the stage prompt: top lib_id, real pin table (number=name/type), MULTI-UNIT flag, alternative candidates, explicit NO-INSTALLED-SYMBOL lines. Recomputed per stage attempt; size-capped with named overflow; degrades to no block on any error, and yields **no** dossier when no library is readable — "could not check" must never render as "checked and absent".
2. **`symbol_pins` tool** — pins + unit count for a lib_id, following `extends`; misses answer with in-library closest names and cross-library matches; warns on multi-unit symbols (the drafting engine refuses them).
3. **Stage 3 is instructed to confirm drawability, not just existence** — the prompt directs a `symbol_pins` single-unit check before a BOM row commits (prompt strength only: the stage completion gate is unchanged, as the delta spec states), closing the `SN74LVC2G17` → `PESD5V0L4UG` exists-but-refused trap that burned a stage-attempt.
4. **Rebase reconciliation** (`ef194d7`) — with #186's `found-elsewhere` outcome in the union type, two #198-era callers (`symbolAvailabilityFacts`, `symbol_pins`) rendered a *located* part as "not installed"; both now report the located lib_ids. The #186 Windows-discovery tests also asserted against only fixture dirs and failed on any machine with a real KiCad install; assertions now filter to the fixture root.

Merge-resolution choices, for review: one shared library lister (`listInstalledLibraries`) replaces `allLibraryFiles`; same-file candidates use the shared ranked `closestSymbolNames` (sub-unit filtered, no single-letter garbage) inside #186's evidence ordering (same-file first, cross-library only when the file offers nothing, fuzzy weaker than exact); `findSymbolAcrossLibraries` keeps its edit-distance guard (`SHT40`→`SHT4x` catchable, `TPS22860`≠`TPS22810`).

## Why

Measured on the lemondrop harness runs (#201): the best stage-4 run to date spent ~15 of 19 turns (~10 of 15 minutes) paging raw `.kicad_sym` geometry to recover pin names, and an earlier attempt probed drafts with deliberately bogus pins to make the validator echo the real pin lists. The BOM freezes before stage 4, so all of it is computable before turn 1.

## Spec

`openspec/changes/add-symbol-pin-dossier/` — proposal, tasks, delta specs for `agent-core` and `create-pipeline`. `openspec validate` passes.

## Testing

| Check | Result |
| --- | --- |
| Typecheck | pass |
| Full suite | 754 passed, 0 failed |
| `openspec validate add-symbol-pin-dossier` | valid |
| Build + `copperhead --version` | pass |

New coverage in `test/dossier.test.ts` (14 tests): MPN/Value resolution, passive omission, no-symbol lines, multi-unit flag, size-cap disclosure, unreadable-library silence, unit counts through `extends`, and all `symbol_pins` outcomes.

## Invariant checklist

- [ ] N/A **Spec-gated in**: `symbol_pins` is read-only, `requiresUnlock: false`, same tier as `search_symbols`.
- [ ] N/A **Verification-gated out**: no gate changes; the dossier is prompt content only.
- [x] **`check`/`verify` stays LLM-free and network-free**: all new paths are filesystem-only.
- [x] **No sexp serialization**: read-only parsing; nothing written or round-tripped.
- [ ] N/A **Secrets**: none.
- [x] **Spec coherence**: the refusal contract is preserved — the dossier informs, never guesses a pinout; false absence is spec-forbidden in both delta specs.

## Summary by CodeRabbit

- **New Features**
  - Added symbol search across installed KiCad libraries, with ranked matches and cross-library suggestions.
  - Added symbol pin lookup showing pin details, electrical types, unit counts, alternatives, and multi-unit warnings.
  - Schematic generation now verifies symbol availability and provides a bounded pin dossier for selected BOM parts.
- **Bug Fixes**
  - Improved handling of missing, unreadable, or incorrectly referenced symbols with actionable diagnostics.
  - Added automatic KiCad symbol-directory discovery on Windows.
- **Tests**
  - Expanded coverage for symbol matching, discovery, resolution, diagnostics, and dossier generation.


animesh-chouhan:
_Automated pr-review pass (Claude Code). Not a human review._

## Verdict: request changes

The design is right and the spec artifacts are coherent, but the availability path that the whole PR rests on has a hole: the function the dossier and `search_symbols` use to decide "is this part installed" lacks the edit-distance rule that its sibling resolver has, so it prints `NO INSTALLED SYMBOL matches` for parts that are installed. Measured against the real library set on this machine, `STM32F103C8T6` and `SHT40` both come back absent while `MCU_ST_STM32F1:STM32F103C8Tx` and `Sensor_Humidity:SHT4x` are sitting there. That is the false-absence the PR body and both delta specs forbid, on the default path, inside a block labelled "machine-verified".

**Review scope and disclosures**

- Reviewed at head `4c5f088`. The branch was force-pushed during this review (`ef194d7` to `4c5f088`, a rebase onto the moved base); the six reviewed source files are content-identical across that rebase, only `src/commands/create.ts` line numbers shifted, and all line references below are at `4c5f088`.
- The local checkout has uncommitted in-flight edits to `recovery.ts`, `tools.ts`, `dossier.ts`, `symlib.ts` and a staged deletion of `node_modules`. None of that is in the PR, and this review does not cover it.
- Every commit under review is agent-co-authored, and `cc3f2ae` (formerly `5f08bc4`) is co-authored by the same model writing this review. Findings 1 and 5 land on that commit, so treat them as self-review and check them yourself rather than taking them on trust; both come with reproductions you can run.
- No earlier pr-review pass exists on this PR, so this is a full report. CodeRabbit's pass was still in progress at fetch time.

## Metrics

```
lines: +1445 / -29 (net 1416) across 18 files  ·  src +654/-27, test +516/-2, docs +0/-0, spec +274/-0, other +1/-0
new vs net: 11 new files (1 in src), 7 modified; 127 new src lines (the net-new surface)
tests: +516 test lines; suite not measured -> 750/20/0 (pass/skip/fail, base -> head)
diff coverage: 87.2% of changed executable src lines exercised (381/437, measured via vitest v8); uncovered: src/agent/recovery.ts:157-159,174-187,190,234, src/agent/tools.ts:359-367,402-405,407-409, src/commands/create.ts:858-861,865, src/kicad/dossier.ts:85-87,99,107-109,125-126, src/kicad/draft/symsource.ts:317-320, src/kicad/symlib.ts:48-49,127
ci: 8 pass / 0 fail / 0 pending; required checks: none reported (gh pr checks)
mergeable: MERGEABLE, not a draft (gh pr view)
```

Method: change size and area split from `git diff --numstat` against merge base `ff4fe56`, generated files excluded. Suite from an actual `npx vitest run --reporter=json` at head: 750 passed, 20 skipped, 0 failed, 770 total. The PR body reports 754 passed; `test/draft-reference-boards.test.ts:21` registers one test per directory under `manual-tests/reference-boards/`, so the total moves with that directory, and every run in this session had 0 failures. Base suite not run (no `--base-tests`; base CI reports it). Diff coverage measured with `@vitest/coverage-v8@3.2.7` intersected with the changed src lines: the version-matched pin is needed, since plain `npm i -D @vitest/coverage-v8` resolves to 4.1.10 and ERESOLVEs against the repo's vitest 3.2.7. No dependency changes in the diff (`package.json` untouched).

## Invariant checks performed

| Invariant | Result |
| --- | --- |
| Spec-gated in (`edit_file`/`write_file` structurally absent until a proposal validates) | pass: both new tools are `requiresUnlock: false` on the read-only tier, `edit_file`/`write_file` keep `requiresUnlock: true`, and `availableTools` still filters by omission |
| Verification-gated out (ERC/DRC, repair to `maxRepairCycles`, rollback) | pass: `finish` and the repair/rollback paths are untouched, `src/agent/loop.ts` is not in the diff |
| `check`/`verify` LLM-free and network-free | pass: the transitive import graph of `src/commands/check.ts` reaches neither `symlib.ts`, `dossier.ts`, nor `recovery.ts`, and the new files contain no `fetch`/`http`/URL |
| No sexp serialization | pass: all new parsing is read-only, the `symsource.ts` diff touches error-suggestion branches only |
| Sync-obligations ledger | pass: both new tools are read-only and touch no file, so no hook entry is owed, consistent with `list_symbols`/`verify_symbols` |
| Secrets | pass: `src/util/redact.ts` unchanged, `.gitignore` keeps `.env` and `.copperhead/runs/`, dossier content is lib_ids and pin names only |
| Spec coherence | pass: `openspec validate add-symbol-pin-dossier` and `add-symbol-search-tool` both valid; the delta specs match the code, including the deliberate "prompt strength, gate unchanged" choice for stage 3 |

## Findings

### 1. high: `searchInstalledSymbols` has no edit-distance tier, so the machine-verified dossier declares installed parts absent

**Where:** `src/kicad/symlib.ts:309-339` (ranking via `rankSymbolNames`, `symlib.ts:139-165`), consumed at `src/kicad/dossier.ts:90-93` and `src/agent/tools.ts:362-364`.

`findSymbolAcrossLibraries` matches on containment **and** `editDistanceWithin(q, lower, 1)` (`symlib.ts:284-293`). `searchInstalledSymbols` ranks only exact / prefix / substring / name-inside-query. A datasheet suffix that differs mid-string from KiCad's family name therefore never matches, and the two resolvers disagree about the same part.

**Fails when:** any BOM row names an MPN whose stock symbol is a family variant. Against the real `/usr/share/kicad/symbols` on this machine:

```
searchInstalledSymbols('STM32F103C8T6')   -> []
findSymbolAcrossLibraries('STM32F103C8T6') -> [{ lib: 'MCU_ST_STM32F1', name: 'STM32F103C8Tx', exact: false }]
resolveLibrarySymbol('MCU_ST_STM32F1:STM32F103C8T6') -> found-elsewhere ['MCU_ST_STM32F1:STM32F103C8Tx']
searchInstalledSymbols('SHT40')            -> []
findSymbolAcrossLibraries('SHT40')         -> [{ lib: 'Sensor_Humidity', name: 'SHT4x', exact: false }]
```

A 15-row BOM run through `bomSymbolDossier` against those libraries produced, under the heading "Installed-symbol pin dossier (machine-verified)":

```
- U1 (STM32F103C8T6): NO INSTALLED SYMBOL matches — not capturable as named; substitute a part whose symbol exists (search_symbols to find one)
- U4 (SHT40): NO INSTALLED SYMBOL matches — not capturable as named; substitute a part whose symbol exists (search_symbols to find one)
```

`search_symbols` says the same thing ("The part is not capturable on this machine as named"), and the new stage-3 prompt (`src/commands/create.ts:188`) instructs the agent to substitute the part on exactly that evidence. So the run drops a correctly chosen STM32 and SHT40 for something else, and stage 4 opens with false pin facts. `add-symbol-pin-dossier/specs/create-pipeline/spec.md` requires that a machine-verified block never render "could not check" as "checked and absent"; the guard for that only covers the no-readable-library case (`dossier.ts:63`), not the search-miss case. Note the irony: `SHT40 -> SHT4x` is the example the PR body cites as fixed, and it is fixed, through the other function.

**Fix:** give `rankSymbolNames` a fourth tier using the same `editDistanceWithin(q, c, 1)` rule with the existing `MIN_EDIT_MATCH_LEN` guard, so `searchInstalledSymbols`, the dossier, and `search_symbols` inherit it; or, at minimum, have `dossier.ts:92` and `tools.ts:363` fall back to `findSymbolAcrossLibraries(query, dirs)` before emitting any absence line, and reword `dossier.ts:93` to state a search miss rather than an absence. Regression test: a fixture library containing only `STM32F103C8Tx`, asserting `searchInstalledSymbols('STM32F103C8T6', [fixtureDir])` is non-empty and that the dossier line for that row is not `NO INSTALLED SYMBOL`.

### 2. medium: `symbolAvailabilityFacts` has no "no readable library" guard, so it reports absence as ground truth when nothing was readable

**Where:** `src/agent/recovery.ts:160`.

The guard checks `searchDirs.length`, not whether any `.kicad_sym` was found. `bomSymbolDossier` has the stronger guard (`dossier.ts:63`, `if (!(await listInstalledLibraries(dirs)).size) return ''`); recovery does not.

**Fails when:** a symbol directory exists but holds no libraries (a stale `KICAD_SYMBOL_DIR`, or a packaging split where the symbols package is absent). Reproduced against an existing empty directory:

```
facts:
- Device:R: no library of that nickname is installed; no installed symbol matches "R" in any library
- Audio:TLV320AIC3100: no library of that nickname is installed; no installed symbol matches "TLV320AIC3100" in any library
dossier: ""
```

Those lines are injected under "Each line reported below is ground truth and overrides anything the transcript claims about that symbol's availability" (`recovery.ts:230`), and the prompt then tells the diagnostician that a contradicted absence means retry. Here the block confirms a bogus refusal instead, pushing the verdict to abort. The whole rendering branch is uncovered (`recovery.ts:174-187`).

**Fix:** after `recovery.ts:160`, add `if (!(await listInstalledLibraries(searchDirs)).size) return '';` and a test that asserts an empty-but-existing symbol dir yields `''`.

### 3. medium: `verify_symbols` reports "0 issue(s) to reconcile" while printing an issue

**Where:** `src/kicad/symlib.ts:496-503` and `src/agent/tools.ts:432`.

The new `found-elsewhere` branch pushes a finding tagged `kind: 'no-library'` and `continue`s without incrementing `checked` or `skipped`. `SymbolFinding['kind']` (`symlib.ts:441`) has no `found-elsewhere` member, and `tools.ts:432` computes `mismatches = findings.filter((f) => f.kind !== 'no-library').length`, which excludes it.

**Fails when:** a schematic's only divergent `lib_symbols` entry names an uninstalled nickname for a part installed elsewhere. The tool returns:

```
verify_symbols: 0 verified, 0 unverifiable (library not installed), 0 issue(s) to reconcile:
  - [no-library] "X:Y" names the wrong library, not the wrong part: use A:Y instead.
```

The agent is told there is nothing to reconcile in the same breath as the actionable finding, and a model that reads the summary line rather than the list moves on.

**Fix:** add `'found-elsewhere'` to `SymbolFinding['kind']` and emit the finding with that kind so `tools.ts:432` counts it as a mismatch. Regression test: assert the summary reports `1 issue(s)` for a found-elsewhere entry.

### 4. medium: the new pre-diagnosis probe runs outside the diagnosis watchdog

**Where:** `src/commands/create.ts:487`.

`symbolAvailabilityFacts(...)` is awaited on the line before `withTimeout(...)`, so the provider turn it feeds has a deadline and it does not. It probes up to 8 lib_ids, and each miss triggers up to two full-library sweeps (`findSymbolAcrossLibraries` inside `resolveLibrarySymbol`, then `searchInstalledSymbols`), at roughly 1 s per sweep on this machine, while the provider opened at `create.ts:482` is held.

**Fails when:** a stage fails on a machine whose symbol libraries sit on a slow or stalled network mount. Recovery hangs with no deadline instead of failing safe to `abort`, and every other provider step in that function is bounded.

**Fix:** move the call inside the `withTimeout` closure, or wrap it in its own `withTimeout` that degrades to `''`.

### 5. medium: a machine-local absolute `node_modules` symlink is committed on the branch

**Where:** repository root, mode `120000`, added by `cc3f2ae` (formerly `5f08bc4`).

The tracked blob is the literal string `/home/animesh/Github/chouhan-industries/copperhead/node_modules`. Confirmed present on the PR head via the contents API (`type: symlink`, `target: /home/animesh/...`). `.gitignore:5` is `node_modules/` with a trailing slash, which matches directories only, so a symlink of that name slipped past. A staged deletion exists in the author's local checkout but is not part of the PR.

**Fails when:** anyone clones the branch. They get a dangling absolute symlink at the repo root pointing into the author's home directory; `npm ci` happens to survive because it removes `node_modules` first, which is why CI is green, but `npm install` and any tool that follows the link do not. On the author's own machine a second clone silently resolves its dependencies from the original checkout. It also publishes a local home path.

**Fix:** `git rm --cached node_modules` and change `.gitignore:5` to `node_modules` with no trailing slash.

### 6. medium: building the dossier costs about a second of uncached library I/O per BOM part, on every stage-4 attempt

**Where:** `src/commands/create.ts:851-865` calling `src/kicad/dossier.ts:90`, backed by `src/kicad/symlib.ts:309-339`.

Neither `listInstalledLibraries` nor `searchInstalledSymbols`/`findSymbolAcrossLibraries` caches anything: each call re-reads every installed `.kicad_sym` (222 files, 221 MB here). Measured on a warm page cache with a single search dir: 0.91 s for one `searchInstalledSymbols`, and **15.7 s** for a 15-row BOM through `bomSymbolDossier`. It runs serially before the first agent turn, with no progress output, recomputed on every stage attempt.

**Fails when:** a 40 to 50 part BOM enters stage 4, or a retry loop re-enters it, on a cold cache or with several search dirs. Compounding it, the size-cap early exit at `dossier.ts:84` (`if (spent >= maxChars)`) is effectively unreachable, since `spent` only grows while `spent + line.length <= maxChars`: the cap bounds the rendered output but not the work, so parts past it still pay a full scan each.

**Fix:** build the per-directory `Map<lib, names[]>` index once (module-level, keyed on the resolved dirs) and share it across the dossier, `findSymbolAcrossLibraries`, and the fact block; and make the cap check `spent + estimated > maxChars` so overflow parts short-circuit before the scan.

### 7. low: the `found-elsewhere` message contradicts itself when the match is in the same library

**Where:** `src/kicad/symlib.ts:500` and `src/kicad/draft/symsource.ts:198`.

`findSymbolAcrossLibraries` skips only the *exact* check for `excludeLib` (`symlib.ts:252`), so a fuzzy hit inside the guessed library comes back as `found-elsewhere`.

**Fails when:** the library nickname is right and the name is a variant. Verified: `resolveLibrarySymbol('MCU_ST_STM32F1:STM32F103C8T6')` returns `found-elsewhere ['MCU_ST_STM32F1:STM32F103C8Tx']`, which renders as `"MCU_ST_STM32F1:STM32F103C8T6" names the wrong library, not the wrong part: use MCU_ST_STM32F1:STM32F103C8Tx instead`. The library was not wrong.

**Fix:** in both message builders, branch on whether every suggested lib_id shares the queried nickname and say "does not exist in that library, closest real name: ..." in that case.

### 8. low: a model-supplied lib_id can escape the symbol search directories

**Where:** `src/kicad/symlib.ts:111` (`path.join(dir, `${lib}.kicad_sym`)`), reachable from `symbol_pins` (`src/agent/tools.ts:371-410`), which is the first tool to let the model control that path segment directly.

**Fails when:** the model passes `../outside/Private:SECRET_PART`. Verified with a fixture outside the search dir:

```
resolveLibrarySymbol('../outside/Private:SECRET_PART', [libsDir])
-> {"status":"ok","pins":[{"number":"1","name":"TOP","type":"passive"}],"units":1}
```

The blast radius is small (the target must be a parseable `.kicad_sym`, and only symbol and pin names are echoed), and these tools legitimately read outside the repo, so this is not an AC-4.2 sandbox break. It is still an unnormalized model-controlled path segment.

**Fix:** reject a `lib` containing `/`, `\`, or `..` in `findLibraryFile` before joining.

### 9. low: per-part probe errors are reported to the model as size-cap overflow

**Where:** `src/kicad/dossier.ts:107` versus the label at `dossier.ts:119-121`.

The per-part `catch` pushes onto `overflow`, which renders as `NOT INCLUDED (size cap 24000 chars)`. A part whose library was unreadable is described as one omitted for length, and if every part throws, the entire block is a single line blaming a cap that was never approached. Separately `overflow.join('; ')` is not counted against `maxChars`, so the block can exceed the bound it names. Both paths are uncovered (`dossier.ts:107-109`).

**Fix:** track error-skipped parts in their own list with a "could not be checked here" line, preserving the "could not check is not absent" distinction, and bound the overflow line with an explicit "+N more".

### 10. low: BOM rows with a query under three characters are dropped with no disclosure

**Where:** `src/kicad/dossier.ts:72`.

`if (query.length < 3) continue;` drops the row entirely: not rendered, not in `overflow`, not in the NOT INCLUDED line.

**Fails when:** a row like `| Y1 | 8M | Crystal:X | | 8 MHz crystal |` has a short value and no MPN. It vanishes, which contradicts the module docstring ("never silently dropped, because absence from the dossier must never read as absence from the libraries") and the delta spec's disclosure requirement. The passive filter at `dossier.ts:69` is disclosed in the prompt header; this filter is not.

**Fix:** push short-query rows onto a disclosed list instead of `continue`, or mention the rule in the prompt header.

### 11. low: the new comment claims a `KICAD_SYMBOL_DIR` exclusivity the code does not provide

**Where:** `src/kicad/symlib.ts:81-84` versus `symlib.ts:69`.

The comment states that a caller pointing `KICAD_SYMBOL_DIR` at an isolated directory "must get only that directory back". The `fromEnv.length > 0` check at `symlib.ts:87` skips only the Windows discovery; line 69 still iterates `[...fromEnv, ...defaults]`, so `/usr/share/kicad/symbols` is appended anyway. The behavior is pre-existing; the claim is new.

**Fails when:** `KICAD_SYMBOL_DIR=/opt/pinned-symbols` on a machine with stock KiCad, where the dossier then resolves parts out of the libraries the caller pinned away from. It also makes the new `symbol_pins` tests (`test/dossier.test.ts:149-208`) non-hermetic: they set the env var but still scan the host's 221 MB of real libraries, and pass only because the fixture dir is searched first. This is the same class the PR fixed for the Windows-discovery tests in the last commit.

**Fix:** either `for (const dir of fromEnv.length ? fromEnv : defaults)` at `symlib.ts:69` (matching the comment and making the tests hermetic), or soften the comment and stop claiming isolation.

### 12. low: the non-versioned Windows install path was removed with no replacement

**Where:** `src/kicad/symlib.ts:60-67` and the discovery loop at `symlib.ts:94`.

`C:/Program Files/KiCad/share/kicad/symbols` was dropped from the defaults, and the loop only probes `${root}/${version}/share/kicad/symbols`. An install laid out directly under `C:\Program Files\KiCad\share\...` now resolves to `.../share/share/kicad/symbols` and is never found, where it previously was.

**Fix:** probe `${kicadRoot}/share/kicad/symbols` in the same loop, after the version directories.

### 13. low: the PR body overstates the stage-3 drawability check

**Where:** PR description item 3 ("`symbol_pins` single-unit check **required** before a BOM row commits") versus `src/commands/create.ts:165-184`.

The stage-3 completion predicate is unchanged: it still only requires one non-`UNVERIFIED` MPN. The drawability check lives entirely in the prompt string at `create.ts:188`. The delta spec is accurate and deliberate here ("The availability and drawability requirements stay at prompt strength; the stage completion gate is unchanged (fuzzy MPN-to-symbol matching in a gate would refuse valid BOMs)"), and given finding 1 that is clearly the right call. Only the PR body reads as enforcement.

**Fix:** reword the PR body to "stage 3 is instructed to confirm drawability", so the merge commit does not claim a gate that does not exist.

### 14. low: untested surface

Diff coverage is 87.2%; the 56 uncovered executable lines reconcile with the uncovered list in the metrics block as follows, and each is a new branch or error path no test reaches:

- `src/agent/tools.ts:359-367`: the entire `search_symbols` handler, including the no-search-dirs message and the no-hits absence message from finding 1. The tool has no test at all.
- `src/agent/tools.ts:402-405,407-409`: `symbol_pins` rendering for `no-symbol` with in-library candidates, and for an uninstalled nickname. The four existing tests all land on `ok` or `found-elsewhere`.
- `src/agent/recovery.ts:174-187`: the entire not-found rendering branch of the fact block, which is where finding 2 lives; plus `157-159` (the `symbolSearchDirs` failure and empty-dirs returns) and `190`, `234`.
- `src/commands/create.ts:858-861,865`: the dossier injection into the stage prompt and its degrade-to-nothing catch. Nothing tests that a dossier failure leaves the stage prompt intact, which is the change's central safety claim.
- `src/kicad/dossier.ts:85-87` (the unreachable cap early exit, finding 6), `99` (the "pins unreadable here" re-resolve failure), `107-109` (the per-part catch, finding 9), `125-126` (the outer catch).
- `src/kicad/draft/symsource.ts:317-320`: the no-library to cross-library-suggestion path.
- `src/kicad/symlib.ts:48-49,127`: the `winRoot` default and a `findLibraryFile` early return.

Add tests for the two tool handlers and for the recovery fact block's absence branch first: those carry findings 1 and 2.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 202, branch feat/symbol-pin-dossier, based on
feat/deterministic-schematic-drafting. Reviewed at head 4c5f088; line numbers are at
that commit.

Check it out first:
  gh pr checkout 202

FINDING 1 [high] searchInstalledSymbols has no edit-distance tier, so the
"machine-verified" dossier and search_symbols declare installed parts absent.
  Where: src/kicad/symlib.ts:309-339 (ranking in rankSymbolNames, symlib.ts:139-165),
    consumed at src/kicad/dossier.ts:90-93 and src/agent/tools.ts:362-364
  Fails when: any BOM row names an MPN whose stock symbol is a family variant.
    findSymbolAcrossLibraries matches on containment AND editDistanceWithin(q, lower, 1)
    (symlib.ts:284-293); searchInstalledSymbols ranks only exact/prefix/substring/
    name-inside-query, so the two resolvers disagree. Against the real
    /usr/share/kicad/symbols:
      searchInstalledSymbols('STM32F103C8T6')    -> []
      findSymbolAcrossLibraries('STM32F103C8T6') -> MCU_ST_STM32F1:STM32F103C8Tx
      searchInstalledSymbols('SHT40')            -> []
      findSymbolAcrossLibraries('SHT40')         -> Sensor_Humidity:SHT4x
    The dossier then prints, under the heading "Installed-symbol pin dossier
    (machine-verified)":
      - U1 (STM32F103C8T6): NO INSTALLED SYMBOL matches ... substitute a part whose
        symbol exists
    and the stage-3 prompt (src/commands/create.ts:188) tells the agent to substitute
    the part on exactly that evidence, so a correctly chosen STM32 or SHT40 gets
    dropped and stage 4 opens with false pin facts. The delta spec
    add-symbol-pin-dossier/specs/create-pipeline/spec.md forbids rendering "could not
    check" as "checked and absent"; the only guard for that (dossier.ts:63) covers the
    no-readable-library case, not the search-miss case.
  Fix: give rankSymbolNames a fourth tier using the same editDistanceWithin(q, c, 1)
    rule with the existing MIN_EDIT_MATCH_LEN guard, so searchInstalledSymbols, the
    dossier and search_symbols all inherit it; or at minimum have dossier.ts:92 and
    tools.ts:363 fall back to findSymbolAcrossLibraries(query, dirs) before emitting
    any absence line, and reword dossier.ts:93 to state a search miss rather than an
    absence. Regression test: a fixture library containing only STM32F103C8Tx,
    asserting searchInstalledSymbols('STM32F103C8T6', [fixtureDir]) is non-empty and
    that the dossier line for that row is not NO INSTALLED SYMBOL.

FINDING 2 [medium] symbolAvailabilityFacts has no "no readable library" guard, so it
reports absence as ground truth when nothing was readable.
  Where: src/agent/recovery.ts:160
  Fails when: a symbol directory exists but holds no libraries (a stale
    KICAD_SYMBOL_DIR, or a packaging split without the symbols package). The guard
    checks searchDirs.length only, while bomSymbolDossier has the stronger
    listInstalledLibraries(dirs).size guard at dossier.ts:63. Reproduced against an
    existing empty directory, the fact block emits:
      - Device:R: no library of that nickname is installed; no installed symbol
        matches "R" in any library
    while bomSymbolDossier on the same dirs correctly returns "". Those lines are
    injected under "Each line reported below is ground truth and overrides anything
    the transcript claims about that symbol's availability" (recovery.ts:230), so the
    diagnostician confirms a bogus refusal and aborts. The whole rendering branch is
    uncovered (recovery.ts:174-187).
  Fix: after recovery.ts:160 add
    if (!(await listInstalledLibraries(searchDirs)).size) return '';
    plus a test asserting an empty-but-existing symbol dir yields ''.

FINDING 3 [medium] verify_symbols reports "0 issue(s) to reconcile" while printing an
issue.
  Where: src/kicad/symlib.ts:496-503 and src/agent/tools.ts:432
  Fails when: a schematic's only divergent lib_symbols entry names an uninstalled
    nickname for a part installed elsewhere. The found-elsewhere branch pushes a
    finding tagged kind 'no-library' and continues without incrementing checked or
    skipped; SymbolFinding['kind'] (symlib.ts:441) has no found-elsewhere member, and
    tools.ts:432 computes mismatches as findings.filter(f => f.kind !== 'no-library'),
    which excludes it. Output:
      verify_symbols: 0 verified, 0 unverifiable (library not installed), 0 issue(s)
      to reconcile:
        - [no-library] "X:Y" names the wrong library, not the wrong part: use A:Y
          instead.
  Fix: add 'found-elsewhere' to SymbolFinding['kind'] and emit the finding with that
    kind so tools.ts:432 counts it as a mismatch. Regression test: assert the summary
    reports 1 issue(s) for a found-elsewhere entry.

FINDING 4 [medium] The new pre-diagnosis probe runs outside the diagnosis watchdog.
  Where: src/commands/create.ts:487
  Fails when: a stage fails on a machine whose symbol libraries sit on a slow or
    stalled network mount. symbolAvailabilityFacts(...) is awaited on the line before
    withTimeout(...), so the provider turn it feeds has a deadline and it does not. It
    probes up to 8 lib_ids, each miss triggering up to two full-library sweeps
    (findSymbolAcrossLibraries inside resolveLibrarySymbol, then
    searchInstalledSymbols) at about 1 s per sweep, while the provider opened at
    create.ts:482 is held. Recovery hangs instead of failing safe to abort, and every
    other provider step in that function is bounded.
  Fix: move the call inside the withTimeout closure, or wrap it in its own withTimeout
    that degrades to ''.

FINDING 5 [medium] A machine-local absolute node_modules symlink is committed on the
branch.
  Where: repository root, mode 120000, added by commit cc3f2ae (formerly 5f08bc4)
  Fails when: anyone clones the branch. The tracked blob is the literal string
    /home/animesh/Github/chouhan-industries/copperhead/node_modules, confirmed present
    on the PR head. .gitignore:5 is "node_modules/" with a trailing slash, which
    matches directories only, so the symlink slipped past. A clone gets a dangling
    absolute symlink at the repo root; npm ci survives because it removes node_modules
    first (which is why CI is green), but npm install and anything following the link
    do not, and a second clone on the author's machine silently resolves dependencies
    from the original checkout. It also publishes a local home path.
  Fix: git rm --cached node_modules, and change .gitignore:5 to "node_modules" with no
    trailing slash.

FINDING 6 [medium] Building the dossier costs about a second of uncached library I/O
per BOM part, on every stage-4 attempt.
  Where: src/commands/create.ts:851-865 calling src/kicad/dossier.ts:90, backed by
    src/kicad/symlib.ts:309-339
  Fails when: a 40 to 50 part BOM enters stage 4, or a retry re-enters it, on a cold
    cache or with several search dirs. Neither listInstalledLibraries nor
    searchInstalledSymbols/findSymbolAcrossLibraries caches: each call re-reads every
    installed .kicad_sym (222 files, 221 MB on the review machine). Measured warm with
    one search dir: 0.91 s for a single searchInstalledSymbols, 15.7 s for a 15-row
    BOM through bomSymbolDossier, serial, before the first agent turn, with no
    progress output. The size-cap early exit at dossier.ts:84 (spent >= maxChars) is
    effectively unreachable, since spent only grows while spent + line.length <=
    maxChars, so the cap bounds rendered output but not work and parts past it still
    pay a full scan.
  Fix: build the per-directory Map<lib, names[]> index once (module-level, keyed on
    the resolved dirs) and share it across the dossier, findSymbolAcrossLibraries and
    the fact block; and make the cap check spent + estimated > maxChars so overflow
    parts short-circuit before the scan.

FINDING 7 [low] (optional) The found-elsewhere message contradicts itself when the
match is in the same library.
  Where: src/kicad/symlib.ts:500 and src/kicad/draft/symsource.ts:198
  Fails when: the library nickname is right and only the name is a variant.
    findSymbolAcrossLibraries skips only the exact check for excludeLib
    (symlib.ts:252), so a fuzzy hit inside the guessed library returns
    found-elsewhere. Verified: resolveLibrarySymbol('MCU_ST_STM32F1:STM32F103C8T6')
    returns found-elsewhere ['MCU_ST_STM32F1:STM32F103C8Tx'], rendered as
    "MCU_ST_STM32F1:STM32F103C8T6" names the wrong library, not the wrong part: use
    MCU_ST_STM32F1:STM32F103C8Tx instead. The library was not wrong.
  Fix: in both message builders, branch on whether every suggested lib_id shares the
    queried nickname and say "does not exist in that library, closest real name: ..."
    in that case.

FINDING 8 [low] (optional) A model-supplied lib_id can escape the symbol search
directories.
  Where: src/kicad/symlib.ts:111, reachable from symbol_pins (src/agent/tools.ts:371-410)
  Fails when: the model passes a lib segment containing "..". Verified with a fixture
    outside the search dir: resolveLibrarySymbol('../outside/Private:SECRET_PART',
    [libsDir]) returns status ok with that file's pins. The blast radius is small (the
    target must be a parseable .kicad_sym, only symbol and pin names are echoed) and
    these tools legitimately read outside the repo, so this is not an AC-4.2 sandbox
    break; it is an unnormalized model-controlled path segment.
  Fix: reject a lib containing "/", "\" or ".." in findLibraryFile before joining.

FINDING 9 [low] (optional) Per-part probe errors are reported to the model as size-cap
overflow.
  Where: src/kicad/dossier.ts:107 versus the label at dossier.ts:119-121
  Fails when: a part's library is unreadable or malformed. The per-part catch pushes
    onto overflow, which renders as "NOT INCLUDED (size cap 24000 chars)", describing
    an unreadable part as one omitted for length; if every part throws, the whole block
    is one line blaming a cap that was never approached. Separately
    overflow.join('; ') is not counted against maxChars, so the block can exceed the
    bound it names. Both paths are uncovered (dossier.ts:107-109).
  Fix: track error-skipped parts in their own list with a "could not be checked here"
    line, preserving the "could not check is not absent" distinction, and bound the
    overflow line with an explicit "+N more".

FINDING 10 [low] (optional) BOM rows with a query under three characters are dropped
with no disclosure.
  Where: src/kicad/dossier.ts:72
  Fails when: a row like | Y1 | 8M | Crystal:X | | 8 MHz crystal | has a short value
    and no MPN. "if (query.length < 3) continue;" drops it entirely: not rendered, not
    in overflow, not in the NOT INCLUDED line, contradicting the module docstring
    ("never silently dropped, because absence from the dossier must never read as
    absence from the libraries") and the delta spec's disclosure requirement. The
    passive filter at dossier.ts:69 is disclosed in the prompt header; this one is not.
  Fix: push short-query rows onto a disclosed list instead of continuing, or state the
    rule in the prompt header.

FINDING 11 [low] (optional) The new comment claims a KICAD_SYMBOL_DIR exclusivity the
code does not provide.
  Where: src/kicad/symlib.ts:81-84 versus symlib.ts:69
  Fails when: KICAD_SYMBOL_DIR=/opt/pinned-symbols on a machine with stock KiCad. The
    comment says such a caller "must get only that directory back", but the
    fromEnv.length > 0 check at symlib.ts:87 skips only the Windows discovery, and
    line 69 still iterates [...fromEnv, ...defaults], so /usr/share/kicad/symbols is
    appended anyway and the dossier resolves parts out of the libraries the caller
    pinned away from. It also makes the new symbol_pins tests
    (test/dossier.test.ts:149-208) non-hermetic: they set the env var but still scan
    the host's real libraries and pass only because the fixture dir is searched first.
  Fix: either "for (const dir of fromEnv.length ? fromEnv : defaults)" at symlib.ts:69
    (matching the comment and making the tests hermetic), or soften the comment and
    stop claiming isolation.

FINDING 12 [low] (optional) The non-versioned Windows install path was removed with no
replacement.
  Where: src/kicad/symlib.ts:60-67 and the discovery loop at symlib.ts:94
  Fails when: KiCad is installed directly under C:\Program Files\KiCad\share\... The
    fixed path was dropped from the defaults and the loop only probes
    ${root}/${version}/share/kicad/symbols, so that layout resolves to
    .../share/share/kicad/symbols and is never found, where it previously was.
  Fix: also probe ${kicadRoot}/share/kicad/symbols in the same loop, after the version
    directories.

FINDING 13 [low] (optional) The PR body overstates the stage-3 drawability check.
  Where: PR description item 3 versus src/commands/create.ts:165-184
  Fails when: a reader takes "symbol_pins single-unit check required before a BOM row
    commits" as enforcement. The stage-3 completion predicate is unchanged (it still
    only requires one non-UNVERIFIED MPN); the check lives entirely in the prompt
    string at create.ts:188. The delta spec is accurate and deliberate here ("the
    stage completion gate is unchanged"), and given finding 1 that is the right call.
  Fix: reword the PR body to "stage 3 is instructed to confirm drawability", so the
    merge commit does not claim a gate that does not exist.

For every confirmed correctness bug, add the regression test that fails before the fix
and passes after it; this repo pairs bug fixes with tests.

These changed lines are reached by no test: src/agent/tools.ts:359-367 (the entire
search_symbols handler, including the absence message from finding 1),
src/agent/tools.ts:402-405,407-409 (symbol_pins rendering for no-symbol with
candidates, and for an uninstalled nickname), src/agent/recovery.ts:157-159,174-187,190,234
(the whole not-found rendering branch of the fact block, where finding 2 lives),
src/commands/create.ts:858-861,865 (dossier injection and its degrade-to-nothing
catch), src/kicad/dossier.ts:85-87,99,107-109,125-126 (the unreachable cap early exit,
the "pins unreadable here" path, the per-part catch, the outer catch),
src/kicad/draft/symsource.ts:317-320, src/kicad/symlib.ts:48-49,127. Add tests for the
new exported symbols, branches and error paths they contain, starting with the two
tool handlers and the recovery fact block.

`openspec validate add-symbol-pin-dossier` and `openspec validate add-symbol-search-tool`
have to pass. If the fix for finding 1 changes what the dossier promises, update the
create-pipeline delta spec's false-absence scenario with it.

    by prompt text.
    dependence on execution order, and every test asserts something. Tests that touch
    symbol libraries must resolve from fixtures, not from the host's installed KiCad
    (see finding 11).

Then verify:
  npm run typecheck
  npm run build
  npm test

and report it.

````

Issue #163, opened 2026-08-01
Stage 4 deadlock: the BOM value cross-check freezes the only lever that can clear the legibility gate


animesh-chouhan:
## Summary

Stage 4 (`schematic`) can be **unsatisfiable**. Two gates that are each reasonable in isolation contradict each other, and the agent has no legal move that satisfies both:

- [`src/kicad/draft/ir.ts:210`](https://github.com/chouhanindustries/copperhead/blob/3bc7d6d/src/kicad/draft/ir.ts#L210) pins the IR's part value to `docs/BOM.md` by **exact string equality**.
- [`src/commands/create.ts:311`](https://github.com/chouhanindustries/copperhead/blob/3bc7d6d/src/commands/create.ts#L311) fails the stage while any error-severity legibility finding remains — and `edit_file` is refused on a drafted sheet, so the IR is the agent's only lever.

When stage 3 writes a *description* into BOM.md's Value column — which it does routinely — the engine renders that string as the symbol's Value field, it collides with neighbouring symbols, and the only fix (shortening it) is rejected by the BOM cross-check.

The run stalls at 4/8 and burns all three stage attempts. This is a design bug in the drafting contract, not a flaky run: any brief whose BOM carries descriptive values hits it.

## Reproduction

```
copperhead create --brief brief.md --model claude-code
```

Brief: an RF game joystick (61 parts, 53 nets, 8 subsystems). copperhead 0.9.0, branch `feat/deterministic-schematic-drafting` @ `0d23858` (+5 uncommitted files — the binary under test exists in no commit), node v24.18.0, kicad-cli 10.0.4.

Result: stages 1–3 committed, stage 4 exhausted 3/3 attempts, **exit 1**, 3h11m, 144 turns, 956.3k output tokens.

## The deadlock, concretely

`BT1`'s value in the stage-3 BOM is:

```
| BT1 | 1S Li-Po cell, 500 mAh, bare leads | Connector_JST:JST_PH_S2B-PH-K_1x02_P2.00mm_Horizontal | ...
```

34 characters rendered as the Value field next to a battery symbol overruns into Q4 → error-severity legibility finding → `create.ts:311` refuses the stage.

Putting `500mAh Li-Po` in the IR instead fails first, at `ir.ts:210`:

```js
else if (bomValue !== p.value) add(`${p.ref} value "${p.value}" differs from BOM.md's "${bomValue}"`)
```

[...]

PR #215, opened 2026-08-08
fix(kicad): verify vendored symbols, refuse IR power parts, survive a stale symbol-dir override

### LINKED ISSUE #212: Post-merge review of #160/#202: verifier misreports engine-owned power symbols, IR power parts are silently dropped, and a stale symbol-dir env var disables search
## What happened

Post-merge review of the `deterministic-schematic-drafting` surface (#160 and #202, both landed as merge commit `8b92152`) turned up three defects that survive on `main` at `c208680`. All three are reproduced below against committed fixtures, no live LLM involved.

Two of them share a root cause worth stating up front: **the symbol verifier and the drafting engine disagree about who owns power symbols.** The engine replaces power nets with its own generated `copperhead_power:*` symbols vendored into the project's `sym-lib-cache/`, but `verifySchematicSymbols` only ever searches installed system directories, and the engine silently discards any `power:*` part the model puts in the IR. The agent is therefore told to fix symbols that are correct, and gets no signal at all about symbols it asked for and did not get.

### 1. `verify_symbols` reports the engine's own power symbols as `wrong-library`

`verifySchematicSymbols` (src/kicad/symlib.ts) resolves lib_ids against `symbolSearchDirs()` only. It has no notion of the project's vendored `sym-lib-cache/`, so every `copperhead_power:*` symbol the engine generated fails to resolve, then fuzzy-matches its way to a stock lib_id.

On the committed `ldo-demo` reference board this yields four bogus findings out of six parts. Three are reported as `wrong-library` with a confident "use X instead" instruction, and the fourth as `no-library`:

```
checked 5 skipped 1
wrong-library | copperhead_power:3V3       -> "use Converter_DCDC:JTD1524S3V3 or Diode:BZV55B3V3 or Power_Protection:D3V3X8U9LP3810 or power:+3V3 instead"
wrong-library | copperhead_power:GND       -> "use power:GND or Device:Crystal_GND2 instead"
wrong-library | copperhead_power:PWR_FLAG  -> "use power:PWR_FLAG instead"
no-library    | copperhead_power:VIN       -> "library ... is not installed on this machine; cannot verify its pins"
```

Every one of these is wrong: the symbols exist and are correct, in `manual-tests/reference-boards/ldo-demo/sym-lib-cache/copperhead_power.kicad_sym`. `3V3` matching to a DC-DC converter and a Zener diode shows how far the fuzzy fallback reaches once resolution fails.

This is worse than noise, because the suggested fix is unreachable. The agent cannot act on it: the IR exposes no lever for power symbol lib_ids, and `edit_file` is refused on engine-drafted sheets. So stage 4 hands the model four "issues to reconcile" that have no legal resolution. That is the same shape as the deadlock in #163.

`symbolAvailabilityFacts` already knows to skip the `copperhead_power` namespace; this path does not. Fix is either to teach `symbolSearchDirs`/`verifySchematicSymbols` about the project cache, or to skip the engine-owned namespace the way the availability path does. The first is better: it also makes verification work for any vendored library, not just power.

### 2. Power parts in the IR are accepted, then silently dropped

`validateIntent` accepts an IR part whose symbol resolves with `isPower`, but the engine filters those parts out of every placement path (`src/kicad/draft/engine.ts`, the `!symbols.get(p.ref)!.isPower` guards around group naming, label extents, and placement). The part is never placed, never added to `lib_symbols`, never given a net endpoint, and never mentioned in the report.

Repro: take the committed `ldo-demo` intent, add one power part, and draft it.

```json
{ "ref": "PG1", "libId": "power:GND", "value": "GND", "footprint": "", "group": "Indicator" }
```
with `"PG1.1"` appended to the `GND` net's pins.

```
ok=true | PG1 present in sheet? false | power:GND placed? false
notes: []
groups: [{"name":"Indicator","members":["D1","R1"]},{"name":"Power","members":["C1","C2","J1","U1"]}]
vendoredLibIds includes power:GND? true
```

The draft reports success. `PG1` appears nowhere in the emitted sheet, it is absent from its own group's member list, and `notes` is empty. The engine even vendors `power:GND` into `sym-lib-cache/` on the way past, so the project accumulates a cached symbol for a part that was never drawn.

Dropping the part is arguably the right behavior, since the engine synthesizes its own per-pin power symbols and a model-authored one would be redundant. The bug is that it happens silently. A model that writes `power:GND` into the IR, sees `ok: true`, and finds no note has no way to learn that its mental model is wrong, so it will keep doing it. Either reject it in `validateIntent` with a finding that names the correct approach, or place it, or at minimum emit a note.

### 3. A stale `KICAD*_SYMBOL_DIR` disables symbol search entirely

`symbolSearchDirs` (src/kicad/symlib.ts) treats an env override as exclusive, which is intentional and documented in the comment above it: a caller pinning an isolated directory should not get the machine's stock install appended. But the exclusivity test is `fromEnv.length` measured **after** the entries are filtered for existence, not on whether the variable was set. So a variable pointing at a directory that no longer exists collapses the search path to nothing instead of falling back:

```
no env    -> 1 dirs  ['/usr/share/kicad/symbols']
stale env -> 0 dirs  []
```

With zero search dirs, every symbol tool reports "not installed on this machine" for every part, which is a false environmental claim rather than an empty result. That is a plausible mechanism behind the fabricated "reduced symbol library" constraint in #197, and it would make `search_symbols` (#200) and near-miss candidates (#195) fail the same way.

Note this one predates both PRs: it traces to `7140262`, so it is not a regression from the drafting work. It surfaced during this review.

The fix should preserve the documented exclusivity, so it needs to distinguish "set" from "set and resolvable". Failing loudly when an override is set but resolves to nothing seems better than either silently falling back or silently searching nothing.

## Steps to reproduce

All three run offline against committed fixtures. Findings 1 and 3 need no scratch setup:

```ts
import { symbolSearchDirs, verifySchematicSymbols } from './src/kicad/symlib.js';

// Finding 3
await symbolSearchDirs({});                                          // -> ['/usr/share/kicad/symbols']
await symbolSearchDirs({ KICAD_SYMBOL_DIR: '/nonexistent/stale' });   // -> []

// Finding 1
await verifySchematicSymbols('manual-tests/reference-boards/ldo-demo/reference/ldo-demo.kicad_sch');
```

Finding 2: copy `manual-tests/reference-boards/ldo-demo/schematic.intent.json` into a scratch repo, add the `PG1` part and net pin shown above, then call `draftSchematicToText({ repoRoot, schematic: 'reference/ldo.kicad_sch', today: '2026-08-08' })`.

## Which command

create (stage 4 drafting and symbol verification), plus `check` for finding 3.

## Model backend

Not applicable. All three reproduce offline with no provider configured.

## Version and environment

copperhead 0.9.0 at `c208680`; node v24.18.0, kicad-cli 10.0.4, Pop!_OS 22.04 LTS.

## Related

- #204 (router net merge), found in the same review pass, already fixed by #205.
- #163: finding 1 produces the same unresolvable-gate shape.
- #197, #200, #195: finding 3 is a candidate mechanism for the false "library not installed" claims.
- #61: closed, but the same power-symbol-namespace assumption.

thegoodengineer:
## What

Closes #212.

Fixes the three defects in it: `verify_symbols` no longer misreports the engine's own power symbols, an IR power-port part is refused instead of silently dropped, and a stale `KICAD*_SYMBOL_DIR` no longer disables symbol search entirely.

## Why

All three were found by the post-merge review of #160/#202 and reproduced here against committed fixtures, no live LLM.

**1. Engine-owned power symbols reported as `wrong-library`.** [verifySchematicSymbols](src/kicad/symlib.ts) resolved lib_ids against installed directories only, so the engine's generated `copperhead_power:*` symbols never resolved and the cross-library fallback fuzzy-matched them to whatever stock name looked close:

```
checked=5 skipped=1 findings=4
  [wrong-library] copperhead_power:3V3       -> "use Converter_DCDC:JTD1524S3V3 or Diode:BZV55B3V3 ... instead"
  [wrong-library] copperhead_power:GND       -> "use power:GND or Device:Crystal_GND2 instead"
  [wrong-library] copperhead_power:PWR_FLAG  -> "use power:PWR_FLAG instead"
  [no-library]    copperhead_power:VIN       -> "not installed on this machine; cannot verify its pins"
```

Four confident findings out of six parts, none of them reachable: the IR exposes no lever for power-symbol lib_ids and `edit_file` is refused on a drafted sheet, so stage 4 was handed issues with no legal resolution. That is the deadlock shape of #163.

**2. IR power parts accepted, then dropped.** [validateIntent](src/kicad/draft/ir.ts) accepted a power-port part and the engine filtered it out of every placement path, so it was never placed, never given an endpoint, absent from its own group's member list, and missing from `notes`, while the draft still returned `ok: true`. A model that writes `power:GND` had no signal at all and would keep doing it.

**3. Stale env override disabled search.** [symbolSearchDirs](src/kicad/symlib.ts) tested exclusivity on whether an override was **set**, while its output collects only directories that **exist**. A variable left by an uninstalled or relocated KiCad made those diverge and returned an empty search path. That is not "search nothing", it is "every symbol is absent": `search_symbols`, `symbol_pins`, the dossier, and the recovery fact-check all degrade at once, and stage 3 substitutes parts that never needed substituting.

## Design

**Ordering is the load-bearing decision in defect 1.** A project's `sym-lib-cache/` is discovered by walking up from the schematic (`vendoredCacheDirs`) and appended **after** the installed directories, never before. This function exists to catch a `lib_symbols` entry whose pins have drifted from the real part, so resolving against the cache the entry was copied from would compare a file with itself and report clean forever. Appending means a stock lib_id still verifies against the stock library while a lib_id no installed directory provides falls through to the vendored copy. A test pins this directly: a tampered `Device:R` pin is still caught as `pin-mismatch` in the same pass where the power symbols verify clean.

I took the "teach it about the cache" option over "skip the `copperhead_power` namespace", per the issue's reasoning: it also makes verification work for any library a project vendors, not only the namespace the engine owns today.

`SYM_CACHE_DIR` moved to [symlib.ts](src/kicad/symlib.ts) as canonical and is re-exported from [symsource.ts](src/kicad/draft/symsource.ts), because symlib cannot import the drafting layer without a cycle. No call sites changed.

For defect 2 the part is recorded in `symbols` **before** the refusal is added, so the group and BOM cross-checks below take their existing `isPower` skips and the model gets one finding about one mistake instead of three unrelated-looking ones.

The only invariant this engages is spec coherence, handled in the section below. It changes advisory reporting and IR validation only: no new tool, no change to the gate ordering, no schematic serialization.

## Spec / OpenSpec

Defect 2 is a spec-level change, corrected after review: an earlier version of this description claimed it was not.

The drafting-engine delta spec enumerates exactly which inputs IR validation refuses and named the multi-unit refusal explicitly. The power-port refusal added here belongs in that enumeration, and the spec's "non-power part" phrasing (in the IR shape, the validation list, and the ungrouped-part scenario) presumed power parts may appear in the IR and are merely exempt from grouping, which stopped being true.

Moved together:

- `openspec/changes/deterministic-schematic-drafting/specs/schematic-drafting-engine/spec.md`: the power-port refusal added to the validation enumeration with its rationale, a new `Scenario: Power-port part is rejected`, and the two "non-power part" clauses reworded.
- `openspec/changes/deterministic-schematic-drafting/tasks.md`: 6.2 reworded to match, 6.2a added for the refusal.
- `openspec/specs/SPEC.md`: unchanged, and deliberately so. It does not describe `schematic.intent.json` or IR validation at all (zero references), so the delta spec is the only artifact that governs this behavior.

`openspec validate deterministic-schematic-drafting` passes: `Change 'deterministic-schematic-drafting' is valid`.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | pass, 0 new failures (see note) |
| Live-LLM (opt-in) | n/a, no provider path touched | n/a |

New tests: 14. Five on the stale-override fallback (all-stale, partially-stale where the surviving override still wins, and an override naming a regular file rather than a directory), four on vendored-cache verification (discovery by walking up, clean verification, the non-vacuity guard, and an unvendored/uninstalled lib_id still reported), and one on the power-part refusal beside its sibling multi-unit case. Four more were added after review: the project-root anchor (a foreign cache above a project is not adopted), the maxDepth bound, a `.git` file as a worktree carries it, and an assertion on the reworded no-symbol diagnostic so it cannot silently regress.

**Known pre-existing failures unrelated to this PR.** This branch was developed on Windows, where 20 tests fail on `main` before any change here: CRLF normalization (the repo has no `.gitattributes`, so committed `.kicad_sch` files check out CRLF while the emitter writes LF) plus a few platform assumptions in the tests themselves. I compared failure sets **by name** before and after rather than counting, so the "0 new failures" claim is independent of that baseline. Those 20 are not addressed here. CI is Linux-only and is green on this PR.

Sandbox variant used: the committed reference boards under `manual-tests/reference-boards/`, driven through the same entry points the tools use.

Commands run and outcome:

```text
$ node -e "verifySchematicSymbols('manual-tests/reference-boards/<board>/reference/<board>.kicad_sch')"
before:  ldo-demo  checked=5 skipped=1 findings=4   (the four bogus findings quoted above)
after:   ldo-demo         checked=9  skipped=0 findings=0
         buck-12v-5v      checked=12 skipped=0 findings=0
         npn-switch       checked=10 skipped=0 findings=0
         usb-atmega-node  checked=17 skipped=0 findings=0

$ node -e "draftSchematic(ldo-demo intent + {ref:PG1, libId:'power:GND'} on the GND net)"
before:  ok = true | PG1 in sheet? false | notes: [] | PG1 absent from its group's members
after:   ok = false | findings = 1
         PG1: "power:GND" is a power-port symbol, which the drafting engine supplies itself. ...

$ node -e "symbolSearchDirs({ KICAD_SYMBOL_DIR: 'C:/definitely/not/here' })"
before:  []
after:   ["C:/Program Files/KiCad/10.0/share/kicad/symbols"]
$ node -e "symbolSearchDirs({ KICAD_SYMBOL_DIR: <a real isolated dir> })"
after:   [<that dir>]     (exclusivity preserved)
```

Non-vacuity checked by hand as well as in a test: tampering a `Device:LED` pin name in a copy of the reference sheet still reports `pin-mismatch` after the change.

Retention in defect 2 checked by mutation: replacing `symbols.set(p.ref, sym)` with a non-power-only version makes the new test fail with `expected [ …(2) ] to have a length of 1 but got 2`, confirming the assertion guards the property rather than passing incidentally.

## Docs

None needed. The changes are internal to the symbol layer and the IR validator; no CLI surface, configuration key, or environment variable changed. Behavior is documented in code comments naming #212 at each site.

---

## Invariant checklist

- [ ] **Spec-gated in**: N/A. No change to the tool list or the unlock gate.
- [ ] **Verification-gated out**: N/A. No change to ERC/DRC sequencing, repair cycles, or rollback.
- [x] **`check`/`verify` stays LLM-free and network-free**: preserved. The touched modules are pure filesystem reads; no provider, key, or network access added. `vendoredCacheDirs` does bounded `stat` calls only.
- [x] **No sexp serialization**: preserved. The parser stays read-only; this PR reads `lib_symbols` and never writes a KiCad file.
- [ ] **Sync-obligations ledger**: N/A. Not touched.
- [x] **Secrets**: preserved. No new logging, transcript output, or environment reads beyond the existing `KICAD*_SYMBOL_DIR` names.
- [x] **Spec coherence**: the power-port refusal is recorded in the drafting-engine delta spec (validation enumeration, a new scenario, and the reworded "non-power part" clauses) and in `tasks.md` 6.2/6.2a. `openspec validate deterministic-schematic-drafting` passes. `SPEC.md` does not describe the IR, so it correctly does not move.

## Review feedback addressed

CodeRabbit raised three items, all valid and all fixed in the second commit:

- **Major**: `access()` also succeeds on a regular file, so an override naming a file counted as resolved and suppressed the fallback while satisfying no lookup. Now `stat().isDirectory()`, with a regression test.
- **Minor**: the `no-symbol` diagnostic still said "the installed library", which names the wrong file to fix once a vendored library can be the match. Now "installed or vendored symbol library".
- **Nitpick**: the power-part test did not pass `docsDir`, so the BOM and group paths never ran and the "exactly one finding" assertion would have held even without the retention. Real `SUBSYSTEMS.md` and `BOM.md` fixtures added and `docsDir` passed; the mutation check above confirms the assertion now bites.

An automated pr-review pass then raised three more, all valid and all fixed in the third commit:

- **Medium, spec coherence**: the power-port refusal was absent from the delta spec's validation enumeration (which names the sibling multi-unit refusal explicitly), and this description claimed no spec-level change. Spec, scenario, and tasks updated; description corrected above; `openspec validate` passes.
- **Low**: `vendoredCacheDirs` ascended six levels with no project anchor, so a board nested under a directory holding a `sym-lib-cache` adopted that foreign copy, which could let verification borrow another project's evidence and report clean. The walk now stops at the project root (`.git` or `.copperhead`), inclusive; `.git` is matched by existence because a linked worktree stores it as a file. Confirmed by mutation: removing the anchor makes the new test fail with `expected [ Array(1) ] to deeply equal []`. The same cause had made an existing test depend on what else lived in the temp tree, fixed by giving its fixture a `.copperhead` marker.
- **Low**: the reworded `no-symbol` diagnostic was asserted by no test, so the inaccurate wording could have come back green. Now asserted, alongside cases for the `maxDepth` bound and the `.git`-as-a-file case.

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering project-vendored symbol caches, including nested project locations.
  * Improved symbol-library selection across installed libraries and project caches, with installed libraries taking precedence.
* **Bug Fixes**
  * Drafting now clearly rejects explicitly supplied power-port symbols and directs users to represent power through nets.
  * Improved handling of stale symbol search locations, with reliable fallback discovery.
  * Enhanced missing-symbol diagnostics to identify relevant library sources.


thegoodengineer:
_Automated pr-review pass. Not a human review._

**Reviewer authored this PR.** This is a self-review, so treat the clean invariant results with more skepticism than an independent pass would warrant. It ran in an isolated `git worktree` at commit `bcf1938`; the main checkout was never switched.

## Verdict: approve with nits

The three defects in #212 are genuinely fixed, and each fix is backed by a before/after I re-derived independently in the worktree rather than taking from the description. The ordering decision in defect 1 (installed dirs first, vendored cache appended) is the right call and is guarded by a real non-vacuity test. One medium finding on spec coherence, two low.

## Metrics

```text
lines: +394 / -36 (net 358) across 5 files  ·  src +140/-36, test +254/-0, docs +0, spec +0
new vs net: 0 new files, 5 modified; 0 net-new src files (all edits to existing modules)
tests: +254 test lines; suite (base main) 750P/20F/20S of 790 -> (head) 760P/20F/20S of 800
diff coverage: not measured (@vitest/coverage-v8 not installed and the offline install failed);
               untested surface mapped by hand instead, enumerated in finding 3
ci: 5 pass / 0 fail / 0 pending; required checks: none reported
```

Method: measured in the review worktree, merge base `c208680` (`git merge-base origin/main HEAD`). Change size from `git diff --numstat`. Suite from `npx vitest run --reporter=json` run inside the worktree, compared to a `main` run **by test name**, not by count: the 20 failures are identical at base and head, so **0 new failures**. They are pre-existing Windows/CRLF failures unrelated to this PR. Diff coverage is genuinely not measured, so finding 3 is a hand-mapped substitute, not a percentage.

## Invariant checks performed

| Invariant | Result | Evidence |
| --- | --- | --- |
| Spec-gated in | pass | diff touches no file under `src/agent/`; tool list and `requiresUnlock` unchanged |
| Verification-gated out | pass | no change to ERC/DRC sequencing, repair cycles, or rollback |
| `check` LLM-free / network-free | pass | `draft-tools.test.ts -t "LLM-free"` module-graph test passes at head |
| No sexp serialization | pass | diff adds no `writeFile` or serializer to `src/kicad/`; reads only |
| Sync-obligations ledger | pass | `src/agent/ledger.ts` untouched |
| Secrets | pass | no `console.*` and no new `process.env` reads added |
| Spec coherence | **fail** | see finding 1 |

## Findings

### 1. [medium] The new power-port refusal is absent from the delta spec, and the PR description states the opposite

**Where:** `src/kicad/draft/ir.ts:219`, against `openspec/changes/deterministic-schematic-drafting/specs/schematic-drafting-engine/spec.md:21`

**Fails when:** the delta spec enumerates exactly which inputs validation refuses, and it names the multi-unit refusal explicitly: "Validation SHALL also refuse a part resolving to a multi-unit library symbol". That is the sibling this change was modeled on and placed directly beside. The new power-port refusal adds a class to that enumeration and is not recorded in it. Worse, lines 7, 21, and 35 all use "non-power part" phrasing ("group assignments naming a subsystem per non-power part", "every non-power part must belong to exactly one group", "WHEN a non-power part has no group assignment"), which only makes sense if power parts may legitimately appear in the IR and are merely exempt from grouping. After this change they cannot appear at all, so those carve-outs and the `isPower` skips they correspond to (`ir.ts:238`, `ir.ts:368`) are reachable only on a run that has already failed validation. A reader implementing against the spec would build the wrong validator.

Separately, the PR description's "Spec / OpenSpec" section reads "n/a. No spec-level behavior changes", which the above contradicts. #212 offered three options (reject, place, or note); choosing reject is defensible, but it is a spec-level choice and needs recording.

**Fix:** add the refusal to the validation enumeration at `spec.md:21` in the same voice as the multi-unit clause, add a WHEN/THEN scenario for it, reword the "non-power part" clauses that are now unreachable, and run `openspec validate deterministic-schematic-drafting`. Correct the PR description's Spec / OpenSpec section to describe the change instead of claiming none.

### 2. [low] `vendoredCacheDirs` walks out of the project and can adopt a foreign cache

**Where:** `src/kicad/symlib.ts:172-187`, consumed at `src/kicad/symlib.ts:591`

**Fails when:** the walk ascends up to six levels with no project anchor, so any `sym-lib-cache/` in an ancestor directory is adopted. Reproduced in the worktree: a project with no cache of its own, nested one level under a directory that has one, returns the outer cache.

```
project boardA has its own cache? no
vendoredCacheDirs returned : ["...\Temp\foreign-Ujj0LG\sym-lib-cache"]
escaped the project?         true
```

The consequence is a weakened verification rather than a wrong board: `verifySchematicSymbols` could resolve a lib_id against a different project's vendored copy and report clean where it should have reported the library missing. It needs an unusual layout (nested projects, or a workspace directory where a draft once ran), which is why this is low and not medium.

The same root cause makes the new test environment-dependent: `test/symlib.test.ts` asserts `toEqual([path.join(repo, 'sym-lib-cache')])` on a path under `$TMPDIR`, and the walk reaches six ancestors of it, so a `sym-lib-cache` directory anywhere in `C:\Users`, `C:\Users\<name>`, `%LOCALAPPDATA%`, or `%TEMP%` would fail that assertion on an otherwise healthy machine.

**Fix:** stop ascending at the project boundary: after checking a level, break if that directory contains `.git` or `.copperhead`. Verified this does not regress the reference boards, whose caches sit below the repo's own `.git`. Add a test for the escape case (a foreign cache one level above a project with none) asserting an empty result.

### 3. [low] Untested surface in the changed code

**Where:** `src/kicad/symlib.ts:618`, `src/kicad/symlib.ts:619`, `src/kicad/symlib.ts:175`, `src/kicad/symlib.ts:183`

**Fails when:** diff coverage could not be measured, so this is a hand-walk of the changed lines rather than a tool result. Three changed paths are reached by no test:

- `symlib.ts:618-619`, the reworded `no-symbol` diagnostics ("installed or vendored symbol library"). `grep` over `test/` finds no assertion on either string, so a future edit could revert the wording or reintroduce the inaccurate "the installed library" with the suite still green. This wording was itself a review fix, which makes an unasserted change the weaker outcome.
- `symlib.ts:175`, the `maxDepth` bound: no test places a cache seven or more levels up to confirm the walk stops.
- `symlib.ts:183`, the filesystem-root break: no test drives a path to the root.

**Fix:** assert the new diagnostic substring in the existing `no-symbol` test in `test/symlib.test.ts`. Add one `vendoredCacheDirs` case with a cache beyond `maxDepth` asserting an empty result, which covers the loop bound; the root break is exercised incidentally by any path that runs out of ancestors.

## Notes, not findings

- The non-vacuity guard on defect 1 is the right instinct and I confirmed it independently: tampering a `Device:R` pin in the worktree still reports `pin-mismatch` while the power symbols verify clean, so appending the cache did not turn the cross-check into a self-comparison.
- The retention of `symbols.set(p.ref, sym)` before the power refusal is confirmed load-bearing by mutation: removing it produces two findings instead of one and the test fails. Good.
- Scope is tight. The `SYM_CACHE_DIR` relocation is necessary to avoid an import cycle and changes no call site. No unrelated refactors.
- New tests are deterministic: no network, no `Date.now()`, no `Math.random()`, no order dependence, and all assert. The single environmental coupling is the one described in finding 2.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 215, branch fix/212-vendored-symbol-verification,
based on main.

Check it out first:
  gh pr checkout 215

FINDING 1 [medium] The new power-port refusal in IR validation is absent from the
delta spec, and the PR description claims no spec-level change.
  Where: src/kicad/draft/ir.ts:219, against
         openspec/changes/deterministic-schematic-drafting/specs/schematic-drafting-engine/spec.md:21
  Fails when: spec.md line 21 enumerates exactly which inputs validation refuses and
    names the multi-unit refusal explicitly ("Validation SHALL also refuse a part
    resolving to a multi-unit library symbol"). The new power-port refusal adds a
    class to that enumeration and is not recorded. Lines 7, 21 and 35 also use
    "non-power part" phrasing that only makes sense if power parts may appear in the
    IR and are merely exempt from grouping; after this change they cannot appear at
    all, so those clauses and the isPower skips at ir.ts:238 and ir.ts:368 are
    reachable only after validation has already failed. The PR description's
    "Spec / OpenSpec" section says "n/a. No spec-level behavior changes", which
    contradicts this.
  Fix: add the power-port refusal to the validation enumeration at spec.md line 21 in
    the same voice as the multi-unit clause, add a WHEN/THEN scenario for it, reword
    the "non-power part" clauses that are now unreachable, and run
    `openspec validate deterministic-schematic-drafting`. Update the PR description's
    Spec / OpenSpec section to describe the change instead of claiming none.

FINDING 2 [low] (optional) vendoredCacheDirs walks out of the project and can adopt a
foreign sym-lib-cache.
  Where: src/kicad/symlib.ts:172-187, consumed at src/kicad/symlib.ts:591
  Fails when: the walk ascends up to six levels with no project anchor, so a
    sym-lib-cache in any ancestor directory is adopted. A project with no cache of its
    own, nested one level under a directory that has one, returns the outer cache.
    verifySchematicSymbols can then resolve a lib_id against a different project's
    vendored copy and report clean where it should have reported the library missing.
    The same cause makes the new test environment-dependent: test/symlib.test.ts
    asserts toEqual([path.join(repo, 'sym-lib-cache')]) on a path under the OS temp
    dir, and the walk reaches six of its ancestors, so a sym-lib-cache directory
    anywhere in C:\Users, the user home, LOCALAPPDATA, or TEMP fails that assertion.
  Fix: stop ascending at the project boundary: after checking a level, break if that
    directory contains .git or .copperhead. This does not regress the reference
    boards, whose caches sit below the repo's own .git. Add a test for the escape case
    (a foreign cache one level above a project with none) asserting an empty result.

FINDING 3 [low] (optional) Untested surface in the changed code.
  Where: src/kicad/symlib.ts:618, src/kicad/symlib.ts:619, src/kicad/symlib.ts:175,
         src/kicad/symlib.ts:183
  Fails when: no test reaches these changed paths. Lines 618 and 619 are the reworded
    no-symbol diagnostics ("installed or vendored symbol library"); no test asserts
    either string, so a future edit could reintroduce the inaccurate "the installed
    library" with the suite still green. Line 175 is the maxDepth loop bound, with no
    test placing a cache beyond it. Line 183 is the filesystem-root break, with no
    test driving a path to the root.
  Fix: assert the new diagnostic substring in the existing no-symbol test in
    test/symlib.test.ts. Add one vendoredCacheDirs case with a cache beyond maxDepth
    asserting an empty result, which covers the loop bound; the root break is
    exercised incidentally by any path that runs out of ancestors.

`openspec validate deterministic-schematic-drafting` has to pass.

    by prompt text.

Then verify:
  npm run typecheck
  npm run build
  npm test

All three must pass. Note that on Windows checkouts 20 tests fail on main before any
change, from CRLF normalization and platform assumptions in the tests; compare failure
sets by test name rather than by count so you do not mistake those for regressions.

````

**Comment by thegoodengineer** [comment]
_Automated pr-review pass (second). Not a human review._

**Reviewer authored this PR.** Self-review again, so the clean results below deserve more scepticism than an independent pass. Ran in an isolated `git worktree` at commit `38cf212`; the main checkout was never switched.

Incremental pass. The [first pass](https://github.com/chouhanindustries/copperhead/pull/215#issuecomment-5227330709) covered `bcf1938`; this one covers the one commit added since, `38cf212`, which is the commit that fixed that pass's findings and which nothing had reviewed (CodeRabbit reported "Review rate limited" on it).

## Verdict: approve with nits

All three findings from the first pass are resolved, and I re-verified each rather than trusting the commit message. Two new low findings, both in the spec text that `38cf212` itself edited: the rewording overshot and made one scenario contradict the implementation.

## Status of the previous pass's findings

| # | Finding | State | How verified |
| --- | --- | --- | --- |
| 1 | Power-port refusal absent from the delta spec | resolved | refusal added to the enumeration, `Scenario: Power-port part is rejected` added, `tasks.md` 6.2a added, `openspec validate` returns valid |
| 2 | `vendoredCacheDirs` escapes the project | resolved | anchor at `.git`/`.copperhead`; mutation test confirms removing the `break` fails the new test with `expected [ Array(1) ] to deeply equal []` |
| 3 | Untested surface | mostly resolved | diagnostic string and `maxDepth` now asserted; the filesystem-root break is still untested, see notes |

Finding 2's second symptom is also gone: every new `vendoredCacheDirs` test now anchors at depth 0 or 1, so none of them depend on what else lives in the temp tree.

## Metrics

```text
lines: +524 / -40 (net 484) across 7 files  ·  src +182/-36, test +332/-0, docs +0, spec +10/-4
new vs net: 0 new files, 7 modified; 0 net-new src files (all edits to existing modules)
tests: +332 test lines; suite (base main) 750P/20F/20S of 790 -> (head) 766P/18F/20S of 804
diff coverage: not measured (@vitest/coverage-v8 unavailable in the worktree);
               untested surface hand-mapped instead, reconciled in the notes below
ci: 5 pass / 0 fail / 0 pending; required checks: none reported
```

Method: measured in the review worktree at `38cf212`, merge base `c208680`. Change size from `git diff --numstat`. Suite from `npx vitest run --reporter=json` inside the worktree with a real `npm ci`, compared to a `main` run **by test name**: **0 new failures**. The 18 remaining are a subset of `main`'s 20 pre-existing Windows/CRLF failures (two of that set are flaky and passed this run; I am not claiming this PR fixed them). Diff coverage genuinely not measured, so the untested-surface note is a hand-walk, not a percentage.

## Invariant checks performed on `38cf212`

| Invariant | Result | Evidence |
| --- | --- | --- |
| Spec-gated in | pass | zero files under `src/agent/` touched by this commit |
| Verification-gated out | pass | no change to ERC/DRC sequencing, repair cycles, or rollback |
| `check` LLM-free / network-free | pass | no `fetch`, provider, or HTTP reference added; commit touches `src/kicad/symlib.ts` only |
| No sexp serialization | pass | no `writeFile` or serializer added; reads only |
| Sync-obligations ledger | pass | untouched |
| Secrets | pass | no `console.*` and no new `process.env` reads |
| Spec coherence | pass with nits | artifacts moved together and `openspec validate` passes; wording issues are findings 1 and 2 |

## Findings

### 1. [low] The reworded ungrouped-part scenario now contradicts the implementation

**Where:** `openspec/changes/deterministic-schematic-drafting/specs/schematic-drafting-engine/spec.md:35`, against `src/kicad/draft/ir.ts:238`

**Fails when:** the scenario reads "**WHEN** a part has no group assignment / **THEN** validation fails naming the refdes and the available groups from SUBSYSTEMS.md". Drive exactly that input with a power-port part and the THEN does not happen. Reproduced in the worktree with an IR containing `{"ref": "PG1", "libId": "power:GND"}` and **no `group` field at all**:

```
power part WITHOUT a group -> ok=false
   - PG1: "power:GND" is a power-port symbol, which the drafting engine supplies itself. Remove PG1 ...
```

One finding, and it is the power-port one. No group finding is produced, because `ir.ts:238` still does `if (sym?.isPower) continue;` in the group loop. That skip is deliberate and load-bearing (it is what keeps the power refusal to a single finding, guarded by a mutation-checked assertion in `draft-hardening.test.ts`), so the code is right and the spec is what moved too far.

The previous wording, "a non-power part has no group assignment", was accurate for precisely this reason. Removing the qualifier to reflect that power parts can no longer appear in a valid IR lost the distinction that the validator still makes internally. The same overshoot is in the requirement sentence at `spec.md:21`, "every part must belong to exactly one group present in SUBSYSTEMS.md": true as a property of valid IRs, but not a description of a check the validator performs on every part.

**Fix:** restore the qualifier in the scenario at `spec.md:35` ("a non-power part has no group assignment") and at `spec.md:21`, and if the intent was to record that power parts cannot appear at all, say that where it belongs, in the power-port refusal clause that is already in the same paragraph, rather than by weakening the group clause.

### 2. [low] `PWR_FLAG` is listed among lib_id examples but is not a lib_id

**Where:** `openspec/changes/deterministic-schematic-drafting/specs/schematic-drafting-engine/spec.md:21`

**Fails when:** the new clause reads "a part resolving to a power-port library symbol (`power:GND`, `power:+3V3`, `PWR_FLAG`)". The first two are lib_ids in `library:symbol` form; the third is a bare symbol name. The engine's own generated flag is `copperhead_power:PWR_FLAG` and the stock one is `power:PWR_FLAG`, so a reader copying the example writes a lib_id that resolves to nothing. Minor, but this is an example list in a normative clause, which is exactly the text a reader copies.

**Fix:** write `power:PWR_FLAG` so all three examples are the same shape.

## Notes, not findings

- **Correcting my own claim from the first pass.** I wrote that the filesystem-root break would be "exercised incidentally by any path that runs out of ancestors". It is not: every new test anchors before reaching the root, so `symlib.ts:216` is still unreached. I am leaving it as a note rather than a finding because `maxDepth` already bounds the loop independently, so the break is an optimization rather than a safety property, and `path.dirname` is idempotent at the root. Worth knowing that the line is decorative rather than load-bearing.
- **The unanchored fallback is a deliberate behavior change worth a second opinion.** When no `.git` or `.copperhead` is found within `maxDepth`, only a cache adjacent to the sheet counts, where the previous code collected every cache up the walk. For a real project this cannot trigger (a git repo is required before any run starts and the config lives in `.copperhead/`), and all four reference boards carry both markers, which I verified. But it is a silent narrowing for a schematic outside a project, and it is documented only in the function's doc comment.
- The four new `vendoredCacheDirs` tests are deterministic and each anchors at depth 0 or 1, so unlike the fixture they replaced they cannot be perturbed by unrelated directories in the temp tree.
- `openspec validate deterministic-schematic-drafting` returns `Change 'deterministic-schematic-drafting' is valid`. Run via `npx @fission-ai/openspec` because the CLI is not installed locally.
- Scope stayed tight: the commit touches one source file, one test file, and two spec artifacts, with no drive-by edits.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 215, branch fix/212-vendored-symbol-verification,
based on main.

Check it out first:
  gh pr checkout 215

Both findings are in spec prose, not in code. Fix each below. Keep the change scoped
to these findings: no unrelated refactors, renames, reformatting, or dependency
changes. If you disagree with a finding, say so and leave it alone rather than doing
a partial fix.

FINDING 1 [low] (optional) The reworded ungrouped-part scenario now contradicts the
implementation.
  Where: openspec/changes/deterministic-schematic-drafting/specs/schematic-drafting-engine/spec.md:35,
         against src/kicad/draft/ir.ts:238
  Fails when: the scenario reads "WHEN a part has no group assignment / THEN
    validation fails naming the refdes and the available groups from SUBSYSTEMS.md".
    Drive that input with a power-port part carrying no group field and the THEN does
    not happen: validation produces exactly one finding, the power-port refusal, and
    no group finding, because ir.ts:238 still does `if (sym?.isPower) continue;` in
    the group loop. That skip is deliberate and load bearing, since it keeps the
    power refusal to a single finding, and it is guarded by a mutation-checked
    assertion in test/draft-hardening.test.ts. The code is right; the spec rewording
    went too far. The same overshoot is in the requirement sentence at spec.md:21,
    "every part must belong to exactly one group present in SUBSYSTEMS.md": true as a
    property of valid IRs, but not a description of a check the validator runs on
    every part.
  Fix: restore the "non-power part" qualifier in the scenario at spec.md:35 and in
    the requirement sentence at spec.md:21. If the intent was to record that power
    parts cannot appear in a valid IR at all, state that in the power-port refusal
    clause already present in the same paragraph, rather than by weakening the group
    clause.

FINDING 2 [low] (optional) PWR_FLAG is listed among lib_id examples but is not a
lib_id.
  Where: openspec/changes/deterministic-schematic-drafting/specs/schematic-drafting-engine/spec.md:21
  Fails when: the clause reads "a part resolving to a power-port library symbol
    (power:GND, power:+3V3, PWR_FLAG)". The first two are lib_ids in library:symbol
    form; the third is a bare symbol name. The engine's generated flag is
    copperhead_power:PWR_FLAG and the stock one is power:PWR_FLAG, so a reader
    copying the example writes a lib_id that resolves to nothing. This is an example
    list inside a normative clause, which is exactly the text a reader copies.
  Fix: write power:PWR_FLAG so all three examples have the same shape.

Neither finding is a correctness bug, so no regression test is required for them. Do
not weaken or delete the existing assertions in test/draft-hardening.test.ts that pin
the single-finding behavior, since finding 1 depends on that behavior being correct.

This PR already changes spec-level behavior, and openspec/specs/SPEC.md correctly does
not move (it does not describe schematic.intent.json or IR validation at all). After
editing the delta spec, `openspec validate deterministic-schematic-drafting` must still
pass. The CLI may not be installed locally; run it with
`npx @fission-ai/openspec validate deterministic-schematic-drafting`.

    by prompt text.

Then verify:
  npm run typecheck
  npm run build
  npm test

All three must pass. Note that on Windows checkouts 20 tests fail on main before any
change, from CRLF normalization and platform assumptions in the tests; compare failure
sets by test name rather than by count so you do not mistake those for regressions.

````

**Comment by animesh-chouhan** [comment]
_Automated pr-review pass (third). Not a human review._

Incremental pass, and the first on this PR not authored by the PR author (the reviewer has no commits under review; the two earlier passes were the author's self-reviews). The [second pass](https://github.com/chouhanindustries/copperhead/pull/215#issuecomment-5227593247) covered `38cf212`; this one covers the single commit added since, `519b6a4`, which changes two OpenSpec markdown files and no source or test code. The review ran in an isolated `git worktree` at commit `519b6a4`; the main checkout was never switched.

## Verdict: approve

Both findings from the second pass are resolved, each re-verified against the artifacts and the implementation rather than taken from the commit message. No new findings. Across the three passes the PR has now absorbed CodeRabbit's three items and five pr-review findings, and each fix is pinned by a test or by `openspec validate`; the branch is in mergeable shape.

## Status of the previous pass's findings

| # | Finding | State | How verified |
| --- | --- | --- | --- |
| 1 | Reworded ungrouped-part scenario contradicted the implementation | resolved | "non-power part" qualifier restored at the requirement (`spec.md:21`) and the scenario (`spec.md:35`), with the intent of the earlier rewrite recorded in the power-port refusal clause instead ("No IR that passes validation therefore contains a power-port part at all; the qualifier scopes which parts the group check itself examines"). That sentence is accurate against the code: the group loop still skips power parts at `src/kicad/draft/ir.ts:238`, so the qualifier describes real validator behavior. `tasks.md` 6.2 was updated to the same wording. |
| 2 | `PWR_FLAG` listed bare among lib_id examples | resolved | now `power:PWR_FLAG`; all three examples in the refusal clause are in `library:symbol` form |

`openspec validate deterministic-schematic-drafting` returns `Change 'deterministic-schematic-drafting' is valid` at head, re-run in the worktree.

## Metrics

```text
lines: +522 / -38 (net 484) across 7 files  ·  src +182/-36, test +332/-0, docs +0/-0, spec +8/-2, other +0/-0
new vs net: 0 new files (0 in src), 7 modified; 0 new src lines (the net-new surface)
tests: +332 test lines; suite (head) 784 pass / 20 skip / 0 fail; base suite not run (base CI green)
diff coverage: 98.2% of changed executable src lines exercised (55/56, measured via vitest v8); uncovered: src/kicad/symlib.ts:660
```

Method: measured by the metrics script in the review worktree at `519b6a4`, merge base `c208680` (`git merge-base origin/main HEAD`). Change size from `git diff --numstat` with generated files excluded. Suite from `npx vitest run --reporter=json` inside the worktree (Linux, so the Windows/CRLF baseline failures the author documented do not appear; zero failures of any kind at head). Coverage from vitest's v8 provider intersected with the changed src lines. The base suite was not re-run locally; base CI is green.

**Reconciling the one uncovered line.** `src/kicad/symlib.ts:660` is the with-candidates branch of the `no-symbol` diagnostic. Its only change versus the merge base is the phrase "installed or vendored symbol library", which is byte-identical in the covered sibling branch on line 661 and asserted there by `test/symlib.test.ts:700` (with a negative assertion against the old wording on line 701). No changed behavior on that line is unasserted, so it is a note rather than an untested-surface finding.

## Invariant checks performed on `519b6a4`

The commit touches only `openspec/changes/deterministic-schematic-drafting/specs/schematic-drafting-engine/spec.md` and `.../tasks.md`, so the code invariants are trivially preserved; each was still checked against the diff rather than assumed.

| Invariant | Result | Evidence |
| --- | --- | --- |
| Spec-gated in | pass | no file under `src/` touched by this commit; tool list and unlock gate unchanged |
| Verification-gated out | pass | no change to ERC/DRC sequencing, repair cycles, or rollback |
| `check` LLM-free / network-free | pass | no source change; nothing new reachable from `src/commands/check` |
| No sexp serialization | pass | no source change; the parser stays read-only |
| Sync-obligations ledger | pass | untouched |
| Secrets | pass | markdown-only diff; no logging or env reads |
| Spec coherence | pass | the two artifacts moved together, the restored qualifier now matches `ir.ts:238`, and `openspec validate` passes |

## Findings

None. Both candidate concerns this pass examined (the accuracy of the new explanatory sentence in the refusal clause, and the uncovered diagnostic line) were refuted by reading the implementation and the tests, as documented above.

## Fix prompt

No findings survived, so there is no fix prompt.

adhavan18 (commented):
Checked this on `519b6a4` rather than reading the diff, since defect 1's fix turns on an ordering decision that is easy to state and easy to get backwards.

`npm run typecheck` clean. `test/symlib.test.ts` 41/41. `test/draft-hardening.test.ts` 32 with one failure, `a drafted sheet with a derived symbol loads in KiCad and matches its vendored copy`, which fails identically on the merge-base with `kicad-cli not found on PATH`: environmental here, not yours.

## The ordering is right, and it is the part that matters

`verifySchematicSymbols` exists to catch a `lib_symbols` entry whose pins have drifted from the real part. Resolving against the vendored cache first would compare the entry with the file it was copied from and report clean forever, which would quietly turn this function into a no-op while still printing reassuring output. Confirmed the search path is built the safe way round:

```ts
const dirs = [...(await symbolSearchDirs(env)), ...(await vendoredCacheDirs(schPath))];
```

installed first, vendored appended. And the test at [symlib.test.ts:658](test/symlib.test.ts) pins exactly the failure mode that ordering protects against: a tampered `Device:R` pin is still caught in the same pass where the project's own symbols verify clean. That is the right test to have written, because it is the one that fails if someone later "simplifies" the concatenation order.

Agreed with taking the vendored-cache route over special-casing the `copperhead_power` namespace. The namespace fix would have been three lines and would have needed doing again the first time a project vendored anything else.

## Defect 3 is the one I would have shipped wrong

`symbolSearchDirs` testing exclusivity on whether an override is *set* while its output collects only directories that *exist* is a genuinely nasty divergence, because the failure is silent and total rather than partial: an empty search path is not "search nothing", it is "every symbol is absent", and `search_symbols`, `symbol_pins`, the dossier and the recovery fact-check all degrade at once while still returning confident answers.

The `addIfPresent` comment about `access` succeeding on a regular file is a good catch on top of that: an override naming a file would have counted as resolved, suppressed the fallback, and then satisfied no lookup, since every search joins `<lib>.kicad_sym` onto these entries. That is the same bug one layer down and it would have survived the obvious fix.

## Defect 2, and the thing I liked most about this PR

Recording the part in `symbols` *before* adding the refusal, so the group and BOM cross-checks take their existing `isPower` skips, is a small ordering choice with a real effect on what the model sees: one finding about one mistake instead of three unrelated-looking ones. A model that gets three findings for one error will usually try to fix three things.

Also worth saying: the PR description corrects an earlier version of itself on whether defect 2 was a spec-level change, and then moves the delta spec accordingly. That is the right call and not a common one.

## No blocking findings

I tried to find a case where a vendored symbol shadows a stock one and could not: containment is by directory order and the stock library wins whenever it provides the lib_id. The `no-library` versus `wrong-library` distinction still behaves as before for genuinely missing libraries.

One small note, not a request: `vendoredCacheDirs` walks up to `maxDepth = 5` and only anchors on a project root, falling back to the starting directory alone when it never finds one. The tests cover the unanchored case and the `maxDepth = 12` case explicitly, so the behaviour is deliberate. Worth a line in the doc comment about what happens on a schematic sitting deeper than 5 levels inside a large monorepo, since the failure there is a silent revert to old behaviour rather than an error.

Nice piece of work. Defect 1 in particular was a deadlock shape, not just noisy output: four confident findings out of six parts, none of them reachable, which is exactly what #163 looks like from the model's side.

thegoodengineer:
Thanks for checking it on `519b6a4` rather than reading the diff. The ordering in defect 1 was the thing I most wanted a second pair of eyes on, precisely because it reads fine either way round and only one of them leaves the function doing its job, so having someone independently reconstruct that argument is worth a lot.

Acted on your note, and it turned out to deserve more than the doc line you asked for. You called it "a silent revert to old behaviour", and that is exactly what it is: past the budget the walk falls back to the sheet's own directory, misses the project's `sym-lib-cache`, and every `copperhead_power:*` symbol goes back to being unresolvable. That is #212's headline bug returning, with no error to notice it by. Measured before the change:

```
 4 levels below repo root -> finds the project cache
 5 levels below repo root -> finds the project cache
 6 levels below repo root -> MISSES it (defect 1 returns)
 7 levels below repo root -> MISSES it (defect 1 returns)
```

You were generous in assuming `maxDepth = 5` was deliberate. It was not: I picked the number with nothing behind it, which is the whole problem. `path.dirname` strictly shortens until the filesystem root, so the walk terminates on its own and the bound was never load-bearing. It is now set far past any real tree and documented as a guard against a pathological path rather than a tuning knob, so exhausting it is not a state a real project can reach. Every depth out to 20 finds the cache in `d0930e8`.

The anchor is unchanged, so the escape it was added for is still refused (a foreign cache above a project still yields nothing), and the deep-tree test now asserts the default reaches a seven-level sheet while exercising the guard with an explicitly tight budget instead of leaning on the default being small.

On the derived-symbol failure: agreed, that is `kicad-cli` missing on your PATH. It fails the same way on the merge base, and passes here where KiCad 10 is installed. Nothing in this PR touches it.

thegoodengineer:
_Automated pr-review pass (third). Not a human review._

**Reviewer authored every commit under review.** Self-review, so a clean result here is weaker evidence than an independent one. Ran in an isolated `git worktree` at commit `d0930e8`; the main checkout was never switched.

Incremental. [Pass 1](https://github.com/chouhanindustries/copperhead/pull/215#issuecomment-5227330709) covered `bcf1938`, [pass 2](https://github.com/chouhanindustries/copperhead/pull/215#issuecomment-5227593247) covered `38cf212`. This covers the two commits since: `519b6a4` and `d0930e8`, the second of which came from @adhavan18's review note.

## Verdict: approve

No findings. Both prior automated passes are fully resolved, and the human review's one note is resolved beyond what it asked for.

## Status of everything raised so far

| Raised by | Item | State | How verified here |
| --- | --- | --- | --- |
| pass 1 | power-port refusal absent from the delta spec | resolved | `openspec validate` returns valid; refusal, scenario, and `tasks.md` 6.2a all present |
| pass 1 | `vendoredCacheDirs` escapes the project | resolved | foreign cache above a project still yields `[]` at `d0930e8` |
| pass 1 | untested surface | resolved | see the coverage note below |
| pass 2 | reworded scenario contradicted the implementation | resolved | re-ran both directions in the worktree, output below |
| pass 2 | `PWR_FLAG` not a lib_id | resolved | now `power:PWR_FLAG` |
| @adhavan18 | `maxDepth = 5` silently reverts to old behaviour deep in a tree | resolved | depth table below, out to 20 levels |

Pass 2's finding re-verified against the worktree build rather than taken from the commit message:

```
SCENARIO "non-power part has no group" -> R2 has no group assignment; available groups: ...
power part, no group                  -> findings=1 :: PG1: "power:GND" is a power-port symbol ...
```

The first is the restored scenario's THEN actually firing. The second is why the qualifier has to be there: the group loop still skips `isPower` parts, so a power part yields the power finding alone.

## Metrics

```text
lines: +543 / -38 (net 505) across 7 files  ·  src +200/-36, test +335/-0, docs +0, spec +8/-2
new vs net: 0 new files, 7 modified; 0 net-new src files (all edits to existing modules)
tests: +335 test lines; suite (base main) 750P/20F/20S of 790 -> (head) 764P/20F/20S of 804
diff coverage: not measured (@vitest/coverage-v8 unavailable in the worktree);
               untested surface hand-mapped instead, reconciled below
ci: 5 pass / 0 fail / 0 pending; required checks: none reported
```

Method: measured in the review worktree at `d0930e8`, merge base `c208680`. Change size from `git diff --numstat`. Suite from `npx vitest run --reporter=json` inside the worktree after a real `npm ci`, compared to a `main` run **by test name**: **0 new failures**. The 20 at head are exactly `main`'s 20 pre-existing Windows/CRLF failures, same names. Diff coverage is genuinely not measured, so the coverage note below is a hand-walk, not a percentage.

## Invariant checks performed on `519b6a4..d0930e8`

| Invariant | Result | Evidence |
| --- | --- | --- |
| Spec-gated in | pass | zero files under `src/agent/` touched |
| Verification-gated out | pass | no change to ERC/DRC sequencing, repair cycles, or rollback |
| `check` LLM-free / network-free | pass | no `fetch`, provider, or URL added |
| No sexp serialization | pass | no `writeFile` or serializer added; reads only |
| Sync-obligations ledger | pass | untouched |
| Secrets | pass | no `console.*` and no new `process.env` reads |
| Spec coherence | pass | `openspec validate deterministic-schematic-drafting` returns valid; `SPEC.md` correctly unmoved |

## Findings

None. Three candidates were raised and all three were refuted:

- **Does the deeper walk let a test escape into `TMPDIR`'s ancestors?** No. Walked the real chain: no `.git`, `.copperhead`, or `sym-lib-cache` exists anywhere from `%TEMP%` up to `C:\`. More importantly it cannot matter structurally: all four `vendoredCacheDirs` tests anchor inside their own fixture, so the walk never leaves it, and an unanchored walk only ever checks depth 0. Environment-independent either way.
- **Does raising the bound reintroduce the escape pass 1 found?** No. The anchor is what refuses that, and it is unchanged. Re-confirmed at `d0930e8`: a foreign cache one level above a project still yields `[]`.
- **Does the new explanatory sentence in the spec misdescribe the mechanism?** No. "A power-port part is refused by this rule before the group rule could apply" is accurate in program order: the refusal is recorded in the parts loop, the group loop runs later and skips `isPower`.

## Coverage note, reconciled with pass 1 and pass 2

Pass 1 raised three untested paths. Two were fixed in `38cf212` (the reworded diagnostic, the `maxDepth` bound). The third, the filesystem-root break at `symlib.ts:216`, I noted in pass 2 as still unreached, correcting my own earlier claim that it was incidental.

It is now genuinely reached, as a side effect of `d0930e8` rather than by design: with the walk no longer stopping at 5, an unanchored sheet in a temp directory runs all the way up until `path.dirname` stops shortening, which is exactly that branch. Several existing `verifySchematicSymbols` tests drive that path. Worth recording that this was luck, not intent.

## Notes, not findings

- `MAX_PROJECT_WALK = 64` is now a named constant that documents itself as a guard rather than a tuning knob, which is the substantive part of the fix. The previous `5` was a literal with nothing behind it, and that is what made the regression possible.
- The deep-tree test was previously asserting `[]` **via the default**, meaning it encoded the bug as expected behaviour. It now asserts the default reaches a seven-level sheet and exercises the guard with an explicit tight budget. That inversion is the more valuable half of `d0930e8`.
- Scope stayed tight across both commits: one source file, one test file, two spec artifacts, no drive-by edits.
- Still worth stating plainly: three of the four review rounds on this PR were self-review. @adhavan18's pass is the only independent one, and it found the thing three automated passes had all walked past.

No findings, so no fix prompt.

animesh-chouhan:
ack, checking

Issue #204, opened 2026-08-08
Deterministic router merges two distinct nets (multiple_net_names): HSE_DRV shorted to TOUCH_IRQ survives every intent-level lever


animesh-chouhan:
## Observed

Harness run lemondrop attempt-05 (`copperhead-test-harness/run-logs/2026-08-07T17-52-03/attempt-05`, branch `feat/symbol-pin-dossier` @ `a2ba5f5`, 160-part STM32H7 granular synth). Stage 4 drafted a full netlist (160 parts, ~108 nets, 513 wire segments, 248 labels, 62 no-connects, nine group boxes; `verify_symbols` clean, `check_drift` clean) and ERC consistently failed with exactly one electrical error:

```
multiple_net_names: Both HSE_DRV and TOUCH_IRQ are attached to the same items
```

tying U3.13 (PH1, 24.576 MHz crystal drive via R6) to the TOUCH_IRQ net (U3.97/PE0, J8.12, R44). Shipping it would push a netlist into layout with the HSE drive shorted to the touch interrupt.

## Why this looks like an engine routing artifact, not an intent error

The stage-4 agent tested and disproved four repair hypotheses, methodically, before refusing to certify (its full write-up is preserved in the failed-run stash `copperhead failed run 2026-08-07T23-34-46-178Z` in the workspace, file `docs/STAGE4-STATUS.md`):

1. Net rename: the collision reproduced verbatim under different names.
2. Sheet size: reproduced on A2, A1, and A0.
3. Group ordering: five `hints.groupOrder` permutations plus the engine default.
4. Part declaration order: byte-identical output coordinates, so declaration order does not affect placement at all.

The intent JSON names two disjoint nets; the drafted geometry connects them. A wire segment routed by the engine appears to make contact with both nets' items, which KiCad then reports as one net carrying two names.

Relation to I2 (`fix e5c4f34`): that work made net-merge a hard ERC failure, and the gate held here (the run refused rather than certifying). This issue is the generative half: the router can still produce the merged geometry the gate exists to catch.

## Cheapest known unblock

The agent's proposed topology workaround: drop R6 (a 0 ohm placeholder in the crystal drive) and wire Y1 direct, removing the pass-through the router routes badly. That unblocks the brief; the engine defect remains.

## Repro material

- Constraint recorded by the run: `mcu.hse_drv_touch_irq_net_merge`
[...]

PR #216, opened 2026-08-08
test(draft): run the engine against real designs via their KiCad netlists

animesh-chouhan:
## What

Points the drafting engine at real KiCad designs instead of fixtures. Adds an opt-in sweep, [draft-real-corpus.test.ts](test/draft-real-corpus.test.ts), and the netlist reader it runs on, [netlist.ts](test/support/netlist.ts). Tests only, no production source changes, default `npm test` unchanged.

## Why

Every existing draft suite feeds the engine intents we wrote. That makes them good regression tests and poor discovery tests: an intent written to exercise the engine tends to stay inside what the engine already does. Real designs do not. They bring multi-unit opamps, project-local symbol libraries, hierarchies eight sheets deep, and net counts in the hundreds.

This is the corpus half of what [#141](https://github.com/chouhanindustries/copperhead/issues/141) asks for, applied to drafting rather than layout.

## Design

**Input is the netlist export, not the schematic.** `kicad-cli sch export netlist` makes KiCad flatten the hierarchy, resolve the buses, and name every net before we see it, including the ones the designer never named (`Net-(D15-1)`). Reading the `.kicad_sch` directly would mean re-deriving connectivity, and unnamed nets are the majority on a real board: `interf_u` has 303 of 373 pins on nets with no label. The export is an s-expression, so [sexp.ts](src/kicad/sexp.ts) reads it and no new parser enters the tree.

`libsource` supplies the lib_id and `sheetpath` supplies the group, so the IR extraction is deterministic with no LLM in the loop.

**The contract is not "every board drafts".** Most real boards legitimately will not: the engine refuses multi-unit symbols by design, because a symbol's units share symbol-space pin coordinates and drafting one would silently merge nets. Asserting that every board drafts would just encode the wrong goal.

The contract is the weaker, more important one: draft faithfully or refuse legibly, never silently wrong. Per project, drafting either succeeds or returns findings and never throws; when it succeeds the drawn netlist must partition pins exactly as the source did; when it refuses every finding must carry an attributable detail.

**Comparison is by pin partition, not net name.** What must survive drafting is which pins end up joined to which. Comparing names would also churn against KiCad's auto-naming, which numbers unnamed nets from the first pin it meets and so shifts whenever placement does.

**Nothing is vendored, and the sweep is opt-in.** Gated on `COPPERHEAD_TEST_CORPUS=1`, skipped when the corpus is absent, `COPPERHEAD_CORPUS_DIR` to point it elsewhere. The projects are not committed because the interesting ones carry licences a permissive repo cannot redistribute: the one board that currently round-trips clean, stickhub, is CC BY-NC-SA (NonCommercial). Reading them from a local KiCad install redistributes nothing. Of the corpus, only jetson-agx-thor-baseboard and tiny_tapeout (Apache-2.0) and royalblue54L_feather (CERN-OHL-P) would be vendorable, and all three currently refuse, so a vendored tier would start as a known-limitation ledger rather than a suite.

## Spec / OpenSpec

No spec-level behavior change: tests only, no production source touched.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | pass (770 passed, 21 skipped, 57 files; corpus skipped by default) |
| Corpus sweep (opt-in) | `COPPERHEAD_TEST_CORPUS=1 npx vitest run test/draft-real-corpus.test.ts` | pass (16 tests, 87s) |
| Live-LLM (opt-in) | keyed on `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `COPPERHEAD_TEST_CODEX` | n/a, this PR adds no LLM-touching test |

New tests: 16 in [draft-real-corpus.test.ts](test/draft-real-corpus.test.ts), all skipped unless opted in. No changed or deleted tests. `npm test` run with `COPPERHEAD_MIN_FREE_MB=100` for an environmental disk-preflight trip unrelated to this change.

Exercised against KiCad 10.0.4's 15 demo projects on Linux.

```text
$ COPPERHEAD_TEST_CORPUS=1 npx vitest run test/draft-real-corpus.test.ts

real-design corpus (/usr/share/kicad/demos):
  cm5_minima:                refused parts=84   nets=95   merged-net=1
  complex_hierarchy:         refused parts=68   nets=50   multi-unit=2
  ecc83:                     refused parts=15   nets=9    multi-unit=1
  interf_u:                  refused parts=24   nets=110  merged-net=1
  jetson-agx-thor-baseboard: refused parts=1123 nets=893  symbol-unresolved=1123
  kit-dev-coldfire-xilinx:   refused parts=160  nets=209  multi-unit=2
  multichannel:              refused parts=114  nets=79   multi-unit=5
  openair-max:               refused parts=204  nets=120  symbol-unresolved=194 missing-pin=6
  pic_programmer:            refused parts=63   nets=34   multi-unit=1
  royalblue54L_feather:      refused parts=71   nets=68   symbol-unresolved=4 missing-pin=1
  sonde xilinx:              refused parts=25   nets=26   multi-unit=2
  stickhub:                  DRAFTED parts=94   nets=45   paper=A0 lost=0 gained=0
  tiny_tapeout:              refused parts=150  nets=108  multi-unit=2
  video:                     refused parts=189  nets=371  symbol-unresolved=10
  vme-wren:                  refused parts=1507 nets=1307 symbol-unresolved=1494

 Test Files  1 passed (1)
      Tests  16 passed (16)      Duration 87.77s

$ npx vitest run test/draft-real-corpus.test.ts     # not opted in
 Test Files  1 skipped (1)
      Tests  1 skipped (1)
```

A real 94-part board round-trips with connectivity exact: stickhub in, engine-drafted sheet out, zero nets lost and zero gained.

### What the sweep found

Two engine limitations, filed separately rather than fixed here:

- Router produces a merged net on dense real boards (cm5_minima shorts +5V to GND, interf_u shorts /PC-RD to /WR_REG). The guard catches both and refuses, so nothing ships wrong, but neither board can be drafted. Same class as [#204](https://github.com/chouhanindustries/copperhead/issues/204), which is closed.
- Multi-unit symbols block 7 of 15 real designs. Known and deliberate, now quantified.

## Docs

None. The two new files carry their reasoning in file-level doc comments, including why the netlist export is the input and why nothing is vendored.

---

## Invariant checklist

- [x] **`check`/`verify` stays LLM-free and network-free**: unaffected. The sweep is itself LLM-free; it shells out to `kicad-cli` only, which `check` already does.
- [x] **No sexp serialization**: preserved. The netlist reader parses and never serializes; the only file written is emitter output from the drafting path.
- [x] **Sync-obligations ledger**: N/A, no hook or ledger change.
- [x] **Secrets**: N/A, no transcript, summary, or key handling touched.
- [x] **Spec coherence**: N/A, no spec-level behavior change.

## Summary by CodeRabbit

* **New Features**
  * Added support for drafting multi-unit KiCad symbols as separate, correctly connected instances.
  * Improved oversized group layouts with bounded columns and deterministic wrapping.
  * Added a command for drafting and comparing real KiCad design examples.

* **Bug Fixes**
  * Prevented unintended electrical contacts between wires, pins, power stubs, and labels.
  * Improved power-stub clearance and label placement.

* **Tests**
  * Added real-project connectivity checks, netlist validation, multi-unit, layout, and clearance coverage.

* **Documentation**
  * Added guides and reports for real-design validation examples.


animesh-chouhan:
Pushed the #217 fix onto this branch (251929d), as requested, so the PR now carries both the corpus and an engine change.

## Root cause

Two passes decide where a stub ends by asking a question about **text**, then move the wire endpoint as a side effect. The power pass picks a length that keeps the rail name from colliding ([engine.ts](src/kicad/draft/engine.ts), power stub pass); the label nudge rides a stub outward until its label reads clean (`rideTo` moves both the label and `stub.x2/y2`). Neither asked where the endpoint landed, so both could park a live end on another net's wire.

The merged-net gate caught the result and refused, which is why nothing shipped wrong. The cost was capability: the board could not be drafted at all.

## What the fix does

Both passes now get the clearance check the signal passes already had, with three corrections that the first attempt got wrong and the test suite caught:

1. **A pin sits on its own part's outline**, so *every* stub length crosses a body, including the preferred one shipped today. Treating any crossing as disqualifying vetoed the whole ladder. The test is now no-worse-than-baseline: a rung may not enter a body the preferred length was clear of.
2. **The ladder may shorten, not only lengthen.** Two pins facing each other cannot be separated by growing the stub, only by pulling it back.
3. **The power pass must not react to predicted signal stub ends.** Signal stubs already adapt around power stubs (the pull-up idiom relies on exactly this: SIG grows to 3 units to clear the VCC wire, then the two cross interior-to-interior, which does not connect). Predicting them made the power stub move first and dragged its rail name into R1. Prediction is now opt-out, and off for the two passes that run when the real geometry is known.

## Result

| | before | after |
| --- | --- | --- |
| cm5_minima | refused, +5V shorted to GND | DRAFTED, lost=0 gained=0 |
| interf_u | refused, /PC-RD shorted to /WR_REG | DRAFTED, lost=0 gained=0 |
| corpus boards drafted | 1 | 3 |

The four byte-pinned reference boards and the Tier C golden are **unchanged**, which is the intended property: the ladder only fires where the old engine would have refused outright.

`npm test` 773 passed / 21 skipped / 58 files. Typecheck and build pass.

## One caveat on the regression tests

`test/draft-stub-clearance.test.ts` adds a `FacingRails` fixture symbol reproducing the cm5_minima geometry (two rails one column apart, facing across four grid units). That test **fails on main and passes with the fix**, so it genuinely pins the power half.

The label-nudge test does not. Two sweeps over fixture-scale intents (varying part count, part type, net-name length, and the mix of wired pairs to labelled runs) failed to recreate interf_u's geometry, so that test passes both before and after. It is labelled in the file as an invariant guard rather than a reproduction, and the reproducing case for that half lives only in the opt-in corpus sweep.

animesh-chouhan:
Added the manual half (d3dbcf3), since the automated sweep answers whether the engine drew the right circuit but not whether anyone could read the result.

`npm run realdesigns` drafts every project in the corpus and leaves the engine's drawing next to the one a person made of the same circuit. Docs in [manual-tests/real-designs/README.md](manual-tests/real-designs/README.md), following the `reference-boards/` convention.

No byte comparison here, deliberately. There the golden is the engine's own pinned output so a diff answers the question; here the golden was drawn by hand and the engine places by its own rules, so the two sheets never match and diffing them says nothing. Netlist by machine, drawing by eye. Nothing is vendored: projects are read from a local KiCad install and outputs land in the gitignored `manual-tests/runs/`, which matters because the board that drafts cleanest (stickhub) is CC BY-NC-SA.

## What it immediately found: #219

The renders showed something the numbers had hidden. All three boards that draft are **correct and unreadable**:

| Board | Parts | Netlist | Composite | True error findings |
| --- | --- | --- | --- | --- |
| stickhub | 94 | exact | 40 | **367 out-of-frame** |
| cm5_minima | 84 | exact | 40 | 20 text-collision |
| interf_u | 24 | exact | 40 | 13 text-collision |
| buck-12v-5v (reference) | 24 | n/a | 85.6 | 0 |
| ldo-demo (reference) | 12 | n/a | 92 | 0 |

stickhub reproduces its netlist exactly and puts 94 parts in one horizontal ribbon on an A0 sheet that is roughly 85% empty, where the designer drew the same circuit densely on A3. 367 out-of-frame findings means content placed outside the drawing frame, so it would not plot.

`interf_u` is the useful counterexample: same part count as buck-12v-5v but 110 nets against 11, and it scores 40 too. The trigger is density, not size.

This is invisible to the automated sweep, correctly: its contract is draft faithfully or refuse legibly, and faithfulness is not legibility. Filed as #219 rather than fixed here.

## Also: a search papercut

`checkLegibility` reports exactly 10 findings for every board because `familyCap` truncates per family per sheet; true counts live in `suppressed`. Establishing that took longer than it should have, because [src/kicad/legibility.ts](src/kicad/legibility.ts) held a **literal NUL byte** as a map-key separator, which made git-grep, ripgrep and most editors classify the whole file as binary and return **empty with no error** for every search, indistinguishable from no match. Now written as a unicode escape: identical string at runtime, file searchable again (`grep -c "severity"` went 0 to 13).

## Full sweep

15 projects, all processed. 3 draft with connectivity exact, 12 refuse (7 multi-unit, 5 symbol resolution). `npm test` 773 passed / 21 skipped / 58 files; typecheck and build pass.

PR #226, opened 2026-08-09
fix(sync): report a configured schematic that is missing from disk

### LINKED ISSUE #188: check exits 0 when the configured schematic is missing, and says it is not configured

adhavan18:
## What

`copperhead sync` no longer reports a clean run when `config.schematic` names a file that is not on disk. It reports the missing schematic and exits 2.

## Why

Closes #188. This is the `sync` half; #191 is the `check` half and is already approved. Splitting them because the fix is a different shape in each command, per the review discussion on #191.

Drift and the forbidden-pin constraint check are both gated on the schematic existing, written out twice in [sync.ts](src/commands/sync.ts). When the file is absent the two skip together and nothing else notices, so the command reports clean having verified nothing.

Reproduced on `a2471b3`, with an init-ed repo whose `schematic` points at a path that does not exist:

```text
$ copperhead --repo <sandbox> sync --dry-run
sync: no inconsistencies
exit=0
```

The contrast with `check` on the same repo state (with #191 applied) is the reason this is worth fixing rather than leaving to the exit code of the other command:

```text
check -> exit 1, "schematic configured at hardware/board.kicad_sch but the file is missing"
sync  -> exit 0, "sync: no inconsistencies"
```

## Design

The missing schematic goes into `violations[]`, not `resolvable[]`.

That is the load-bearing decision here, so the reasoning: [cli.ts](src/cli.ts) hands every resolvable item to `syncResolve`, which runs the agent loop against them. A schematic that is not on disk is not something an agent should try to write its way out of, and asking it to would be inviting it to invent one. Restoring the file or correcting the path is a human decision, which is exactly what `violations[]` already means in this command (AC-7.3, "never auto-resolved"). It also produces the right exit code for free, since violations exit 2 before the resolve phase is reached.

`SyncViolationItem['kind']` widens from the single `'requirement-violation'` to a union with `'schematic-missing'`, so the two can be told apart by a caller reading `--json`. The `formatSyncReport` heading is unchanged: "not auto-resolved; human decision needed" is accurate for this case too.

The two guards are resolved once into `schematicUsable`, the same consolidation #191 does in `check.ts`.

## Spec / OpenSpec

No spec-level behavior change. AC-7.3 is preserved and this leans on it: the new item is a violation precisely so it is never auto-resolved.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | targeted, see below | 6 new pass, no regressions |

New: [sync-missing-schematic.test.ts](test/sync-missing-schematic.test.ts), 6 tests, all pass here.

On the parent commit 4 of the 6 fail. The other 2 are the negative controls (no schematic configured, and schematic present), which correctly pass on both sides:

```text
× reports it, where it previously found nothing at all      -> expected [] to have a length of 1
× says the checks did not run, not that nothing was wrong   -> expected 'sync: no inconsistencies' not to be 'sync: no inconsistencies'
× is a violation, not a resolvable item                     -> expected 0 to be greater than 0
× behaves the same when the file is deleted after config    -> expected [] to have a length of 1
✓ stays silent when no schematic is configured
✓ stays silent when the configured schematic is present
```

Regressions: `gating-sync.test.ts` is the only existing suite that exercises `syncVerify`. Ran it on this branch and on the parent and diffed the failing-test lists: 1 failure each, identical, pre-existing in this environment.

Sandbox: fixture schematic copied into a fresh git repo, `copperhead init` run, then `.copperhead/config.json` pointed at a path that does not exist.

**Disclosure:** KiCad is not installed on this machine, so `COPPERHEAD_KICAD_CLI` points at a stub that answers `version`, purely to get past preflight. It is never invoked in these runs, because every schematic-reading step is skipped by the condition under test.

```text
$ copperhead --repo /tmp/ch-sync188 sync --dry-run
1 REQUIREMENT VIOLATION(S) (not auto-resolved; human decision needed):
- schematic configured at hardware/board.kicad_sch but the file is missing, so drift and constraint checks did not run
    governed by: .copperhead/config.json
exit=2
```

Control, restoring the file at the configured path and changing nothing else:

```text
$ cp fixture.kicad_sch hardware/board.kicad_sch
$ copperhead --repo /tmp/ch-sync188 sync --dry-run
sync: no inconsistencies
exit=0
```

## Docs

None. The behavior is a report line and an exit code, both shown above. Happy to add the `schematic-missing` kind to the configuration reference if you want the `--json` shape written up.

---

## Invariant checklist

- [ ] N/A **Spec-gated in**: agent tool list untouched.
- [ ] N/A **Verification-gated out**: no mutation or repair path touched. `syncResolve` is unchanged.
- [ ] N/A **`check`/`verify` stays LLM-free**: `check.ts` untouched in this PR (that is #191).
- [x] **No sexp serialization**: `src/kicad/` untouched. `pinNets` stays read-only and now takes the already-resolved path.
- [ ] N/A **Sync-obligations ledger**: not touched. This is `commands/sync.ts`, not the agent ledger.
- [ ] N/A **Secrets**: no transcript, summary, or env path touched. The new line contains only a config-relative path.
- [x] **Spec coherence**: no spec-level behavior change; AC-7.3 is relied on rather than modified.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sync verification now detects configured schematics that are missing and reports their paths.
  * Missing schematics are shown as unchecked, with dependent drift, constraint, and pin checks skipped.
  * Missing schematics are excluded from automatic resolution, including after a configured file is deleted.
  * Sync reports distinguish missing schematics from other requirement violations and explain deferred resolution items.
  * Sync and check commands now fail when a required configured schematic is unavailable.
  * Interactive sync inspection now separates missing schematics from other requirement violations.

### DIFF
--- a/openspec/changes/build-copperhead-phase-1/specs/cli-surface/spec.md
+++ b/openspec/changes/build-copperhead-phase-1/specs/cli-surface/spec.md
@@ -58,6 +58,10 @@ With `--json`, commands SHALL emit machine-readable results with stable keys.
 - **WHEN** `sync` runs on a consistent repo, or runs twice in a row
 - **THEN** it exits 0 reporting no inconsistencies, and the second consecutive run makes no edits and no commit
 
+#### Scenario: Configured schematic absent (AC-7.6)
+- **WHEN** `sync` runs on a repo whose configured `schematic` names a file that is not on disk
+- **THEN** it reports the missing file and the checks it disarmed (drift, forbidden pins), exits non-zero rather than reporting no inconsistencies, and does not hand the missing file to the resolve phase
+
 ### Requirement: Dry-run mode
 With `--dry-run`, `do` SHALL print the proposed diff and write no files.
 
--- a/openspec/changes/build-copperhead-phase-1/specs/cli-surface/spec.md
+++ b/openspec/changes/build-copperhead-phase-1/specs/cli-surface/spec.md
@@ -61,6 +61,7 @@ With `--json`, commands SHALL emit machine-readable results with stable keys.
 #### Scenario: Configured schematic absent (AC-7.6)
 - **WHEN** `sync` runs on a repo whose configured `schematic` names a file that is not on disk
 - **THEN** it reports the missing file and the checks it disarmed (drift, forbidden pins), exits non-zero rather than reporting no inconsistencies, and does not hand the missing file to the resolve phase
+- **AND** any resolvable items found alongside it are reported as deferred rather than unfixable
 
 ### Requirement: Dry-run mode
 With `--dry-run`, `do` SHALL print the proposed diff and write no files.
--- a/openspec/changes/build-copperhead-phase-1/tasks.md
+++ b/openspec/changes/build-copperhead-phase-1/tasks.md
@@ -97,6 +97,7 @@
 - [x] 10.3 Implement the resolve phase: spec-gated agent run seeded with the report, truth precedence (KiCad = as-built facts, specs/budgets = requirements), single commit with DECISIONS/CHANGELOG entries (AC-7.1, AC-7.2)
 - [x] 10.4 Implement violation flagging: inconsistencies that imply a requirement violation are reported with both sides and the governing spec, never rewritten, exit non-zero (AC-7.3)
 - [x] 10.6 Implement the configured-but-absent schematic case: report the missing file and the disarmed checks, exit non-zero, never hand it to the resolve phase (AC-7.6)
+- [x] 10.7 Report resolvable items found alongside a missing schematic as deferred rather than unfixable (AC-7.6)
 - [ ] 10.5 Tests: doc-drift fixture resolved then `check` clean; dual-write gap repaired both directions; violation fixture flagged not rewritten; clean repo no-op and double-run idempotence (AC-7.5)
 
 ## 11. Submission readiness
--- a/openspec/changes/build-copperhead-phase-1/tasks.md
+++ b/openspec/changes/build-copperhead-phase-1/tasks.md
@@ -96,6 +96,7 @@
 - [x] 10.2 Implement `sync --dry-run`: print the full report (doc, claim, actual, proposed resolution), write nothing (AC-7.4)
 - [x] 10.3 Implement the resolve phase: spec-gated agent run seeded with the report, truth precedence (KiCad = as-built facts, specs/budgets = requirements), single commit with DECISIONS/CHANGELOG entries (AC-7.1, AC-7.2)
 - [x] 10.4 Implement violation flagging: inconsistencies that imply a requirement violation are reported with both sides and the governing spec, never rewritten, exit non-zero (AC-7.3)
+- [x] 10.6 Implement the configured-but-absent schematic case: report the missing file and the disarmed checks, exit non-zero, never hand it to the resolve phase (AC-7.6)
 - [ ] 10.5 Tests: doc-drift fixture resolved then `check` clean; dual-write gap repaired both directions; violation fixture flagged not rewritten; clean repo no-op and double-run idempotence (AC-7.5)
 
 ## 11. Submission readiness
--- a/openspec/specs/SPEC.md
+++ b/openspec/specs/SPEC.md
@@ -475,7 +475,7 @@ Format: Given / When / Then. "Fixture" = the open-telegraph repo (or the tiny te
 - **AC-7.3 (never resolve a violation silently)** With an inconsistency that reflects a requirement violation (e.g. a part whose leakage breaks the sleep-current budget), `sync` does **not** rewrite either side to agree; it flags the violation with both sides and the governing spec/budget, and exits non-zero.
 - **AC-7.4 (dry run)** `sync --dry-run` prints every detected inconsistency (doc, claim, actual, proposed resolution) and writes nothing (`git status` unchanged).
 - **AC-7.5 (clean and idempotent)** On a consistent repo, `sync` exits 0 with "no inconsistencies", makes no edits and no commit; running `sync` twice in a row makes the second run a no-op.
-- **AC-7.6 (configured schematic absent)** With `schematic` set in config to a path that does not exist, `sync` reports the missing file and the checks it disarmed (drift, forbidden pins) and exits non-zero; it does not report "no inconsistencies", and the missing file is not handed to the resolve phase. `check` fails the same repo state for the same reason.
+- **AC-7.6 (configured schematic absent)** With `schematic` set in config to a path that does not exist, `sync` reports the missing file and the checks it disarmed (drift, forbidden pins) and exits non-zero; it does not report "no inconsistencies", and the missing file is not handed to the resolve phase.
 
 ### AC-15 · Turn-budget continue & loop efficiency (issue #15)
 
--- a/openspec/specs/SPEC.md
+++ b/openspec/specs/SPEC.md
@@ -475,7 +475,7 @@ Format: Given / When / Then. "Fixture" = the open-telegraph repo (or the tiny te
 - **AC-7.3 (never resolve a violation silently)** With an inconsistency that reflects a requirement violation (e.g. a part whose leakage breaks the sleep-current budget), `sync` does **not** rewrite either side to agree; it flags the violation with both sides and the governing spec/budget, and exits non-zero.
 - **AC-7.4 (dry run)** `sync --dry-run` prints every detected inconsistency (doc, claim, actual, proposed resolution) and writes nothing (`git status` unchanged).
 - **AC-7.5 (clean and idempotent)** On a consistent repo, `sync` exits 0 with "no inconsistencies", makes no edits and no commit; running `sync` twice in a row makes the second run a no-op.
-- **AC-7.6 (configured schematic absent)** With `schematic` set in config to a path that does not exist, `sync` reports the missing file and the checks it disarmed (drift, forbidden pins) and exits non-zero; it does not report "no inconsistencies", and the missing file is not handed to the resolve phase.
+- **AC-7.6 (configured schematic absent)** With `schematic` set in config to a path that does not exist, `sync` reports the missing file and the checks it disarmed (drift, forbidden pins) and exits non-zero; it does not report "no inconsistencies", and the missing file is not handed to the resolve phase. Resolvable items found alongside it are reported as deferred rather than unfixable, so the run says why it stopped short of resolving them.
 
 ### AC-15 · Turn-budget continue & loop efficiency (issue #15)
 
--- a/openspec/specs/SPEC.md
+++ b/openspec/specs/SPEC.md
@@ -475,6 +475,7 @@ Format: Given / When / Then. "Fixture" = the open-telegraph repo (or the tiny te
 - **AC-7.3 (never resolve a violation silently)** With an inconsistency that reflects a requirement violation (e.g. a part whose leakage breaks the sleep-current budget), `sync` does **not** rewrite either side to agree; it flags the violation with both sides and the governing spec/budget, and exits non-zero.
 - **AC-7.4 (dry run)** `sync --dry-run` prints every detected inconsistency (doc, claim, actual, proposed resolution) and writes nothing (`git status` unchanged).
 - **AC-7.5 (clean and idempotent)** On a consistent repo, `sync` exits 0 with "no inconsistencies", makes no edits and no commit; running `sync` twice in a row makes the second run a no-op.
+- **AC-7.6 (configured schematic absent)** With `schematic` set in config to a path that does not exist, `sync` reports the missing file and the checks it disarmed (drift, forbidden pins) and exits non-zero; it does not report "no inconsistencies", and the missing file is not handed to the resolve phase. `check` fails the same repo state for the same reason.
 
 ### AC-15 · Turn-budget continue & loop efficiency (issue #15)
 
--- a/src/commands/repl-inspect.ts
+++ b/src/commands/repl-inspect.ts
@@ -11,7 +11,7 @@ import { loadConfig } from '../config.js';
 import { checkDrift } from '../memory/drift.js';
 import { loadConstraints } from '../memory/constraints.js';
 import { parseBomTable } from '../memory/bom-table.js';
-import { syncVerify, formatSyncReport } from './sync.js';
+import { syncVerify, formatSyncReport, type SyncViolationItem } from './sync.js';
 import { listSymbols, listNets, pinNets } from '../kicad/sexp.js';
 import { openspecValidate } from '../openspec/cli.js';
 import { copper, dim, ok, warn, err } from '../agent/theme.js';
@@ -47,6 +47,23 @@ async function resolveSchematic(
   return { schPath };
 }
 
+/**
+ * Headline wording for the non-auto-resolved items.
+ *
+ * The headline shows counts without the descriptions that follow it, so
+ * calling a missing file a "requirement violation" here has nothing to correct
+ * it. AC-7.3 uses that term for a design fact that breaks a budget or a spec,
+ * and the two are worth telling apart at a glance.
+ */
+function summariseViolations(violations: SyncViolationItem[]): string {
+  const missing = violations.filter((v) => v.kind === 'schematic-missing').length;
+  const requirement = violations.length - missing;
+  const parts: string[] = [];
+  if (requirement) parts.push(`${requirement} requirement violation(s)`);
+  if (missing) parts.push(`${missing} schematic missing`);
+  return parts.join(', ');
+}
+
 /** `/sync` — verify-only inconsistency report (never auto-resolves). */
 export async function formatSyncInspect(repoRoot: string): Promise<string> {
   const report = await syncVerify(repoRoot);
@@ -55,7 +72,7 @@ export async function formatSyncInspect(repoRoot: string): Promise<string> {
     !report.resolvable.length && !report.violations.length
       ? ok('  sync: clean')
       : report.violations.length
-        ? err(`  sync: ${report.violations.length} requirement violation(s), ${report.resolvable.length} resolvable`)
+        ? err(`  sync: ${summariseViolations(report.violations)}, ${report.resolvable.length} resolvable`)
         : warn(`  sync: ${report.resolvable.length} resolvable inconsistency(ies)`);
   return ['', copper('  Sync'), traceRule(12), '', headline, '', ...body.split('\n').map((l) => `  ${l}`), ''].join(
     '\n',
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -25,7 +25,14 @@ export interface SyncItem {
 }
 
 export interface SyncViolationItem {
-  kind: 'requirement-violation';
+  /**
+   * `schematic-missing` sits with the violations rather than the resolvable
+   * items on purpose: every resolvable item is handed to the LLM resolve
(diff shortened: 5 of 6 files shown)


animesh-chouhan:
_Automated `pr-review` pass (not a human review)._

## Verdict: approve with nits

The fix is correct and minimal, the two duplicated guards are consolidated the same way [#191](https://github.com/chouhanindustries/copperhead/pull/191) does it in `check.ts`, and routing the missing schematic to `violations[]` rather than `resolvable[]` is the right call for the reason the description gives: the resolve phase would hand it to the agent. Diff coverage is 100%, the suite is green, and every repo invariant I checked holds. The four nits below are about how the new condition is labelled and what else it now blocks, not about whether it works.

Reviewed in an isolated worktree at head `673134c` (merge base `a2471b3`); the main checkout was never entered and stays on its own branch.

## Metrics

```
lines: +109 / -4 (net 105) across 2 files  ·  src +25/-4, test +84/-0, docs +0/-0, spec +0/-0, other +0/-0
new vs net: 1 new files (0 in src), 1 modified; 0 new src lines (the net-new surface)
tests: +84 test lines; suite not measured -> 778/20/0 (pass/skip/fail, base -> head)
diff coverage: 100% of changed executable src lines exercised (12/12, measured via vitest v8)
```

Method: measured in the review worktree; areas from `git diff --numstat` against the merge base; suite via `npx vitest run --reporter=json` at head; coverage via vitest v8 intersected with the changed src lines from `git diff -U0`; CI state from `gh pr checks`; `mergeable=MERGEABLE`, `draft=false`. Base suite not run (base CI is green and no base-sensitive metric depends on it), so the base column of the suite line is unmeasured, hence "suite not measured -> 778/20/0". Uncovered changed lines: none, so there is no untested-surface list to reconcile below.

## Invariant checks

| Invariant | Result |
| --- | --- |
| Spec-gated in (`edit_file`/`write_file` structurally absent) | pass, agent tool list untouched |
| Verification-gated out (ERC/DRC, repair, rollback) | pass, `syncResolve` and the loop untouched; the change is stricter, not weaker |
| `check`/`verify` stays LLM-free and network-free | pass, `check.ts` untouched and nothing new is imported into its graph |
| No sexp serialization | pass, `pinNets` stays read-only and now just takes the pre-resolved path |
| Sync-obligations ledger | pass, not touched |
| Secrets (`sk-` redaction, `.env`, `.copperhead/runs/`) | pass; the new description carries only a config-relative path, and it never reaches a prompt or transcript because violations exit 2 before the resolve phase |

## Findings

### 1. low: a missing schematic now blocks the resolve phase for schematic-independent items

[src/commands/sync.ts:56-62](src/commands/sync.ts#L56-L62), consumed at [src/cli.ts:330-333](src/cli.ts#L330-L333)

Putting the item in `violations[]` does more than set the exit code. `cli.ts` exits 2 on any violation before it reaches `syncResolve`, so a repo whose `config.schematic` points at a path that does not exist can no longer auto-resolve the inconsistencies that have nothing to do with the schematic: the `coverage` items for a missing `DECISIONS.md` or `CHANGELOG.md`, the `dual-write` audit, the `pins-h` comparison, and the `openspec` validation failure all still get detected at lines 78-151 and then all get skipped.

Failure scenario: a repo with a renamed schematic and a missing `docs/DECISIONS.md`. Before this change, `sync` scaffolded the doc and exited 0 (having silently verified nothing about the schematic, which is the bug being fixed). After it, `sync` exits 2 and the DECISIONS.md item is reported but never resolved until a human fixes the config path first.

This is defensible, arguably it is the point, but it is a second behavior change and the description only claims the exit code "for free". Either state it in the description as intended, or note it in the report line so the user knows the other items are deferred rather than unfixable.

### 2. low: the new condition is rendered and counted as a "requirement violation"

[src/commands/sync.ts:181](src/commands/sync.ts#L181), [src/commands/repl-inspect.ts:58](src/commands/repl-inspect.ts#L58)

`formatSyncReport` prints the whole `violations[]` array under `N REQUIREMENT VIOLATION(S) (not auto-resolved; human decision needed)`, and the REPL `/sync` headline reads `sync: 1 requirement violation(s), 0 resolvable`. A file that is absent from disk is not a requirement violation, it is a check that could not run, and the distinction matters here because SPEC.md AC-7.3 defines requirement violation narrowly (a part whose leakage breaks a budget, the design disagreeing with a spec).

The description argues the heading is still accurate because the item genuinely needs a human. That holds for the heading's parenthetical but not for the noun, and the REPL headline shows only the count, without the clarifying description text that rescues it in the full report. Since the `kind` union now distinguishes the two, both sites can split the count, or the heading can be widened to something like "NOT AUTO-RESOLVED (human decision needed)".

Neither site has a test covering the new kind (both are pre-existing lines whose output changes), which is why 100% diff coverage does not catch this.

### 3. low: the `--json` shape for this condition diverges from the `check` half

[src/commands/sync.ts:35](src/commands/sync.ts#L35) vs `check.ts` in [#191](https://github.com/chouhanindustries/copperhead/pull/191)

The two halves of #188 represent the same fact two different ways: `check` adds a top-level `schematicMissing: string | null` to `CheckResult`, while `sync` widens `SyncViolationItem['kind']` and buries the path inside a prose `description`. A `--json` consumer that wants "is the configured schematic missing" has to read a field in one command and pattern-match a union member plus parse a sentence in the other, and only `check` exposes the path as data.

Cheapest fix that keeps both designs: add the path as its own field on the violation item (an optional `schematic?: string`, or reuse `governedBy`'s sibling slot) so `sync --json` exposes it structurally, matching what `check --json` gives. Worth settling while both PRs are open, since they land as one user-visible fix.

### 4. low: SPEC.md AC-7 gains a non-zero exit condition with no scenario

`openspec/specs/SPEC.md:471-478`

AC-7 enumerates the `sync` outcomes (AC-7.1 resolve drift, AC-7.2 dual-write, AC-7.3 never resolve a violation, AC-7.4 dry run, AC-7.5 clean and idempotent). This PR adds a new way for `sync` to exit non-zero that none of them describes. The description's claim that AC-7.3 already covers it is close but not exact: AC-7.3 is about requirement violations specifically, which is the same gap as finding 2.

A one-line AC-7.6 ("a configured schematic that is absent from disk is reported and exits non-zero, and the schematic-gated checks are reported as not run") would keep code and spec moving together per CLAUDE.md, and would give the sibling `check` fix a spec anchor too. Judgment call, but a code-only PR that adds a user-visible exit path is exactly the case the repo's convention is aimed at.

## Notes, not findings

- The `if (schematicUsable && config.schematic)` at line 64 looks redundant but is not: the aliased-condition narrowing gives you `schematicPath`, not `config.schematic`, and `checkDrift` needs the latter as a `string`. Leave it.
- `config.schematic === ''` still falls through silently (empty string is falsy, so no violation is raised). That is pre-existing `loadConfig` behavior and not a regression from this PR.
- `config.board` configured but absent has the same class of gap, but `syncVerify` does not read the board at all, so nothing is introduced here. Out of scope for #188.
- The test file is deterministic: no timing, no network, real fixture, `tempFixtureRepo` cleanup in a `finally`. The two negative controls passing on both sides of the change is the right way to have written it.

## Fix prompt

````text
Resolve the following review findings on PR #226 (branch fix/sync-missing-schematic) of the copperhead repo. All four are low severity; none blocks merge, so if you disagree with one, say why rather than forcing a change.

CONTEXT
PR #226 makes `copperhead sync` report a `config.schematic` that names a file which is not on disk, as a `violations[]` item with kind `schematic-missing`, instead of silently skipping drift and the forbidden-pin check and exiting 0. It is the `sync` half of issue #188; PR #191 is the `check` half and is still open. Files touched: src/commands/sync.ts, test/sync-missing-schematic.test.ts.

FINDING 1 (low): a missing schematic now blocks the resolve phase for schematic-independent items
src/commands/sync.ts:56-62, consumed at src/cli.ts:330-333.
Because the item lands in violations[], cli.ts exits 2 before syncResolve runs. A repo with a missing schematic path can therefore no longer auto-resolve the items that do not depend on the schematic at all: the `coverage` items for a missing docs/DECISIONS.md or docs/CHANGELOG.md, the `dual-write` audit, the `pins-h` comparison, and the `openspec` validation failure. They are all still detected at sync.ts:78-151, then all skipped.
Repro: a repo whose config.schematic points at a renamed path AND whose docs/DECISIONS.md is missing. Before: sync scaffolded the doc and exited 0. After: exit 2, the doc item is printed but never resolved.
Fix: this is probably intended, so either state it explicitly in the PR description, or add a line to the report when resolvable items coexist with the schematic-missing violation, telling the user those items are deferred until the schematic path is fixed rather than unfixable.

FINDING 2 (low): the new condition is rendered and counted as a "requirement violation"
src/commands/sync.ts:181 and src/commands/repl-inspect.ts:58.
formatSyncReport prints all of violations[] under "N REQUIREMENT VIOLATION(S) (not auto-resolved; human decision needed)", and the REPL /sync headline reads "sync: 1 requirement violation(s), 0 resolvable". A file that is absent is not a requirement violation; SPEC.md AC-7.3 defines that term narrowly (a design fact that breaks a budget or a spec). The full report's description text softens this, but the REPL headline shows only the count with no description.
Fix: now that kind distinguishes the two, split the counts at both sites, or widen the heading to something like "NOT AUTO-RESOLVED (human decision needed)". Add a test for whichever you choose; neither site currently has coverage of the new kind, which is why 100% diff coverage did not catch it.

FINDING 3 (low): the --json shape for this condition diverges from the check half
src/commands/sync.ts:35 versus check.ts in PR #191.
#191 adds a top-level `schematicMissing: string | null` to CheckResult. #226 widens SyncViolationItem['kind'] and leaves the path only inside a prose `description`. A --json consumer must read a field in one command and pattern-match a union member plus parse a sentence in the other, and only check exposes the path as data.
Fix: expose the path structurally on the sync violation item too, e.g. an optional `schematic?: string` on SyncViolationItem, set alongside kind: 'schematic-missing'. Keep the description sentence as is. Coordinate with #191 so both halves of #188 land with a consistent --json contract.

FINDING 4 (low): SPEC.md AC-7 gains a non-zero exit condition with no scenario
openspec/specs/SPEC.md:471-478.
AC-7.1 through AC-7.5 enumerate the sync outcomes and none covers a configured-but-absent schematic. The PR claims AC-7.3 already covers it, but AC-7.3 is about requirement violations specifically, which is the same gap as finding 2.
Fix: add a one-line AC-7.6 along the lines of "a configured schematic that is absent from disk is reported and exits non-zero, and the schematic-gated checks are reported as not run". Per CLAUDE.md, SPEC.md and the active change artifacts move together, so if this touches openspec/changes/build-copperhead-phase-1/, update those too and re-run `openspec validate build-copperhead-phase-1`.

DO NOT CHANGE
- Keep the item in violations[], not resolvable[]. The reasoning in the PR description is correct: every resolvable item is handed to the LLM resolve phase, and a schematic that is not on disk is not something an agent should write its way out of.
- Keep the schematicUsable consolidation and keep `if (schematicUsable && config.schematic)` at sync.ts:64. It looks redundant but is not: the aliased-condition narrowing gives you schematicPath, not config.schematic, and checkDrift needs the latter typed as string.
- Do not widen scope to config.board; syncVerify does not read the board.

REPO INVARIANTS (must still hold after your changes)
1. Spec-gated in: edit_file/write_file stay structurally absent from the agent tool list until an OpenSpec proposal validates. Gate by omission, never by prompt text.
2. Verification-gated out: every mutation still ends in ERC passing (and DRC if the board changed), repairing up to maxRepairCycles, then rolling back to the git snapshot.
3. check/verify stays LLM-free and network-free: nothing in src/commands/check.ts or its import graph may touch a provider, an API key, or the network.
4. No sexp serialization: src/kicad/ stays read-only; KiCad files are edited only by anchored exact-match text replace.
5. Sync-obligations ledger: post-tool-call hooks keep feeding it, and commit keeps refusing while any obligation is open.
6. Secrets: transcripts and summaries redact sk-[A-Za-z0-9_-]+ at write time; keys live only in env vars; .gitignore keeps .env and .copperhead/runs/.

VERIFY
npm run typecheck
npm run build
npx vitest run test/sync-missing-schematic.test.ts test/gating-sync.test.ts test/check-missing-schematic.test.ts
npx vitest run
openspec validate build-copperhead-phase-1   # only if you touched openspec/
Report the actual command output. If a test fails, say so with the output rather than describing it as passing.
````

**Comment by adhavan18** [comment]
all four addressed in `1b4c19a`. taking them in the order that changed the most code.

## 2, the naming

you were right and I had talked myself into the wrong side of it. I argued the heading was rescued by the parenthetical, but the REPL headline shows counts with no description at all, and I had not looked at that site. both fixed:

```
before:  1 REQUIREMENT VIOLATION(S) (not auto-resolved; human decision needed):
         - schematic configured at hardware/board.kicad_sch but the file is missing ...

after:   1 NOT AUTO-RESOLVED (human decision needed):
         - [schematic-missing] schematic configured at hardware/board.kicad_sch but the file is missing ...
```

each line now carries its kind, and `formatSyncInspect` splits the counts through a small `summariseViolations` helper, so `/sync` reads `2 requirement violation(s), 1 schematic missing` rather than lumping them.

## 1, the deferred items

intended, and you are right that "for free" was the wrong way to describe it. rather than only fixing the description I added the report line, since the description is not what the user reads:

```
1 resolvable inconsistency(ies):
- [coverage] docs/DECISIONS.md: claims "exists (transparency layer)" but actual is "missing"
    resolution: scaffold it (copperhead init creates it)
1 NOT AUTO-RESOLVED (human decision needed):
- [schematic-missing] schematic configured at hardware/board.kicad_sch but the file is missing, so drift and constraint checks did not run
    governed by: .copperhead/config.json
note: the 1 resolvable item(s) above are deferred, not unfixable. sync stops before the resolve phase while anything is unresolved above; clear that first, then re-run.
```

worth noting this was already true for requirement violations before this PR; a missing schematic just makes it reachable far more often. the note is not conditional on the schematic case for that reason.

## 3, the --json shape

agreed, and this was the best of the four. `SyncViolationItem` gains an optional `schematic`:

```json
{
  "kind": "schematic-missing",
  "description": "schematic configured at hardware/board.kicad_sch but the file is missing, so drift and constraint checks did not run",
  "governedBy": ".copperhead/config.json",
  "schematic": "hardware/board.kicad_sch"
}
```

so a consumer reads a field in both commands instead of a field in one and a sentence in the other. #191 needs no change: its top-level `schematicMissing` was already the structural form, this half was the one out of step.

## 4, the spec

added AC-7.6 to SPEC.md, plus the matching scenario in the phase-1 change's `cli-surface` delta and a task 10.6. **`openspec validate` I could not run**: `npx openspec` reports "could not determine executable to run" on this machine, so that claim is unverified by me rather than verified. the edits follow the shape of the AC-7.4/7.5 scenarios exactly, but worth a real validate on your side.

## verification

typecheck clean, build clean, 10/10 in the sync suite (4 new). `test/repl.test.ts` has 3 failures both before and after my change; I re-ran to be sure, since one run came back clean and made it look like I had fixed something. they are the known Windows path/timeout cases and are flaky, not affected by this.

## what I did not change

kept it in `violations[]`, kept `if (schematicUsable && config.schematic)`, and left `config.board` alone, per your note. the empty-string `config.schematic` gap you spotted is real but is `loadConfig` behaviour and predates this, so I left it; happy to file it separately if you want it tracked.

adhavan18:
@coderabbitai you're right, and the split was backwards in a way worth naming.

AC-7.6 ended with "`check` fails the same repo state for the same reason", which this change doesn't implement — that's #191. removed in `0508306`, so AC-7.6 and the delta scenario now both describe `sync` only, one to one.

the other half of it: #191 changes `check`'s behaviour and had **no** SPEC.md entry, no delta scenario and no task at all. so this PR documented what it doesn't do while #191 did what it doesn't document. pushed `cfff699` there adding AC-2.6, a matching scenario under the `check` requirement, and task 7.4.

numbered it AC-2.6 under `check` rather than extending AC-7.6, so the two are independent and correct in either merge order.

`tsc` clean, `lint:md` clean, 10/10 here and 5/5 on #191.

adhavan18:
@coderabbitai fair, pushed `0a26baf`.

the code already reports the deferred items — `formatSyncReport` emits `note: the N resolvable item(s) above are deferred, not unfixable` — but AC-7.6 didn't say so, so the spec promised less than the run delivers. that's the same class of gap as the `check` sentence, just in the other direction.

AC-7.6 now ends with it, with a matching `AND` clause on the delta scenario and task 10.7.

`lint:md` clean, 10/10 tests still pass.

adhavan18:
Already addressed in `0a26baf` (Aug 15): AC-7.6 now states resolvable items alongside the missing schematic are reported as deferred rather than unfixable, with matching updates to the phase-1 cli-surface spec and tasks. Thread was just never marked resolved.

adhavan18:

Issue #208, opened 2026-08-08
Checkpoint-commit accepted drafts during stage 4, then repair on top — users get output continuously and rollback stops destroying work


animesh-chouhan:
## Why

Stage 4 is all-or-nothing today: the stage commits only when ERC + drift + legibility are all clean, and a failed attempt rolls the tree back (stash + clean). The lemondrop campaign (`run-logs/2026-08-07T17-52-03`) shows the cost: attempt-05 drafted the full 160-part sheet, was ERC-clean at one point, found and fixed two genuine electrical bugs — and the repo ended the day empty. The work survived only in a git stash (tracked files) while the agent's untracked handoff doc was destroyed outright. Retries restarted from zero, twice. The harness tracks this class as R6/I16/R9.

## Proposal

Commit an initial draft as soon as one is accepted, then let the repair loop commit on top:

1. **Flagged WIP checkpoints on the branch**: after each accepted `draft_schematic` (and each stage-attempt boundary), commit with an unmistakable marker, e.g. `copperhead: WIP schematic draft — ERC FAILING, 4 findings, NOT FOR FABRICATION`. The existing stage-complete commit remains the only certification. Rollback becomes reset-to-last-checkpoint instead of stash-and-clean, so a retry starts from the best known state by construction and the retry boundary can no longer discard the artifacts it exists to preserve.
2. Alternative, more conservative: a **parallel WIP ref** (`refs/copperhead/wip/<stage>`) so the main branch stays certified-only while every accepted draft is still committed, visible, and diffable.

## Invariant analysis

Verification-gated-out ("every mutation ends ERC-clean or rolls back") exists so the repo never holds an uncertified schematic mistakable for approved output. Option 1 preserves the guarantee by moving the certification claim from "it is in git" to "the stage-complete commit exists" — which the pipeline and harness already distinguish (`definition_of_done.every_stage_committed`) — with the WIP marker making status unmistakable to humans and tools. Option 2 preserves it verbatim. Either is a spec-level change to the rollback contract and should go through an OpenSpec proposal.

## What users gain

A reviewable KiCad sheet minutes into stage 4 instead of only at stage completion, monotonic progress across retries, and an end to the class of loss where hours of accepted work exist only in a stash (or, for untracked files, nowhere).

PR #243, opened 2026-08-22
feat(create): checkpoint a failed stage attempt to a side ref

adhavan18:
## What

A failed stage attempt now records its work to `refs/copperhead/wip/<stage>` before the rollback, so accepted work stays recoverable instead of being destroyed.

## Why

Closes the loss described in #208. Today a failed attempt rolls the tree back and the work survives only as a stash of *tracked* files: anything the agent left untracked goes outright. The lemondrop campaign lost a 160-part sheet that had been ERC-clean, and its handoff doc existed nowhere afterwards. Retries restart from zero.

## Design

This is the issue's second, conservative option. The branch is untouched, so verification-gated-out holds verbatim: only stage-complete commits land on it and the checkpoint is reachable solely by its own ref, so nothing uncertified can be mistaken for approved output.

`checkpoint()` writes to a temporary index (`GIT_INDEX_FILE`), seeded from HEAD, then `commit-tree` and `update-ref`. That captures tracked changes and untracked files together with `.gitignore` still respected, and leaves the caller's staged state alone, which matters because this runs mid-pipeline rather than between commands.

It returns `null` rather than throwing on any failure. A checkpoint is a safety net, and one that fails the stage it exists to protect is worse than none.

## Spec / OpenSpec

The issue says either option is a spec-level change to the rollback contract and wants an OpenSpec proposal. I have not added one, because on this option no documented behaviour changes: the rollback still happens, the tree still returns to the snapshot, and the branch still carries only certified commits. The side ref is additive, in the same spirit as the existing `preserveFailedRun` stash.

If you read it as spec-level anyway, say so and I will add the proposal, delta specs and tasks entry before you spend review time on the code.

## Testing

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | 753 passed, 25 failed, 20 skipped |
| Markdown lint | `npm run lint:md` | pass (0 issues) |

New: `test/wip-checkpoint.test.ts`, 6 cases.

The one that matters reproduces the loss from the issue: a tracked edit plus an untracked handoff doc, then a real `restore()`. The rollback deletes the untracked file, and the checkpoint still holds both.

The others pin that HEAD, the branch and the working tree are untouched, that the caller's staged state survives, that the checkpoint is not reachable from the branch, that a clean tree records nothing, that each stage keeps its own ref advancing to the latest attempt, and that an unusable repo yields `null` rather than an exception.

Pre-existing failures unrelated to this PR: the 25 above are identical on `main` in my environment (Windows, no `kicad-cli`). Baseline is 25 failed / 747 passed / 792 total; this branch is 25 failed / 753 passed / 798 total, i.e. the same failures plus the 6 new tests.

### Manual test log

`create` needs `kicad-cli` and a provider credential, neither available here, so the checkpoint path is exercised directly against a real repo rather than through a full pipeline run:

```text
$ git show --name-only --format= <checkpoint>
HANDOFF.md
docs/BOM.md

$ git log --format=%H        # the branch does not contain it
<base commit only>

$ git rev-parse refs/copperhead/wip/schematic
<checkpoint>
```

## Docs

None yet. If the ref naming is right I will follow up with a line in the workflows docs on recovering from a failed stage.

---

## Invariant checklist

- [x] **Spec-gated in**: unchanged, no tool surface touched.
- [x] **Verification-gated out**: preserved verbatim. The rollback still runs, the branch still takes only stage-complete commits, and the checkpoint lives outside it.
- [x] **`check`/`verify` stays LLM-free and network-free**: N/A, nothing under `src/commands/check` touched.
- [x] **No sexp serialization**: N/A, the KiCad parser is untouched.
- [ ] **Sync-obligations ledger**: N/A, no tool-call hooks changed.
- [x] **Secrets**: N/A. The checkpoint commits the same paths `commitAll` would, with `GIT_ADD_EXCLUDES` applied.
- [x] **Spec coherence**: see the OpenSpec note above.

## Summary by CodeRabbit

* **New Features**
  * Unsuccessful stage attempts now save work-in-progress checkpoints for later inspection or retry.
  * Checkpoints preserve both tracked and untracked changes without altering the active branch or working files.
  * Each stage maintains its own checkpoint, making prior attempts easier to review.

* **Reliability**
  * Checkpoint creation failures are handled gracefully, allowing the workflow to continue without interruption.

### DIFF
--- a/src/agent/loop.ts
+++ b/src/agent/loop.ts
@@ -22,7 +22,16 @@ import { collectRunMeta, renderCliHeader, type RunMeta, type RunMetaInput } from
 import { plainRenderer, fmtDuration, fmtTokens, type ProgressRenderer } from './render.js';
 import { styleHeaderLines } from './theme.js';
 import { ObligationsLedger } from './ledger.js';
-import { gitPreflight, isDirty, snapshot, restore, commitAll, changedFiles, preserveFailedRun } from '../util/git.js';
+import {
+  gitPreflight,
+  isDirty,
+  snapshot,
+  restore,
+  commitAll,
+  changedFiles,
+  preserveFailedRun,
+  checkpointTree,
+} from '../util/git.js';
 import { withRetry, isRateLimit, sessionLimit } from '../util/retry.js';
 import { openspecArchive } from '../openspec/cli.js';
 import { existsSync } from 'node:fs';
@@ -75,6 +84,16 @@ export interface RunResult {
   transcriptDir: string;
   filesTouched: string[];
   commit: string | null;
+  /**
+   * A checkpoint commit captured from the dirty tree just before an
+   * unsuccessful run rolled it back (issue #208), or null on a successful run
+   * or when there was nothing to capture. Not yet pointed at by any ref: the
+   * caller decides where, since only it knows the stage (see
+   * util/git.js#recordWipRef). Capturing here, before restore(), is what
+   * makes this the failed attempt's own work rather than whatever the tree
+   * happens to hold after rollback.
+   */
+  failedAttemptCommit: string | null;
   /** Cost/telemetry for this run. Surfaced by the create pipeline's per-stage
    *  cost table (5.2) so the expensive stages are obvious across runs. */
   stats: RunStats;
@@ -380,6 +399,11 @@ async function runWithProviders(opts: RunOptions, providers: Set<Provider>): Pro
     // it, so a budget-exhaustion (or any) failure is recoverable (issue #15).
     const preserved = await preserveFailedRun(repoRoot, ctx.runId);
     if (preserved) await transcript.event('work-preserved', { stash: preserved });
+    // Capture the dirty tree as a checkpoint commit before the rollback below
+    // destroys it, so the create pipeline can point a stage's WIP ref at the
+    // failed attempt itself rather than the clean, already-restored tree it
+    // would otherwise see (issue #208).
+    const failedAttemptCommit = await checkpointTree(repoRoot, `copperhead: WIP checkpoint — NOT CERTIFIED (${reason})`);
     // The rollback itself can fail (git in a bad state). That must not become
     // an unhandled throw that skips run-end and summary.md — the summary is
     // most valuable exactly when the tree is left in an unknown state.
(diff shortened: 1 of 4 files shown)


adhavan18:
Fixed all three in `3c818b1`.

**Checkpoint timing (create.ts:928).** Confirmed by reading the actual control flow: `runAgentLoop`'s `fail()`/`refused` paths call `restore()` — a hard reset — before returning, so by the time `create.ts` called `checkpoint()` afterward, there was nothing of the failed attempt left; it recorded the clean, already-restored tree. Split `checkpoint()` into `checkpointTree()` (captures the dirty tree, returns a sha, touches no ref) and `recordWipRef()` (points a stage's ref at an already-captured commit). `runAgentLoop` now calls `checkpointTree()` in both rollback paths *before* `restore()` runs, and returns the sha as `RunResult.failedAttemptCommit`; `create.ts` records the ref from that once it knows the stage, instead of re-deriving from a tree that's already gone.

**`.gitignore` mutation (git.ts:365-387, git.ts:368-374 — same root cause).** `ensureIgnored()` wrote to the real `.gitignore` to keep `.history/` out. `checkpointTree()` now strips `GIT_ADD_EXCLUDES` from the temporary index directly after `git add -A`, so nothing touches the worktree, and it treats a resulting tree identical to HEAD's as nothing worth capturing — closing the "checkpoint made of nothing but the `.gitignore` edit" case too.

**Temp-dir allocation (git.ts:365-387).** `mkdtemp` moved inside the `try`; only cleaned up in `finally` when it actually succeeded.

4 new tests: the exact timing bug reproduced and fixed in one test (capture-before-restore vs. the old capture-after-restore order), the temp-dir failure path, and both `.gitignore`-exclusion cases (including the "only excluded content was dirty" edge case my first attempt at that test got wrong — it produced a same-tree commit rather than skipping entirely, so I tightened `checkpointTree` to compare against `HEAD^{tree}` and return null there too). 10/10 pass, typecheck clean, full suite otherwise unchanged from `main`.

animesh-chouhan:
Cross-reference for #258 (run protocol stream): this PR is a concrete case for it. The checkpoint is reachable solely by its own ref by design, which is right for the verification gate, and it also means nothing outside the process can discover that the work survived. A consumer watching a failed stage has no way to tell someone their 160-part sheet is still there.

No change requested. When the protocol lands, a `notice` carrying the ref, plus `resumeSafe` and the last completed stage in the result document, is what turns this from recoverable-in-principle into recoverable-in-the-UI.

adhavan18:

Issue #252, opened 2026-08-26
Layout stops at a deterministic grid: pick the placement/routing engine boundary (DSN/SES vs circuit-json) and adopt an external router


animesh-chouhan:
### The problem

The board half of `create` now has a deterministic front end and nothing behind it.

`src/kicad/board.ts` (in flight on `test/real-netlist-corpus`) closes #227: before every attempt of stage 5, `ensureBoardPopulated` exports the schematic's netlist with `kicad-cli sch export netlist --format kicadsexpr`, resolves each part to an installed footprint, shelf-packs the footprints inside an Edge.Cuts rectangle sized to fit, assigns pad nets, and verifies the board loads and passes DRC. The agent no longer refuses for lack of readable pad geometry.

```
schematic.intent.json
     |  draft engine + emit.ts
.kicad_sch
     |  kicad-cli sch export netlist (kicadsexpr)
netlist (parts, nets, Footprint fields)
     |  resolveFootprint: schematic field > symbol default > generic table > ki_fp_filters
     |  shelfPack (area-descending, 1.5 mm gaps)
     |  instantiateFootprint (text surgery on the library .kicad_mod)
.kicad_pcb   all on F.Cu, pads carry nets, zero tracks, zero vias
     |  kicad-cli pcb drc  ->  clean, N unrouted
stage 5 agent: nudges (at X Y) lines, writes LAYOUT.md "## Draft quality"
     |
stage 6: gerbers, drill, DXF, STEP, BOM.csv
```

Two gaps follow from that picture:

**Placement is a grid, not a layout.** `shelfPack` orders by footprint area and packs left to right. It knows nothing about connectivity, so a decoupling cap can land on the far side of the board from its IC. The stage prompt asks the agent to fix this by editing `(at X Y)` lines a few parts at a time with `run_drc` between batches, and explicitly offers the grid as a fallback when that fails. On any real part count that fallback is what ships.

**Routing does not exist and no gate asks for it.** `normalizeReport` deliberately keeps `unconnected_items` in a separate `unrouted` bucket that never counts against `ok`, which is right for a draft. But `grep -rn "unrouted" src/` returns only `check.ts` and the stage-5 prompt: the fab release gate in `src/kicad/fab.ts` and `copperhead export` never look at it. A board with zero copper passes stage 6 and produces gerbers. Related: #228.

So the question this issue is for: **what is the engine boundary, and which engines go behind it.**

### Findings on the integration boundary

[...]

PR #253, opened 2026-08-27
Implement end-to-end PCB layout and routing for #252

ShashankT-ECE:
## What

Implements the end-to-end PCB layout and routing pipeline described in #252, from schematic netlist export through deterministic board population, placement, optional external routing, routed-board import, and verification.

The agent can now populate a real KiCad board from the schematic, resolve installed footprints, place components, optionally route through Freerouting, import the routed result, and verify the resulting board with DRC and the routing-completeness gate.

## Why

Issue #252 requires a clear placement/routing engine boundary and a functional path from schematic-derived connectivity to a routed PCB. This PR closes that implementation gap while keeping routing failures explicit and avoiding false routing success when the external router is unavailable.

## Design

The pipeline is:

`kicad-cli` schematic netlist export → netlist parsing → installed footprint resolution → board population → deterministic placement → Specctra DSN generation → optional Freerouting → SES parsing → routed track/via import → DRC verification → routing-completeness gate.

The implementation adds a real board model populated from schematic connectivity and installed KiCad footprint data rather than a hand-authored board model.

Placement and routing are separated behind `PlacementEngine` and `RoutingEngine` interfaces. The current placement implementation is connectivity-aware: a deterministic Prim-style ordering over an inverse-fanout-weighted net graph clusters components that share a low-fanout signal net before grid packing, so the router's nets stay short instead of spanning the board. True 2-D cluster-growth placement remains the remaining #141 generation work.

Freerouting failures are surfaced explicitly and a board is never marked routed when routing fails. After routing, hard DRC violations cause the result to roll back to the placed board.

The routing-completeness gate distinguishes unrouted connections from hard DRC violations and is reused by the fabrication/export path.

## Spec / OpenSpec

This PR changes only the design and tasks artifacts of the existing `openspec/changes/add-fab-release-gate/` change. It does not modify SPEC.md, `proposal.md`, or any delta spec.

- `design.md`: adds an "Issue #252 follow-on" note documenting the placement/routing engine boundary and the gate wiring.
- `tasks.md`: checks off the routing-completeness gate (task 1.1) and the `--fab`/`--strict` CLI wiring (tasks 3.1, 3.2, 3.3) as delivered via #252. The remaining gate checks (BOM readiness 1.2, schematic-to-PCB match 1.3, output freshness 1.4) stay unchecked.

The board population, placement, DSN/SES bridge, and Freerouting adapter implement #252 directly and do not have a dedicated OpenSpec change. Connectivity-aware *ordering* is now implemented; true 2-D cluster-growth placement remains the remaining #141 work.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | 840 passed, 23 skipped, 4 pre-existing environment failures |

The four failing reference-board tests reproduce on clean `origin/main` and are caused by the installed KiCad 9.0.9 environment failing to load those reference schematics. They are unrelated to this PR.

#252-specific tests: 35 passed, 1 skipped (the gated Freerouting integration test).

### Live Freerouting validation

The Freerouting integration test (`test/layout-freerouting.test.ts`) now runs against a real Freerouting release and passes end-to-end: Copperhead DSN generation, real Freerouting process, real SES output, Copperhead SES parsing, routed PCB import, with the resulting board DRC-clean (0 unrouted nets, 0 hard violations).

- JRE: Temurin OpenJDK 25.0.4.1+1 (LTS)
- Freerouting: v2.3.0 (official `freerouting-2.3.0.jar`)
- Enabled by `COPPERHEAD_TEST_FREEROUTING_JAR` (test gate) + `COPPERHEAD_FREEROUTING_JAR` (jar path) + `java` on PATH.

### SES coordinate-scale fix

Live validation surfaced a real SES-parsing bug: Freerouting's `SesWriter` emits session coordinates in resolution units (0.1 µm per unit for `(resolution um 10)`, i.e. 10 000 units/mm), but the parser read them as raw micrometres, scaling every routed track/via 10× and leaving routes dangling off-pad. `unitsPerMmOf` now applies the resolution factor, and the captured `sample.ses` fixture was corrected to Freerouting's real output format. DSN emission was already correct (Freerouting reads raw-µm DSN on ingest).

The alternative-router evaluation below surfaced the mirror-image interop bug in the other direction: the `freeroute` Python port writes SES coordinates in raw micrometres while still declaring `(resolution um 10)`, so `parseSes` now also accepts `{ resolutionUnits: false }` for raw-unit writers (the Freerouting default is unchanged).

### Real-design routing (complex_hierarchy)

`scripts/route-real-design.ts` routes the real `complex_hierarchy` design (68 parts, 50 nets, all through-hole) and compares it against its golden `.kicad_pcb` (365 tracks, 0 vias, 0 DRC, 0 unrouted). The connectivity-aware placer (above) clusters the parts, so Freerouting finishes with 2 vias and 374 track segments — within 2% of the golden — versus 27 vias and 642 segments on the old blind grid. **It still does not route DRC-clean**: Freerouting leaves 26 error-severity violations (solder-mask bridges, optimizer shorts, clearance) and one unrouted net on the finished board, because it does not model the solder-mask aperture on a dense through-hole board. `runBoardLayout` therefore correctly rolls the result back to the placed (unrouted) board, so `res.routed` is `false` for this design and the board is never reported as routed. The DRC ceiling is Freerouting's dense-THT model, not the placer; the one unrouted net is the remaining 1-D-ordering scatter that true 2-D placement (#141) addresses.

### Connectivity-aware placement (#141 ordering)

`src/kicad/layout/placement.ts` now orders footprints by schematic connectivity (a deterministic Prim-style growth over an inverse-fanout-weighted net graph) before grid packing, rather than a connectivity-blind area sort. Measured on `complex_hierarchy`: net-locality spread 29.2 → 8.7 (3.4×), vias 27 → 2, track segments 642 → 374 (golden 365). Covered by `test/layout-placement-connectivity.test.ts` (5 tests: consecutive placement, inverse-fanout weighting, determinism, net-less fallback, no-overlap). A net-less board still places exactly as before (area-descending fallback).

### Alternative router evaluation (#253 review)

The founder asked to actually run alternative open-source routing packages, not merely document them. `scripts/route-compare.ts` (new, `npm run route:compare`) emits the DSN once and routes it through each engine, then DRC-checks every import. Executed against `complex_hierarchy` (golden: 365 tracks, 0 vias, 0 DRC, 0 unrouted):

| Router | Routed | Unrouted | Hard DRC | Vias | Runtime |
| --- | --- | --- | --- | --- | --- |
| Freerouting 2.3.0 (jar, production) | 50/50 | 0 | 25 | 32 | ~100 s |
| freeroute 2026.7.14 (grid) | 50/50 | 0 | 1255 | 102 | 23 s |
| freeroute 2026.7.14 (exact) | — | 40 | 646 | 203 | 18.5 s |
| freeroute 2026.7.14 (room) | — | — | — | — | >15 min (stopped) |
| KiCadRoutingTools 0.21.3 | 50/50 | 0 | 0 | 190 | ~58 s |
| kicad-tools 0.20.0 | 45/47 | — | — | — | >5 min (stopped) |

- **Freerouting** remains the best DSN/SES engine (25 vs 1255 violations for its own Python port).
- **KiCadRoutingTools** is the strongest for raw completion/cleanliness but achieves 0 violations by self-selecting finer geometry and writing a `.kicad_pro` that lowers the clearance rule and demotes `solder_mask_bridge` to "ignore" — so it is **not adopted**.
- **freeroute** is the only JRE-free DSN/SES drop-in but is dramatically worse on a dense through-hole board.
- **kicad-tools**' C++ backend gives up on dense nets and falls back to slow Python.
- Investigated but **not executable here**: `topola` (Rust, no prebuilt binary), `routing-rs` (Rust, self-declared non-functional), `ACORN` (missing `libgd-dev`/`libpng-dev` headers), `qautorouter` (Qt4, abandoned). Full commands, versions, and tradeoffs are in `manual-tests/ROUTING-ENGINES.md`.

### Manual test log

The manual-tests sandbox CLI exercise (see `manual-tests/README.md`) was not performed in this verification, so no `create`/`edit` sandbox variant is claimed. What was actually run:

```text
$ npm run typecheck
PASS

$ npm run build
PASS

$ npm test
840 passed, 23 skipped, 4 pre-existing environment failures

$ npm test -- test/board-layout.test.ts test/layout-dsn.test.ts test/layout-freerouting.test.ts test/layout-placement.test.ts test/fab-routing-gate.test.ts test/report.test.ts test/fab.test.ts
32 passed, 0 skipped
```

A throwaway end-to-end script (not committed) also exercised DSN generation → real Freerouting → SES parsing → routed PCB import → `kicad-cli` DRC on a two-resistor board, yielding 0 unrouted nets and 0 hard DRC violations (only the expected lib-footprint-mismatch and silk-edge-clearance warnings).

## Docs

No README, CHANGELOG, or DECISIONS entries changed. `manual-tests/ROUTING-ENGINES.md` was rewritten to record which routers were actually executed vs investigated, with exact versions, commands, and results. Source doc comments were added to the new KiCad layout modules (`board.ts`, `dsn.ts`, `freerouting.ts`, `layout.ts`, `placement.ts`, `types.ts`), and the `add-fab-release-gate` OpenSpec artifacts (`design.md`, `tasks.md`) were updated.

---

## Reference artifacts (founder's #253 review)

The founder asked to "attach the files as reference in the PR". The reference files are committed to this branch and linked here:

- [`manual-tests/ROUTING-ENGINES.md`](https://github.com/copperheadhq/copperhead/blob/ec64c4381e2717bc30c3929947eadf909df96147/manual-tests/ROUTING-ENGINES.md) — the alternative-router evaluation: which routers were actually executed vs investigated, exact versions/commands, results, and why Freerouting stays the default.
- [`scripts/route-real-design.ts`](https://github.com/copperheadhq/copperhead/blob/ec64c4381e2717bc30c3929947eadf909df96147/scripts/route-real-design.ts) — the real-design routing harness (`npm run route:realdesign`): populates/places/routes a real KiCad project and writes a golden-vs-copperhead comparison.
- [`scripts/route-compare.ts`](https://github.com/copperheadhq/copperhead/blob/ec64c4381e2717bc30c3929947eadf909df96147/scripts/route-compare.ts) — the alternative-router comparison harness (`npm run route:compare`): one DSN routed through each engine, DRC on each import.
- [`manual-tests/real-designs/README.md`](https://github.com/copperheadhq/copperhead/blob/ec64c4381e2717bc30c3929947eadf909df96147/manual-tests/real-designs/README.md) — the pre-existing real-design corpus infrastructure this work builds on.

The per-run comparison reports (golden `.kicad_pcb` metrics vs copperhead's routed board; per-engine unrouted/DRC/connectivity) are regenerated by the two scripts into `manual-tests/runs/` (gitignored); the committed `ROUTING-ENGINES.md` carries the executed results.

---

## Invariant checklist

- N/A **Spec-gated in**: `edit_file`/`write_file` gating is unchanged by this PR (a new `layout_board` tool is added, not `edit_file`/`write_file`).
- [x] **Verification-gated out**: board mutations via `layout_board` are DRC-gated and roll back to the placed board on hard violations.
- [x] **`check`/`verify` stays LLM-free and network-free**: the `--fab`/`--strict` additions to `check` are deterministic and offline (no provider or network).
- [x] **No sexp serialization**: the new DSN/SES bridge parses read-only and emits text (no round-trip).
- N/A **Sync-obligations ledger**: the ledger/commit mechanism is unchanged; `layout_board` only reports and clears the DRC obligation.
- N/A **Secrets**: no transcript/redaction or key-handling changes.
- N/A **Spec coherence**: no spec-level behavior change in this PR (SPEC.md, `proposal.md`, and delta specs are not modified).

### DIFF
--- a/manual-tests/ROUTING-ENGINES.md
+++ b/manual-tests/ROUTING-ENGINES.md
@@ -1,0 +1,128 @@
+# Routing engines evaluated for #252
+
+The board pipeline emits Specctra DSN and reads SES back
+(`src/kicad/layout/dsn.ts`), and also emits a full `.kicad_pcb`
+(`src/kicad/board.ts`). Both formats are on the table: a candidate engine either
+speaks DSN/SES (the Freerouting ecosystem) or consumes the emitted `.kicad_pcb`
+directly.
+
+This file records what was **actually executed** in this environment (KiCad 9.0.9,
+Temurin JRE 25, no Rust toolchain) against the `complex_hierarchy` reference
+design (68 parts, 50 nets, all through-hole, golden `.kicad_pcb` in
+`/usr/share/kicad/demos/complex_hierarchy/`) — not what a README claims. A router
+is listed as "executed" only where a binary actually ran and a routed output was
+produced; "investigated" means the repo/docs were examined but it could not be
+run here, with the concrete reason.
+
+Reproduce everything with:
+
+```bash
+# the comparison harness: DSN once, then each engine, DRC on each import
+COPPERHEAD_FREEROUTING_JAR=/path/to/freerouting.jar \
+COPPERHEAD_FREEROUTE_BIN=/path/to/freeroute \
+  npm run route:compare                 # → manual-tests/runs/router-compare/
+```
+
+## Results on `complex_hierarchy` (68 parts, 50 nets, 2 signal layers)
+
+Golden reference: 365 tracks, 0 vias, 0 DRC violations, 0 unrouted, board
+100.7 × 80.0 mm. "Hard DRC" counts error-severity violations excluding the
+ratsnest, via the same `kicad-cli pcb drc` on every board.
+
+| Router | Version | Executed | I/O | Routed | Unrouted | Hard DRC | Tracks | Vias | Runtime |
+| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
+| **Freerouting** (jar) | 2.3.0 | yes | DSN→SES | 50/50 | 0 | **25** | 752 | 32 | ~100 s |
+| KiCadRoutingTools | 0.21.3 | yes | PCB→PCB | 50/50 | 0 | **0** | 1305 | 190 | ~58 s |
+| kicad-tools | 0.20.0 | yes¹ | PCB→PCB | 45/47² | — | — | — | — | >5 min |
+| freeroute grid | 2026.7.14 | yes | DSN→SES | 50/50 | 0 | 1255 | 2697 | 102 | 23 s |
+| freeroute exact | 2026.7.14 | yes | DSN→SES | — | 40 | 646 | 1975 | 203 | 18.5 s |
+| freeroute room | 2026.7.14 | yes³ | DSN→SES | — | — | — | — | — | >15 min |
+
+¹ kicad-tools' C++ backend gave up on the denser nets ("C++ A* open set
+exhausted") and fell back to pure-Python A*; it was still routing after 5 min and
+was stopped. ² 45 of its 47 net-set at the stop point. ³ the `room` engine did
+not finish a 68-part board in 15 minutes and was stopped.
+
+The two numbers that matter are **Routed** (did it finish every net) and
+**Hard DRC** (did the finished board pass). By that pair:
+
+- **KiCadRoutingTools** is the strongest here: 50/50 and 0 violations — but it
+  does so by self-selecting finer geometry (0.127 mm tracks, ~0.158 mm clearance,
+  190 vias) and writing a `.kicad_pro` that *lowers the clearance rule and demotes
+  `solder_mask_bridge` to "ignore"*. The 0 above is measured against KiCad's
+  default board rules, i.e. *before* applying its own relaxation.
+- **Freerouting** finishes every net but leaves 25 violations (6 shorts,
+  7 clearance, 12 solder-mask bridges) — the known cost of routing a
+  connectivity-blind placement on a dense through-hole board.
+- **freeroute** (the Python port) is dramatically worse on dense boards: its grid
+  engine emits imprecise, grid-quantised geometry (397 same-layer crossings,
+  511 clearance, 191 hole-clearance), and its "exact" engine drops 40 nets rather
+  than violate clearance.
+
+## The one real interop bug this surfaced: SES coordinate units
+
+Freerouting's `SesWriter` writes session coordinates in Specctra *resolution
+units* — 0.1 µm each for `(resolution um 10)`, i.e. 10 000 units/mm. The `freeroute`
+port echoes raw micrometre coordinates back unchanged (1 000 units/mm) **while
+still declaring `(resolution um 10)`**. Both files look identical in their
+header and differ by an exact factor of 10 in every coordinate; reading a
+`freeroute` session with the Freerouting scale shrinks the board 10×.
+
+`parseSes` now accepts `{ resolutionUnits: false }` for a raw-unit writer, and
+the default (Freerouting) convention is unchanged. See
+`src/kicad/layout/dsn.ts` and the pinned test in `test/layout-dsn.test.ts`.
+
+## Investigated but NOT executable here
+
+| Project | License | Why not run |
+| --- | --- | --- |
+| [topola](https://github.com/mikwielgus/topola) | MIT | The only other serious headless DSN→SES autorouter (`topola-cli`), but Rust-only with no prebuilt binary and no crates.io publish; no `cargo` here. Revisit if a Rust toolchain is added. |
+| [routing-rs](https://github.com/dilawar/routing-rs) | AGPL-3.0 | Rust-only and self-declared non-functional ("vibecoded, doesn't work yet"). |
+| [ACORN](https://github.com/danboyne/ACORN) | MIT | Build fails (`png.h: No such file or directory`; needs `libgd-dev`/`libpng-dev`, no sudo). Even built, it consumes/emits its own text+HTML format, not DSN/SES/`.kicad_pcb`. |
+| [qautorouter](https://github.com/udif/qautorouter-svn) | GPL-3.0 | Qt4 GUI app, abandoned; Qt4 unavailable on Ubuntu 22.04. |
+
+Not routers at all: `pcbflow` (layout scripting), FD-Autoplacer (placement),
+`kicad-freerouting-plugin-alt` (DSN export only, still calls Freerouting), TopoR /
+Specctra / Electra / DeepPCB (proprietary or cloud).
+
+## Why Freerouting stays the default
+
+For the DSN/SES path it is the only mature, licence-compatible engine, and it
+beats its own Python port by an order of magnitude on this board (25 vs 1255
+violations). KiCadRoutingTools is genuinely better at raw completion/cleanliness
+but is a `.kicad_pcb`-in/`.kicad_pcb`-out engine that quietly relaxes the DRC
+floor to get there, which cuts against copperhead's verification-gated-out
+contract (`AC-2.1`): copperhead will not import a board that only passes because
+the router rewrote the rules. It remains a strong candidate to revisit as a
+second, opt-in engine once its rule-relaxation is made explicit and opt-out.
+
+## Known limitations
+
+- **The ceiling is placement, not the router.** copperhead places on a
+  deterministic, connectivity-blind grid (`src/kicad/layout/placement.ts`). A
+  decoupling cap can land far from its IC, so nets span the whole board and the
+  router congests. Testing a squarer aspect ratio (1.5× vs 2.5× `sqrt(area)`) did
+  *not* reduce violations (66 vs 25 in back-to-back runs), so the wider row is
+  retained; fixing this properly is connectivity-aware placement (#141), out of
+  scope here.
+- **Freerouting's multi-threaded optimisation is known to introduce clearance
+  violations** (it prints this warning itself and recommends `-mt 1`). `-mt 1`
+  did not change the result on this board (identical internal score), so the
+  adapter is left unchanged; it is a lever to pull if a board shows
+  optimisation-only violations.
+- **Connectivity vs golden is name-based and low (24/50)** on every engine,
+  because the flattened netlist uses hierarchical net names the 2022-vintage
+  golden does not. It is a net-naming artifact, not a routing defect; the
+  load-bearing metric is `unrouted == 0`.
+- Freerouting is non-deterministic (multi-threaded); violation counts can vary
+  run-to-run.
+
+## What remains
+
+1. **Connectivity-aware placement (#141)** — the single largest quality lever;
+   the router comparison above shows no router rescues a blind grid placement.
+2. **A KiCad-native engine adapter** behind the same `RoutingEngine` boundary,
+   gated so it can never relax the DRC floor copperhead verifies against.
+3. **THT-hole awareness in the DSN padstacks** — neither Freerouting nor
+   `freeroute` models the drill, so tracks can crowd holes (freeroute's 191
+   hole-clearance violations come from this).
--- a/manual-tests/ROUTING-ENGINES.md
+++ b/manual-tests/ROUTING-ENGINES.md
@@ -38,8 +38,8 @@ ratsnest, via the same `kicad-cli pcb drc` on every board.
 | freeroute exact | 2026.7.14 | yes | DSN→SES | — | 40 | 646 | 1975 | 203 | 18.5 s |
 | freeroute room | 2026.7.14 | yes³ | DSN→SES | — | — | — | — | — | >15 min |
 
-¹ kicad-tools' C++ backend gave up on the denser nets ("C++ A* open set
-exhausted") and fell back to pure-Python A*; it was still routing after 5 min and
+¹ kicad-tools' C++ backend gave up on the denser nets ("C++ A\* open set
+exhausted") and fell back to pure-Python A\*; it was still routing after 5 min and
 was stopped. ² 45 of its 47 net-set at the stop point. ³ the `room` engine did
 not finish a 68-part board in 15 minutes and was stopped.
 
@@ -96,15 +96,44 @@ contract (`AC-2.1`): copperhead will not import a board that only passes because
 the router rewrote the rules. It remains a strong candidate to revisit as a
 second, opt-in engine once its rule-relaxation is made explicit and opt-out.
 
+## Placement: connectivity-aware ordering (was the ceiling)
+
+The first pass of this work placed on a deterministic, connectivity-blind grid —
+the exact "grid, not a layout" fallback #141 names. `src/kicad/layout/placement.ts`
+now orders the board's footprints by schematic connectivity before packing, using
+a deterministic Prim-style growth over an **inverse-fanout-weighted** net graph:
+a two-pin signal net contributes weight 1 between its pair, while a 26-pin GND
+star contributes ~0.04, so power rails stop diluting the ordering. The packer is
+unchanged, so the no-overlap and board-bounds guarantees are untouched.
+
+Measured on `complex_hierarchy` (68 parts, 50 nets):
+
+| Metric | Blind grid (before) | Connectivity order (after) |
+| --- | --- | --- |
+| Net locality (avg ordering spread) | 29.2 | **8.7** (3.4× better) |
+| Vias | 27 | **2** |
+| Track segments | 642¹ | 374 (golden: 365) |
+| Hard DRC | 21–66 (run-dependent) | 26 |
+| Unrouted nets | 0 | 1 |
+
+¹ the earlier committed run; the before/after are not the same invocation.
+
+The locality win is real and machine-checkable (see `test/layout-placement-connectivity.test.ts`):
+fewer/shorter nets mean the router needs almost no vias (27 → 2) and produces a
+track count within 2% of the golden. But it is **not** enough to make Freerouting
+route this dense THT board hard-clean — see the next section.
+
 ## Known limitations
 
-- **The ceiling is placement, not the router.** copperhead places on a
-  deterministic, connectivity-blind grid (`src/kicad/layout/placement.ts`). A
-  decoupling cap can land far from its IC, so nets span the whole board and the
-  router congests. Testing a squarer aspect ratio (1.5× vs 2.5× `sqrt(area)`) did
-  *not* reduce violations (66 vs 25 in back-to-back runs), so the wider row is
-  retained; fixing this properly is connectivity-aware placement (#141), out of
-  scope here.
+- **Placement is now connectivity-aware, but the DRC ceiling is Freerouting, not
+  placement.** Even with connected components clustered, Freerouting leaves 26
+  error-severity violations (solder-mask bridges, optimizer shorts, clearance) on
+  the dense THT board: it does not model the solder-mask aperture, and its
+  multi-threaded optimisation is known to introduce clearance violations. The
+  connectivity order reduces copper (374 segments vs 642, 2 vias vs 20) but
+  trades it for one net Freerouting fails to join — a dense multi-channel THT
+  output stage is beyond what a 1-D ordering can fully cluster. True 2-D
+  connectivity placement remains the next lever (#141, see below).
 - **Freerouting's multi-threaded optimisation is known to introduce clearance
   violations** (it prints this warning itself and recommends `-mt 1`). `-mt 1`
   did not change the result on this board (identical internal score), so the
@@ -119,8 +148,13 @@ second, opt-in engine once its rule-relaxation is made explicit and opt-out.
 
 ## What remains
 
-1. **Connectivity-aware placement (#141)** — the single largest quality lever;
-   the router comparison above shows no router rescues a blind grid placement.
+1. **True 2-D connectivity placement (#141)** — the connectivity *ordering* is
+   implemented and is a real win (see above), but it is still a 1-D sequence laid
+   into a 2-D grid. A multi-channel design whose output stage bridges two clusters
+   can still scatter one of its nets across the board (the single unrouted net
+   above). A 2-D cluster-growth placer — placing each component adjacent to its
+   strongest connected neighbour, not merely in sequence order — is the remaining
+   generation half of #141.
 2. **A KiCad-native engine adapter** behind the same `RoutingEngine` boundary,
    gated so it can never relax the DRC floor copperhead verifies against.
 3. **THT-hole awareness in the DSN padstacks** — neither Freerouting nor
--- a/openspec/changes/add-fab-release-gate/design.md
+++ b/openspec/changes/add-fab-release-gate/design.md
@@ -41,3 +41,7 @@ Purely additive flag; no behavior change without `--fab`. Rollback is removing t
 ## Open Questions
 
 - Whether stage 6 should start writing the export hash record immediately (task here) or wait for a dedicated `copperhead export` command (Phase 3 idea); this change writes it in stage 6 and any future export path inherits the contract.
+
+## Issue #252 follow-on (placement/routing engine boundary)
+
+The routing-completeness gate (task 1.1) and the `--fab`/`--strict` CLI wiring (tasks 3.1–3.3) were delivered under issue #252; the remaining checks (BOM readiness, schematic-to-PCB match, output freshness — tasks 1.2–1.4) are unchanged and out of #252's scope. Issue #252 also landed the board half of `create` end to end: board population from the exported netlist (`src/kicad/board.ts`, `netlist.ts`, `fplib.ts`), a deterministic grid placer (`PlacementEngine`), a TypeScript Specctra DSN/SES bridge plus a local Freerouting adapter (`RoutingEngine`), and the `layout_board` agent tool driving populate → place → route → DRC → rollback. The routing-completeness gate is wired into the export step (`export_outputs`) so an unrouted board cannot silently reach fabrication without an explicit override. Connectivity-aware placement remains a deferred follow-up (#141), not part of #252.
--- a/openspec/changes/add-fab-release-gate/tasks.md
+++ b/openspec/changes/add-fab-release-gate/tasks.md
@@ -2,7 +2,7 @@
 
 ## 1. Gate module
 
-- [ ] 1.1 Implement `src/kicad/fab.ts`: routing-completeness check over the normalized DRC report (unconnected items → named nets with locations)
+- [x] 1.1 Implement `src/kicad/fab.ts`: routing-completeness check over the normalized DRC report (unconnected items → named nets with locations) — delivered via issue #252
 - [ ] 1.2 Implement BOM-readiness check: parse BOM.md rows, flag missing MPN/footprint as failures, `UNVERIFIED` rows as warnings
 - [ ] 1.3 Implement schematic-to-PCB match: refdes + footprint join via `list_symbols` and a board-side footprint enumerator added to `src/kicad/sexp.ts` (read-only)
 - [ ] 1.4 Implement output-freshness check: read the export hash record from `.copperhead/config.json`, recompute SHA-256 of `.kicad_pcb`, compare; distinct messages for missing outputs, missing record, and stale hash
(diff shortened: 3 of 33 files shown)


animesh-chouhan:
take a look at this folder: https://github.com/copperheadhq/copperhead/tree/main/manual-tests/real-designs

and try to implement something similar or just reuse this and try to route already existing reference pcb design using your implementation and compare it with golden reference. fix issues and improve the routing. also do try out various routing open source packages available online. and attach the files as reference in the pr as well. great work!

animesh-chouhan:
_Automated `pr-review` pass (Claude Code). Not a human review; no approval implied._

## Verdict: request changes

The layout/routing pipeline itself is well-built: the engine boundary is clean, the DSN/SES bridge is carefully documented against Freerouting's real dialect, the SES resolution-unit fix is a genuine catch, and the connectivity-aware placer is a real improvement with real measurements behind it. What blocks merge is smaller and mostly outside the new modules: the change to `hardViolations` quietly relaxes the DRC commit gate past the ratsnest it documents, the routing gate passes vacuously on the exact "zero-copper board" case its own comment says it exists to catch, and the agent-tool wiring where all of that lives has no test coverage.

Reviewed in an isolated worktree at head `c0b9f75`; the main checkout stayed on its branch and was never entered.

## Metrics

```
lines: +3732 / -19 (net 3713) across 33 files  ·  src +2035/-14, test +898/-0, docs +162/-0, spec +8/-4, other +629/-1
new vs net: 20 new files (8 in src), 13 modified; 1817 new src lines (the net-new surface)
tests: +898 test lines; head suite 838 passed / 6 failed / 23 skipped (base not measured)
diff coverage: not measured (@vitest/coverage-v8 could not be installed offline)
deps: +[none] -[none] ~[none] (package.json changed: two new scripts only)
ci: 3 pass / 0 fail / 0 pending; required checks: none reported (gh pr checks)
```

**Method.** Measured in a detached review worktree of `refs/pull/253/head`. Merge base `7e9b788` (`git merge-base origin/main HEAD`). Areas from `git diff --numstat`, generated files excluded. Suite from an actual `vitest run` in the worktree, not from the description.

Two caveats on the suite number, both stated rather than smoothed over:
- The 6 failures are 60s timeouts and legibility-gate assertions in `test/init-check.test.ts` and `test/stage-completion.test.ts`, all of them `kicad-cli`-heavy. A second full suite (another PR's worktree) was running on the same box, so these are very likely contention artifacts rather than PR defects.
- I did not run the base suite, so I cannot confirm the description's claim of "4 pre-existing environment failures" on `origin/main`. Treat both numbers as unreconciled.
- Diff coverage is genuinely not measured, so the untested-surface finding below comes from reading the test files for references to the changed symbols, not from a coverage report.

## Invariant checks

| Invariant | Result |
| --- | --- |
| Spec-gated in | **pass** — `layout_board` is added with `requiresUnlock: true`; the `edit_file`/`write_file` omission mechanism is untouched |
| Verification-gated out | **partial** — the populate/route/DRC/rollback path is sound, but the commit gate was weakened (findings 1, 4, 7) |
| `check`/`verify` LLM-free and network-free | **pass** — `src/commands/check.ts` imports only `fab.js`, `report.js`, and fs; the routing gate reuses the DRC already run rather than shelling out again |
| No sexp serialization | **pass** — `board.ts` and `dsn.ts` emit via string builders; `netlist.ts`, `fplib.ts`, `parseSes` are read-only |
| Sync-obligations ledger | **pass** — `layout_board` and `run_drc` both feed `ctx.ledger`; commit still refuses on open obligations |
| Secrets | **pass** — no transcript, redaction, or key handling touched; `.gitignore` keeps `.env`, `.copperhead/runs/`, adds `manual-tests/runs/` |
| Spec coherence | **partial** — `openspec validate add-fab-release-gate` passes, but SPEC.md was not updated for the DRC gate change and `tasks.md` over-claims (findings 1, 5) |

---

## Findings

### 1. [high] The DRC commit gate silently stopped blocking on warning-severity violations

**Where:** `src/kicad/report.ts:91`, `src/agent/tools.ts:551`, `src/agent/tools.ts:860`

The gate moved from `report.ok` (zero violations, any severity) to `hardViolations(report).length === 0`, which filters `severity === 'error'` **and** subtracts the ratsnest. The ratsnest half is justified and defensible: `SPEC.md:183` explicitly allows a first-draft layout to leave nets as ratsnest, and the code comments argue that case well. The severity half is not mentioned anywhere in the PR, the design note, or SPEC.md, and it changes behavior on its own. `run_erc` keeps the strict `.ok` test at `tools.ts:857`, so ERC and DRC now apply different standards with no stated reason.

**Fails when:** the agent edits a `.kicad_pcb`, DRC returns warning-severity findings (silk over copper, overlapping courtyards) and zero errors. `run_drc` clears the `drc` obligation, `finish` reports no problems, and the run commits. `copperhead check` on that same repo still uses `drc.ok` (`src/commands/check.ts:80`), so `do` commits a board that `check` immediately rejects.

**Fix:** subtract only the ratsnest and leave severity alone: `report.violations.filter((v) => !report.unrouted.includes(v))`. If dropping warnings is intended, make it explicit in SPEC.md and apply the same rule to ERC. Add a regression test that feeds `finish` a DRC report with a warning-severity violation and asserts the gate refuses.

### 2. [medium] The routing gate passes vacuously on the zero-copper board it exists to catch

**Where:** `src/kicad/fab.ts:152-153`, `src/agent/tools.ts` (`export_outputs` handler), `src/commands/check.ts:155`

`checkRoutingCompleteness` infers "fully routed" from the *absence* of `unconnected_items`. A board with no footprints and no nets has none, so it passes. The `export_outputs` comment states "a zero-copper / unrouted board must not silently proceed to fabrication", which is precisely the case that gets through. The PR's own test codifies it: `test/init-check.test.ts:111` asserts `res.fab!.routing.status` is `'pass'` with the comment "bare outline: nothing unrouted". Separately, when `config.board` names a file that does not exist, `check.ts:78` leaves `drc` null and `fab.ts:153` returns a vacuous pass.

**Fails when:** `copperhead check --fab` on a repo whose board is a bare Edge.Cuts outline, or whose configured board file is missing, reports `fab routing ✓` and exits 0. `export_outputs` exports gerbers for a board with no copper without triggering the `allow_unrouted` prompt.

**Fix:** fail (or at minimum warn) when the board carries no nets at all, and treat a configured-but-missing or un-runnable board as a gate failure rather than a vacuous pass. Update the `init-check` assertion to match the corrected semantics.

### 3. [medium] `emitDsn` emits net pins for components that have no image whenever a footprint is unresolved

**Where:** `src/kicad/board.ts:81-82`, `src/kicad/layout/dsn.ts:319`, `src/kicad/layout/layout.ts:95`

`buildBoard` builds `board.nets` from every netlist node, but drops footprints it cannot resolve into `unresolved`. `runBoardLayout` still calls the router when `unresolved` is non-empty. The emitted DSN then references pins on components that have no `(image …)` and no `(place …)` record.

**Fails when:** any footprint is missing from the installed libraries. Confirmed by running the branch's own code:

```
unresolved: [{"ref":"U1","footprint":"Missing:NOPE"}]
images: [ 'R1' ]
--- network section ---
    (net "NET1"
      (pins R1-1 U1-5)
    )
```

This is the repo's own fixture path, not a hypothetical: `test/board-layout.test.ts:132` asserts that `open-key`'s U1 is unresolved against the installed KiCad 9 footprint set. Freerouting rejects the dangling reference, so the run surfaces as `process-failed` rather than as a wrong board: the consequence is a confusing error, not bad copper.

**Fix:** filter net pins down to refs present in `board.footprints` (dropping nets left with fewer than two pins), or refuse to route while `unresolved` is non-empty and say why.

### 4. [medium] `runBoardLayout` returns pre-rollback DRC counters beside the post-rollback report

**Where:** `src/kicad/layout/layout.ts:116-117`, `:128`, `:154-155`, `:181`

`drcViolations` and `unroutedNets` are computed from the routed board at lines 116-117. On the rollback path the code re-runs DRC into `finalDrc` (line 128) and returns that as `drc`, but returns the two stale scalars unchanged (lines 154-155). The doc comment at `layout.ts:74-79` promises the returned report always matches the final on-disk file; the numbers printed next to it do not.

**Fails when:** the router produces hard violations and the board is rolled back, which the PR description says is the actual outcome for `complex_hierarchy`. `formatBoardLayoutResult` then prints `0 net(s) unrouted` for a board that is a complete ratsnest, and `DRC: 26 hard violation(s)` for a board that has none. The agent reads that text.

**Fix:** recompute both counters from `finalDrc` after the rollback, and add a test asserting that a rolled-back result reports the placed board's unrouted count.

### 5. [medium] `--strict` is accepted and documented but does nothing, and `tasks.md` marks it delivered

**Where:** `src/cli.ts:184`, `src/commands/check.ts:38`, `openspec/changes/add-fab-release-gate/tasks.md` task 3.1

The flag is threaded from the CLI into `CheckOptions.strict` and never read. Task 3.1 is checked off as delivered including its "warnings non-fatal unless `--strict`" clause, which is not implemented. `checkDocumentationPresence` can return `'warn'`, and `check.ts:185` only fails on `'fail'`, so nothing escalates.

**Fails when:** a team puts `copperhead check --fab --strict` in CI expecting warnings to be fatal. They get a silently permissive gate with no indication the flag was ignored.

**Fix:** either implement the escalation (a `'warn'` status fails the gate under `--strict`) or drop the flag from `--help` until task 1.2 lands, and un-check the part of task 3.1 that is not delivered.

### 6. [medium] `layout_board` regenerates the whole board from the netlist, discarding existing board content

**Where:** `src/kicad/layout/layout.ts:82-84`, `src/kicad/board.ts` (`emitBoard`)

`emitBoard` writes a `.kicad_pcb` from scratch: grid-placed footprints, a fresh Edge.Cuts rectangle, no zones, no pre-existing tracks, no board setup beyond `pad_to_mask_clearance`. Nothing scopes the tool to a fresh board or to `create` stage 5.

**Fails when:** the agent calls `layout_board` during a `do` run on a board that already has manual placement, routing, zones, or a non-rectangular outline. All of it is overwritten. The run's git snapshot makes it recoverable within the run, but a run that then passes DRC commits the loss, and neither the tool description nor the stage 5 prompt warns about it.

**Fix:** refuse when the target board already carries footprints or tracks unless an explicit `replace: true` argument is passed, and state the destructive behavior in the tool description.

### 7. [medium] The run summary contradicts the commit it just made

**Where:** `src/agent/loop.ts:651-652`, `src/agent/loop.ts:752-753`

These still format DRC state from `ctx.lastDrc.ok`, which the gate no longer uses.

**Fails when:** a `layout_board` run leaves a legitimate ratsnest. `finish` allows the commit (correctly, per the ratsnest rationale), and the transcript and run summary record `DRC FAILING` alongside `outcome: success`. The persisted record of a successful run says its verification failed.

**Fix:** use `hardViolations` in `loop.ts` too, or report both parts explicitly ("DRC: 0 hard violations, 12 unrouted").

### 8. [low] `hardViolations` doc claims `schematic_parity` is excluded; the code includes it

**Where:** `src/kicad/report.ts:82-91`

The comment says "`unconnected_items` (and `schematic_parity`) stay out of this count on purpose", but the filter only subtracts `report.unrouted`. `schematic_parity` violations are pushed to `violations` only, so error-severity parity findings do count as hard violations. Someone reading the comment would conclude a schematic/PCB mismatch does not block the gate, when it does.

**Fix:** correct the comment, or exclude parity findings for real if that was the intent.

### 9. [low] Untested surface: the changed gate and tool wiring has no test reaching it

**Where:** `src/agent/tools.ts:551`, `:582`, `:620` (export gate), `:860`; `src/kicad/layout/layout.ts:116-156`

No test file references `layout_board`, `export_outputs`, `allow_unrouted`, or the changed `run_drc`/`finish` gate. `hardViolations` is exercised only inside `test/route-real-design.test.ts`, which is double-gated on `COPPERHEAD_TEST_CORPUS` and `COPPERHEAD_TEST_FREEROUTING_JAR` and therefore skipped on any normal run. The 898 new test lines are real and good, but they all cover the pure modules (netlist, fplib, board, dsn, placement, the fab gate function). Every behavior in findings 1, 2 (export path), 4, 6, and 7 is unexecuted by CI.

This reconciles with the "diff coverage: not measured" line above: the gap was found by reading the test files for references to the changed symbols, not from a coverage report, so treat it as a floor rather than an exact list.

**Fix:** the two regression tests worth writing first are finding 1 (a warning-severity DRC report at `finish` must refuse) and finding 4 (a rolled-back layout result must report the placed board's counters).

### 10. [low] The two committed measurement sets for the same engine and design disagree

**Where:** `manual-tests/ROUTING-ENGINES.md:37`, PR description "Real-design routing" section

The description reports Freerouting on `complex_hierarchy` at 374 track segments, **2 vias**, 26 violations. `ROUTING-ENGINES.md` reports the same engine, same design, same placement at 752 tracks, **32 vias**, 25 violations. Both harnesses call the same `populateBoard` (`scripts/route-compare.ts:220`, `scripts/route-real-design.ts:178`), so the placement is identical. Track counts could differ by polyline-versus-segment convention; 2 versus 32 vias cannot. Since `manual-tests/runs/` is gitignored, neither figure is reproducible from the repo, and these are the numbers that justify the placer change.

**Fix:** re-run one harness and reconcile, or state in `ROUTING-ENGINES.md` which run each table came from and at which commit.

---

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 253, branch issue-252-layout-routing,
based on main.

Check it out first:
  gh pr checkout 253

FINDING 1 [high] The DRC commit gate silently stopped blocking on warning-severity violations.
  Where: src/kicad/report.ts:91, src/agent/tools.ts:551, src/agent/tools.ts:860
  Fails when: the gate moved from report.ok (zero violations, any severity) to
    hardViolations(report).length === 0, which filters severity === 'error' AND
    subtracts the ratsnest. The ratsnest half is justified: SPEC.md:183 allows a
    first-draft layout to leave nets as ratsnest. The severity half is not
    mentioned anywhere in the PR, the design note, or SPEC.md. run_erc keeps the
    strict .ok test at tools.ts:857, so ERC and DRC now apply different standards
    with no stated reason. Concretely: the agent edits a .kicad_pcb, DRC returns
    warning-severity findings (silk over copper, overlapping courtyards) and zero
    errors, run_drc clears the drc obligation, finish reports no problems, and the
    run commits. copperhead check on that same repo still uses drc.ok
    (src/commands/check.ts:80), so do commits a board that check immediately rejects.
  Fix: subtract only the ratsnest and leave severity alone:
    report.violations.filter((v) => !report.unrouted.includes(v)). If dropping
    warnings is intended, make it explicit in SPEC.md and apply the same rule to
    ERC. Add a regression test that feeds finish a DRC report with a
    warning-severity violation and asserts the gate refuses.

FINDING 2 [medium] The routing gate passes vacuously on the zero-copper board it exists to catch.
  Where: src/kicad/fab.ts:152-153, src/agent/tools.ts (export_outputs handler),
    src/commands/check.ts:155
  Fails when: checkRoutingCompleteness infers "fully routed" from the absence of
    unconnected_items. A board with no footprints and no nets has none, so it
    passes. The export_outputs comment states "a zero-copper / unrouted board must
    not silently proceed to fabrication", which is precisely the case that gets
    through. The PR's own test codifies it: test/init-check.test.ts:111 asserts
    routing.status is 'pass' with the comment "bare outline: nothing unrouted".
    Separately, when config.board names a file that does not exist, check.ts:78
    leaves drc null and fab.ts:153 returns a vacuous pass. So copperhead check
    --fab on a bare-outline repo, or one whose board file is missing, reports
    "fab routing ok" and exits 0, and export_outputs exports gerbers for a board
    with no copper without triggering the allow_unrouted prompt.
  Fix: fail (or at minimum warn) when the board carries no nets at all, and treat
    a configured-but-missing or un-runnable board as a gate failure rather than a
    vacuous pass. Update the init-check assertion to match the corrected semantics.

FINDING 3 [medium] emitDsn emits net pins for components that have no image whenever a footprint is unresolved.
  Where: src/kicad/board.ts:81-82, src/kicad/layout/dsn.ts:319,
    src/kicad/layout/layout.ts:95
  Fails when: buildBoard builds board.nets from every netlist node but drops
    footprints it cannot resolve into unresolved, and runBoardLayout still calls
    the router when unresolved is non-empty. The emitted DSN then references pins
    on components that have no (image ...) and no (place ...) record. Reproduced
    against this branch with one resolved part R1 and one unresolved part U1: the
    network section emits (pins R1-1 U1-5) while the library section contains only
    image R1. This is the repo's own fixture path, not a hypothetical:
    test/board-layout.test.ts:132 asserts that open-key's U1 is unresolved against
    the installed KiCad 9 footprint set. Freerouting rejects the dangling
    reference, so the run surfaces as process-failed rather than as a wrong board:
    the consequence is a confusing error, not bad copper.
  Fix: filter net pins down to refs present in board.footprints (dropping nets left
    with fewer than two pins), or refuse to route while unresolved is non-empty and
    say why.

FINDING 4 [medium] runBoardLayout returns pre-rollback DRC counters beside the post-rollback report.
  Where: src/kicad/layout/layout.ts:116-117, :128, :154-155, :181
  Fails when: drcViolations and unroutedNets are computed from the routed board at
    lines 116-117. On the rollback path the code re-runs DRC into finalDrc (line
    128) and returns that as drc, but returns the two stale scalars unchanged
    (lines 154-155). The doc comment at layout.ts:74-79 promises the returned
    report always matches the final on-disk file; the numbers printed next to it do
    not. This fires when the router produces hard violations and the board is
    rolled back, which the PR description says is the actual outcome for
    complex_hierarchy: formatBoardLayoutResult then prints "0 net(s) unrouted" for
    a board that is a complete ratsnest, and "DRC: 26 hard violation(s)" for a
    board that has none. The agent reads that text.
  Fix: recompute both counters from finalDrc after the rollback, and add a test
    asserting that a rolled-back result reports the placed board's unrouted count.

FINDING 5 [medium] --strict is accepted and documented but does nothing, and tasks.md marks it delivered.
  Where: src/cli.ts:184, src/commands/check.ts:38,
    openspec/changes/add-fab-release-gate/tasks.md task 3.1
  Fails when: the flag is threaded from the CLI into CheckOptions.strict and never
    read. Task 3.1 is checked off as delivered including its "warnings non-fatal
    unless --strict" clause, which is not implemented. checkDocumentationPresence
    can return 'warn', and check.ts:185 only fails on 'fail', so nothing escalates.
    A team putting copperhead check --fab --strict in CI expecting warnings to be
    fatal gets a silently permissive gate with no indication the flag was ignored.
  Fix: either implement the escalation (a 'warn' status fails the gate under
    --strict) or drop the flag from --help until task 1.2 lands, and un-check the
    part of task 3.1 that is not delivered.

FINDING 6 [medium] layout_board regenerates the whole board from the netlist, discarding existing board content.
  Where: src/kicad/layout/layout.ts:82-84, src/kicad/board.ts (emitBoard)
  Fails when: emitBoard writes a .kicad_pcb from scratch: grid-placed footprints, a
    fresh Edge.Cuts rectangle, no zones, no pre-existing tracks, no board setup
    beyond pad_to_mask_clearance. Nothing scopes the tool to a fresh board or to
    create stage 5. If the agent calls layout_board during a do run on a board that
    already has manual placement, routing, zones, or a non-rectangular outline, all
    of it is overwritten. The run's git snapshot makes it recoverable within the
    run, but a run that then passes DRC commits the loss, and neither the tool
    description nor the stage 5 prompt warns about it.
  Fix: refuse when the target board already carries footprints or tracks unless an
    explicit replace: true argument is passed, and state the destructive behavior in
    the tool description.

FINDING 7 [medium] The run summary contradicts the commit it just made.
  Where: src/agent/loop.ts:651-652, src/agent/loop.ts:752-753
  Fails when: these still format DRC state from ctx.lastDrc.ok, which the gate no
    longer uses. On a layout_board run that leaves a legitimate ratsnest, finish
    allows the commit (correctly, per the ratsnest rationale) and the transcript and
    run summary record "DRC FAILING" alongside outcome: success. The persisted
    record of a successful run says its verification failed.
  Fix: use hardViolations in loop.ts too, or report both parts explicitly
    ("DRC: 0 hard violations, 12 unrouted").

FINDING 8 [low] (optional) hardViolations doc claims schematic_parity is excluded; the code includes it.
  Where: src/kicad/report.ts:82-91
  Fails when: the comment says "unconnected_items (and schematic_parity) stay out of
    this count on purpose", but the filter only subtracts report.unrouted.
    schematic_parity violations are pushed to violations only, so error-severity
    parity findings do count as hard violations. Someone reading the comment would
    conclude a schematic/PCB mismatch does not block the gate, when it does.
  Fix: correct the comment, or exclude parity findings for real if that was the intent.

FINDING 9 [low] (optional) Untested surface: the changed gate and tool wiring has no test reaching it.
  Where: src/agent/tools.ts:551, :582, :620 (export gate), :860;
    src/kicad/layout/layout.ts:116-156
  Fails when: no test file references layout_board, export_outputs, allow_unrouted,
    or the changed run_drc/finish gate. hardViolations is exercised only inside
    test/route-real-design.test.ts, which is double-gated on COPPERHEAD_TEST_CORPUS
    and COPPERHEAD_TEST_FREEROUTING_JAR and therefore skipped on any normal run. The
    898 new test lines all cover the pure modules (netlist, fplib, board, dsn,
    placement, the fab gate function). Every behavior in findings 1, 2 (export
    path), 4, 6, and 7 is unexecuted by CI.
  Fix: the two regression tests worth writing first are finding 1 (a
    warning-severity DRC report at finish must refuse) and finding 4 (a rolled-back
    layout result must report the placed board's counters).

FINDING 10 [low] (optional) The two committed measurement sets for the same engine and design disagree.
  Where: manual-tests/ROUTING-ENGINES.md:37, PR description "Real-design routing" section
  Fails when: the description reports Freerouting on complex_hierarchy at 374 track
    segments, 2 vias, 26 violations. ROUTING-ENGINES.md reports the same engine,
    same design, same placement at 752 tracks, 32 vias, 25 violations. Both
    harnesses call the same populateBoard (scripts/route-compare.ts:220,
    scripts/route-real-design.ts:178), so the placement is identical. Track counts
    could differ by polyline-versus-segment convention; 2 versus 32 vias cannot.
    Since manual-tests/runs/ is gitignored, neither figure is reproducible from the
    repo, and these are the numbers that justify the placer change.
  Fix: re-run one harness and reconcile, or state in ROUTING-ENGINES.md which run
    each table came from and at which commit.

These lines in the diff are reached by no test: src/agent/tools.ts:551, :582, :620,
:860 and src/kicad/layout/layout.ts:116-156. Add tests for the new exported symbols,
branches, and error paths they contain.

This PR changes spec-level behavior (the DRC gate semantics in finding 1), so
openspec/specs/SPEC.md and the active change artifacts (proposal.md, design.md,
delta specs, tasks.md) have to move together, and
`openspec validate add-fab-release-gate` has to pass.

    by prompt text.

Then verify:
  npm run typecheck
  npm run build
  npm test

and report it.

````

animesh-chouhan:
Thank you for this. It was the first working path from a schematic to a routed board in this repo, and it shaped what came after. Closing it now because the layout work landed on a different footing, with credit where the pieces came from.

**What was re-landed from here** (branch `feat/pcb-layout-framework`, change `add-pcb-layout-framework`, RFC 11 in `copperhead-rfcs`):

- The Specctra DSN/SES bridge, including the conventions this PR established (KiCad's own exporter as the dialect, Freerouting's resolution-unit coordinates in SES), is the seed of `router-freerouting`. Task 3.3 and the proposal name this PR as the source.
- The routing-completeness idea became the routing gate in the verification layer, measured against the board's own fabrication profile.
- The router comparison script became `copperbench`, which now runs golden microboards and PCBench boards through the same code paths and publishes milestone reports (B1 to B4).

**Why the interface here was not merged as is.** RFC 11 settled the division: copperhead owns the harness and wraps engines; it does not write placers or routers. The engine boundary on the branch therefore carries what an in-process `RoutingEngine`/`PlacementEngine` pair cannot: a manifest per engine (capabilities, license, network need, determinism), a registry, out-of-process execution with a scrubbed environment, an immutable input snapshot checked after the run, wall and engine budgets, provenance, and typed failure kinds. Users choose engines through `pcb.routers` / `pcb.placers`, and the bench decides what ships. The connectivity-aware placer here is the in-house engine work the RFC rules out (§3.8), and the bench later measured that the human placement beats every wrapped placer on real boards, which is the evidence that decision rests on.

**Two things from your engine notes are being carried forward** rather than lost: topola as the second DSN-native router to wrap (it was the only headless DSN/SES router besides Freerouting you found), and the `freeroute` raw-micrometre SES quirk, which any second DSN engine's adapter must handle.

If you want to pick up the topola adapter behind the new contract, it is a manifest plus a small adapter over the existing bridge, and I would be glad to review it.

PR #254, opened 2026-08-28
fix(agent): guard restore() on the refuse and dry-run exit paths too

### LINKED ISSUE #107: a failed rollback takes the run report with it on the refuse and dry-run paths

adhavan18:
## What

A rollback failure on the refuse or dry-run exit paths crashed `runAgentLoop` with an unhandled throw instead of reporting it, exactly when the working tree is in an unknown state and the report matters most.

## Why

Closes #107. `fail()` already guards its `restore()` call: if the rollback throws, it catches the error, records it on the transcript and in `summary.md`, and still finishes the run. The refuse and dry-run paths call `restore()` unguarded, so the identical failure there escapes `runAgentLoop` entirely.

@DPS0340 found and diagnosed this (with the original fix in the now-withdrawn #108): [snapshot()](src/util/git.ts) records an unreferenced `git stash create` object (never `stash store`d, so nothing points at it), and a routine `gc` that runs between the snapshot and the rollback collects it, making the later `git stash apply` fatal. They withdrew #108 for queue reasons unrelated to the fix, and didn't respond when I asked on the issue whether they wanted to reclaim it, so sending it now rather than leaving the gap open, with full credit for the diagnosis.

## Design

Both paths now wrap `restore()` in the same try/catch `fail()` uses: catch the error, emit a `restore-failed` transcript event, and fold it into `summary.md`'s `detail` field with the same "working tree may be in a partial state" wording, before still calling `run-end`/`writeSummary`/returning. Neither path's own outcome classification changes on a rollback failure, matching how `fail()` treats it as an orthogonal, logged condition rather than a new exit path.

## Spec / OpenSpec

Not spec-level: `fail()`'s own restore-guard was never encoded as a numbered AC either, so this extends existing implementation-level defensive behavior to the two paths that lacked it rather than changing anything SPEC.md documents.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | 787 passed, 25 failed, 21 skipped |
| Live-LLM (opt-in) | keyed on `ANTHROPIC_API_KEY`/`OPENAI_API_KEY` | not run (no key in this environment) |

Two new tests in [test/observability.test.ts](test/observability.test.ts), alongside the existing coverage for `fail()`'s identical guard: one for the refuse path, one for the dry-run path. Per DPS0340's own warning on the issue that a naive regression test can pass against the unfixed code for the wrong reason, both reproduce the actual mechanism rather than a shortcut: dirtying only an untracked file never produces a stash object to begin with, since `git stash create` captures tracked changes only, so the dry-run test specifically dirties a tracked fixture file, then runs `git gc --prune=now` mid-run to collect the resulting unreferenced stash object, exactly reproducing the issue's repro. Verified both fail against unfixed `loop.ts` (temporarily reverted) with the exact uncaught errors the issue describes (`fatal: not a git repository` / `fatal: '...' is not a stash-like commit`), and pass clean against the fix.

All 25 pre-existing failures are unrelated to this change: `kicad-cli` not installed and two Windows-only `chmod` semantics assertions, in this environment. Diffed the full failure set against a clean `main` worktree in the same environment: identical set, zero failures introduced by this branch.

n/a: this environment has neither `kicad-cli` on PATH nor a live LLM API key, so the CLI sandboxes in `manual-tests/` (which need both) aren't runnable here. The regression tests above exercise the actual failure mechanism directly at the git level rather than through a mock, which is arguably a more reliable reproduction of this specific bug than a manual run would be, since the underlying trigger is a `gc` timing race the manual sandbox wouldn't deterministically hit either. Flagging the gap rather than skipping it silently.

## Docs

None needed: no user-facing behavior changes (a rollback failure was always reported by two of the three paths; this makes it uniform), and no spec artifacts to keep in sync per above.

---

## Invariant checklist

- [ ] Spec-gated in: N/A, does not touch the agent tool list.
- [ ] Verification-gated out: N/A, does not touch ERC/DRC gating.
- [ ] `check`/`verify` stays LLM-free and network-free: N/A, this PR does not touch `src/commands/check`.
- [ ] No sexp serialization: N/A, does not touch `src/kicad/`.
- [ ] Sync-obligations ledger: N/A, does not touch the ledger.
- [ ] Secrets: N/A, no new transcript/summary fields carry anything but the existing `restoreError` message, which was already surfaced by `fail()`'s identical path.
- [ ] Spec coherence: N/A per the Spec/OpenSpec section above.

### DIFF
--- a/src/agent/loop.ts
+++ b/src/agent/loop.ts
@@ -613,7 +613,17 @@ async function runWithProviders(opts: RunOptions, providers: Set<Provider>): Pro
       const { outcome, summary } = ctx.finishRequest;
       const files = [...ctx.filesTouched];
       if (outcome === 'refuse') {
-        await restore(repoRoot, snap);
+        // The rollback itself can fail (git in a bad state, or — as issue #107
+        // found — the snapshot's stash object already collected by a gc that ran
+        // between snapshot() and here). Same guard as fail(): that must not become
+        // an unhandled throw that skips run-end and summary.md.
+        let restoreError: string | null = null;
+        try {
+          await restore(repoRoot, snap);
+        } catch (err) {
+          restoreError = (err as Error).message;
+          await transcript.event('restore-failed', { error: restoreError });
+        }
         await transcript.event('run-refused', { summary });
         const runStats = stats('refused');
         await transcript.event('run-end', runStats);
@@ -629,11 +639,16 @@ async function runWithProviders(opts: RunOptions, providers: Set<Provider>): Pro
           tokensOut,
           outcome: 'aborted',
           openObligations: null,
-          detail: `REFUSED: ${summary}`,
+          detail: restoreError
+            ? `REFUSED: ${summary}\n\nROLLBACK FAILED: ${restoreError} — the working tree may be in a partial state; inspect it with git status/git diff before rerunning`
+            : `REFUSED: ${summary}`,
           env: meta,
           stats: runStats,
         });
         log(`refused: ${summary}`);
+        if (restoreError) {
+          log(`WARNING: rollback failed (${restoreError}); the working tree may be in a partial state`);
+        }
         r.finish(outcomeLine(runStats));
         return {
           outcome: 'refused',
@@ -662,7 +677,15 @@ async function runWithProviders(opts: RunOptions, providers: Set<Provider>): Pro
         log('--- dry run: proposed diff ---');
         log(diff || '(no diff)');
         if (untracked) log(`new files:\n${untracked}`);
-        await restore(repoRoot, snap);
+        // Same guard as fail(): a rollback failure here must not become an
+        // unhandled throw that skips run-end and summary.md (issue #107).
+        let restoreError: string | null = null;
+        try {
+          await restore(repoRoot, snap);
+        } catch (err) {
+          restoreError = (err as Error).message;
+          await transcript.event('restore-failed', { error: restoreError });
+        }
         const runStats = stats('done');
         await transcript.event('run-end', runStats);
         await transcript.writeSummary({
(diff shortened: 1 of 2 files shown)


animesh-chouhan:
Cross-reference for #258 (run protocol stream), not a review comment: this lands `restore-failed` on two more exit paths, and that event is a good example of what #258 tracks. It is one of 18 transcript event types, and the cloud consumer maps 4, so a refusal whose rollback failed reads downstream as a clean refusal with no sign that the working tree may be partial. The wording you fold into `summary.md`'s `detail` is exactly what should travel in the event payload.

Nothing to change here. Noting it so `restore-failed` is in the vocabulary when the protocol lands, rather than being rediscovered as an unmapped row.

adhavan18:

Issue #246, opened 2026-08-24
Tool-calling framework + skills stream: formalize tool system, ship part-research tools, expose via MCP


Devesh36:
## Summary

This stream owns the tool system as a framework and its outward-facing surfaces. It formalizes copperhead's currently-flat tool registry into a real framework (schema registry, tool-name registry, schema versioning, gating, cross-provider adapters, result envelopes, error/retry semantics, view hints), delivers **part-research tools** as the first payload, and then reuses the same schemas to power the Copperhead MCP server and the Claude Code / Codex skills.

## Non-negotiable invariant

Edit tools (`edit_file`, `write_file`, and the other `requiresUnlock` tools) must remain **absent from the model's tool list until an OpenSpec proposal validates** — enforced through code and structural tests, not prompt wording. This is enforced today in `availableTools()` / `dispatchTool()` in `src/agent/tools.ts` and covered by `test/gating-sync.test.ts`. Every change here keeps that test green and adds a matching structural test on any new surface (registry, envelope path, MCP surface).

`check` / `verify` stays LLM-free and network-free (AC-2.1). The `test/init-check.test.ts` module-graph guard is extended, never weakened.

## Current state (baseline)

- Tool system is a flat `TOOLS: ToolDef[]` array with **plain-string results**, working structural spec-gating, and cross-provider adapters in `src/agent/providers/` (native OpenAI/Anthropic; `tool-protocol.ts` for reasoning-only Claude Code / Cursor; structured-JSON for Codex).
- **No** schema-registry abstraction, tool-name registry, schema versioning, structured result envelope, or per-tool view-hint metadata yet.
- `add-part-research-tools` OpenSpec proposal is complete (4/4 artifacts) but 0/47 tasks; no `src/research/` code on `main`.
- `add-mcp-server` OpenSpec proposal exists, unimplemented.

### Phase 0 — Framework foundations (incremental)
- [ ] `ToolRegistry` keyed by name (wrap `TOOLS`); `get`/`list(ctx)`/gating metadata
- [ ] Schema versioning: `version` on `ToolSchema` + registry protocol version
- [ ] Structured `ToolResult` envelope (`{ ok, summary, detail?, data?, error?, viewHint? }`) that flattens to today's plain-string for backward compatibility
- [ ] Typed tool-error kinds (`unavailable`/`validation`/`refusal`/`exception`) + unified retry policy with `src/util/retry.ts`
[...]

Issue #41, opened 2026-07-21
Add part-research tools: web/part search + datasheet cache, sourceable as a checked constraint


animesh-chouhan:
Tracking issue for the **`add-part-research-tools`** OpenSpec change. Full spec: [`openspec/changes/add-part-research-tools/`](openspec/changes/add-part-research-tools/) ([proposal](openspec/changes/add-part-research-tools/proposal.md), [design](openspec/changes/add-part-research-tools/design.md), [tasks](openspec/changes/add-part-research-tools/tasks.md), [specs/](openspec/changes/add-part-research-tools/specs/)).

## Why

The agent designs from trained knowledge alone: it cannot look up a datasheet, confirm a part exists, or check availability, so every new part is guesswork flagged `UNVERIFIED` and constraint sources cite datasheets nobody fetched. Grounding part selection in real, current data removes the biggest credibility gap in `create`/`do` output and pulls forward the "sourceable becomes a checked constraint" roadmap item (SPEC.md §8).

## What ships

- New **research tool family**, available to the agent loop in `do`/`create` alongside file/KiCad/memory tools:
  - `web_search`: reference designs, app notes, errata
  - `search_parts`: structured part search (MPN, parametrics, availability, pricing) via a supplier API
  - `fetch_datasheet`: download a datasheet PDF and cache it under `.copperhead/datasheets/`, recorded with URL, retrieval date, and content hash for reproducibility
- **Sourceable becomes a checked constraint**: search results (availability, lifecycle) land in `.copperhead/constraints.json` as cached snapshots with `source`/`retrieved`; `check` validates against the snapshot offline and reports staleness.
- **`UNVERIFIED` upgrade path**: a part confirmed against a cached datasheet is flagged `VERIFIED(datasheet)` with a pointer to the cached file and confirming section. No fetched evidence, stays `UNVERIFIED`. The agent still never invents MPNs.
- Fetched facts are cited (cache path + section), not pasted, so the registry is human-verifiable offline.

## BREAKING spec change to watch

The "no network tools in Phase 1" safety rail (SPEC.md §7) is **replaced** by a scoped network boundary: research tools are the only network surface, restricted to an allowlisted set of hosts, every request logged to the transcript. **`check`/`verify` stay LLM-free and network-free (AC-2.1 unchanged).** This modifies the `safety-rails` capability, so review that delta spec carefully.

## Invariants that MUST hold

[...]

Issue #40, opened 2026-07-21
Add MCP server: expose the gated copperhead pipeline to MCP hosts


animesh-chouhan:
Tracking issue for the **`add-mcp-server`** OpenSpec change. Full spec: [`openspec/changes/add-mcp-server/`](openspec/changes/add-mcp-server/) ([proposal](openspec/changes/add-mcp-server/proposal.md), [design](openspec/changes/add-mcp-server/design.md), [tasks](openspec/changes/add-mcp-server/tasks.md), [specs/](openspec/changes/add-mcp-server/specs/)).

## Why

Coding agents (Claude Code, Codex, every MCP host) are already pointed at KiCad repos through raw tool servers that expose ungated file edits, so the host can rewrite a `.kicad_sch` with no spec gate, no verification, no rollback. copperhead's invariants only protect users whose entry point is the CLI. A thin MCP wrapper makes the gated pipeline the thing the host agent calls, so the invariants survive the integration.

**Decision (see proposal): this lives in copperhead, not a separate repo.** The MCP layer's value is the gated pipeline, not raw KiCad access; it dispatches into the existing command implementations so all invariants are inherited by construction.

## What ships

- New command `copperhead mcp`: a stdio MCP server exposing only opaque, high-level tools:
  - `copperhead_check` (optional `fab`/`strict` inputs) returns the JSON report
  - `copperhead_do` runs the full gated `do` loop and returns the run summary; the host never drives intermediate steps
  - `copperhead_sync` (verify always; resolve only when `resolve: true`)
  - `copperhead_init` scaffolds docs-memory
- **Structural opacity**: no file-edit, no raw KiCad, no partial-loop tools exposed. Nothing to reach around.
- Companion Claude Code / Codex skill under `integrations/` telling agents to use these tools instead of editing `.kicad_*` directly.

## Invariants that MUST hold (CLAUDE.md, SPEC.md §1.3 / §7)

- Spec-gated in: mutating tools stay structurally absent until a proposal validates.
- Verification-gated out: mutations are not done until ERC (and DRC if the board changed) passes, then repair or rollback.
- Sandbox to repo root; secret redaction (`sk-[A-Za-z0-9_-]+`); keys in env vars only.
- No new network surface: stdio only; LLM calls happen exactly where the CLI already makes them. `copperhead_do` / `sync --resolve` refuse when no API key env var is present.

## Impact

[...]

PR #255, opened 2026-08-29
feat(cli): add skill commands for listing and running registered skills

Devesh36:
## What

Adds a two-tier tool catalog (atomic tools plus skills) and a `copperhead skill` CLI. Users can run `copperhead skill list` (no LLM) and `copperhead skill run generate-report` for a read-only design report. `check`, `create` stages, and spec-gated edits are unchanged.

## Why

Closes Phase 0 of [#246](https://github.com/copperheadhq/copperhead/issues/246) (option 2 from the issue discussion). The tool system was a flat string-returning array; part-research (#41) and MCP (#40) need one registry, `gate()` predicates, versioned schemas, and structured envelopes. This PR is that frame plus one proof skill. It does not implement part-research or MCP.

## Design

One catalog, two tiers. [registry.ts](src/agent/registry.ts) `list(ctx)` applies each entry's `gate()`. Dispatch in [tools.ts](src/agent/tools.ts) runs a tool handler or a nested skill sub-run and returns one `ToolResult`. Providers still get a flattened string (same wire shape). Decisions: [design.md](openspec/changes/formalize-tool-registry/design.md) D1–D10.

Load-bearing choices:
- `gate()` replaces `requiresUnlock`. A skill's default gate is the conjunction of its tools' gates, so a skill that lists `edit_file` is absent until a proposal validates (D2, D3).
- Nested skill is a turn-loop extract, not a second `do`: shared ledger/transcript/`editsUnlocked`, no git snapshot, no commit, no `openspec archive`, `finish` not in the child catalog (D6).
- Envelope `{ ok, summary, detail?, data?, error?, viewHint? }` is sealed (redacted) at construction in [envelope.ts](src/agent/envelope.ts). [theme.ts](src/agent/theme.ts) `toolLine` uses `ok` for the glyph, not a pass/fail regex on prose (D4, D10).
- Proof skill [`generate_report`](src/capabilities/skills/generate-report.ts): read-only tools only, `maxTurns` 8, report is the envelope. [`STAGES`](src/commands/create.ts) stay in `create`.
- Atomic tools stay the [`HANDLERS`](src/capabilities/handlers.ts) table wrapped in the barrel; skills are one file each under `src/capabilities/skills/`.

Invariants: spec-gating stays absence from `registry.list(ctx)` ([gating-sync.test.ts](test/gating-sync.test.ts) keeps `"not available"` / `"unlock"` plus additive `error.kind === 'unavailable'`). Verification-gating stays on the parent `finish`. [`check.ts`](src/commands/check.ts) still imports no capability module.

## Spec / OpenSpec

Change: `formalize-tool-registry` (`openspec validate formalize-tool-registry` passes).

Delta specs: [tool-registry](openspec/changes/formalize-tool-registry/specs/tool-registry/spec.md), [spec-gating](openspec/changes/formalize-tool-registry/specs/spec-gating/spec.md), [agent-core](openspec/changes/formalize-tool-registry/specs/agent-core/spec.md), [cli-surface](openspec/changes/formalize-tool-registry/specs/cli-surface/spec.md), [safety-rails](openspec/changes/formalize-tool-registry/specs/safety-rails/spec.md).

Main spec: [SPEC.md](openspec/specs/SPEC.md) §3 (`skill list` / `skill run`) and §4.2 (`generate_report` in the catalog). AC-2.1 (check module graph excludes `src/capabilities/` and provider SDKs) and AC-4.1 (redaction, including envelope construction) are the acceptance criteria this change is responsible for.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | fail (833 pass, 4 fail, 21 skip) |
| Live-LLM (opt-in) | `COPPERHEAD_MODEL=claude-code` (`skill run`) | pass (see manual log) |

New tests (all pass): [envelope.test.ts](test/envelope.test.ts), [capability-conformance.test.ts](test/capability-conformance.test.ts), [tool-registry.test.ts](test/tool-registry.test.ts) (27 tests). Gating lock, observability glyph, and check module-graph tests also pass.

The 4 `npm test` failures are this machine's KiCad 10.0.5 `lib_symbol_mismatch` warnings on the open-key fixture (`Symbol 'R' doesn't match copy in library 'Device'`). The fixture ignores `lib_symbol_issues` but not the KiCad 10 `lib_symbol_mismatch` check, so ERC is not clean and finish/`check` AC-2.1 fail. Not caused by this PR; live-LLM agent-integration tests were skipped (no `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `COPPERHEAD_TEST_CODEX`).

`npm run docs:build` was not run.

- Sandbox variant used (`create` / `edit`): `edit` (temp copy of `test/fixtures/open-key`, same fixture as `./manual-tests/setup.sh edit`, then `init`)
- Commands run and outcome:

```text
$ npx tsx src/cli.ts --help
Commands include: skill    run a registered skill (nested tool loop; no git commit)

$ npx tsx src/cli.ts skill --help
Commands: list, run [options] <name>

$ npx tsx src/cli.ts --repo <edit-sandbox> skill list
· generate-report  Read-only report of the current design: ERC, DRC (if a board exists), drift, nets, and an optional SVG path. Does not edit files.

$ npx tsx src/cli.ts --json --repo <edit-sandbox> skill list
[{ "name": "generate_report", "available": true, "description": "..." }]

$ npx tsx src/cli.ts --repo <edit-sandbox> skill run banana
unknown skill "banana"
exit 1 (no stack trace)

$ env -u OPENAI_API_KEY -u ANTHROPIC_API_KEY -u COPPERHEAD_MODEL npx tsx src/cli.ts --repo <edit-sandbox> skill run generate-report
no model configured: ... skill run needs OPENAI_API_KEY, ANTHROPIC_API_KEY, or --model for a saved-login provider (codex, claude-code, cursor).
exit 1 (no stack trace)

$ npx tsx src/cli.ts --repo <edit-sandbox> --plain skill run generate-report
# COPPERHEAD_MODEL=claude-code
design report
... run_erc / run_drc / check_drift output ...
exit 0
git diff -- hardware: empty (schematic/PCB untouched; no report file written)
```

## Docs

- [README.md](README.md): `skill list` / `skill run generate-report`; note that `create` stages are not yet skills
- [SPEC.md](openspec/specs/SPEC.md) §3 and §4.2 (drafted for archive)

---

## Invariant checklist

- [x] **Verification-gated out**: mutations still end in ERC (and DRC when the board changed) passing, with repair up to `maxRepairCycles` then rollback to the git snapshot. Skill sub-runs cannot call `finish` or commit.
- [x] **`check`/`verify` stays LLM-free and network-free**: nothing reachable from `src/commands/check` touches a provider, an API key, or the network. Module-graph guard now also excludes `src/capabilities/`.
- [N/A] **No sexp serialization**: this PR does not touch the `src/kicad/` parser or edit path.
- [x] **Sync-obligations ledger**: post-tool-call hooks keep feeding the ledger, and commit keeps refusing while obligations are open. Nested skills share the parent ledger and cannot clear it via `finish`.
- [x] **Secrets**: transcripts and summaries redact `sk-[A-Za-z0-9_-]+` at write time; envelope `data`/`summary`/`detail` are also redacted at construction (`*_KEY`/`*_SECRET`/`*_TOKEN` env values); keys live only in env vars; `.gitignore` keeps `.env` and `.copperhead/runs/`.
- [x] **Spec coherence**: SPEC.md and the change artifacts (`proposal.md`, `design.md`, delta specs, `tasks.md`) moved together and `openspec validate formalize-tool-registry` passes.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,8 @@ Spec-gated in, verification-gated out: the design can't drift from its requireme
 ```text
 copperhead init [--path hardware/]   # scaffold docs/ from an existing schematic; idempotent
 copperhead do "<change request>"     # the core loop: propose, edit, verify, propagate, commit
+copperhead skill list                # registered skills (no LLM)
+copperhead skill run generate-report # read-only design report (needs a model)
 copperhead check                     # ERC + DRC + doc-drift + spec validation; no LLM calls (alias: verify)
 copperhead doctor                    # env preflight: node, kicad-cli, git, openspec, provider credential; no LLM/network
 copperhead sync [--dry-run]          # verify the whole design state, resolve drift
@@ -122,7 +124,9 @@ copperhead create --brief brief.md   # brief → full output package
 copperhead export bom --supplier jlcpcb   # supplier-ready ordering file from docs/BOM.md
 ```
 
-Global flags: `--repo <path>` (default: cwd) and `--json` for machine-readable output. `--model` is available on `do`, `sync`, `create`, and `doctor`; `--interactive` only on `do` and `create`; `do` also takes `--dry-run`, `--max-turns`, and `--allow-dirty`.
+Global flags: `--repo <path>` (default: cwd) and `--json` for machine-readable output. `--model` is available on `do`, `sync`, `create`, `skill run`, and `doctor`; `--interactive` only on `do` and `create`; `do` also takes `--dry-run`, `--max-turns`, and `--allow-dirty`.
+
+`copperhead create` still uses its own hardcoded stages; those are not yet skills.
 
 `--model` accepts `gpt-5` (OpenAI), `claude` / `claude-<id>` (Anthropic API), `claude-code` / `claude-code:<id>` (Claude Code, saved login), `cursor` / `cursor:<id>` (Cursor Agent CLI, saved login), and `codex` / `codex:<id>` (Codex CLI, saved login). Routing is by prefix; `claude-code` is matched before the `claude` prefix. `compat:<id>` targets any OpenAI-compatible endpoint (Groq, OpenRouter, Gemini, or a local Ollama) via `COPPERHEAD_BASE_URL` and `COPPERHEAD_API_KEY_ENV` - worked examples for each in [`.env.example`](.env.example) and the [configuration reference](https://docs.copperhead.sh/reference/configuration/#model-selection).
 
--- a/openspec/changes/formalize-tool-registry/.openspec.yaml
+++ b/openspec/changes/formalize-tool-registry/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-08-29
--- a/openspec/changes/formalize-tool-registry/design.md
+++ b/openspec/changes/formalize-tool-registry/design.md
@@ -1,0 +1,167 @@
+# formalize-tool-registry: Design
+
+## Context
+
+Today there are two capability systems that do not know about each other. `TOOLS: ToolDef[]` in `src/agent/tools.ts` is 23 entries, `requiresUnlock: boolean`, handlers returning `Promise<string>`. `STAGES` in `src/commands/create.ts` is a skill registry under another name: each stage has `name`, `isComplete`, and `prompt`, and `create` executes one by calling `runAgentLoop`. The model sees only the flat tool list. `theme.ts` classifies a tool result by regex on its first line. There is no schema version, no structured envelope, and no `gate()` predicate — so the research tools' condition (`research.enabled` AND keys present) and any future skill would each grow a second gating path.
+
+Issue #246 owns this stream. This change is Phase 0, option 2: framework plus one proof skill. Part-research (#41), MCP (#40), and migrating `STAGES` are out of scope.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- One catalog the model sees; dispatch picks atomic tool vs skill.
+- `gate()` as the single gating mechanism; spec-gating of `edit_file`/`write_file` stays an absence-from-list invariant, proven by test.
+- Structured `ToolResult` envelopes; renderer reads `ok` / `viewHint`.
+- Envelope `data` redacted at construction; render/dock never see a raw secret-shaped payload.
+- Proof skill `generate_report` plus nested sub-run, with CLI `copperhead skill run generate-report`.
+- `check`/`verify` remain LLM-free and network-free (AC-2.1).
+
+**Non-Goals:**
+
+- Migrating `STAGES` onto the skill registry (follow-up; `stage-completion.test.ts` and the `create-*.test.ts` files stay put).
+- Part-research tools, MCP server, Claude Code / Codex companion skills.
+- Changing provider wire formats beyond flattening the envelope to today's string `Msg` tool content.
+- A new network surface, a new LLM SDK, or any change to the `check` module graph.
+
+## Decisions
+
+### D1. One catalog, two tiers; dispatch picks the tier
+
+```
+catalog:  [read_file, edit_file, run_erc, ..., generate_report]
+             |______ tier 1: atomic ______|   |____ tier 2 ____|
+```
+
+`registry.list(ctx)` returns every entry whose `gate(ctx)` is true. Names are unique across both tiers. A skill presents as a tool in the provider's schema list (name, description, parameters). `dispatch(name)` looks up the entry: tier 1 runs `handler`; tier 2 runs the nested sub-run (D6) and returns one envelope.
+
+Alternative considered: keep skills out of the model catalog and inject their prompts into the current loop. Rejected — that forfeits budget isolation and a real completion contract, which is exactly what makes create stages recoverable today.
+
+### D2. `gate()` predicate replaces `requiresUnlock`
+
+```ts
+gate: (ctx: RunContext) => boolean
+```
+
+Edit tools: `gate: (ctx) => ctx.editsUnlocked`. Research tools (later): `gate: (ctx) => enabled && keysPresent`. A boolean cannot express the second without a parallel filter. One predicate is the project's established "absent, not discouraged" idiom.
+
+`availableTools(ctx)` becomes `registry.list(ctx)` filtered to the requested tier, or the combined catalog. Off-catalog rejection in `parseToolCalls` is unchanged: it still takes a `Set` of names from `list(ctx)`.
+
+### D3. A skill's default gate is the conjunction of its tools' gates
+
+```ts
+gate?: (ctx) => boolean  // default: every declared tool's gate(ctx) is true
+```
+
+Load-bearing. A skill that lists `edit_file` is structurally absent from `list(ctx)` until the proposal validates, by construction. Without this default the skill tier is the hole in invariant 1. A skill MAY supply a stricter `gate`; it MUST NOT supply a weaker one. The conformance test (D8) asserts the effective gate is at least as strict as every declared tool.
+
+Worked example: `generate_report` declares `[read_file, search, list_nets, run_erc, run_drc, export_svg, check_drift]` — no mutation tools — therefore always available and provably unable to write.
+
+### D4. Envelope at the handler boundary; flatten for providers
+
+```ts
+interface ToolResult {
+  ok: boolean;
+  summary: string;     // one line the renderer shows
+  detail?: string;     // what the model reads (flattened)
+  data?: unknown;      // structured payload, redacted at construction
+  error?: { kind: 'unavailable' | 'validation' | 'refusal' | 'exception'; message: string };
+  viewHint?: 'diagnostic' | 'mutation' | 'query' | 'export';
+}
+```
+
+Handlers return `ToolResult`. `dispatchTool` returns the envelope to the loop. The loop (1) records the envelope on the transcript, (2) flattens `summary` + `detail` (or `error.message`) into the string `Msg` tool content providers already consume, (3) hands the envelope to the renderer.
+
+Flattening preserves today's gating-sync substrings: an unavailable edit tool's `error.message` still contains `"not available"` and `"unlock"`. The absence assertion `expect(registry.list(ctx).map(e => e.name)).not.toContain('edit_file')` is the invariant; the envelope `kind === 'unavailable'` assertion is additive, never a replacement.
+
+`data` is passed through `redactSecrets` (and the `*_KEY`/`*_SECRET`/`*_TOKEN` value set) at envelope construction. Renderers receive the already-redacted envelope. They never call `JSON.stringify` on unredacted handler output.
+
+Alternative: keep handlers returning strings and wrap at dispatch. Rejected — view hints and typed errors would be reconstructed from prose, which is the regex heuristic this change deletes.
+
+### D5. Schema version on each entry; protocol version on the registry
+
+Each tool and skill carries `version: number` (schema of its parameters + result). The registry exposes a protocol version for the envelope shape. Provider adapters keep sending `{ name, description, parameters }` over the wire; `version` stays in our catalog so MCP hosts can cache a tool list later without this change taking a dependency on MCP.
+
+### D6. Nested sub-run, not a second `do`
+
+A skill call is **not** `runAgentLoop` as `do`/`create` call it. The outer loop owns git preflight, snapshot, commit, and OpenSpec archive. A nested skill that went through that path would snapshot mid-run, commit out from under the parent, or archive a change the parent still holds.
+
+The nested path is a turn-loop extract:
+
+- Shares the parent's `RunContext` ledger, transcript, and `editsUnlocked`.
+- Own messages, own `maxTurns` (skill default or arg), own tool subset (`skill.tools` resolved against the registry).
+- No git snapshot, no commit, no `openspec archive`.
+- Completes when `isComplete(ctx, args)` is true or the turn budget is exhausted; the parent receives one envelope.
+- `finish` is forbidden in a skill's `tools` array (conformance test). The nested loop does not put `finish` in the child catalog. Even if a future caller forced it, `finishRequest` on the shared ctx MUST NOT trigger the parent's commit; the nested path never reads it for that purpose.
+
+Verification-gating stays with the parent: a skill that somehow mutated would open obligations on the shared ledger, and the parent's `finish` would still refuse. `generate_report` cannot mutate, so this is a structural backstop for later skills, not a generate_report concern.
+
+Alternative: inject skill instructions into the current loop. Rejected (D1). Alternative: spawn a full `runAgentLoop`. Rejected because of snapshot/commit.
+
+### D7. `generate_report` is the proof skill; `STAGES` stay put
+
+Read-only report of the current design state. Parameters: `{ scope: 'power' | 'all' }` (default `all`). The nested loop uses the declared tools and returns `summary` + `detail` as the report (ERC/DRC/drift status, net count, SVG path if exported). It does not call `write_file`; the report is the envelope. The CLI prints that envelope; the parent transcript records it.
+
+`STAGES` remain hardcoded in `create.ts`. Migrating them is a later change so this one does not move `stage-completion.test.ts` and the six `create-*.test.ts` files in the same diff as the registry.
+
+### D8. Handlers table + skill modules, identity `defineTool`/`defineSkill`, barrel + conformance
+
+```
+src/capabilities/
+  handlers.ts                atomic tools (schema + string handler + requiresUnlock)
+  index.ts                   wrap HANDLERS via defineTool; import each skill
+  skills/generate-report.ts  export default defineSkill({ ... })
+```
+
+Atomic tools stay one table so a new tool is one `HANDLERS` entry, not a stub file plus a barrel line. Skills remain one module each — they have a prompt, tool subset, and completion contract that a table row cannot hold. `defineTool` / `defineSkill` are identity functions for inference and defaults (`gate` conjunction, `version` required). `src/agent/tools.ts` keeps `RunContext`, dispatch, and the nested skill sub-run.
+
+Conformance test over every registered entry:
+
+- names unique across both tiers
+- `version` and `viewHint` set
+- catalog tool names match `HANDLERS` 1:1
+- every skill `tools` name resolves
+- a skill's effective gate is at least as strict as each declared tool
+- every file under `src/capabilities/skills/` is imported by the barrel (an unregistered skill file fails CI)
+
+Existing tools keep their current names, schemas, and observable string summaries after flattening, so provider integration tests and `gating-sync` stay green.
+
+### D9. CLI form is the same definition
+
+```
+copperhead skill list
+copperhead skill run generate-report [--scope power|all]
+```
+
+`skill list` is LLM-free and network-free: it prints registered skills and whether `gate(ctx)` would include them. `skill run` invokes D6's nested sub-run (it needs a model, same as `do`) and prints the envelope. Tests call the exported runner with a `ScriptedProvider` (already used in `test/observability.test.ts`) so the suite stays offline. Missing key on `skill run` is a typed error naming the env var, not a stack trace.
+
+`check`/`verify` do not import `src/capabilities/`. The module-graph guard in `test/init-check.test.ts` keeps scanning outward from `src/commands/check.ts`.
+
+### D10. Renderer reads `ok` / `viewHint`; retry maps from error kind
+
+`toolLine` uses `result.ok` for the glyph, not `/\b(clean|ok|pass(?:ed)?|success|done)\b/i`. `viewHint` selects layout (diagnostic vs mutation vs query vs export). `validation` and `refusal` are not retried; `exception` uses `src/util/retry.ts` the way provider 429s already do; `unavailable` is not retried (the catalog did not contain the tool).
+
+## Risks / Trade-offs
+
+- [Envelope refactor weakens `gating-sync`] → keep `expect(registry.list(ctx).map(e => e.name)).not.toContain('edit_file')`; keep `"not available"` / `"unlock"` in the unavailable message; add `error.kind === 'unavailable'` alongside, never instead.
+- [Nested `runAgentLoop` snapshots or commits] → D6 is a turn-loop extract, not `do`. Tests assert a skill call does not create a git snapshot, a commit, or an OpenSpec archive.
+- [Skill lists `finish` and launders an unverified mutation] → conformance forbids `finish` in `tools`; nested path never consults `finishRequest` for commit; parent ledger still blocks.
+- [File split of `tools.ts` is a large mechanical diff] → one commit-worth of moves with behavior-preserving flatten; existing handler strings become `summary`/`detail` so golden tests that match result text stay stable.
+- [Third-party bytes on the render path in later phases] → redaction at envelope construction now, with a test that a key-shaped token in `data` never reaches the renderer verbatim. Phase 2 rides this.
+- [`check` accidentally imports the registry] → module-graph guard stays pointed at `src/commands/check.ts`; `src/capabilities/` is not on that graph.
+
+## Migration Plan
+
+1. Land types, envelope helpers (including redaction-at-construction), and `ToolRegistry` wrapping the current `TOOLS` array with `gate: (ctx) => !requiresUnlock || ctx.editsUnlocked`. No behavior change.
+2. Wrap `HANDLERS` into the catalog, add `skills/*`, barrel + conformance. Flattened strings stay identical.
+3. Nested sub-run + `generate_report` + `copperhead skill list`/`run`.
+4. Renderer switches from regex to `ok`/`viewHint`.
+5. On archive: SPEC.md §4.2 documents the catalog (tools + skills) and §3 gains `skill`.
+
+Rollback: revert the change. No config flag — this is the new tool frame, not an opt-in. Existing `do`/`create`/`check` contracts are preserved by flattening and by leaving `STAGES` in place.
+
+## Open Questions
+
+- Skill names in the model catalog: `generate_report` (underscore, matches tools) vs `generate-report` (CLI kebab). Leaning underscore in the catalog and kebab in the CLI, mapped by the skill command, so the model sees one naming convention.
+- Whether inner skill turns share the parent's JSONL transcript as nested events or write a sibling `skill-<name>.jsonl`. Leaning nested events in the same transcript so one run dir stays the audit trail.
+- Default `maxTurns` for `generate_report`. Leaning 8 — enough for ERC+DRC+drift+a handful of reads, far under `do`'s 40.
--- a/openspec/changes/formalize-tool-registry/design.md
+++ b/openspec/changes/formalize-tool-registry/design.md
@@ -70,13 +70,13 @@ interface ToolResult {
 }
 ```
 
-Handlers return `ToolResult`. `dispatchTool` returns the envelope to the loop. The loop (1) records the envelope on the transcript, (2) flattens `summary` + `detail` (or `error.message`) into the string `Msg` tool content providers already consume, (3) hands the envelope to the renderer.
+Catalog handlers return `ToolResult`. The legacy `HANDLERS` table may return prose for operations whose outcome is the operation itself, or `{ ok, text }` when the handler already knows a separate domain outcome (ERC/DRC/drift/symbol/legibility diagnostics and multi-artifact exports). The barrel converts either form to one envelope before dispatch returns it. This prevents a known failing diagnostic from being re-inferred from prose. The loop (1) records the envelope on the transcript, (2) flattens `summary` + `detail` (or `error.message`) into the string `Msg` tool content providers already consume, (3) hands the envelope to the renderer.
 
 Flattening preserves today's gating-sync substrings: an unavailable edit tool's `error.message` still contains `"not available"` and `"unlock"`. The absence assertion `expect(registry.list(ctx).map(e => e.name)).not.toContain('edit_file')` is the invariant; the envelope `kind === 'unavailable'` assertion is additive, never a replacement.
 
 `data` is passed through `redactSecrets` (and the `*_KEY`/`*_SECRET`/`*_TOKEN` value set) at envelope construction. Renderers receive the already-redacted envelope. They never call `JSON.stringify` on unredacted handler output.
 
-Alternative: keep handlers returning strings and wrap at dispatch. Rejected — view hints and typed errors would be reconstructed from prose, which is the regex heuristic this change deletes.
+Unsuccessful domain outcomes without an invocation error keep their `detail` when flattened. Typed invocation errors still flatten to `error.message`.
 
 ### D5. Schema version on each entry; protocol version on the registry
 
@@ -92,6 +92,8 @@ The nested path is a turn-loop extract:
 - Own messages, own `maxTurns` (skill default or arg), own tool subset (`skill.tools` resolved against the registry).
 - No git snapshot, no commit, no `openspec archive`.
 - Completes when `isComplete(ctx, args)` is true or the turn budget is exhausted; the parent receives one envelope.
+- Provider turns inherit the main loop's bounded turn timeout and 429 retry policy. An exhausted timeout, rate limit, or other provider exception is converted by dispatch into a failed `exception` envelope; it never escapes the parent loop.
+- Inner tool results are retained in call order, including repeated calls to the same tool. An incomplete skill returns every partial result it gathered.
 - `finish` is forbidden in a skill's `tools` array (conformance test). The nested loop does not put `finish` in the child catalog. Even if a future caller forced it, `finishRequest` on the shared ctx MUST NOT trigger the parent's commit; the nested path never reads it for that purpose.
 
 Verification-gating stays with the parent: a skill that somehow mutated would open obligations on the shared ledger, and the parent's `finish` would still refuse. `generate_report` cannot mutate, so this is a structural backstop for later skills, not a generate_report concern.
@@ -133,13 +135,13 @@ copperhead skill list
 copperhead skill run generate-report [--scope power|all]
 ```
 
-`skill list` is LLM-free and network-free: it prints registered skills and whether `gate(ctx)` would include them. `skill run` invokes D6's nested sub-run (it needs a model, same as `do`) and prints the envelope. Tests call the exported runner with a `ScriptedProvider` (already used in `test/observability.test.ts`) so the suite stays offline. Missing key on `skill run` is a typed error naming the env var, not a stack trace.
+`skill list` is LLM-free, network-free, and filesystem-side-effect-free: it prints registered skills and whether `gate(ctx)` would include them without initializing a transcript. `skill run` invokes D6's nested sub-run (it needs a model, same as `do`) and prints the envelope. Tests call the exported runner with a `ScriptedProvider` (already used in `test/observability.test.ts`) so the suite stays offline. Missing key on `skill run` is a typed error naming the env var, not a stack trace.
 
 `check`/`verify` do not import `src/capabilities/`. The module-graph guard in `test/init-check.test.ts` keeps scanning outward from `src/commands/check.ts`.
 
 ### D10. Renderer reads `ok` / `viewHint`; retry maps from error kind
 
-`toolLine` uses `result.ok` for the glyph, not `/\b(clean|ok|pass(?:ed)?|success|done)\b/i`. `viewHint` selects layout (diagnostic vs mutation vs query vs export). `validation` and `refusal` are not retried; `exception` uses `src/util/retry.ts` the way provider 429s already do; `unavailable` is not retried (the catalog did not contain the tool).
+`toolLine` uses `result.ok` for the glyph, not `/\b(clean|ok|pass(?:ed)?|success|done)\b/i`. Diagnostics that already have a structured result set `ok` from that result (for example `CheckReport.ok`), never from a prefix in formatted prose. `viewHint` selects layout (diagnostic vs mutation vs query vs export). `validation` and `refusal` are not retried; thrown 429s use `src/util/retry.ts`; `unavailable` is not retried.
 
 ## Risks / Trade-offs
 
--- a/openspec/changes/formalize-tool-registry/proposal.md
+++ b/openspec/changes/formalize-tool-registry/proposal.md
@@ -1,0 +1,36 @@
+# formalize-tool-registry: Proposal
+
+## Why
+
+The tool system is a flat `TOOLS` array with boolean gating and plain-string results. Skills already exist as `STAGES` in `create`, but they are a second, hardcoded registry the model cannot see. Issue #246's later phases (part-research, MCP, host skills) need one catalog, versioned schemas, structured results, and a `gate()` predicate — none of which exist yet. Formalizing that frame now is what those surfaces reuse; building them against the flat list would freeze the wrong shape.
+
+## What Changes
+
+- **One two-tier catalog.** Atomic tools (tier 1) and composed skills (tier 2) register through the same path and appear as one tool list to the model. Dispatch picks the tier: a tool runs its handler; a skill runs a nested `runAgentLoop` and returns one envelope to the parent.
+- **`ToolRegistry` replaces the flat array.** Lookup by name, `list(ctx)` applying each entry's `gate()`, schema `version`, and a registry protocol version. `requiresUnlock: boolean` becomes `gate: (ctx) => boolean`.
+- **Structured `ToolResult` envelope** `{ ok, summary, detail?, data?, error?, viewHint? }`. Providers still receive a flattened string (backward compatible). The renderer reads `ok` / `viewHint` instead of regex-guessing pass/fail from prose.
+- **Typed error kinds** `unavailable` / `validation` / `refusal` / `exception`, with envelope construction as the redaction boundary (`data` never carries a secret-shaped payload to render or dock).
+- **Skill contract.** A skill declares its tool subset, prompt, completion gate, and optional turn budget. Its default `gate` is the conjunction of those tools' gates, so a skill that lists `edit_file` is structurally absent until the proposal validates — by construction.
+- **Proof skill `generate_report`.** Read-only (`read_file`, `search`, `list_nets`, `run_erc`, `run_drc`, `export_svg`, `check_drift`). Always available; cannot write. Nested loop returns the report as the envelope. `create`'s `STAGES` are **not** migrated in this change.
+- **CLI form** `copperhead skill run generate-report`, the same definition as the model-facing tool, so the skill is testable without a provider key.
+
+## Capabilities
+
+### New Capabilities
+
+- `tool-registry`: the two-tier catalog, versioned schemas, `ToolResult` envelope, `gate()` predicates, skill sub-run contract, `generate_report` as the proof skill, and the `copperhead skill run` CLI.
+
+### Modified Capabilities
+
+- `agent-core`: the loop consumes the registry; dispatch returns envelopes that flatten to today's string for providers; a skill call is a nested `runAgentLoop` whose parent sees one result, not the inner transcript.
+- `cli-surface`: the CLI gains `skill run <name>`.
+- `spec-gating`: gating is a `gate()` predicate on each catalog entry; a skill's effective gate is at least as strict as every tool it declares; the structural test remains absence from `list(ctx)`, not a dispatch-time `unavailable` error.
+- `safety-rails`: envelope `data` is redacted at construction; the render/dock path never receives a raw secret-shaped payload. `check`/`verify` stay LLM-free and network-free (AC-2.1).
+
+## Impact
+
+- **Code:** new `src/capabilities/` (`HANDLERS` table wrapped by `defineTool`, one file per skill, barrel); `src/agent/tools.ts` is registry + dispatch + skill sub-run; loop, providers, `render.ts` / `dock-renderer.ts` / `theme.ts` consume envelopes; `src/commands/skill.ts` for the CLI form.
+- **Tests:** `gating-sync` keeps the absence assertion (`edit_file` not in `list(ctx)`); envelope assertions are additive; conformance test over every registered entry; offline `skill run` via stub provider; redaction test that a key-shaped token never reaches the renderer.
+- **Docs:** SPEC.md §4.2 tool table becomes the catalog (tools + skills); skill CLI lands in §3.
+- **Unchanged contracts:** spec-gated edit tools remain structurally absent until a proposal validates; verification-gated completion is unchanged; a skill sub-run cannot clear parent obligations via `finish`; `check` imports no capability module; `STAGES` stay in `create.ts`.
+- **Out of scope:** part-research tools (#41 / `add-part-research-tools`), MCP server and host skills (#40 / `add-mcp-server`), migrating `create` stages onto the skill registry.
--- a/openspec/changes/formalize-tool-registry/specs/agent-core/spec.md
+++ b/openspec/changes/formalize-tool-registry/specs/agent-core/spec.md
@@ -3,7 +3,7 @@
 ## ADDED Requirements
 
 ### Requirement: Loop consumes the registry and flattens envelopes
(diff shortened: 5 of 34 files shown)

### CODE REVIEW

**Comment by animesh-chouhan** [comment]
🤖 **Automated pr-review pass** (Claude Code `/pr-review`) — not a human review, and not a formal approval or change request on the merge gate.

## Verdict: request changes

The two-tier catalog is a genuinely good frame — `gate()` as a predicate, the skill-conjunction default, sealing the envelope at construction, and keeping the provider wire shape as a flattened string are all the right calls, and the spec-gating invariant survives the refactor intact (I checked it, see below). Two things block: a skill sub-run's provider error escapes as an unhandled exception and skips the snapshot rollback, and the new envelope-driven glyph reports a **failing** ERC with a ✓.

Review ran in an isolated `git worktree` at commit `da565539`; the main checkout was never entered and stays on its current branch.

---

## Metrics

```
lines: +2433 / -837 (net 1596) across 34 files  ·  src +1425/-831, test +521/-5, docs +5/-1, spec +482/-0
new vs net: 21 new files (9 in src), 13 modified; 1250 new src lines (the net-new surface)
tests: +521 test lines; head suite 834 pass / 21 skip / 5 fail (local, environment-dependent — see note)
diff coverage: not measured (@vitest/coverage-v8 is not installed and could not be fetched)
ci: 4 pass / 0 fail / 0 pending (cla, label, template, test); required checks: none reported
mergeable: MERGEABLE · draft: false
```

**Method**

- measured in: `/tmp/pr-review-worktrees/copperhead-eea95c6f/pr-255` (review worktree, detached at `da565539`)
- merge base: `7e9b788be686866a6dc1863c8ac9dadb3e988a4f` (`git merge-base origin/main HEAD`)
- areas: `git diff --numstat` against the merge base, classified `src/` `test/` `openspec/` `docs+*.md`; generated files excluded
- head suite: `npx vitest run` in the worktree, 285s, actual run — **not** assumed
- base suite: not run; base CI is the reference
- diff coverage: **not measured.** `@vitest/coverage-v8` is absent from `node_modules` and installing it was not possible here. The untested-surface list below is therefore by inspection, not by instrument — treat it as a floor on the gap, not a ceiling.

**On the 5 local failures**: `test/init-check.test.ts` (ERC timeout) and `test/stage-completion.test.ts` (legibility gate) fail on this machine's KiCad build, matching the class of failure the PR description already documents. CI's `test` job passed on this exact head commit, which is what I'm treating as authoritative — I am not attributing these to the PR.

---

## Invariant checks

| Invariant | Result |
| --- | --- |
| **Spec-gated in** | **pass.** `edit_file`/`write_file` are absent from `registry.list(ctx)` while locked, by omission (`gate: (ctx) => ctx.editsUnlocked` in `src/capabilities/index.ts:42`), not by prompt text. The skill-conjunction in `src/agent/registry.ts:42-47` closes the obvious smuggling route, and `test/tool-registry.test.ts` proves a skill declaring `edit_file` is absent while locked. |
| **Verification-gated out** | **FAIL** — see finding 1. `finish` stays on the parent and is excluded from every skill catalog (`src/agent/registry.ts:52`), which is correct, but the rollback path is now bypassable. |
| **`check`/`verify` LLM-free and network-free** | **pass.** `src/commands/check.ts` imports no capability module, no provider, no SDK; the module-graph guard in `test/init-check.test.ts:170` now also rejects `capabilities`. |
| **No sexp serialization** | **pass** (N/A — `src/kicad/` untouched). |
| **Sync-obligations ledger** | **pass.** `markTouched` still feeds `ledger.onKicadEdit`/`onDocEdit`; sub-runs share the parent ledger and cannot reach `finish`. |
| **Secrets** | **pass.** `.gitignore` unchanged; `redactSecrets` still applied at transcript write time, and `seal()` now redacts summary/detail/data/error at envelope construction as well — a real tightening. |
| **Spec coherence** | **pass.** `openspec validate formalize-tool-registry` → valid. SPEC.md §3 and §4.2 moved with the delta specs. |

---

## Findings

### 1. high · correctness / invariant — a provider error inside a skill sub-run escapes as an unhandled exception, skipping the snapshot rollback

`src/agent/tools.ts:32-35`

The skill branch returns **before** the `try` that wraps atomic-tool dispatch:

```ts
if (entry.kind === 'skill') {
  if (!opts.provider) return failResult(...);
  return runSkillSubRun({ ctx, skill: entry, args, provider: opts.provider });  // outside try
}
try {
  return await withRetry(() => entry.handler(ctx, args), ...);
} catch (err) {
  return failResult('exception', ...);
}
```

`runSkillSubRun` awaits `provider.chat` at `src/agent/tools.ts:75` with no guard. The dispatch call site at `src/agent/loop.ts:592` is not wrapped either, and `runAgentLoop` (`src/agent/loop.ts:206-219`) has only a `finally` for provider cleanup — no `catch`.

**Failure scenario**: a `do` run unlocks edits, the model edits `hardware/board.kicad_sch`, then calls `generate_report` on a later turn. The provider returns a 502 (or the socket drops) during the nested `provider.chat`. The rejection propagates out of `dispatchToolResult` → out of the turn loop → out of `runWithProviders` → out of `runAgentLoop`. `fail()` never runs, so `preserveFailedRun` never stashes the work, `restore(repoRoot, snap)` never rolls back, and no `run-end` event or `summary.md` is written. The user is left with unverified edits in the working tree and no transcript explaining why. An identical 502 during any atomic tool becomes a tidy `failResult('exception', …)` and the run continues.

**Fix**: move the skill branch inside the `try`, so every dispatch path funnels through the same exception→envelope conversion.

**Repro test** (mirrors the existing `test/tool-registry.test.ts` 429 case):

```ts
it('a provider throw inside a skill sub-run becomes an envelope, not an unhandled rejection', async () => {
  const { repo, cleanup } = await tempFixtureRepo();
  try {
    await runInit({ repoRoot: repo });
    const ctx = await makeCtx(repo);
    const exploding: Provider = {
      name: 'exploding',
      async chat() { throw new Error('provider 502'); },
    };
    const env = await dispatchToolResult(ctx, 'generate_report', {}, { provider: exploding });
    expect(env.ok).toBe(false);
    expect(env.error?.kind).toBe('exception');
  } finally {
    await cleanup();
  }
});
```

### 2. high · correctness — every diagnostic tool now renders the ✓ success glyph even when it fails

`src/agent/envelope.ts:86-103`, `src/agent/theme.ts:68-78`

`textResult` marks a handler string as a failure only when it starts with `error:`, `rejected:`, `edit REVERTED`, `validation FAILED`, `refused:`, `cannot finish yet`, or `proposal validated but human declined`. `formatViolations` (`src/kicad/report.ts:69-81`) returns none of those on failure — a failing ERC is `"ERC: 3 violation(s)\n  error unconnected_pin: …"`. So `ok` is `true`, and `toolLine` picks its glyph straight off `ok`.

**Failure scenario**: `run_erc` finds 3 unconnected pins. Scrollback prints `✓ run_erc  ERC: 3 violation(s)`. The previous regex — `clean|ok|pass|success|done` AND not `fail|error|violat` — got this right and printed `▸`. The same inversion hits `run_drc`, `check_drift`, `score_schematic`, `verify_symbols`, and `check_legibility`: every diagnostic reads ✓ regardless of outcome, and the only ones that can ever read ▸ are the handful with a hardcoded error prefix. On a repo whose whole thesis is verification visibility, the pass/fail signal in the transcript is now inverted for the tools that matter most. `finish` still refuses correctly — this is display, not gating — but the user watching the run is told the opposite of the truth.

The PR description names this as a design win (D4/D10: "`toolLine` uses `ok` for the glyph, not a pass/fail regex on prose"). The idea is right; the `ok` it reads is wrong, because `textResult` infers success from the absence of a prefix rather than from the tool's actual outcome.

**Fix**: let handlers report their own outcome rather than inferring it from prose. The narrow version is to give `run_erc`/`run_drc`/`check_drift` (and the other diagnostics) real envelopes returning `ok: report.ok`; the structural version is to widen the `HANDLERS` signature to return `{ ok, text }`. Either way, `textResult`'s prefix sniffing should not be the only failure detector for a diagnostic that already knows whether it passed.

**Repro test**:

```ts
it('a failing ERC does not render the success glyph', () => {
  setColorEnabled(false);
  const r = textResult('ERC: 3 violation(s)\n  error unconnected_pin: pin 2 of U1', 'diagnostic');
  expect(r.ok).toBe(false);
  expect(toolLine('run_erc', r.summary, r.ok)).toContain('▸');
});
```

### 3. medium · weakened verification — the skill sub-run's provider call has no timeout and no rate-limit retry

`src/agent/tools.ts:72-75`

The main loop wraps every provider call in `withRetry(() => withTimeout(() => provider.chat(...), config.turnTimeoutMs, () => provider.close?.()))` (`src/agent/loop.ts:504-513`), with a bounded `maxTurnTimeouts` retry on `TurnTimeoutError`. The nested loop calls `await provider.chat(messages, tools)` bare.

**Failure scenario**: the provider hangs mid-stream on turn 3 of a `generate_report` sub-run. `config.turnTimeoutMs` does not apply inside the sub-run, so nothing aborts the call or tears down the subprocess; the whole `do` run hangs indefinitely with no watchdog and no recovery. Separately, a 429 that the main loop would back off and retry instead throws straight through (compounding finding 1). The sub-run is a turn loop — it should inherit the turn loop's recovery policy.

### 4. medium · correctness — an incomplete skill report silently discards every result it gathered

`src/agent/envelope.ts:73-79`, `src/agent/tools.ts:105-111`

`runSkillSubRun` returns `seal({ ok: complete, summary, detail, viewHint })` with **no `error` field**. `flatten` short-circuits on `!ok`:

```ts
if (!result.ok) return result.error?.message ?? result.summary;
```

so it falls through to `summary` and drops `detail` entirely.

**Failure scenario**: `generate_report` runs ERC, DRC and `list_nets` but burns its 8 turns before calling `check_drift`, so `isComplete` is false. `copperhead skill run generate-report` prints exactly `error: design report incomplete` and exits 1 — every ERC violation, DRC violation and net it actually collected is thrown away. Inside a `do` run the parent model gets the same single useless line. The user has no way to see the partial report, and no way to tell "the model stalled" from "the design is broken."

**Fix**: either give the incomplete envelope an `error` whose message carries the detail, or have `flatten` append `detail` when a failed envelope has one.

### 5. low · correctness — repeated calls to the same tool inside a skill overwrite each other in the report

`src/agent/tools.ts:65, 84, 100-104`

`results` is a `Map` keyed by tool **name**, so the second `read_file` erases the first, and the report is assembled in `skill.tools` declaration order rather than call order. `generate_report` lists `read_file` and `search`; a model that reads three files contributes only the last one, with no indication anything was dropped. Key by call id (or push to an array) and preserve call order.

### 6. low · side effect — `copperhead skill list` writes a run directory despite being documented LLM-free and side-effect-free

`src/commands/skill.ts:20-21, 67-90`

`listSkills` builds a `minimalCtx`, which constructs a `Transcript` and awaits `init()` — `mkdir(.copperhead/runs/<ts>)` plus an empty `transcript.jsonl` (`src/agent/transcript.ts:81-84`). The transcript is never used by the listing. SPEC.md §3 advertises `skill list` as "LLM-free, network-free" alongside `check`, the CI-safe command; every CI invocation now litters an empty run directory. Give `listSkills` a context without a transcript, or lazily init.

### 7. low · misleading message — `unavailable()` blames the spec lock for tools that are merely outside the skill's list

`src/agent/tools.ts:29`

When a skill sub-run rejects an off-list call, it returns `unavailable(name, ctx.editsUnlocked)`, whose message is `tool "list_symbols" is not available (edit tools unlock after the proposal validates)`. Inside `generate_report` the tool is not gated at all — it just is not in the skill's seven. Telling the model a proposal will unlock it invites a `propose_change` attempt that the skill cannot make. Pass a reason so the sub-run path says "not part of this skill".

### 8. low · fragility — `ToolRegistry`'s constructor mutates the shared catalog entries

`src/agent/registry.ts:16-22`

The constructor assigns `e.gate = conj` / `e.ownGate = conj` on entry objects it does not own, with `conj` closing over `this`. `catalog` is a module singleton, and `registry.all()` hands back those same objects — so `new ToolRegistry([...registry.all(), smuggle])` (exactly what `test/tool-registry.test.ts:120` does) rebinds the global `generate_report`'s gate to the *new* registry's conjunction. It's benign today because both registries hold the same tools, but any registry built over a subset silently corrupts the singleton, and it makes the tests order-dependent. Clone the entry, or hold the binding in a registry-side `Map` instead of on the entry.

### 9. low · test-coverage — the skill-gate conformance test is vacuous for real skills

`test/capability-conformance.test.ts:73-84`

`if (s.ownGate(ctx)) expect(conj, s.name).toBe(true)` cannot fail for any catalog skill: finding 8's constructor already set `ownGate` to that very conjunction, so this asserts `conj === conj`. Only the synthetic `weak` case at line 86 exercises the real check, and it never goes through `registry.list`. Assert on `registry.list(ctx)` with a weak-gated skill registered, so the test would actually catch a skill that widens its own gate.

### 10. low · test-coverage — untested surface (by inspection; coverage was not measurable)

No test reaches any of these new paths:

- `src/agent/tools.ts:33` — the `skill "…" requires a provider` validation branch.
- `src/agent/tools.ts:77-81` — the tool-less-turn nudge inside the sub-run. `ScriptedProvider` always returns a turn with tool calls, so this never runs.
- `src/agent/tools.ts:105-111` — the `ok: false` / "design report incomplete" path (finding 4).
- `src/agent/envelope.ts:95-101` — `textResult`'s entire refusal branch. `test/envelope.test.ts:34` covers only the `error:` prefix; no test asserts what `textResult` does with a *failing diagnostic* string, which is why finding 2 shipped.
- `src/agent/registry.ts:13` — the duplicate-name throw.
- `src/agent/registry.ts:25-27` — `protocolVersion` has no caller anywhere in `src/` and no test. Either wire it into the envelope/transcript or drop it until #40 needs it.
- `src/commands/skill.ts:40-42` — the `skill "…" is not available in this repo` branch.
- `src/commands/skill.ts:62` — `formatSkillEnvelope`'s `--json` branch.
- `src/cli.ts:325-329, 351-354` — both `catch` branches of the new subcommands.

---

## Fix prompt

````text
You are fixing review findings on PR #255 (branch fix/246) of the copperhead repo — a
TypeScript CLI agent that edits real KiCad projects. Fix the findings below, most severe
first. Add a regression test for each correctness finding, matching the repo's existing
test style in test/tool-registry.test.ts and test/envelope.test.ts.

REPO INVARIANTS — do not break any of these while fixing:
1. Spec-gated in: edit_file/write_file must stay structurally ABSENT from the agent tool
   list until an OpenSpec proposal validates. Gated by omission from registry.list(ctx),
   never by prompt text.
2. Verification-gated out: no mutation is "done" until ERC (and DRC if the board changed)
   passes; repair up to maxRepairCycles, then roll back to the git snapshot. No path may
   skip verification or the rollback.
3. `check`/`verify` is contractually LLM-free and network-free. Nothing reachable from
   src/commands/check.ts may touch a provider, an API key, or the network.
4. The sexp parser in src/kicad/ is read-only. KiCad files are edited only via anchored
   exact-match text replace. No round-tripping.
5. Post-tool-call hooks keep feeding the sync-obligations ledger, and commit keeps
6. Transcripts and summaries redact sk-[A-Za-z0-9_-]+ at write time; keys live only in
   env vars; .gitignore keeps .env and .copperhead/runs/.
7. If you change spec-level behavior, SPEC.md and the change artifacts under
   openspec/changes/formalize-tool-registry/ must move together.

--- FINDING 1 (high, correctness/invariant) ---
src/agent/tools.ts:32-35 — a provider error inside a skill sub-run escapes as an
unhandled exception, skipping the snapshot rollback.

The `if (entry.kind === 'skill')` branch returns runSkillSubRun(...) BEFORE the try that
wraps atomic-tool dispatch. runSkillSubRun awaits provider.chat at src/agent/tools.ts:75
with no guard; the dispatch call site at src/agent/loop.ts:592 is not wrapped; and
runAgentLoop (src/agent/loop.ts:206-219) has only a `finally` for provider cleanup, no
`catch`.

Failure scenario: a `do` run unlocks edits, the model edits hardware/board.kicad_sch, then
calls generate_report on a later turn. The provider returns a 502 during the nested
provider.chat. The rejection propagates out of dispatchToolResult -> out of the turn loop
-> out of runWithProviders -> out of runAgentLoop. fail() never runs, so preserveFailedRun
never stashes the work, restore(repoRoot, snap) never rolls back, and no run-end event or
summary.md is written. The user is left with unverified edits in the working tree and no
transcript. The identical 502 during any atomic tool becomes a failResult('exception', …)
and the run continues.

Fix: move the skill branch inside the try so every dispatch path funnels through the same
exception->envelope conversion.

Regression test (add to test/tool-registry.test.ts):

  it('a provider throw inside a skill sub-run becomes an envelope, not an unhandled rejection', async () => {
    const { repo, cleanup } = await tempFixtureRepo();
    try {
      await runInit({ repoRoot: repo });
      const ctx = await makeCtx(repo);
      const exploding: Provider = {
        name: 'exploding',
        async chat() { throw new Error('provider 502'); },
      };
      const env = await dispatchToolResult(ctx, 'generate_report', {}, { provider: exploding });
      expect(env.ok).toBe(false);
      expect(env.error?.kind).toBe('exception');
    } finally {
      await cleanup();
    }
  });

--- FINDING 2 (high, correctness) ---
src/agent/envelope.ts:86-103 and src/agent/theme.ts:68-78 — every diagnostic tool now
renders the ✓ success glyph even when it fails.

textResult marks a handler string as a failure only when it starts with `error:`,
`rejected:`, `edit REVERTED`, `validation FAILED`, `refused:`, `cannot finish yet`, or
`proposal validated but human declined`. formatViolations (src/kicad/report.ts:69-81)
returns none of those on failure — a failing ERC is "ERC: 3 violation(s)\n  error
unconnected_pin: …". So ok is true, and toolLine picks its glyph straight off ok.

Failure scenario: run_erc finds 3 unconnected pins. Scrollback prints
"✓ run_erc  ERC: 3 violation(s)". The previous regex (clean|ok|pass|success|done AND not
fail|error|violat) printed ▸ correctly. The same inversion hits run_drc, check_drift,
score_schematic, verify_symbols and check_legibility: every diagnostic reads ✓ regardless
of outcome. finish still refuses correctly — this is display, not gating — but the user
watching the run is told the opposite of the truth.

The D4/D10 idea (glyph from `ok`, not a regex on prose) is right; the `ok` it reads is
wrong, because textResult infers success from the absence of a prefix rather than from the
tool's actual outcome.

Fix: let handlers report their own outcome rather than inferring it from prose. Narrow
version: give run_erc/run_drc/check_drift (and the other diagnostics) real envelopes
returning ok: report.ok. Structural version: widen the HANDLERS signature to return
{ ok, text }. Either way, textResult's prefix sniffing must not be the only failure
detector for a diagnostic that already knows whether it passed.

Regression test (add to test/envelope.test.ts):

  it('a failing ERC does not render the success glyph', () => {
    setColorEnabled(false);
    const r = textResult('ERC: 3 violation(s)\n  error unconnected_pin: pin 2 of U1', 'diagnostic');
    expect(r.ok).toBe(false);
    expect(toolLine('run_erc', r.summary, r.ok)).toContain('▸');
  });

--- FINDING 3 (medium, weakened verification) ---
src/agent/tools.ts:72-75 — the skill sub-run's provider call has no timeout and no
rate-limit retry.

The main loop wraps every provider call in
withRetry(() => withTimeout(() => provider.chat(...), config.turnTimeoutMs, () => provider.close?.()))
(src/agent/loop.ts:504-513), with a bounded maxTurnTimeouts retry on TurnTimeoutError. The
nested loop calls `await provider.chat(messages, tools)` bare.

Failure scenario: the provider hangs mid-stream on turn 3 of a generate_report sub-run.
config.turnTimeoutMs does not apply inside the sub-run, so nothing aborts the call or tears
down the subprocess; the whole `do` run hangs indefinitely with no watchdog. Separately, a
429 that the main loop would back off and retry instead throws straight through
(compounding finding 1).

Fix: the sub-run is a turn loop — give it the same withRetry + withTimeout recovery policy
as the main loop, reading config.turnTimeoutMs.

--- FINDING 4 (medium, correctness) ---
src/agent/envelope.ts:73-79 and src/agent/tools.ts:105-111 — an incomplete skill report
silently discards every result it gathered.

runSkillSubRun returns seal({ ok: complete, summary, detail, viewHint }) with NO error
field. flatten short-circuits on !ok:
    if (!result.ok) return result.error?.message ?? result.summary;
so it falls through to summary and drops detail entirely.

Failure scenario: generate_report runs ERC, DRC and list_nets but burns its 8 turns before
calling check_drift, so isComplete is false. `copperhead skill run generate-report` prints
exactly "error: design report incomplete" and exits 1 — every ERC violation, DRC violation
and net it actually collected is thrown away. Inside a `do` run the parent model gets the
same single useless line, and cannot tell "the model stalled" from "the design is broken."

Fix: either give the incomplete envelope an error whose message carries the detail, or have
flatten append detail when a failed envelope has one. Add a test asserting the flattened
incomplete report still contains the ERC output it gathered.

--- FINDING 5 (low, correctness) ---
src/agent/tools.ts:65, 84, 100-104 — repeated calls to the same tool inside a skill
overwrite each other in the report.

`results` is a Map keyed by tool NAME, so the second read_file erases the first, and the
report is assembled in skill.tools declaration order rather than call order.
generate_report lists read_file and search; a model that reads three files contributes only
the last one, with no indication anything was dropped.

Fix: key by call id (or push to an array) and preserve call order.

--- FINDING 6 (low, side effect) ---
src/commands/skill.ts:20-21, 67-90 — `copperhead skill list` writes a run directory despite
being documented LLM-free and side-effect-free.

listSkills builds a minimalCtx, which constructs a Transcript and awaits init() —
mkdir(.copperhead/runs/<ts>) plus an empty transcript.jsonl (src/agent/transcript.ts:81-84).
The transcript is never used by the listing. SPEC.md §3 advertises `skill list` as
"LLM-free, network-free" alongside `check`, the CI-safe command; every CI invocation now
litters an empty run directory.

Fix: give listSkills a context without a transcript, or make the transcript lazily
initialized.

--- FINDING 7 (low, misleading message) ---
src/agent/tools.ts:29 — unavailable() blames the spec lock for tools that are merely
outside the skill's list.

When a skill sub-run rejects an off-list call it returns unavailable(name,
ctx.editsUnlocked), whose message is
`tool "list_symbols" is not available (edit tools unlock after the proposal validates)`.
Inside generate_report the tool is not gated at all — it just is not in the skill's seven.
Telling the model a proposal will unlock it invites a propose_change attempt the skill
cannot make.

Fix: pass a reason so the sub-run path says "not part of this skill". Keep the existing
"not available" / "unlock" phrasing on the spec-gating path — test/gating-sync.test.ts
asserts both substrings.

--- FINDING 8 (low, fragility) ---
src/agent/registry.ts:16-22 — ToolRegistry's constructor mutates the shared catalog entries.

The constructor assigns e.gate = conj / e.ownGate = conj on entry objects it does not own,
with conj closing over `this`. `catalog` is a module singleton and registry.all() hands back
those same objects — so `new ToolRegistry([...registry.all(), smuggle])` (exactly what
test/tool-registry.test.ts:120 does) rebinds the global generate_report's gate to the NEW
registry's conjunction. Benign today because both registries hold the same tools, but any
registry built over a subset silently corrupts the singleton, and it makes the tests
order-dependent.

Fix: clone the entry, or hold the binding in a registry-side Map instead of on the entry.

--- FINDING 9 (low, test-coverage) ---
test/capability-conformance.test.ts:73-84 — the skill-gate conformance test is vacuous for
real skills.

`if (s.ownGate(ctx)) expect(conj, s.name).toBe(true)` cannot fail for any catalog skill:
finding 8's constructor already set ownGate to that very conjunction, so this asserts
conj === conj. Only the synthetic `weak` case at line 86 exercises the real check, and it
never goes through registry.list.

Fix: assert on registry.list(ctx) with a weak-gated skill registered, so the test would
actually catch a skill that widens its own gate.

--- FINDING 10 (low, test-coverage) ---
Untested surface. Coverage was not measurable (@vitest/coverage-v8 is not installed), so
this list is by inspection and is a floor, not a ceiling. No test reaches:

- src/agent/tools.ts:33 — the `skill "…" requires a provider` validation branch.
- src/agent/tools.ts:77-81 — the tool-less-turn nudge inside the sub-run. ScriptedProvider
  always returns a turn with tool calls, so this never runs.
- src/agent/tools.ts:105-111 — the ok:false / "design report incomplete" path (finding 4).
- src/agent/envelope.ts:95-101 — textResult's entire refusal branch.
  test/envelope.test.ts:34 covers only the `error:` prefix; no test asserts what textResult
  does with a failing diagnostic string, which is why finding 2 shipped.
- src/agent/registry.ts:13 — the duplicate-name throw.
- src/agent/registry.ts:25-27 — protocolVersion has no caller anywhere in src/ and no test.
  Either wire it into the envelope/transcript or drop it until #40 needs it.
- src/commands/skill.ts:40-42 — the `skill "…" is not available in this repo` branch.
- src/commands/skill.ts:62 — formatSkillEnvelope's --json branch.
- src/cli.ts:325-329, 351-354 — both catch branches of the new subcommands.

Add tests for these, prioritising the refusal branch and the sub-run paths.

--- VERIFICATION ---
Run all of these and report actual output; do not claim a pass you did not observe:

  npm run typecheck
  npm run build
  npm test
  npx openspec validate formalize-tool-registry

Note: test/init-check.test.ts and test/stage-completion.test.ts may fail locally on
KiCad 10 (lib_symbol_mismatch on the open-key fixture) — that is pre-existing and unrelated;
CI is the reference for those two. Every other test must pass, and the new regression tests
must pass.
````

**Comment by Devesh36** [comment]
@animesh-chouhan Thanks for the detailed review. Addressed all findings in `7728e4b`.

Key fixes:

- Moved skill dispatch under the common exception boundary, so nested provider failures become typed envelopes and cannot escape the parent rollback flow.
- Added explicit handler outcomes for ERC, DRC, drift, symbol, scoring, and legibility diagnostics, so failing checks no longer render a success glyph.
- Applied bounded timeout and 429 retry handling to nested provider turns.
- Preserved partial detail for incomplete skills and retained repeated tool results in call order.
- Made `skill list` filesystem-side-effect-free and corrected off-list tool messaging.
- Cloned registry entries instead of mutating the shared catalog, and rewrote the gate conformance test to exercise `registry.list(ctx)`.
- Added regression coverage for the previously untested validation, timeout, retry, incomplete-run, nudge, JSON, CLI catch, duplicate-name, protocol-version, and availability paths.

The behavior and test changes are reflected in `openspec/specs/SPEC.md` and the `formalize-tool-registry` change artifacts.

**Comment by Devesh36** [comment]
this one is working easy but the follow up pr for this is going to be fun and hard part .

**Comment by animesh-chouhan** [comment]
🤖 **Automated pr-review pass** (Claude Code `/pr-review`) — not a human review, and not a formal approval or change request on the merge gate.

## Verdict: approve with nits

This is a **follow-up pass** on the [earlier review](https://github.com/copperheadhq/copperhead/pull/255#issuecomment-automated) of `da565539`. `7728e4b4` resolves **all ten** prior findings, including both blockers — the skill branch now sits inside the exception boundary, and diagnostics carry their own `ok` instead of having it sniffed off their prose. I re-verified each one against the code rather than taking the summary at its word; the table below records that.

What is left is one medium (a provider handle the new `skill run` command never closes) and five lows, none of which block.

Review ran in an isolated `git worktree` at commit `7728e4b4`; the main checkout was never entered and stays on its current branch.

---

## Metrics

```
lines: +2820 / -836 (net 1984) across 34 files  ·  src +1471/-830, test +825/-5, docs +5/-1, spec +519/-0
new vs net: 21 new files (9 in src), 13 modified; 1280 new src lines (the net-new surface)
tests: +825 test lines; head suite 855 pass / 0 fail / 21 skip
diff coverage: not measured (@vitest/coverage-v8 could not be installed here)
ci: 4 pass / 0 fail / 0 pending (cla, label, template, test); required checks: none reported
mergeable: MERGEABLE (mergeable_state: clean) · draft: false
```

**Method**

- measured in: `/tmp/pr-review-worktrees/copperhead-eea95c6f/pr-255` (review worktree, detached at `7728e4b4`, `npm ci` run against the PR's own `package-lock.json`)
- merge base: `7e9b788be686866a6dc1863c8ac9dadb3e988a4f` (`git merge-base origin/main HEAD`)
- areas: `git diff --numstat` against the merge base, classified `src/` `test/` `openspec/` `docs+*.md`; generated files excluded
- head suite: `npx vitest run` in the worktree — **actual run, not assumed**. A bare run shows **3 failures** (`init-check` "clean fixture: everything green", `stage-completion` "legibility gate", `gating-sync` "finish blocks on open obligations"). All three are this machine's user-level `~/.config/kicad/10.0/sym-lib-table` adding `lib_symbol_mismatch` ERC warnings to the open-key fixture, which CI deliberately never has. Re-run with an isolated `KICAD_CONFIG_HOME` holding only `fp-lib-table`, the suite is **855 pass / 0 fail / 21 skip**. The 855/0/21 figure is the one to trust; the PR is not responsible for the 3.
- base suite: not run; base CI is the reference
- diff coverage: **not measured** — `@vitest/coverage-v8` is absent and could not be fetched. The coverage-shaped finding below is by inspection, so treat it as a floor, not a ceiling.

---

## Prior findings — verified resolved

| # | Prior finding | Status |
| --- | --- | --- |
| 1 | high · provider error in a skill sub-run escapes the rollback | **fixed.** The skill branch moved inside the `try` (`src/agent/tools.ts:33-37`) with `return await`, so a throw lands in the `failResult('exception', …)` catch. `runSkill` now dispatches through `dispatchToolResult` too (`src/commands/skill.ts:43`), so the CLI path inherits the same boundary. Regression test at `test/tool-registry.test.ts:286`. |
| 2 | high · failing diagnostics render the ✓ glyph | **fixed.** New `outcomeResult` (`src/agent/envelope.ts:82-90`) plus a widened `HandlerDef` returning `string \| {ok, text}`; `run_erc`, `run_drc`, `check_drift`, `verify_symbols`, `score_schematic`, `check_legibility` and the export handler now report their real outcome. Tests at `test/envelope.test.ts:49` and `test/tool-registry.test.ts:236`. |
| 3 | medium · sub-run provider call had no timeout / no 429 retry | **fixed.** `src/agent/tools.ts:78-96` wraps `provider.chat` in the same `withRetry`/`withTimeout` pair as the main loop, with a bounded `TurnTimeoutError` retry and a `skill-turn-timeout` transcript event. Tests at `test/tool-registry.test.ts:307` (429) and `:337` (hang). |
| 4 | medium · incomplete skill report discarded its results | **fixed.** `flatten` now short-circuits on `result.error` rather than `!result.ok` (`src/agent/envelope.ts:74`), so an `ok: false` envelope with no typed error keeps its `detail`. Tests at `test/envelope.test.ts:60` and `test/tool-registry.test.ts:368`. |
| 5 | low · repeated tool calls overwrote each other | **fixed.** `results` is an ordered array of `{name, result}` (`src/agent/tools.ts:66, 105, 121`); the two-`list_nets` case is asserted at `test/tool-registry.test.ts:392`. |
| 6 | low · `skill list` wrote a run directory | **fixed.** `minimalCtx(repoRoot, false)` skips `transcript.init()` for the listing path (`src/commands/skill.ts:21, 67-69`), asserted both before and after in `test/tool-registry.test.ts:455-459`. |
| 7 | low · `unavailable()` blamed the spec lock for off-list tools | **fixed.** `unavailable(name, editsUnlocked, reason?)` (`src/agent/envelope.ts:120-127`); the sub-run passes `'not part of this skill'`. Test asserts the proposal wording is *absent* (`test/tool-registry.test.ts:136`). |
| 8 | low · registry constructor mutated the shared catalog | **fixed.** Entries are cloned (`src/agent/registry.ts:12-16`) and the stray `e.ownGate = conj` write is gone. `test/capability-conformance.test.ts:109` pins the singleton's `gate` identity across a second registry construction. |
| 9 | low · vacuous skill-gate conformance test | **materially fixed.** `test/capability-conformance.test.ts:102-103` now builds an isolated registry with the weak-gated `smuggle_edits` skill and asserts it is absent from `list(ctx)` — the real check. (See new finding 3 for what is left on the older assertion.) |
| 10 | low · untested surface | **mostly fixed.** New tests reach the provider-validation branch, the tool-less-turn nudge, the incomplete-run path, the duplicate-name throw, `protocolVersion`, the `not available in this repo` branch and `formatSkillEnvelope`'s `--json` branch. The two `catch` branches in `src/cli.ts` remain uncovered (process-exiting CLI shells; acceptable). |

---

## Invariant checks

| Invariant | Result |
| --- | --- |
| **Spec-gated in** | **pass.** `edit_file`/`write_file` are absent from `registry.list(ctx)` while locked, by omission (`gate: (ctx) => ctx.editsUnlocked`, `src/capabilities/index.ts:42`), not by prompt text. The skill conjunction (`src/agent/registry.ts:42-47`, re-applied in `list` at `:52`) still closes the smuggling route, and the clone in the constructor means an isolated registry can no longer rebind the singleton's gate. |
| **Verification-gated out** | **pass** (was the failing one). Every dispatch path — atomic tool, skill sub-run, CLI `skill run` — now funnels through the `try`/`catch` in `dispatchToolResult`, so a provider fault becomes an envelope and the run reaches `fail()` → `preserveFailedRun` → `restore()`. `finish` stays on the parent and is excluded from every skill catalog (`src/agent/registry.ts:52`); `cannot finish yet` gating is unchanged (`src/capabilities/handlers.ts:759`). |
| **`check`/`verify` LLM-free and network-free** | **pass.** `src/commands/check.ts` imports only `node:*`, config, `kicad/`, `memory/`, `openspec/` — no provider, no SDK, no capability module. The transitive module-graph guard at `test/init-check.test.ts:170` rejects `capabilities` as well as `providers/`/`openai`/`@anthropic-ai`. |
| **No sexp serialization** | **pass** (N/A — `src/kicad/` parser and edit path untouched). |
| **Sync-obligations ledger** | **pass.** `markTouched` still feeds `ledger.onKicadEdit`/`onDocEdit`; `run_erc`/`run_drc`/`check_drift` still `clear()` their obligations and bump `repairCycles`; sub-runs share the parent ledger and cannot reach `finish`. |
| **Secrets** | **pass.** `.gitignore` still has `.env` (:1) and `.copperhead/runs/` (:3); `redactSecrets` at transcript write time is intact, and `seal()` additionally redacts summary/detail/data/error at envelope construction. |
| **Spec coherence** | **pass.** `openspec validate formalize-tool-registry` → *valid* in the worktree; SPEC.md §3/§4.2 and all five delta specs moved with the code, and `tasks.md` was updated in the fix commit. |

---

## Findings

### 1. medium · resource leak — `copperhead skill run` never closes its provider

`src/cli.ts:337-360`, `src/commands/skill.ts:46-59`

`providerForSkillRun` builds a provider with `makeProvider(model)` and hands it to `runSkill`; nothing ever calls `provider.close()`. The command then falls through to `process.exit(result.ok ? 0 : 1)`.

The main loop treats this as load-bearing: `runAgentLoop` wraps the whole run in `try { … } finally { for (const provider of providers) await provider.close?.() }` (`src/agent/loop.ts:206-219`), and the comment on `close()` in `src/agent/providers/claude-code.ts:245-251` names the reason — *"without it the process is orphaned and keeps writing to its temp cwd, which (with KiCad local history) was a source of the disk-fill halt (2.2/4.1, I8)."* `close()` is what removes the provider's `mkdtemp` working directory (`codex.ts:112-118`, `claude-code.ts:251+`, `cursor.ts:121+`).

**Failure scenario**: a user runs `copperhead skill run generate-report --model claude-code` — exactly the invocation in this PR's own manual test log. The sub-run completes, the envelope prints, `process.exit(0)` fires, and the `copperhead-cc-*` temp directory (plus any subprocess the abort would have killed) is left behind. Nothing in the `skill run` lifecycle reclaims it: `sweepStaleTempDirs` is called only from `src/commands/create.ts:770`, so the dirs accumulate until some later `create` run happens to sweep them, age-gated at 2h. Ten report runs leave ten orphaned trees.

**Fix**: close the provider in a `finally` around `runSkill` in the `skill run` action, mirroring `runAgentLoop`'s cleanup.

### 2. low · correctness — `verify_symbols` reports failure when nothing actually failed

`src/capabilities/handlers.ts:370-380`

Any non-empty `findings` array yields `ok: false`, but `no-library` findings are the *unverifiable* bucket, counted into `skipped` and explicitly excluded from `mismatches`. `verifySchematicSymbols` pushes a `no-library` finding for every lib_id whose library is absent (`src/kicad/symlib.ts:532-539`).

**Failure scenario**: on a machine without the referenced KiCad libraries installed (a CI container, a fresh clone), `verify_symbols` emits `verify_symbols: 0 verified, 4 unverifiable (library not installed), 0 issue(s) to reconcile:` and renders `▸`. The text says zero issues; the glyph says failure. That is the mirror image of the bug finding 2 fixed — a false red where there was a false green.

**Fix**: `ok: mismatches === 0`.

### 3. low · test-coverage — the surviving conformance assertion is now a latent false positive

`test/capability-conformance.test.ts:74-85`

`if (s.ownGate(ctx)) expect(conj, s.name).toBe(true)` was vacuous before because the constructor set `ownGate` to the conjunction itself. Dropping that write (correctly) makes it non-vacuous but wrong-directioned: for a skill that omits `gate`, `defineSkill` sets `ownGate = () => true` (`src/capabilities/define.ts:73`), so the assertion demands the conjunction hold under a *locked* ctx. The day someone adds a legitimate mutation skill listing `edit_file`, this test fails — even though `registry.list` correctly omits that skill, which is the behavior the suite should be protecting.

**Fix**: assert on membership instead — `if (registry.list(ctx).some((e) => e.name === s.name)) expect(conj).toBe(true)` — which is the property that actually matters and is what line 102-103 already does for the synthetic case.

### 4. low · test hygiene — the availability test mutates the singleton registry entry

`test/tool-registry.test.ts:482-497`

To reach the `not available in this repo` branch the test does `skill.gate = () => false` on the object `registry.get('generate_report')` returns, restoring it in `finally`. It is contained today (vitest isolates files, and the `finally` runs), but it re-introduces exactly the shared-mutable-catalog coupling finding 8 removed, and it will silently corrupt every later test in the file if an assertion ever throws outside the `try`.

**Fix**: construct an isolated `new ToolRegistry([...])` with a false-gated clone, the way `capability-conformance.test.ts:102` now does.

### 5. low · duplication — the sub-run's timeout budget is a second hardcoded `3`

`src/agent/tools.ts:86`, `src/agent/loop.ts:353`

`timeoutRetries++ < 3` happens to equal the main loop's `const maxTurnTimeouts = 3`, so behavior is consistent today. It is two independent literals for one policy, and the sub-run's is a bare magic number where the loop's is at least named. Lift the constant and import it, so tuning one tunes both. (`test/tool-registry.test.ts:359-360` hardcodes `4` calls/closes against it, so a drift here breaks a test rather than passing silently — which is the mitigating half.)

### 6. low · dead surface — two new exports have no `src/` caller

`src/agent/envelope.ts:116` (`isRetryableToolKind`), `src/agent/registry.ts:25-27` (`protocolVersion`)

Both are exercised only by tests. `isRetryableToolKind` duplicates a policy that `dispatchToolResult` implements inline (typed envelope failures return, exceptions retry via `withRetry`), and `protocolVersion` is forward-looking for #40. The fix commit answered the prior coverage note by adding a test rather than a caller, which pins the surface without making it load-bearing. Either wire them into the envelope/transcript, or drop them until #40 needs them.

---

## Fix prompt

````text
You are fixing review findings on PR #255 (branch fix/246) of the copperhead repo — a
TypeScript CLI agent that edits real KiCad projects. The two blocking findings from the
previous review round are already fixed in 7728e4b4; these are what remains. Fix them most
severe first. Add a regression test for each correctness finding, matching the repo's
existing test style in test/tool-registry.test.ts and test/envelope.test.ts.

REPO INVARIANTS — do not break any of these while fixing:
1. Spec-gated in: edit_file/write_file must stay structurally ABSENT from the agent tool
   list until an OpenSpec proposal validates. Gated by omission from registry.list(ctx),
   never by prompt text.
2. Verification-gated out: no mutation is "done" until ERC (and DRC if the board changed)
   passes; repair up to maxRepairCycles, then roll back to the git snapshot. No path may
   skip verification or the rollback.
3. `check`/`verify` is contractually LLM-free and network-free. Nothing reachable from
   src/commands/check.ts may touch a provider, an API key, or the network.
4. The sexp parser in src/kicad/ is read-only. KiCad files are edited only via anchored
   exact-match text replace. No round-tripping.
5. Post-tool-call hooks keep feeding the sync-obligations ledger, and commit keeps
6. Transcripts and summaries redact sk-[A-Za-z0-9_-]+ at write time; keys live only in
   env vars; .gitignore keeps .env and .copperhead/runs/.
7. If you change spec-level behavior, SPEC.md and the change artifacts under
   openspec/changes/formalize-tool-registry/ must move together.

--- FINDING 1 (medium, resource leak) ---
src/cli.ts:337-360 and src/commands/skill.ts:46-59 — `copperhead skill run` never closes
its provider.

providerForSkillRun builds a provider with makeProvider(model) and hands it to runSkill;
nothing ever calls provider.close(). The command then falls through to
process.exit(result.ok ? 0 : 1).

The main loop treats this as load-bearing: runAgentLoop wraps the whole run in
try { … } finally { for (const provider of providers) await provider.close?.() }
(src/agent/loop.ts:206-219), and the comment on close() in
src/agent/providers/claude-code.ts:245-251 names the reason — "without it the process is
orphaned and keeps writing to its temp cwd, which (with KiCad local history) was a source
of the disk-fill halt (2.2/4.1, I8)." close() is what removes the provider's mkdtemp
working directory (codex.ts:112-118, claude-code.ts:251+, cursor.ts:121+).

Failure scenario: a user runs `copperhead skill run generate-report --model claude-code` —
exactly the invocation in this PR's own manual test log. The sub-run completes, the
envelope prints, process.exit(0) fires, and the copperhead-cc-* temp directory (plus any
subprocess the abort would have killed) is left behind. Nothing in the skill run lifecycle
reclaims it: sweepStaleTempDirs is called only from src/commands/create.ts:770, so the dirs
accumulate until some later `create` run happens to sweep them, age-gated at 2h. Ten report
runs leave ten orphaned trees.

Fix: close the provider in a finally around runSkill in the `skill run` action, mirroring
runAgentLoop's cleanup. Add a test asserting close() is called exactly once on a provider
stub after a skill run, on both the ok and the failing path.

--- FINDING 2 (low, correctness) ---
src/capabilities/handlers.ts:370-380 — verify_symbols reports failure when nothing actually
failed.

Any non-empty findings array yields ok: false, but 'no-library' findings are the
*unverifiable* bucket, counted into `skipped` and explicitly excluded from `mismatches`.
verifySchematicSymbols pushes a 'no-library' finding for every lib_id whose library is
absent (src/kicad/symlib.ts:532-539).

Failure scenario: on a machine without the referenced KiCad libraries installed (a CI
container, a fresh clone), verify_symbols emits "verify_symbols: 0 verified, 4 unverifiable
(library not installed), 0 issue(s) to reconcile:" and renders ▸. The text says zero
issues; the glyph says failure. That is the mirror image of the bug the previous round's
finding 2 fixed — a false red where there was a false green.

Fix: ok: mismatches === 0. Add a test that a findings list containing only 'no-library'
entries yields ok: true.

--- FINDING 3 (low, test-coverage) ---
test/capability-conformance.test.ts:74-85 — the surviving conformance assertion is now a
latent false positive.

`if (s.ownGate(ctx)) expect(conj, s.name).toBe(true)` was vacuous before because the
constructor set ownGate to the conjunction itself. Dropping that write (correctly) makes it
non-vacuous but wrong-directioned: for a skill that omits `gate`, defineSkill sets
ownGate = () => true (src/capabilities/define.ts:73), so the assertion demands the
conjunction hold under a *locked* ctx. The day someone adds a legitimate mutation skill
listing edit_file, this test fails — even though registry.list correctly omits that skill,
which is the behavior the suite should be protecting.

Fix: assert on membership instead —
`if (registry.list(ctx).some((e) => e.name === s.name)) expect(conj).toBe(true)` — which is
the property that actually matters and is what line 102-103 already does for the synthetic
case.

--- FINDING 4 (low, test hygiene) ---
test/tool-registry.test.ts:482-497 — the availability test mutates the singleton registry
entry.

To reach the "not available in this repo" branch the test does `skill.gate = () => false`
on the object registry.get('generate_report') returns, restoring it in finally. It is
contained today (vitest isolates files, and the finally runs), but it re-introduces exactly
the shared-mutable-catalog coupling the previous round's finding 8 removed, and it will
silently corrupt every later test in the file if an assertion ever throws outside the try.

Fix: construct an isolated `new ToolRegistry([...])` with a false-gated clone, the way
capability-conformance.test.ts:102 now does.

--- FINDING 5 (low, duplication) ---
src/agent/tools.ts:86 and src/agent/loop.ts:353 — the sub-run's timeout budget is a second
hardcoded 3.

`timeoutRetries++ < 3` happens to equal the main loop's `const maxTurnTimeouts = 3`, so
behavior is consistent today. It is two independent literals for one policy, and the
sub-run's is a bare magic number where the loop's is at least named.

Fix: lift the constant to one exported binding and import it in both places, so tuning one
tunes both. Note test/tool-registry.test.ts:359-360 hardcodes 4 calls/closes against it.

--- FINDING 6 (low, dead surface) ---
src/agent/envelope.ts:116 (isRetryableToolKind) and src/agent/registry.ts:25-27
(protocolVersion) have no src/ caller.

Both are exercised only by tests. isRetryableToolKind duplicates a policy that
dispatchToolResult implements inline (typed envelope failures return, exceptions retry via
withRetry), and protocolVersion is forward-looking for #40. The previous round's coverage
note was answered by adding a test rather than a caller, which pins the surface without
making it load-bearing.

Fix: either wire them into the envelope/transcript, or drop them until #40 needs them.
State which you chose and why in the PR description.

VERIFICATION — run all of these and report actual output:
  npm run typecheck
  npm run build
  npm test
  npx openspec validate formalize-tool-registry

NOTE ON THE SUITE: a bare `npm test` on a machine with a user-level KiCad sym-lib-table
shows 3 environmental failures (init-check "clean fixture: everything green",
stage-completion "legibility gate", gating-sync "finish blocks on open obligations") from
lib_symbol_mismatch ERC warnings on the open-key fixture. They are not yours. Run with an
isolated KICAD_CONFIG_HOME containing only 10.0/fp-lib-table copied from
/usr/share/kicad/template/fp-lib-table to reproduce CI, where the suite is
855 pass / 0 fail / 21 skip.
````

animesh-chouhan:
🤖 **Automated pr-review follow-up** — merged as `26d9b51`.

Thanks @Devesh36 — `7728e4b4` cleared all ten findings from the first round, including both blockers, and I verified each against the code rather than the summary. CI was 4/4 green and the offline suite ran 855 pass / 0 fail / 21 skip on this head (with an isolated `KICAD_CONFIG_HOME`; the 3 failures a bare `npm test` shows on some machines are the local `sym-lib-table` / `lib_symbol_mismatch` environment issue, not this PR).

The [six remaining findings](https://github.com/copperheadhq/copperhead/pull/255#issuecomment-5560861628) — one medium, five low — are non-blocking, so they did not hold up the merge. They are **fixed and verified but not yet landed**: the patch could not be pushed to the fork branch from the review environment, so it will follow as a separate PR against `main`.

What that patch contains:

- **medium** — `copperhead skill run` never closed the provider it created. `close()` is what removes the saved-login backend's `mkdtemp` working directory and aborts its subprocess; `runAgentLoop` does it in a `finally` for exactly that reason (the I8 disk-fill mode), and `skill run` had no loop to inherit it from. Fixed with a `runSkillCli` that brackets the run — including the unknown-skill throw path — and carries an exit code out, since `process.exit` inside a `try` skips pending `finally` blocks.
- **low** — `verify_symbols` returned `ok: false` for any non-empty findings array, but `no-library` is the *unverifiable* bucket that `verifySchematicSymbols` documents as "never treated as a mismatch". On a machine without the referenced libraries it printed `0 issue(s) to reconcile` under a failure glyph — the mirror of the false green this round fixed. Now reports from the mismatch count.
- **low** — the surviving conformance assertion (`if (s.ownGate(ctx)) expect(conj).toBe(true)`) became a latent false positive once `ownGate` stopped being rebound to the conjunction: it would fail the first legitimate mutation skill someone adds, even though `registry.list` correctly omits it. Now asserts membership ⇒ conjunction open.
- **low** — the availability test's singleton `gate` swap is bracketed so no assertion can leak it.
- **low** — the two hardcoded turn-timeout budgets become one shared `MAX_TURN_TIMEOUTS`.
- **low** — dead `isRetryableToolKind` dropped (the dispatch 429/exception test still pins the real policy). `protocolVersion` **stays** — `specs/tool-registry/spec.md` requires the registry to expose it, so that half of the finding was wrong to raise as removable.

Three regression tests come with it, and the delta specs plus `tasks.md` move with the code (`cli-surface` gains the provider-release guarantee, `tool-registry` gains "unverifiable is not failure"). On merged `main` that patch runs typecheck, build, `openspec validate formalize-tool-registry`, and 892 pass / 0 fail / 21 skip.

Issue #233, opened 2026-08-09
Stacked passives draw as isolated islands: a 16-cap bank carries 26 power symbols where the human used 2


animesh-chouhan:
## What happened

After #224 (frame fit, text collisions, compaction), page size across the corpus floors at roughly 2.5x to 4x the human's total sheet area, with utilization above the 50% floor on most boards. The excess is no longer packing: it is the drawing idiom for stacked passives.

Every capacitor in a decoupling bank draws as an isolated island: its own rail symbol above, its own GND symbol below, no wiring to its neighbours. stickhub's shared VBUS bank is 16 caps that the designer drew hanging off ONE +5V trunk and ONE GND trunk, two power symbols total, with junction dots doing the counting. The engine emits 26 power symbols for the same bank and joins nothing. cm5_minima's top row is 28 such islands; jetson's banks are bigger. Each island is several times the drawn footprint of the human's wired equivalent, which is exactly the residual 2.5x to 4x.

Page-size comparison at #224's head (engine sheet over the human's total sheet area): interf_u 1.0x, ecc83 1.0x, royalblue 1.3x, cm5_minima 2.5x, tiny_tapeout 2.7x, openair-max 2.7x, kit-dev-coldfire 3.2x, stickhub 4.0x, complex_hierarchy 4.0x, multichannel 4.0x, pic_programmer 4.0x, sonde xilinx 4.0x.

## The fix: rail-bank trunks

Caps already placed in a decap row (the `decapOwner` reduction) that sit adjacent, at the same pin height, on the same power net, chain on one trunk:

- one vertical stub per pin (unchanged),
- horizontal trunk segments joining consecutive stub ends (junction dots appear automatically at interior meets),
- ONE power symbol and ONE visible value per run, at the first stub end; the PWR_FLAG stays on the net's first endpoint,
- every stub and trunk segment passes `touchesForeign` and a body-crossing check first; any failure splits the run or falls back to per-cap symbols, so a bank can never buy density with a merged net.

Deterministic in the IR, no LLM, no new config.

## How tscircuit handles the same problem

[...]

PR #224, opened 2026-08-08
Drawing quality: labels inside the frame, text collisions cleared, sheets compacted, passive banks (#220 phases 1, 2, 4 and #233)

animesh-chouhan:
## What

The drawing-quality work from #220 phases 1, 2 and 4, plus the packing levers from #233: edge-facing label text stays inside the frame, the residual text-collision family is cleared by guarded fallbacks, a mostly-empty sheet reflows onto a smaller one, and passive banks chain on one trunk. Plus one connectivity fix the corpus sweep surfaced along the way: the router now treats netless pins as connection points.

Full-corpus error findings, before (branch point e8f9fc0) and after: stickhub 1 to 0 on A2 instead of A1, interf_u 13 to 0 on A3 instead of A1, complex_hierarchy 5 to 0 on A2 instead of A1, cm5_minima 20 to 1, royalblue54L_feather 3 to 1. Every drafted board round-trips exactly.

## Why

#219 measured all three drafting real boards at composite 40, capped by error-severity legibility findings the IR has no lever over. #220 phase 1 closes the out-of-frame half; phase 2 targets the text-collision family (#210's slot-selection half landed earlier in #211); phase 4 is the packing half, where a sheet two sizes above the human's is the finding. #233 folds in the density levers that make those smaller sheets hold. The jetson fix closes a silently-wrong case: a drafted sheet whose netlist differs from the source, which is the one outcome the engine bars.

## Design

Nine commits, each independently verifiable in [engine.ts](src/kicad/draft/engine.ts):

- **Frame fit counts label text** (003e631): the band budget reserves each group's facing-label extents, and the final centering shift clamps against the true extent with label boxes included. Both are no-ops when the sheet has slack, so unbanded layouts do not move.
- **Three text fallbacks** (1569d97), each firing only when the existing passes leave text colliding, so clean boards stay byte-identical:
  - a power-value sweep runs LAST, after labels are final, and moves the one text item with no electrical meaning with the finished sheet as its obstacle set
  - label rides gain a shorten rung (a facing pair in a tight channel can only separate by retreating) and a deeper budget (4 to 8 units, enough to clear the routing channel under a connector)
  - a vertical stub may flip its text to the other side when a parallel trunk blocks every rung; the flip explicitly re-checks the same-point merge case its non-overlapping box would otherwise hide
  - a wired-net label may sit at any interior grid point of its own run, walked along the run's SEGMENTS so a candidate always lies on a wire of its own net
[...]

Issue #220, opened 2026-08-08
Drawing quality: the golden drawings are wired, dense, and anchored; the drafts are labelled islands. Comparison and phased plan


animesh-chouhan:
## What this is

A side-by-side reading of the golden drawings (the schematics humans drew) against what the engine drafts from the same netlists, and a phased plan to close the gap. Produced from `npm run realdesigns` on `test/real-netlist-corpus` at e8f9fc0. Connectivity is already exact on all three drafting boards, so this issue is entirely about the drawing.

This issue is the umbrella record for the comparison and the tracking issue for the pieces that have no home yet: **wiring style (phase 3)** and **the density metrics (phase 6)**. The other phases land against #219, #210, and #207.

## The measured state

All three boards reproduce the source netlist exactly (zero pin groups lost or gained), and all three composite at 40, capped below the known-good floor of 85 by error-severity findings. After de5aa02, stickhub's out-of-frame count fell from 367 to 1; the text-collision family was untouched by design.

| Board | Parts | Nets | Human's paper | Engine's paper | Wires | Labels | Error findings |
| --- | --- | --- | --- | --- | --- | --- | --- |
| stickhub | 94 | 45 | A3, one sheet | A1 | 267 | 197 | 1 out-of-frame |
| cm5_minima | 84 | 95 | 8 sheets, hierarchical | A0, flat | 452 | 205 | 20 text-collision |
| interf_u | 24 | 110 | A3, one sheet | A1 | 322 | 240 | 13 text-collision |

interf_u is the control case: the same part count as the reference board buck-12v-5v, but 110 nets against 11. The trigger for collapse is routing and label density, not part count.

## The same circuit, twice

The human's drawing of cm5_minima's USB subsystem, one of its eight sheets:

![Human's cm5_minima USB sheet: dense, wired, annotated](https://raw.githubusercontent.com/chouhanindustries/copperhead/e8f9fc0f35f0f8b5a390589aee91d09a421e0aa4/manual-tests/real-designs/examples/cm5_minima/render/original-CM5_MINIMA_3-USB.png)

The engine's draft of the whole board, flattened onto one A0:

![Engine's cm5_minima draft: banded label islands on a flat A0](https://raw.githubusercontent.com/chouhanindustries/copperhead/e8f9fc0f35f0f8b5a390589aee91d09a421e0aa4/manual-tests/real-designs/examples/cm5_minima/render/drawn.png)

[...]

Issue #219, opened 2026-08-08
Layout collapses past reference-board scale: 367 out-of-frame findings on a 94-part board, correct netlist, unreadable sheet


animesh-chouhan:
## What happened

The engine draws the right circuit and an unreadable sheet once a board is bigger than the reference corpus. Connectivity is exact; layout is not usable.

Found by drafting real designs (#216) and then looking at the renders next to the originals, which is what `npm run realdesigns` now produces. The automated sweep calls these boards a pass, and by its contract they are: draft faithfully or refuse legibly. Faithfulness is not legibility.

All three real boards that draft come out at composite **40**, capped below the known-good floor of 85, against **85.6 to 92 with zero error findings** for the committed reference boards.

| Board | Parts | Nets | Paper | Netlist | Composite | Error findings (true count) |
| --- | --- | --- | --- | --- | --- | --- |
| stickhub | 94 | 45 | A0 | exact | 40 | **367** out-of-frame |
| cm5_minima | 84 | 95 | A0 | exact | 40 | 20 text-collision |
| interf_u | 24 | 110 | A1 | exact | 40 | 13 text-collision |
| buck-12v-5v (reference) | 24 | 11 | A3 | n/a | 85.6 | 0 |
| ldo-demo (reference) | 12 | 8 | A4 | n/a | 92 | 0 |

Note the counts are the true ones, read from `suppressed`. The findings list shows ten per family per sheet because of `familyCap`, so `checkLegibility` reports "10" for all three and the real figure for stickhub is 367.

## The stickhub case, concretely

94 parts, 45 nets, netlist reproduced exactly (zero pin groups lost or gained). The sheet:

- every part strung out in essentially **one horizontal ribbon** across the top, with a few short rows beneath it
- roughly **85% of the A0 sheet empty**
- the engine grew the paper **four sizes past the A3 the designer used**, rather than wrapping into a block
- **367 out-of-frame findings**: content is placed outside the drawing frame, so it would not plot
- almost every connection reduced to a label (`Net-(D15-1)`, `Net-(D16-1)`) rather than a wire

The human's A3 of the same circuit puts U1 as a central block with pin names visible, USB connectors down the right edge, LED drivers in a column, the decoupling bank under a "Shared VBUS capacitance" note, and the ESD caps in a labelled box. It is dense and reviewable.

## Why this was invisible until now

[...]

PR #264, opened 2026-09-04
Draft compaction, IC-first layout, boxes that enclose their text, and a wiring-style score

animesh-chouhan:
## What

Three commits, each reviewable alone:

1. **Checker measures free text at its declared justification.** Group captions are `justify left top`; measuring them centred flagged a caption 12 mm inside the frame as outside it.
2. **Drafting engine: compact honestly, lead with the IC, keep text inside its box.** The decoupling bank now counts against the column-height budget and may widen under it; a failed compaction says which sheets it tried and by how much they missed, and the report carries a `sheetFit` verdict; columns of four or more cells balance to the IC's height or the group's square side; ICs anchor the first column after the connectors, at its top, with the other members at their signal-net hop distance; group boxes enclose their members' labels, fields and rail symbols; horizontal rail names run along their own row; a module's name stays off its pin names; the engine reserves text at 0.8 of its height and measures a label as the union of the checker's centred box and the box it stands in on paper.
3. **Scorer measures wiring style and scores any sheet on one scale.** `pin-attachment`, `island-parts`, `power-symbol-economy`, `labels-per-part`; an uncapped wiring-style composite; `copperhead score schematic --file`; `npm run scoresheets`; SPEC AC-16.33/34; docs.

## Why

A 286-symbol stage-4 sheet drafted as one ribbon on A0 at 7% ink utilization with an empty notes list, ERC-clean and legibility-clean. The compaction pass had tried A5 through A1, missed A1 by 8 mm because the decoupling bank sits outside the column budget, restored the ribbon silently, and the agent's paper hint was a no-op. Instrumented run:

```
natural A0  util 0.073  ribbon 955×355
try A1 frac 0.5  usable 821×544  wrapped 738×552   power-input=162×329
-> no smaller sheet fits; natural placement restored; notes: []
```

Fixing that exposed the rest in turn, each visible only on a plotted sheet: every group a full-height strip beside a short IC, every IC in the rightmost column because C/D/L/Q/R sort before U, labels running out through box edges, a rail name on the row of the next pin's label, and facing labels overlapping on paper while the checker (0.6 advance, tuned never to false-positive) called them clear.

The scorer then turned out unable to see the biggest remaining gap to a hand-drawn sheet: a resistor wired to its IC pin and one hung on two labels scored the same on every metric, and every hand-drawn sheet in the KiCad demo corpus scores 40, capped for lacking group boxes.

## Measured

Same intent file, before and after:

| | before | after |
| --- | --- | --- |
| paper | A0 | A1 |
| frame covered | 0.37 | 0.84 |
| text overlaps at paper width | 9 pairs | 0 |
| labels crossing a group box | yes | none |
| legibility findings | 1 advisory | 0 |
| gated score | 74.6 | 76.9 (78.4 on the golden with the new metrics) |

Wiring-style calibration on the fifteen KiCad demo projects: people attach 0.90 to 1.00 of their two-pin parts and leave 0.00 to 0.02 as islands; the engine's reference boards attach 0.14 to 0.80. That is the target for the pin-attachment work (#220 phase 3), which this PR measures but does not implement.

Reference boards: all four re-pinned and reviewed; buck-12v-5v moves A3 to A4 at a higher score, the others keep their paper. Golden floors re-pinned (Tier A 99 to 86, 76 to 66; Tier C 90.2 to 78.4) because attachment now weighs in the gated composite. Gates added in `test/draft-sheet-fit.test.ts`: bank in budget, failure note, balance invariants, IC leads its group, no text crosses a group box, no two text items overlap at paper width, every reference board explains a low-ink sheet.

Related: #219, #220 (phases 3 and 6), #224, #233.


animesh-chouhan:

## Verdict: approve with nits

The three commits do what they say, the invariants all hold, and the drafting improvements are gated by real tests. The findings below are one scoring defect worth fixing before the metric is used for calibration, one layout pass that does not compute the distance it documents, a calibration script that measures the wrong sheets, and a set of spec and test-quality nits. None of them block the drafting work.

**One thing I want to state plainly, because it looks alarming in the diff and is not:** the golden floors drop (Tier A 99 to 86 and 76 to 66, Tier C 90.2 to 78.42) purely because `DEFAULT_WEIGHTS` grew from 100 to 115. Both new metrics score exactly 0 on all three goldens, and `100 x 99/115 = 86.09` and `100 x 90.2/115 = 78.43` reproduce the new pins to rounding. The floors are re-pinned as `Math.floor(composite)`, so the headroom is unchanged and Tier C still pins the exact composite with `toEqual`. **This is a faithful rescale, not a hidden regression**, and your description names the cause. See finding 8 for the one thing missing.

## Metrics

```
lines: +2506 / -1366 (net 1140) across 25 files  ·  src +618/-68, test +651/-153, docs +22/-0, spec +2/-0, other +1213/-1145
new vs net: 2 new files (0 in src), 23 modified; 0 new src lines (the net-new surface)
tests: +651 test lines; suite not measured -> 815/21/11 (pass/skip/fail, base -> head)
deps: +[none] -[none] ~[none] (package.json changed: the scoresheets script only, no dependency moved)
```

**Method.** Measured in a throwaway `git worktree` at head `249d3eb`; your checkout was never entered and HEAD was never moved. Merge base `7e9b788` (`git merge-base origin/main HEAD`). Areas from `git diff --numstat`; suite from `npx vitest run --reporter=json` in that worktree.

**Diff coverage: not measured, flagged rather than omitted.** The vitest v8 coverage run over the full suite did not finish here: this container has `kicad-cli` 10.0.6 on PATH and the ERC-backed tests time out before coverage is written. In its place I used the checklist's manual fallback, mapping each new exported symbol, branch and error path to the test that reaches it; findings 5, 11, 12 and the untested-surface list are that mapping's output.

**On the 11 local failures and the failing check.** The failures are environmental and I confirmed the mechanism rather than assuming it: this box has `kicad-cli` 10.0.6 while the repo pins the KiCad 8/9 format (`EMIT_VERSION = '20231120'`, `src/kicad/emit.ts:13-15`), so KiCad 10's `Device:R` does not match the vendored copy in `test/fixtures/open-key/` and a bare `kicad-cli sch erc` on that untouched fixture reports `2 warning | lib_symbol_mismatch`. Every gate built on that fixture then reports not-complete. I also measured 3, 4 and 10 failures on other commits of this stack, so the count is load-dependent too. CI's `test` job passes on this PR, and that is the signal I would trust. The failing `template` check is real: the description is missing the `## Testing` section and the invariant checklist that `.github/pull_request_template.md` requires.

## Findings

### 1. [medium] The wiring-style composite awards full marks for metrics whose denominator is zero, so a sheet with nothing on it outscores every hand-drawn sheet

- **Where:** `src/kicad/score.ts:410-413` (the fallbacks) and `src/kicad/score.ts:456-464` (the fixed `styleTotal`).
- **Fails when:** any sheet with no two-pin parts. Each style ratio falls back to its *perfect* value when the denominator is zero (`twoPin === 0` gives `attachment = 1`, `islands = 0`; no parts gives `powerSymbols/parts = 0` and `labels/parts = 0`), while `styleComposite` always divides by the constant 100. So the sheet collects 50 of the 100 style points for measurements that never happened. **Verified directly:** I ran `scoreSchematic` on a file whose entire content is `(kicad_pcb (version 20240108) (generator "pcbnew"))` and got `wiring style: 100/100`, above every hand-drawn KiCad demo (`complex_hierarchy` tops out at 96.32). It is reachable on real data too: `npm run scoresheets` prints `demo:cm5_minima ... 2-pin attached "-" islands "-" ... style 65.92`, where 50 of those points are for absent measurements. This defeats the composite's stated purpose in AC-16.34 and `schematic-legibility.md` ("a hand-drawn sheet and an engine draft of the same circuit compare on one scale"), and it is the number `--file` and `scoresheets` put in front of a human.
- **Fix:** build `styleMetrics` from only the metrics with a real denominator and renormalize, the same `name in weights` pattern already used for the gated composite at `src/kicad/score.ts:447`: drop `pin-attachment`/`island-parts` when `ws.twoPin === 0` and `power-symbol-economy`/`labels-per-part` when `ws.parts === 0`, then set `styleTotal` to the sum of the surviving weights. Failing test: score a sheet with one 3-pin IC and no wires and assert `report.style.composite` is below 100.

### 2. [medium] The IC-anchored hop-distance pass is a FIFO BFS seeded with unequal distances, so it does not compute distance from the nearest IC and adds a spurious column

- **Where:** `src/kicad/draft/engine.ts:779-787`.
- **Fails when:** two ICs at different depths can both reach the same passive. `dist` is seeded with each anchor's own depth (these differ after IC-to-IC propagation), but the queue is a plain FIFO sorted only once, and `if (dist.has(o)) continue` freezes the first value written, so a member reachable cheaply through a shallow IC gets pinned to a deep IC's distance. Reproduced by running the engine: group `MAIN` = `U1, U2, U3` (MCU8) chained by signal nets `U1.3-U2.3`, `U2.4-U3.4` (depths 1, 2, 3), plus `R1` on `U1.5-R1.1`, `R2` on `R1.2-R2.1` and `R2.2-U3.5`. The block's own comment promises "its signal-net hop distance from the nearest IC", i.e. `R1 = 2`, `R2 = 3` via `R1`. The code dequeues `U3` (dist 3) before it expands `R1` (dist 2), writes `R2 = 4`, then skips the cheaper relaxation. Measured column axes: `53.34 [U1] | 93.98 [R1 U2] | 134.62 [U3] | 168.91 [R2]`, four columns where three are correct. A wider group can cost a paper size, which is what the rest of this PR is trying to reclaim.
- **Fix:** relax instead of freeze, inside the `for (const o of adj.get(k) ?? [])` loop: `const nd = dist.get(k)! + 1; if (icSet.has(o) || isConnector(instByKey.get(o)!.part)) continue; if (dist.has(o) && dist.get(o)! <= nd) continue; dist.set(o, nd); queue.push(o);`. Failing test: the intent above, asserting the distinct x of `/^[UR]\d+$/` symbols is 3.

### 3. [medium] `scripts/score-sheets.ts` scores a sub-sheet instead of the project root for 4 of the 15 demos the calibration is derived from

- **Where:** `scripts/score-sheets.ts:24-29` (`sheetOf`).
- **Fails when:** a KiCad project's root sheet is not named after its directory. `sheetOf` tries `<dir>/<dir>.kicad_sch` and otherwise takes `readdirSync(p).find(f => f.endsWith('.kicad_sch'))`, ignoring the sibling `.kicad_pro` that names the real root. **I verified the corpus layout directly:** `cm5_minima/` holds sub-sheet `CM5.kicad_sch` alongside root `CM5_MINIMA_3.kicad_sch` (named by `CM5_MINIMA_3.kicad_pro`), and `openair-max/` holds `Battery-Holder.kicad_sch` before root `One-Air-Max.kicad_sch`. So the script scores `CM5.kicad_sch` (3 parts, power/part 2.67, labels/part 47.33) instead of `CM5_MINIMA_3.kicad_sch` (81 parts, power/part 1.36, labels/part 3.01), and `Battery-Holder` instead of `One-Air-Max` (37 vs 204 parts). Same for `multichannel` (26 vs 36) and `tiny_tapeout` (25 vs 153). These are the numbers the `STYLE_WEIGHTS` calibration comment (`src/kicad/score.ts:60-68`) and the docs table are built from, and the documented ranges do not survive the correction: with the right root sheets `One-Air-Max` is 0.87 attachment (the docs claim 0.90 to 1.00) and `CM5_MINIMA_3` is 1.36 power/part (the docs claim 0.3 to 1.2). Separately, `readdirSync` is unsorted, so which sub-sheet gets picked is filesystem-order dependent.
- **Fix:** in `sheetOf`, look for `<basename>.kicad_pro` first and return the matching `<basename>.kicad_sch`; keep `<dir>/<dir>.kicad_sch` and a **sorted** first-`.kicad_sch` as later fallbacks. Then regenerate the docs table and state the ranges as observed min-max with the outliers named.

### 4. [medium] SPEC.md's canonical CLI block does not document `--file`, so the two halves of SPEC.md disagree

- **Where:** `openspec/specs/SPEC.md:275` versus `openspec/specs/SPEC.md:563`.
- **Fails when:** a reader uses section 3 as the CLI contract, which is what it is for. Line 275 still reads `copperhead score schematic [--json]`, while AC-16.34 at line 563 asserts `copperhead score schematic --file <sheet>` exists. Section 3 spells out optional flags for every other command (`copperhead init [--path hardware/]`, `copperhead demo [--tour] [--model ...] [--dir <path>]`), so the omission reads as "this flag does not exist".
- **Fix:** change line 275 to `copperhead score schematic [--file <path>] [--json]` and add a sentence to its description: scores any sheet outside a repo, reporting both composites.

### 5. [low] `score schematic --file` accepts any s-expression file, including a `.kicad_pcb`, and reports a confident score for it

- **Where:** `src/cli.ts:237`, with `src/kicad/sexp.ts:124-126`.
- **Fails when:** a user points the new flag at the wrong half of their project. `readSheetGeometry` only requires the root to be a list, never that its head is `kicad_sch`. **Verified:** scoring a file containing only `(kicad_pcb (version 20240108) (generator "pcbnew"))` exits 0 and prints `score: 40/100` and `wiring style: 100/100`. Missing files and plain text are handled correctly (`ENOENT...` and `not a KiCad s-expression file: ...`, both exit 1); only the wrong-format-but-valid-sexp case slips through.
- **Fix:** tighten `sexp.ts:124` to also require `atomAt(root, 0) === 'kicad_sch'` for the root file. Failing test: `await expect(scoreSchematic(pcbPath, { docsDir: null })).rejects.toThrow()`.

### 6. [low] `npm run scoresheets` dies with a raw ENOENT stack trace on a path that does not exist

- **Where:** `scripts/score-sheets.ts:24`, called from `:38`; the per-target `try/catch` that produces the tidy `error` row only starts at `:57`.
- **Fails when:** `npm run scoresheets -- /nonexistent/path.kicad_sch`. `sheetOf` calls `statSync(p)` unguarded and outside the catch, so the run dies with a Node stack and no table at all, while a path that exists but fails to parse prints a one-line `error` row and the rest of the table still renders.
- **Fix:** `if (!existsSync(p)) return null;` as the first statement of `sheetOf` (`existsSync` is already imported).

### 7. [low] The script's `crossings` column is an absolute count, but the docs cite the script as the source for `crossings-per-wire`

- **Where:** `scripts/score-sheets.ts:69`.
- **Fails when:** anyone tries to reproduce the documented calibration. The docs list `crossings-per-wire` as "0.03 on a 90-part board, 0.77 on an 1100-part one" and say the column comes from `npm run scoresheets`, but the script prints `wire-crossings`, an absolute count: `demo:stickhub 89 ... crossings 15` and `demo:jetson-agx-thor-baseboard 1138 ... crossings 3872`. `crossings-per-wire` can never appear in `report.metrics` anyway, because it is not in `DEFAULT_WEIGHTS` and `src/kicad/score.ts:447` filters to `name in weights`.
- **Fix:** use `r.style.metrics.find(x => x.name === 'crossings-per-wire')!.raw` in place of `m('wire-crossings')`, formatted with `f2`.

### 8. [low] The gated composite silently changed scale, and only the PR description records it

- **Where:** `src/kicad/score.ts:42-56` (`DEFAULT_WEIGHTS` now sums to 115, not 100).
- **Fails when:** anyone compares a composite recorded before this PR with one recorded after. Every previously recorded `check --json` composite and AC-16.21 run-summary composite is now on a different scale. As established above the re-pins are arithmetically faithful, so this is a documentation gap rather than a hidden regression, but nothing in `openspec/specs/SPEC.md` or `docs/.../schematic-legibility.md` says the denominator moved. Related: AC-16.33 at `SPEC.md:562` says the four new metrics appear in "the breakdown", using the same phrasing AC-16.32 uses for gated metrics, but `power-symbol-economy` and `labels-per-part` are not in `DEFAULT_WEIGHTS`, so `report.metrics` never contains them; they live only in `report.style.metrics`.
- **Fix:** one line in AC-16.33 and in `schematic-legibility.md:78` recording that the gated denominator went 100 to 115 and why, so the golden diff reads as a rescale. Reword AC-16.33 to say the *wiring-style* breakdown reports those four.

### 9. [low] `isIC` (used for column-balance anchor height) counts connectors while `groupIC` twelve lines earlier excludes them

- **Where:** `src/kicad/draft/engine.ts:892-894`, versus `groupIC` at `:752`.
- **Fails when:** a group holds a tall multi-pin header plus a dozen passives. `anchorH` becomes the header's cell height, so `balanceH = Math.max(anchorH, squareSide)` is tall and `splitColumn(col, balanceH, 4)` never fires, leaving exactly the "tall thin column beside a short IC with the box two-thirds air" shape the pass was added to fix. Reported as a verified code inconsistency with a derived consequence rather than a demonstrated regression: `test/fixtures/symlib` only ships a 3-pin connector, so I could not build the repro.
- **Fix:** `const anchorH = Math.max(0, ...members.filter(m => groupIC(m.key)).map(m => cellDims.get(m.key)!.h));`

### 10. [low] Four new public interface fields and their classifier are computed on every score and read by nothing

- **Where:** `src/kicad/score.ts:113` (`NOT_IC`), `:117`, `:122-127` (`ics`, `icPins`, `icPinsWired`, `icPinsLabelled`), filled at `:217-223`.
- **Fails when:** never at runtime, but they cost a pass over every pin of every 3-or-more-pin part on every `check`, and they ship an unexercised classifier that a future consumer would inherit untested (`NOT_IC` matches `Device:D_Schottky` but not plain `Device:D`, and `Device:Q_` but not `Device:Q`). `grep -rn "icPins\|NOT_IC\|\.ics" src/ test/ scripts/ docs/src` returns only the definition and assignment sites. This is the one piece of scope creep in the PR.
- **Fix:** delete them, or surface an "IC pins wired" column in `scripts/score-sheets.ts` so the classifier is at least exercised.

### 11. [low] The shelf-wrap assertion now compares the model against a note the same code emitted

- **Where:** `test/draft-layout.test.ts:81`.
- **Fails when:** the wrap becomes unbalanced. `expect(rows).toEqual([['G01'..'G04'], ['G05'..'G08']])` became `expect(rows.length).toBe(Number(/\d+/.exec(wrapNote!)![0]))`, but the note text is produced from `new Set(wrap.deltas.map(d => d.dy)).size` at `engine.ts:1596`, the same `wrap.deltas` the rects were shifted by, so the two can never disagree. With only that plus `rows.length >= 2` and the flattened order, a regression that wraps eight equal groups as 7 + 1 passes.
- **Fix:** add a shape assertion the engine's own note cannot satisfy for free: `expect(Math.max(...rows.map(r => r.length)) - Math.min(...rows.map(r => r.length))).toBeLessThanOrEqual(1);`

### 12. [low] The "text inside its box" gate measures labels with the centred half-height box, not the full-height on-paper box the engine reserves

- **Where:** `test/draft-sheet-fit.test.ts:241` and the same expression at `:305`, versus `src/kicad/draft/engine.ts:46-49`.
- **Fails when:** a label's plotted text crosses a group box's top edge by up to 0.635 mm. `labelReserveBox` reserves `minY = y - 1.27, maxY = y + 0.635` (a full height above the anchor, where the glyphs stand), but the test measures `minY = y - 0.635`. So the gate added for this invariant is half a text height weaker than the invariant, despite its comment saying "'inside the box' means on paper".
- **Fix:** in both helpers use `{ minY: l.y - H, maxY: l.y + H / 2 }`, matching `labelReserveBox`.

### 13. [low] A type assertion in production code papers over a test fixture that lies about its type

- **Where:** `src/kicad/draft/draft.ts:169`, forced by `test/draft-bank-guards.test.ts:161-163`.
- **Fails when:** `sheetFit` is later removed or renamed. `sheetFit` is declared non-optional on `SchematicDraftReport` (`engine.ts:139-144`), but the new formatter line reads `report.sheetFit as SchematicDraftReport['sheetFit'] | undefined` and guards with `if (sf)`. The reason is that the test builds `base` as a partial object with `as unknown as SchematicDraftReport`, omitting `sheetFit`. The cast makes the code safe at runtime but removes the compiler's ability to tell you when the field genuinely goes missing.
- **Fix:** give the fixture a real `sheetFit` (`{ paper: 'A4', inkUtilization: 0, compaction: 'not-needed', misses: [] }`) and drop both the cast and the guard.

### 14. [low] `--file` silently drops the repo's scorer config and docs vocabulary, so one file can report two different scores

- **Where:** `src/cli.ts:238`, which passes `{ docsDir: null }` and neither `config.legibility` nor `config.docs`.
- **Fails when:** a user inside a repo with tuned `legibility.score.weights` runs `copperhead score schematic --file hardware/board.kicad_sch` on their own configured schematic. They get a different composite, and different group-caption findings since the caption vocabulary comes from `docs/`, than plain `copperhead score schematic` on the same bytes, with no indication why.
- **Fix:** when `--file` resolves inside the repo and a config loads, reuse `config.legibility` and `config.docs`; otherwise print a one-line note that defaults are in force.

### 15. [low] The new `--file` surface has no test despite being named in an acceptance criterion

- **Where:** `openspec/specs/SPEC.md:563` names it; `grep -rn "\-\-file" test/` returns nothing and `test/cli-surface.test.ts` has no `score` coverage at all.
- **Fails when:** a refactor of the `score` action drops the `--file` branch. The whole suite stays green while AC-16.34 is silently unmet.
- **Fix:** add a `cli-surface` case invoking `score schematic --file test/fixtures/legibility/clean.kicad_sch` outside any repo, asserting exit 0 and a `style` block in `--json`.

## Invariant checks performed

All six checked against the diff, not assumed. `openspec validate build-copperhead-phase-1` returns **"Change 'build-copperhead-phase-1' is valid"** (exit 0); no planning artifacts changed, only `SPEC.md`.

- **Spec-gated in** - pass. No `src/agent/**` file in the diff. `src/agent/tools.ts:820` still reads `TOOLS.filter(t => !t.requiresUnlock || ctx.editsUnlocked)`, and `edit_file`/`write_file` keep `requiresUnlock: true`. The new surface is a CLI flag, not a tool; `score_schematic` remains `requiresUnlock: false` and read-only.
- **Verification-gated out** - pass. `src/agent/loop.ts`, `src/commands/do.ts` and `src/commands/create.ts` are untouched. The schematic stage's `isComplete` (symbols, drift, `runErc`, legibility, byte-exact re-draft) is unchanged; no new early-done path.
- **`check`/`verify` stays LLM-free and network-free** - pass, traced transitively rather than assumed. `src/commands/check.ts` reaches `config.ts`, `kicad/cli.ts` (and `util/preflight.ts`), `kicad/report.ts`, `kicad/legibility.ts`, `kicad/score.ts`, `kicad/sexp.ts`, `memory/constraints.ts`, `memory/drift.ts`, `memory/bom-table.ts`, `openspec/cli.ts`. The only external package in the whole closure is `execa`, pre-existing, for the `kicad-cli` and `openspec` subprocesses. `score.ts`'s new import (`pinsOfUnit`, `pinAbsolute`) comes from `sexp.ts`, already in the closure: zero new edges.
- **No sexp serialization** - pass. All 23 changed lines in `src/kicad/sexp.ts` add readers: `justifyH`/`justifyV` on `TextItem` (`:338-341`) and `(justify ...)` parsing in `readSheetGeometry` (`:557-567`), with `null` for property text. No emit or serialize function added, and `readSheetGeometry` still never writes. `src/kicad/draft/*` emits new files from its own IR through the unchanged `emit.ts`, which is the documented drafting capability, not round-tripping a parsed tree.
- **Sync-obligations ledger** - pass, untouched. The `ctx.ledger.onLegibilityResult(...)` post-tool hook is unchanged and no hook, ledger, or commit-gate file is in the diff.
- **Secrets** - pass. `.gitignore` untouched, still `.env` plus `.copperhead/runs/`; `src/util/redact.ts:7` `/sk-[A-Za-z0-9_-]+/g` untouched; no new transcript or summary writer. `scripts/score-sheets.ts` writes nothing and reads only `COPPERHEAD_CORPUS_DIR`.
- **Test determinism** - pass. No network, no `Date.now()` or `Math.random()` in any assertion (only in `mkdtemp` names), a fixed `'2020-01-01'` date argument, `docsDir: null`, and `new SymbolSource(repo, [])` so `symbolSearchDirs()` is never consulted. **No new test shells out to `kicad-cli` or depends on an installed KiCad.** Every new test asserts, including both arms of the branching "names each smaller sheet" test.
- **Reference-board gate** - worth noting as a strength: `test/draft-reference-boards.test.ts:41` asserts `res.text` is **byte-identical** to the committed reference, plus clean ERC and clean legibility. So the large fixture churn in this PR is genuine regeneration and a hand-edited reference cannot slip through.

**Untested surface** (from the manual mapping, since coverage did not complete): the `--file` branch of the `score` action reaches no test (finding 15); `scripts/score-sheets.ts` is not in `tsconfig.json`'s `include` and has no test at all, so findings 3, 6 and 7 are all in unchecked, unexercised code; the zero-denominator fallbacks at `score.ts:410-413` are reached by no test (finding 1); and the `isIC` branch at `engine.ts:892-894` has no fixture that can reach it (finding 9).

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 264, branch fix/draft-compaction-bank-budget,
based on main.

Check it out first:
  gh pr checkout 264

Context you should not "fix": the golden floor drops in this PR (Tier A 99 to 86
and 76 to 66, Tier C 90.2 to 78.42) are a faithful rescale, not a regression.
DEFAULT_WEIGHTS grew from 100 to 115, both new metrics score exactly 0 on all three
goldens, and 100 * 99/115 = 86.09 and 100 * 90.2/115 = 78.43 reproduce the new pins.
Leave the pins alone. Finding 8 is only about recording the scale change.

FINDING 1 [medium] The wiring-style composite awards full marks for metrics whose
denominator is zero, so a sheet with nothing on it outscores every hand-drawn sheet.
  Where: src/kicad/score.ts:410-413 (fallbacks) and src/kicad/score.ts:456-464
    (the fixed styleTotal)
  Fails when: any sheet with no two-pin parts. Each style ratio falls back to its
    perfect value when the denominator is zero (twoPin === 0 gives attachment = 1
    and islands = 0; no parts gives powerSymbols/parts = 0 and labels/parts = 0),
    while styleComposite always divides by the constant 100, so the sheet collects
    50 of the 100 style points for measurements that never happened. Verified: a
    file containing only "(kicad_pcb (version 20240108) (generator \"pcbnew\"))"
    scores wiring style 100/100, above every hand-drawn KiCad demo
    (complex_hierarchy tops out at 96.32). Reachable on real data:
    npm run scoresheets prints demo:cm5_minima with attachment "-" islands "-" and
    style 65.92.
  Fix: build styleMetrics from only the metrics with a real denominator and
    renormalize, the same "name in weights" pattern already used for the gated
    composite at src/kicad/score.ts:447. Drop pin-attachment and island-parts when
    ws.twoPin === 0, and power-symbol-economy and labels-per-part when
    ws.parts === 0, then set styleTotal to the sum of the surviving weights.
    Failing test: score a sheet with one 3-pin IC and no wires and assert
    report.style.composite is below 100.

FINDING 2 [medium] The IC-anchored hop-distance pass is a FIFO BFS seeded with
unequal distances, so it does not compute distance from the nearest IC and adds a
spurious column.
  Where: src/kicad/draft/engine.ts:779-787
  Fails when: two ICs at different depths can both reach the same passive. dist is
    seeded with each anchor's own depth (these differ after IC-to-IC propagation),
    but the queue is a plain FIFO sorted only once, and "if (dist.has(o)) continue"
    freezes the first value written. Repro: group MAIN = U1, U2, U3 (CopperMCU:MCU8)
    chained by signal nets U1.3-U2.3 and U2.4-U3.4 (depths 1, 2, 3), plus R1 on
    U1.5-R1.1, R2 on R1.2-R2.1 and R2.2-U3.5. The comment promises R1 = 2 and R2 = 3
    via R1; the code dequeues U3 (dist 3) before expanding R1 (dist 2), writes
    R2 = 4, and skips the cheaper relaxation. Measured column axes:
    53.34 [U1] | 93.98 [R1 U2] | 134.62 [U3] | 168.91 [R2], four columns where
    three are correct. A wider group can cost a paper size.
  Fix: relax instead of freeze, inside the "for (const o of adj.get(k) ?? [])" loop:
      const nd = dist.get(k)! + 1;
      if (icSet.has(o) || isConnector(instByKey.get(o)!.part)) continue;
      if (dist.has(o) && dist.get(o)! <= nd) continue;
      dist.set(o, nd); queue.push(o);
    Failing test: the intent above, asserting the number of distinct x values among
    symbols matching /^[UR]\d+$/ is 3.

FINDING 3 [medium] scripts/score-sheets.ts scores a sub-sheet instead of the project
root for 4 of the 15 demos the calibration is derived from.
  Where: scripts/score-sheets.ts:24-29 (sheetOf)
  Fails when: a KiCad project's root sheet is not named after its directory. sheetOf
    tries <dir>/<dir>.kicad_sch and otherwise takes
    readdirSync(p).find(f => f.endsWith('.kicad_sch')), ignoring the sibling
    .kicad_pro that names the real root. Verified against the corpus: cm5_minima/
    holds sub-sheet CM5.kicad_sch alongside root CM5_MINIMA_3.kicad_sch, and
    openair-max/ holds Battery-Holder.kicad_sch before root One-Air-Max.kicad_sch.
    So the script scores CM5.kicad_sch (3 parts, power/part 2.67, labels/part 47.33)
    instead of CM5_MINIMA_3.kicad_sch (81 parts, power/part 1.36, labels/part 3.01),
    and Battery-Holder (37 parts) instead of One-Air-Max (204 parts). Same for
    multichannel (26 vs 36) and tiny_tapeout (25 vs 153). These feed the
    STYLE_WEIGHTS calibration comment at src/kicad/score.ts:60-68 and the docs
    table, and the documented ranges do not survive the correction: One-Air-Max is
    0.87 attachment against a claimed 0.90 to 1.00, and CM5_MINIMA_3 is 1.36
    power/part against a claimed 0.3 to 1.2. readdirSync is also unsorted, so which
    sub-sheet is picked is filesystem-order dependent.
  Fix: in sheetOf, look for <basename>.kicad_pro first and return the matching
    <basename>.kicad_sch; keep <dir>/<dir>.kicad_sch and a sorted first-.kicad_sch
    as later fallbacks. Then regenerate the docs table in
    docs/src/content/docs/reference/schematic-legibility.md and state the ranges as
    observed min-max with the outliers named.

FINDING 4 [medium] SPEC.md's canonical CLI block does not document --file, so the
two halves of SPEC.md disagree.
  Where: openspec/specs/SPEC.md:275 versus openspec/specs/SPEC.md:563
  Fails when: a reader uses section 3 as the CLI contract. Line 275 still reads
    "copperhead score schematic [--json]" while AC-16.34 at line 563 asserts
    "copperhead score schematic --file <sheet>" exists. Section 3 spells out
    optional flags for every other command, so the omission reads as "this flag
    does not exist".
  Fix: change line 275 to "copperhead score schematic [--file <path>] [--json]" and
    add a sentence to its description: scores any sheet outside a repo, reporting
    both composites.

FINDING 5 [low] (optional) score schematic --file accepts any s-expression file,
including a .kicad_pcb, and reports a confident score for it.
  Where: src/cli.ts:237, with src/kicad/sexp.ts:124-126
  Fails when: a user points the flag at the wrong half of their project.
    readSheetGeometry only requires the root to be a list, never that its head is
    kicad_sch. Verified: a file containing only
    "(kicad_pcb (version 20240108) (generator \"pcbnew\"))" exits 0 and prints
    "score: 40/100" and "wiring style: 100/100". Missing files and plain text are
    already handled correctly.
  Fix: tighten sexp.ts:124 to also require atomAt(root, 0) === 'kicad_sch' for the
    root file. Failing test:
    await expect(scoreSchematic(pcbPath, { docsDir: null })).rejects.toThrow()

FINDING 6 [low] (optional) npm run scoresheets dies with a raw ENOENT stack trace on
a path that does not exist.
  Where: scripts/score-sheets.ts:24, called from :38; the per-target try/catch that
    produces the tidy error row only starts at :57
  Fails when: npm run scoresheets -- /nonexistent/path.kicad_sch. sheetOf calls
    statSync(p) unguarded and outside the catch, so the run dies with a Node stack
    and no table at all, while a path that exists but fails to parse prints a
    one-line error row and the rest of the table still renders.
  Fix: add "if (!existsSync(p)) return null;" as the first statement of sheetOf;
    existsSync is already imported.

FINDING 7 [low] (optional) The script's crossings column is an absolute count, but
the docs cite the script as the source for crossings-per-wire.
  Where: scripts/score-sheets.ts:69
  Fails when: anyone tries to reproduce the documented calibration. The docs list
    crossings-per-wire as "0.03 on a 90-part board, 0.77 on an 1100-part one" and
    name npm run scoresheets as the source, but the script prints wire-crossings,
    an absolute count: demo:stickhub 89 parts crossings 15, and
    demo:jetson-agx-thor-baseboard 1138 parts crossings 3872. crossings-per-wire
    can never appear in report.metrics anyway, because it is not in DEFAULT_WEIGHTS
    and src/kicad/score.ts:447 filters to "name in weights".
  Fix: use r.style.metrics.find(x => x.name === 'crossings-per-wire')!.raw in place
    of m('wire-crossings'), formatted with f2.

FINDING 8 [low] (optional) The gated composite silently changed scale, and only the
PR description records it.
  Where: src/kicad/score.ts:42-56 (DEFAULT_WEIGHTS now sums to 115, not 100)
  Fails when: anyone compares a composite recorded before this PR with one recorded
    after. Every previously recorded check --json composite and AC-16.21 run-summary
    composite is now on a different scale. The re-pins are arithmetically faithful,
    so this is a documentation gap, not a hidden regression, but nothing in
    openspec/specs/SPEC.md or docs/src/content/docs/reference/schematic-legibility.md
    says the denominator moved. Related: AC-16.33 at SPEC.md:562 says the four new
    metrics appear in "the breakdown", the same phrasing AC-16.32 uses for gated
    metrics, but power-symbol-economy and labels-per-part are not in
    DEFAULT_WEIGHTS, so report.metrics never contains them.
  Fix: one line in AC-16.33 and in schematic-legibility.md around line 78 recording
    that the gated denominator went 100 to 115 and why. Reword AC-16.33 to say the
    wiring-style breakdown reports those four.

FINDING 9 [low] (optional) isIC (used for column-balance anchor height) counts
connectors while groupIC twelve lines earlier excludes them.
  Where: src/kicad/draft/engine.ts:892-894, versus groupIC at :752
  Fails when: a group holds a tall multi-pin header plus a dozen passives. anchorH
    becomes the header's cell height, so balanceH = Math.max(anchorH, squareSide) is
    tall and splitColumn(col, balanceH, 4) never fires, leaving the tall-thin-column
    shape the pass was added to fix. Reported as a verified code inconsistency with
    a derived consequence: test/fixtures/symlib only ships a 3-pin connector, so no
    repro was buildable.
  Fix: const anchorH = Math.max(0, ...members.filter(m => groupIC(m.key))
        .map(m => cellDims.get(m.key)!.h));

FINDING 10 [low] (optional) Four new public interface fields and their classifier
are computed on every score and read by nothing.
  Where: src/kicad/score.ts:113 (NOT_IC), :117, :122-127 (ics, icPins, icPinsWired,
    icPinsLabelled), filled at :217-223
  Fails when: never at runtime, but they cost a pass over every pin of every
    3-or-more-pin part on every check, and they ship an unexercised classifier a
    future consumer would inherit untested (NOT_IC matches Device:D_Schottky but not
    plain Device:D, and Device:Q_ but not Device:Q). Grepping src/, test/, scripts/
    and docs/src for icPins, NOT_IC and .ics returns only the definition and
    assignment sites.
  Fix: delete them, or surface an "IC pins wired" column in
    scripts/score-sheets.ts so the classifier is exercised.

FINDING 11 [low] (optional) The shelf-wrap assertion now compares the model against
a note the same code emitted.
  Where: test/draft-layout.test.ts:81
  Fails when: the wrap becomes unbalanced.
    expect(rows).toEqual([['G01'..'G04'], ['G05'..'G08']]) became
    expect(rows.length).toBe(Number(/\d+/.exec(wrapNote!)![0])), but the note text
    is produced from new Set(wrap.deltas.map(d => d.dy)).size at engine.ts:1596,
    the same wrap.deltas the rects were shifted by, so the two can never disagree.
    A regression that wraps eight equal groups as 7 + 1 passes.
  Fix: add a shape assertion the engine's own note cannot satisfy for free:
    expect(Math.max(...rows.map(r => r.length)) - Math.min(...rows.map(r => r.length)))
      .toBeLessThanOrEqual(1);

FINDING 12 [low] (optional) The "text inside its box" gate measures labels with the
centred half-height box, not the full-height on-paper box the engine reserves.
  Where: test/draft-sheet-fit.test.ts:241 and the same expression at :305, versus
    src/kicad/draft/engine.ts:46-49
  Fails when: a label's plotted text crosses a group box's top edge by up to
    0.635mm. labelReserveBox reserves minY = y - 1.27, maxY = y + 0.635, where the
    glyphs actually stand, but the test measures minY = y - 0.635. The gate added
    for this invariant is half a text height weaker than the invariant, despite its
    comment saying "'inside the box' means on paper".
  Fix: in both helpers use { minY: l.y - H, maxY: l.y + H / 2 }, matching
    labelReserveBox.

FINDING 13 [low] (optional) A type assertion in production code papers over a test
fixture that lies about its type.
  Where: src/kicad/draft/draft.ts:169, forced by
    test/draft-bank-guards.test.ts:161-163
  Fails when: sheetFit is later removed or renamed. sheetFit is declared
    non-optional on SchematicDraftReport (engine.ts:139-144), but the new formatter
    line reads "report.sheetFit as SchematicDraftReport['sheetFit'] | undefined" and
    guards with "if (sf)". The reason is that the test builds base as a partial
    object with "as unknown as SchematicDraftReport", omitting sheetFit. The cast is
    runtime-safe but removes the compiler's ability to report a genuine loss.
  Fix: give the fixture a real sheetFit
    ({ paper: 'A4', inkUtilization: 0, compaction: 'not-needed', misses: [] }) and
    drop both the cast and the guard.

FINDING 14 [low] (optional) --file silently drops the repo's scorer config and docs
vocabulary, so one file can report two different scores.
  Where: src/cli.ts:238, which passes { docsDir: null } and neither
    config.legibility nor config.docs
  Fails when: a user inside a repo with tuned legibility.score.weights runs
    copperhead score schematic --file hardware/board.kicad_sch on their own
    configured schematic. They get a different composite, and different
    group-caption findings since the caption vocabulary comes from docs/, than plain
    copperhead score schematic on the same bytes, with no indication why.
  Fix: when --file resolves inside the repo and a config loads, reuse
    config.legibility and config.docs; otherwise print a one-line note that
    defaults are in force.

FINDING 15 [low] (optional) The new --file surface has no test despite being named
in an acceptance criterion.
  Where: openspec/specs/SPEC.md:563 names it; grep -rn "\-\-file" test/ returns
    nothing and test/cli-surface.test.ts has no score coverage at all
  Fails when: a refactor of the score action drops the --file branch. The whole
    suite stays green while AC-16.34 is silently unmet.
  Fix: add a cli-surface case invoking
    score schematic --file test/fixtures/legibility/clean.kicad_sch outside any
    repo, asserting exit 0 and a style block in --json.

These parts of the diff are reached by no test: the --file branch of the score
action (finding 15); all of scripts/score-sheets.ts, which is also outside
tsconfig.json's include (findings 3, 6, 7); the zero-denominator fallbacks at
src/kicad/score.ts:410-413 (finding 1); and the isIC branch at
src/kicad/draft/engine.ts:892-894 (finding 9). Add tests for the new exported
symbols, branches and error paths they contain.

artifacts have to stay coherent, and `openspec validate build-copperhead-phase-1`
has to keep passing (it passes today).

    never by prompt text.
    snapshot.
    src/commands/check may touch a provider, an API key, or the network. Its import
    closure is 12 files whose only external package is execa; do not add an edge.

Then verify:
  npm run typecheck
  npm run build
  npm test

not yours to fix. Do not silence any other failure by deleting an assertion,
skipping a test, or loosening a threshold: if something fails and the fix is not
obvious, stop and report it.

````

PR #267, opened 2026-09-05
feat(draft): colour nets by function and hang test points beside their net

animesh-chouhan:
Stacked on #266. Third placement pass, from the KiStack `kicad-schematic` skill's checklist.

## What changes

- **Nets coloured by function** (`netColorsOf` in the engine, `colorOf` in the emitter): rails one colour, grounds another, each prefix family of two or more signals (I2S_, SPI_, BTN_) a colour from a fixed palette in family-name order; lone signals keep the theme default. Colour goes on wires and labels, so a bus reads as one thing across the sheet.
- **Test points beside their net**: a one-pin test point whose net reaches an IC side pin hangs above that pin's row and is wired to it (`isTestPoint`; `orientFor`/`leadsOf`/`spanOf` generalised to one-pin parts).
- **Stage-4 prompt** asks for prefix-named buses, +/- or P/N differential pairs and values with units so the colouring and the convention table have something to bite on.
- **Docs**: `schematic-drafting.md` gains the "Conventions the sheet follows" table with each rule's owner (engine or intent); `schematic-legibility.md` gains the colour rule. Spec AC-16.38, AC-16.39.

## Verification

- `test/draft-sheet-fit.test.ts`: colour test on usb-atmega-node (rails, grounds and the USB family each share one colour; a lone signal has none).
- Reference boards and Tier C golden re-pinned; diff is colour tokens only, every sheet legibility-clean, scores unchanged.
- Suite: 828 passed, 3 failed (gating-sync, init-check, stage-completion, pre-existing on main).
- esp32-amp engagement sheet: 0 legibility findings, 0 text overlaps, ERC 0, 548 colour tokens across 8 colours.


animesh-chouhan:

## Verdict: request changes

The colouring half is small, clean, well-tested and I found nothing wrong with it. The test-point half is the problem: **AC-16.39 is added to SPEC.md but the code path is unreachable for every stock KiCad test point symbol**, and nothing in the suite would have caught it, because the PR's only new test is the colour test.

## Metrics

```
lines: +262 / -145 (net 117) across 16 files  ·  src +85/-14, test +33/-7, docs +18/-0, spec +2/-0, other +124/-124
new vs net: 0 new files (0 in src), 16 modified; 0 new src lines (the net-new surface)
tests: +33 test lines; suite not measured -> 824/21/7 (pass/skip/fail, base -> head)
```

**Method.** Measured in a throwaway `git worktree` at head `64466cc`; your checkout was never entered and HEAD was never moved. Stacked PR, so everything is against `git merge-base origin/feat/inline-parts-and-rotation HEAD` = `2b7357d` (#266's head), not main. Areas from `git diff --numstat`; suite from `npx vitest run --reporter=json` in that worktree.

**Diff coverage: not measured, flagged rather than omitted.** The full-suite vitest v8 coverage run did not finish here (this container's `kicad-cli` 10.0.6 makes the ERC-backed tests time out first). I used the checklist's manual fallback instead: the +33 test lines are entirely the colour test, so the test-point surface added by this PR (`isTestPoint`, the one-pin branches of `orientFor`/`leadsOf`/`spanOf`, and the `claimHang` call at `engine.ts:1105`) is reached by **no test at all**. That is finding 1.

**On the 7 local failures and the failing check.** Environmental, and I confirmed the mechanism rather than assuming it: the five reproducible ones (`draft-tools`, `gating-sync`, `init-check` x2, `stage-completion`) all run ERC or the completion gate over `test/fixtures/open-key/`, a static hand-written project the drafting engine never touches, so neither `netColorsOf` nor the test-point path can reach them. The repo pins the KiCad 8/9 format (`EMIT_VERSION = '20231120'`, `src/kicad/emit.ts:13-15`) while this box has `kicad-cli` 10.0.6, so a bare `kicad-cli sch erc` on that untouched fixture reports `2 warning | lib_symbol_mismatch | Symbol 'R' doesn't match copy in library 'Device'`. CI's `test` job passes on this PR. The failing `template` check is real: the description is missing the `## Testing` section and the invariant checklist required by `.github/pull_request_template.md`.

## Findings

### 1. [high] AC-16.39 is unreachable for every stock KiCad test point, because `hangable()` rejects the `Connector` library one line before the new branch runs

- **Where:** the guard at `src/kicad/draft/engine.ts:1104` (`if (first.key === icKey || !hangable(first.key)) continue;`) runs before the new test-point branch at `src/kicad/draft/engine.ts:1105-1108`. `hangable` at `:964-970` includes `!isConnector(instByKey.get(key)!.part)`, and `isConnector` at `:681` is `p.libId.startsWith('Connector')`.
- **Fails when:** the intent declares a test point using any stock KiCad symbol. **All six** of them live in the `Connector` library: `Connector:TestPoint`, `TestPoint_Probe`, `TestPoint_Flag`, `TestPoint_2Pole`, and friends. `isTestPoint` at `:934` matches them (`/TestPoint/i.test(inst.part.libId)`), but `hangable` has already returned false, so `continue` fires and the branch is dead. The only way to reach it is a non-`Connector` libId containing "TestPoint", which no stock KiCad symbol provides. **Reproduced both ways:** unpatched, `Connector:TestPoint` lands in a column 40mm away with its own duplicate label; with `isConnector` alone amended to exclude test points, it hangs 5.08mm above the pin row and joins one wired run with one label, exactly as AC-16.39 describes. So the AC is a correct description of intended behavior that the code cannot deliver.
- **Aggravating:** `openspec/specs/SPEC.md:542` now asserts this as implemented, and `docs/.../schematic-drafting.md` lists "test points rise beside their net" in the conventions table as an **engine**-owned rule. A code-only divergence from SPEC.md is a finding on its own; here SPEC.md asserts behavior that is structurally impossible.
- **Fix:** exclude test points from the connector gate, e.g. make `hangable`'s clause `(!isConnector(part) || isTestPoint(key))`, or narrow `isConnector` to exclude `TestPoint`. Then add the regression test that does not exist: a `Connector:TestPoint` symbol in `test/fixtures/symlib`, an intent hanging it on an IC side pin, and an assertion that it lands on the pin's row (one wired run, one label). Related and worth fixing in the same pass: `leadsOf` returns `{a: pins[0], b: pins[0]}` for a test point, and `engine.ts:1713` then pushes two connection points for the same net, the second at the far end of the *body* where no pin exists.

### 2. [low] The family palette wraps at 8, so a board with 9 or more signal families gives two buses the same colour

- **Where:** `src/kicad/draft/engine.ts:449` (`FAMILY_PALETTE`, 8 entries) and the assignment `FAMILY_PALETTE[i % FAMILY_PALETTE.length]`.
- **Fails when:** a design declares 9 or more prefix families of 2 or more nets each, which a mid-size board with `I2S_`, `SPI_`, `I2C_`, `UART_`, `BTN_`, `LED_`, `USB_`, `ADC_`, `PWM_` reaches easily. Families 0 and 8 get the identical RGB triple, which is precisely the confusion the feature exists to prevent, and AC-16.38 promises each family "a colour of its own". Related: the prefix regex `^([A-Za-z0-9]+)_` means a net whose prefix contains any non-alphanumeric character forms no family at all.
- **Fix:** either extend the palette and state its size in AC-16.38, or amend AC-16.38 to say families beyond the palette's size share colours in family-name order. If you extend it, keep the entries visually distinct in both light and dark KiCad themes.

## Invariant checks performed

All six checked against the diff, not assumed. `openspec validate build-copperhead-phase-1` returns **"Change 'build-copperhead-phase-1' is valid"** (exit 0), and `npx tsc --noEmit` is clean.

- **Spec-gated in** - pass. `src/agent/` is untouched. Gating is still structural: `src/agent/tools.ts:820` `TOOLS.filter(t => !t.requiresUnlock || ctx.editsUnlocked)`, with `edit_file` and `write_file` omitted from the list rather than prompt-discouraged.
- **Verification-gated out** - pass, and checked specifically because this PR touches `src/commands/create.ts`. That one-line change is **prompt text only**: it appends net-naming guidance to the stage-4 string. The gate above it (`src/commands/create.ts:204-238`: symbols, then drift, then `runErc`, then legibility errors, then the byte-exact re-draft match) is untouched.
- **`check`/`verify` stays LLM-free and network-free** - pass, traced rather than assumed. `src/commands/check.ts`'s full transitive closure is 12 files whose only external import is `execa` (pre-existing, for the `kicad-cli` and `openspec` subprocesses). This PR touches nothing in that closure.
- **No sexp serialization** - pass. `src/kicad/sexp.ts` is untouched. The `src/kicad/emit.ts` change adds `(color r g b 1)` attributes to the from-scratch emitter, which imports only `node:crypto` and builds text from a typed `PlacementModel`; it never sees a parsed tree. This is the documented "the engine emits a brand-new schematic" capability, not round-tripping.
- **Sync-obligations ledger** - pass, untouched.
- **Secrets** - pass. `src/util/redact.ts` unchanged with `sk-[A-Za-z0-9_-]+` intact; `.gitignore` still carries `.env` and `.copperhead/runs/`.
- **AC numbering** - pass. AC-16.38 and AC-16.39 are new and do not collide; across the whole stack SPEC.md carries AC-16.1 through AC-16.64, each defined exactly once, with no gaps and no duplicates.
- **Test determinism** - pass. The colour test uses `mkdtemp` plus `cp`, passes a fixed `'2020-01-01'`, asserts `v.ok` before use, and does not shell out to `kicad-cli`. No network, no `Date.now()`, no `Math.random()`.

**A caveat about the validation result.** `openspec validate` passing carries no signal for these ACs. They were written straight into `openspec/specs/SPEC.md`, and the only drafting change (`deterministic-schematic-drafting`) is archived, so no active change carries an AC-16 delta and validation structurally cannot detect AC-to-code drift. That is the established pattern for this line of work rather than something this PR introduces, but it is why finding 1 went unnoticed.

**Untested surface:** the entire test-point feature. `isTestPoint` (`engine.ts:934`), the one-pin branches of `orientFor` (`:940`, `:948`), `leadsOf` (`:957`), `spanOf` (`:975-982`), the `isTestPoint(cur)` break in the chain walk (`:999`), and the `claimHang` call at `:1105` are reached by no test. There is no `TestPoint` symbol in any fixture library and no reference board uses one; the only real inputs are in `manual-tests/real-designs/examples/tiny_tapeout`, which is not in the automated suite.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 267, branch feat/net-colours-and-testpoints,
based on feat/inline-parts-and-rotation (this PR is stacked on PR 266, not on main).

Check it out first:
  gh pr checkout 267

The net-colouring half of this PR is correct and well-tested. Leave it alone except
for finding 2.

FINDING 1 [high] AC-16.39 is unreachable for every stock KiCad test point, because
hangable() rejects the Connector library one line before the new branch runs.
  Where: the guard at src/kicad/draft/engine.ts:1104
    ("if (first.key === icKey || !hangable(first.key)) continue;") runs before the
    new test-point branch at src/kicad/draft/engine.ts:1105-1108. hangable at
    :964-970 includes !isConnector(instByKey.get(key)!.part), and isConnector at
    :681 is p.libId.startsWith('Connector').
  Fails when: the intent declares a test point using any stock KiCad symbol. All six
    of them live in the Connector library: Connector:TestPoint,
    Connector:TestPoint_Probe, Connector:TestPoint_Flag, Connector:TestPoint_2Pole
    and friends. isTestPoint at :934 matches them via /TestPoint/i on the libId, but
    hangable has already returned false, so continue fires and the branch is dead.
    The only way to reach it is a non-Connector libId containing "TestPoint", which
    no stock KiCad symbol provides. Reproduced both ways: unpatched,
    Connector:TestPoint lands in a column 40mm away with its own duplicate label;
    with isConnector amended to exclude test points, it hangs 5.08mm above the pin
    row and joins one wired run with one label, exactly as AC-16.39 describes.
    Aggravating: openspec/specs/SPEC.md:542 asserts this as implemented, and
    docs/src/content/docs/reference/schematic-drafting.md lists "test points rise
    beside their net" in the conventions table as an engine-owned rule.
  Fix: exclude test points from the connector gate, for example make hangable's
    clause (!isConnector(part) || isTestPoint(key)), or narrow isConnector to
    exclude TestPoint. Then add the regression test that does not exist: add a
    Connector:TestPoint symbol to test/fixtures/symlib, write an intent hanging it
    on an IC side pin, and assert it lands on that pin's row as one wired run with
    one label. Fix in the same pass: leadsOf returns { a: pins[0], b: pins[0] } for
    a test point, and src/kicad/draft/engine.ts:1713 then pushes two connection
    points for the same net, the second at the far end of the body where no pin
    exists.

FINDING 2 [low] (optional) The family palette wraps at 8, so a board with 9 or more
signal families gives two buses the same colour.
  Where: src/kicad/draft/engine.ts:449 (FAMILY_PALETTE, 8 entries) and the
    assignment FAMILY_PALETTE[i % FAMILY_PALETTE.length]
  Fails when: a design declares 9 or more prefix families of 2 or more nets each,
    which a mid-size board with I2S_, SPI_, I2C_, UART_, BTN_, LED_, USB_, ADC_ and
    PWM_ reaches easily. Families 0 and 8 get the identical RGB triple, which is the
    confusion the feature exists to prevent, while AC-16.38 promises each family "a
    colour of its own". Related: the prefix regex ^([A-Za-z0-9]+)_ means a net whose
    prefix contains any non-alphanumeric character forms no family at all.
  Fix: either extend the palette and state its size in AC-16.38, or amend AC-16.38
    to say families beyond the palette's size share colours in family-name order. If
    you extend it, keep the entries visually distinct in both light and dark KiCad
    themes.

fix and passes after it; this repo pairs bug fixes with tests. Finding 1 is a
correctness bug and needs one.

These parts of the diff are reached by no test: the entire test-point feature.
isTestPoint (src/kicad/draft/engine.ts:934), the one-pin branches of orientFor
(:940, :948), leadsOf (:957) and spanOf (:975-982), the isTestPoint(cur) break in
the chain walk (:999), and the claimHang call at :1105. There is no TestPoint symbol
in any fixture library and no reference board uses one. The PR's only new test is
the colour test. Add tests for these branches.

This PR changes spec-level behavior (AC-16.38 and AC-16.39), so openspec/specs/SPEC.md
and the docs pages have to describe what the code actually does, and
`openspec validate build-copperhead-phase-1` has to keep passing (it passes today).
Be aware that validation cannot detect AC-to-code drift here, because these ACs live
directly in SPEC.md and no active change carries an AC-16 delta, so the test in
finding 1 is the only thing that will hold AC-16.39 honest.

    never by prompt text.
    snapshot. This PR's one-line src/commands/create.ts change is prompt text only;
    keep the gate at src/commands/create.ts:204-238 untouched.
    src/commands/check may touch a provider, an API key, or the network. Its import
    closure is 12 files whose only external package is execa; do not add an edge.
    edited only via anchored exact-match text replace, never round-tripped. The
    emitter builds text from a typed PlacementModel and never sees a parsed tree;
    keep it that way.

Then verify:
  npm run typecheck
  npm run build
  npm test

`kicad-cli sch erc` on that untouched fixture reports lib_symbol_mismatch. The five
reproducible failures (draft-tools, gating-sync, init-check twice, stage-completion)
all run over that fixture and are not yours to fix. Do not silence any other failure
by deleting an assertion, skipping a test, or loosening a threshold: if something
fails and the fix is not obvious, stop and report it.

````

PR #268, opened 2026-09-05
feat(draft): put every part on its pin — lanes, anchors, transistors, node hangs, text

animesh-chouhan:
## What

The drafting engine's third to ninth placement passes (AC-16.40 to AC-16.64), and the fixes for this PR's automated review. Every two-lead part is drawn on the pin it serves, text keeps 0.8 mm of air, and the sheet is fitted the way a person lays one out: height-minimising rows, column-major and masonry wraps, the title block as a corner, the squeeze and the look, and one gap between group boxes.

No longer stacked: #267 is merged, and `main` is merged into this branch.

## Why

A group-by-group visual review of the ESP32 amplifier engagement sheet, and a hand-laid A2 version of the same circuit, found parts drawn as labelled islands, values standing on end, an IC name across its pin names, transistors left of their driver, and a sheet two sizes larger than needed. Per the commit messages, that sheet went from 68 % to 100 % of two-lead parts wired, from A1 to A2, with 0 legibility errors. That sheet is not in the repo and was not re-measured for the review fixes.

### Placement passes

| Commit | Pass | ACs |
| --- | --- | --- |
| `a0ffb68` | Every part on its pin: lanes, connector and switch anchors, transistor-ended runs, node hangs, field angles, 0.8 mm text pad, flags along the row | 16.40 to 16.47 |
| `71c10ed` | Denser cells, height-minimising row breaks, finer compaction | 16.48 |
| `e45bcb6` | Column-major wrap, title block as a corner, groups shaped to their band, lanes at the drawn labels' reach, KiCad 10 `(hide yes)` in the reader | 16.49 to 16.52 |
| `7c57e24` | The squeeze (rounds only with `COPPERHEAD_DRAFT_SQUEEZE=1`) | 16.53 |
| `dc5364d` | Anchorless groups shaped, tighter rows, spacing by real labels | 16.54 to 16.56 |
| `7b97f85` | Balanced column splits | 16.57 |
| `78fdefc` | Bridges share a lane with a hook route, shelf infill, banks by rail, aligned boxes | 16.58 to 16.61 |
| `4d2e35e` | Masonry wrap, the look | 16.62, 16.63 |
| `7852ade` | One gap between boxes, caption band | 16.64 |

### Review fixes (`ddb52f1`)

1. **Masonry search was exponential.** [engine.ts](src/kicad/draft/engine.ts) now searches the deals as a tree and drops a branch as soon as a column is too tall or the columns so far are already wider than the best deal. It visits deals in the same order and drops only branches that cannot win, so it keeps the same deal the full enumeration kept. Results are cached per draft by group sizes, with a 200 000 node budget. Drafting 12 groups went from 18.8 s to 0.33 s, and 16 groups from 17.7 s to 1.5 s, with identical models at 4, 8, 12 and 16 groups.
2. **The wrap note counted groups, not columns.** It now counts the columns the boxes are drawn in: `usb-atmega-node` said 6 and draws 3.
3. **Five or more endpoints drew no wires.** A cluster of more than four endpoints is now narrowed to the endpoints bound to one anchor and routed, largest subsets first, within 64 attempts. Four or five parts hung on one pin now draw one wired run with one label.
4. **Some tests passed by skipping.** They now run their assertions (details under Testing).
5. **Some ACs claimed more than the code does.**
   - The final power-value sweep now pads its text checks by 0.8 mm.
   - The title-block corner is re-checked after the boxes grow, using the checker's corner size.
   - AC-16.45, AC-16.50, AC-16.61 and AC-16.62 are reworded to match the code.
6. **Stale board copy.** The old root copy of `usb-atmega-node` (sheet, project, sym-lib-table) is removed.
7. **The nudge test had lost its check.** It is restored and reads the nudge from the trace, because stub lengths also change for electrical reasons.
8. **Abandoned chains stayed turned.** A chain turned half a turn and then abandoned now gets its orientation back, and a turned part keeps its mirror.
9. **Bank width estimate.** Each bank member is measured in its own orientation.
10. **Connector test was too loose.** The connector must now be strictly left of an IC in another column.
11. **Smaller fixes:**
    - `COPPERHEAD_DRAFT_TRACE` prints only when set to `1`.
    - AC-16.53 wording fixed, and AC-16.47 moved into numeric order.
    - Drafting doc: channel width corrected; stale engine comment fixed.
    - A draft kept with boxes still wrong after three re-tilings now says so in its notes.

Also from the Codex notes: a bank trunk never joins members of two groups.

Re-pinned boards: `buck-12v-5v`, `npn-switch` and `usb-atmega-node`. A rail value that touched the GND text beside it moves clear. On `npn-switch` the lower groups also move down one row, following its enlarged box. All three stay ERC- and legibility-clean. `ldo-demo` and the Tier C golden are unchanged.

### Checker change users will see

[legibility.ts](src/kicad/legibility.ts) now measures a global label as a flag two text heights tall, centred on its anchor (it used to measure bare text). It also measures property text at its stored angle plus the symbol's rotation. Existing hand-drawn sheets can therefore gain text-overlap findings. The `open-key` fixture was moved clear of its "Keyer" caption for this reason. [sexp.ts](src/kicad/sexp.ts) changes are read-side only.

### Known issues not addressed here

- On A5 the checker narrows the title block to 95 mm, while KiCad draws it at about 110 mm. `npn-switch`'s MCU box overlaps the drawn title block, as it already did before this PR.
- Open Codex notes:
  - Four-terminal transistors are not recognised as transistors.
  - The column-split count includes inter-chunk gaps.
  - A single-group sheet does not use the title-corner fit.
  - The checker's flag-attachment test does not require the wire to run the same way as the flag.
  - A wired run's first flag can still be placed vertically.

## Spec / OpenSpec

AC-16.40 to AC-16.64 are added to [SPEC.md](openspec/specs/SPEC.md). The review fix rewords AC-16.37, AC-16.45, AC-16.47, AC-16.50, AC-16.53, AC-16.61 and AC-16.62 to match what the code does. As with #264 to #267, no active change carries an AC-16 delta, because the drafting change is archived. The tests are what hold these criteria to the code. `openspec validate build-copperhead-phase-1` passes.

### Automated test status

Run on this branch merged with `main`, with an isolated `KICAD_CONFIG_HOME` holding only `fp-lib-table` (matching CI).

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | pass: 986 passed, 21 skipped, 0 failed (70 files) |
| OpenSpec | `openspec validate build-copperhead-phase-1` | pass |
| Live-LLM (opt-in) | n/a | n/a: no agent, prompt or provider code touched |

New or changed tests:

- [draft-sheet-fit.test.ts](test/draft-sheet-fit.test.ts):
  - 12 groups draft within 10 s (18.8 s before the fix).
  - The wrap note's count equals the lines the boxes are drawn on: rows on the seven-group fixture, columns on `usb-atmega-node`.
  - The squeeze drafts no rounds by default, and keeps a shrinking round on `ldo-demo` with `COPPERHEAD_DRAFT_SQUEEZE=1` (5798 to 5274 mm²).
  - The look test no longer returns early.
  - The connector is strictly left of an IC in another column.
- [draft-idioms.test.ts](test/draft-idioms.test.ts): four and five parts hung on one pin draw one label, and every two-lead part stays attached.
- [draft-layout.test.ts](test/draft-layout.test.ts): the nudge check is restored, read from the trace. The first nudging seeded board is 31.
- The derived-symbol hardening test in [draft-hardening.test.ts](test/draft-hardening.test.ts) now runs in 1.3 s, well inside its 60 s budget (the review measured 52 s).

Known pre-existing failures: none with the isolated KiCad config.

- Sandbox variant used: `edit`, plus the four control boards through `draft schematic`.
- Commands run and outcome:

```text
$ ./manual-tests/setup.sh edit --fresh && npm run dev -- --repo manual-tests/runs/edit init
created docs/DECISIONS.md, docs/CHANGELOG.md, .git/hooks/pre-commit
$ npm run dev -- --repo manual-tests/runs/edit check
ERC ✓  DRC ✓  drift ✓
legibility: 1 error, 5 advisory (error: unlabeled-group "Keyer", same output on main)
$ npm run refboards -- --update
[buck-12v-5v] [ldo-demo] [npn-switch] [usb-atmega-node] drafted, references and renders updated
$ npm run dev -- --repo manual-tests/runs/reference-boards/usb-atmega-node draft schematic
drafted: 6 group(s), 89 wire segment(s), 23 label(s), 19 no-connect(s) on A4
note: groups wrapped onto 3 columns to fit the sheet (declared order runs down each column)
$ npm run dev -- --repo manual-tests/runs/reference-boards/usb-atmega-node check
ERC ✓  DRC skipped (no board configured)  drift ✓  legibility: no findings  score 78.13/100
$ COPPERHEAD_DRAFT_TRACE=1 npm run dev -- --repo manual-tests/runs/reference-boards/npn-switch draft schematic
42 [draft] trace lines; with COPPERHEAD_DRAFT_TRACE=0: 0; the redraft is byte-identical to the re-pinned reference
```

## Docs

- [schematic-drafting.md](docs/src/content/docs/reference/schematic-drafting.md): one section per pass. The review fix updates the text-pad, squeeze, box-alignment, masonry-search and re-tiling paragraphs, and the channel width.
- [schematic-legibility.md](docs/src/content/docs/reference/schematic-legibility.md): flag bounds and text measured as drawn.

---

## Invariant checklist

- [x] **Spec-gated in**: N/A, `src/agent/` is untouched; `edit_file`/`write_file` stay gated by omission.
- [x] **Verification-gated out**: N/A, no command, repair or rollback code changed.
- [x] **`check`/`verify` stays LLM-free and network-free**: `legibility.ts` and `sexp.ts` are in its closure; neither adds an import, and both stay pure.
- [x] **No sexp serialization**: `sexp.ts` changes are read-side only; `emit.ts` builds text from the typed `PlacementModel` and never sees a parsed tree.
- [x] **Sync-obligations ledger**: N/A, untouched.
- [x] **Secrets**: N/A, redaction and `.gitignore` untouched; `COPPERHEAD_DRAFT_TRACE` writes only group, net and ref names and coordinates to stderr.
- [x] **Spec coherence**: SPEC.md and the drafting reference move together; no active change carries AC-16 deltas (archived); `openspec validate build-copperhead-phase-1` passes.


animesh-chouhan:
Second commit 71c10ed answers "why is the sheet so wide": the group boxes summed to 228 k mm² at 6–17 % fill vs A2's 212 k usable. Cells are denser (margin 4, channel 6, row gap 3, group gap 6 → 214 k), the shelf-wrap chooses order-preserving row breaks that minimise height (AC-16.48), and compaction tries height budgets down to a quarter sheet with a trace of every attempt. A plain label's box stands on its line as the checker measures it. esp32-amp stays on A1 with 0 errors and 110/110 attached; A2 misses at 502 × 537 mm vs 574 × 370 at the finest budget, so the groups themselves need to shrink about a third (banks beside circuits, lane pitch, folded test-point columns) before A2 is reachable. Suite 839 passed, 3 pre-existing failures.

animesh-chouhan:
Third commit: page fitting learned from a hand-laid A2 version of the same sheet — column-major wrap (AC-16.49), title block as a corner (AC-16.50), groups shaped to fill their band with a fixed column splitter (AC-16.51), lanes at the drawn labels' reach (AC-16.52), frame-overflow redraft, and a KiCad 10 `(hide yes)` reader fix (ten false text-on-wire errors on eeschema-10 files). buck-12v-5v attachment 0.57 → 0.78. esp32-amp remains A1: columns 242+208+112+gaps = 578 mm vs 574, UI grid 145 mm tall vs 93 by hand; the remaining lever is the hung part's cell height (39 mm vs 25). Suite 840 passed, 3 pre-existing.

animesh-chouhan:
Fourth commit: the squeeze (AC-16.53). Each cell's drawn extent is measured after a draft and further rounds size cells from it; rounds are gated (boxes shrink ≥3 %, no refused hang, no larger sheet, gates hold) and run with `COPPERHEAD_DRAFT_SQUEEZE=1`. On esp32-amp it kept the first round (194 k → 197 k mm²), which locates the remaining gap to the hand-laid A2 in layout structure (the UI grid without a band budget and the hung-part cell height), not in margins. Base drawing byte-identical; suite 841 passed, 3 pre-existing.

animesh-chouhan:
Fifth commit: anchorless groups shaped to a 2.5:1 aspect as grids (AC-16.54), rows tightened with pin-based power reserves (AC-16.55), label reserves only where a flag will exist (AC-16.56), channel and group gap four units. esp32-amp group boxes 208 k → 170 k mm² (the hand-laid A2 has 166 k); still A1 — the one three-row split under A2's height needs 598 mm vs 574. Remaining: lane pitch (11 → 10 mm) and the rails bank wrapping to two rows when the group is width-bound. 110/110 wired, 0 errors; buck reference attachment 0.78 → 0.87. Suite 841 passed, 3 pre-existing.

animesh-chouhan:
Sixth commit: exact minimal-maximum column split (AC-16.57) — eight test points now two balanced columns; esp32-amp boxes 169 k mm² vs 166 k hand-laid. Group by group against the hand-laid v2: rails is 21 mm wider (bootstrap cap along the SW row keeps the bank from wrapping), UART 21 mm wider (D7 not tucked under the right-hand runs). Hanging the bridge cap in a lane was tried and reverted (attachment 110→106, crossings 15→31). Suite 841 passed, 3 pre-existing.

animesh-chouhan:
Seventh commit: group-by-group fixes from the hand-laid sheet — lanes one unit narrower, a bridge cap shares its pin's lane with a hook route for the far lead (AC-16.58; rails 112→92 mm), shelf infill tucks substantial cells under an IC's lanes (AC-16.59; UART 178→134 mm), banks by rail with rail-whole row breaks (AC-16.60), boxes aligned per row/column (AC-16.61). esp32-amp boxes 162 k mm² vs 166 k hand-laid; still A1 (rows 434 mm tall vs 395 with the corner; columns checked next). Suite 841 passed, 3 pre-existing.

animesh-chouhan:
Eighth commit 4d2e35e: masonry wrap (groups dealt to 2–4 columns, every reading-order deal judged; AC-16.62) and a second fit on measured label reach (`sheetFit.look`; AC-16.63). The engagement sheet now drafts on **A2** in three rows: 110/110 two-lead parts wired, 0 legibility errors, 0 text overlaps, 16 crossings, group boxes 162 k mm² against the hand-laid A2 sheet's 166 k. buck-12v-5v and npn-switch re-pinned (same paper and boxes; one group shifted by its measured reach). Suite 843 passed; the three failures are the pre-existing KiCad 10 environment ones.

animesh-chouhan:
Ninth commit 7852ade: uniform 5.08 mm gaps between group boxes (a re-spacing pass after enclosure moves each box with its contents; rows share a top, columns a left; aligned edges stop short of the title block), a full-width caption band with air below it, and the look now measures vertical box growth so the fit stacks boxes (AC-16.64). esp32-amp stays on A2, 110/110, 0 errors. All four reference boards and the Tier C golden re-pinned, legibility clean on each. Suite 843 passed; the three failures are the pre-existing KiCad 10 environment ones.

animesh-chouhan:

## Verdict: request changes

The placement work here is genuinely impressive and the hardest parts are correct: I verified the mirror-times-rotation composition, the field-angle round trip, the chain-walk cycle safety and the re-draft loop's termination, and all of them hold. Three things need settling before merge: a large performance regression that already breaks a test on its own 60s budget, a user-visible draft note that states a false number on a shipped reference board, and a net shape that now emits **zero** connecting wires.

**A process note first, because it shapes how this PR should be read.** The description documents AC-16.40 to AC-16.47 and reports "838 passed", but that describes only the first of **nine** commits. The PR now adds **AC-16.40 through AC-16.64, twenty-five acceptance criteria**, and the measured figures in the body (attachment 0.71 to 0.88, wiring style 80.5 to 95.5, crossings 2 to 15) are all from commit `a0ffb68`. Eight further commits landed after it, each described only in a PR comment. A reviewer reading the description does not learn what they are approving.

## Metrics

```
lines: +10258 / -2037 (net 8221) across 35 files  ·  src +2120/-265, test +994/-148, docs +48/-7, spec +25/-0, other +7071/-1617
new vs net: 3 new files (0 in src), 32 modified; 0 new src lines (the net-new surface)
tests: +994 test lines; suite not measured -> 842/21/4 (pass/skip/fail, base -> head)
ci: 3 pass / 1 fail / 0 pending; required checks: none reported (gh pr checks)
```

**Method.** Measured in a throwaway `git worktree` at head `7852ade`; your checkout was never entered and HEAD was never moved. Stacked PR, so everything is against `git merge-base origin/feat/net-colours-and-testpoints HEAD` = `64466cc` (#267's head), not main. Areas from `git diff --numstat`; suite from `npx vitest run --reporter=json` in that worktree.

**Diff coverage: not measured, flagged rather than omitted.** The full-suite vitest v8 coverage run did not finish here, partly for the environmental reason that affects the whole stack and partly because of finding 1. I used the checklist's manual fallback, mapping the new branches and error paths to the tests that reach them; findings 4, 7 and the untested-surface list are that mapping's output.

**On the failing check.** `template` is real: the description is missing the `## Testing` section and the invariant checklist required by `.github/pull_request_template.md`. Given the description is also nine commits out of date, this one is worth more than the usual box-ticking.

## Findings

### 1. [high] Drafting is 4x slower on the reference boards and up to 500x slower on a mid-size board, and it already fails a test on its own 60s budget

- **Where:** `src/kicad/draft/engine.ts:3069` (`wrapMasonry`), called from `:3157-3161` (k=2..4 x 2 budgets = 6 calls) inside `tryPaperBudgeted`'s 7x4 = 28-combination sweep at `:3192`, per candidate paper, inside a loop that drafts up to 16 times.
- **Fails when:** a group count lands where `k ** n` slips under the `if (total <= 2_000_000)` guard. That guard is a wild over-estimate of the real deal count, so it does not bound the work it is meant to bound. **Measured `draftSchematicPlacement` directly on both branches:**

  | groups | base (`64466cc`) | this PR |
  |---|---|---|
  | 4 | 51 ms | 89 ms |
  | 8 | 57 ms | 3,029 ms |
  | 12 | 103 ms | **53,873 ms** |
  | 16 | 152 ms | 23,839 ms |

  The non-monotonic 12-versus-16 result confirms the attribution exactly: n=12 enumerates k=3 (531k deals) while n=16 trips the guard and skips it. **I reproduced the real-board half of this independently:** drafting the committed reference boards in each worktree, `usb-atmega-node` goes 93ms to 341ms and `buck-12v-5v` goes 41ms to 178ms, both about 4x. The concrete failure today is `test/draft-hardening.test.ts` ("a drafted sheet with a derived symbol loads in KiCad"), which takes 52.2s on this PR against 27.5s on base, inside its own hardcoded 60s budget. Beyond CI, `draft_schematic` is called repeatedly in the stage-4 repair loop, so a 15-to-20-group board burns minutes per turn and can trip the per-turn watchdog in `src/config.ts`.
- **Fix:** memoize `wrapMasonry(k, budgetH)` on the current `groupRects` signature and hoist it out of the frac/div sweep; bound it by deals actually judged rather than by `k ** n`. Disabling only the enumeration removes about 94% of the cost and picks the **same** sheet on every case tested, so the search is buying very little. Add a test that drafts a 12-group intent inside a few seconds.

### 2. [high] The wrap note reports the group count instead of the column count, and says so on a shipped reference board

- **Where:** `src/kicad/draft/engine.ts:3299` and `:3302`, `new Set(wrap.deltas.map(d => d.dx)).size`.
- **Fails when:** any column wrap. Deltas are *target minus the group's own x1*, and the pre-wrap ribbon gives every group a distinct `x1`, so every delta is distinct and the count is always the number of groups. The rows branch three lines below (`:3307`) does it correctly from the absolute y **and carries a comment explaining this exact hazard**: the fix was applied to `dy` and not to `dx`. **Reproduced on a committed board:** drafting `usb-atmega-node` emits `groups wrapped onto 6 columns to fit the sheet`, while the committed reference has exactly **3** distinct group-box x1 values (23.749, 89.154, 196.723) across **6** groups. Masonry only ever builds 2 to 4 columns, so a "6" is impossible by construction. This falsifies AC-16.49's "the draft notes say which was used" and AC-16.62's only mechanically checkable clause, and it is text the agent reads back during stage-4 repair.
- **Fix:** `new Set(wrap.deltas.map((d, i) => Math.round((groupRects[i]!.x1 + d.dx) / U))).size` in both places, matching what the rows branch already does. Add an assertion that the note's k equals the number of distinct box `x1` values, which is the test that would have caught this and finding 6's vacuity together.

### 3. [medium] Dropping the endpoint-count gate leaves a 5-or-more-endpoint local net with zero connecting wires

- **Where:** `src/kicad/draft/engine.ts:1534` (the gate replaced by a `trace`), `:3931` (`anchorKey(a.ep.ref) === anchorKey(b.ep.ref)` in `near()`), `:3939` (`if (cl.length < 2 || cl.length > MAX_WIRED_ENDPOINTS) continue;`).
- **Fails when:** one pin's local net has 5 or more endpoints. The removed gate was `if (eps.length > MAX_WIRED_ENDPOINTS || eps.some(cross-group)) continue;`. The cross-group half is harmless, since `hangable()` still requires `memberKeySet.has(key)`. The endpoint half is not: hangs are now claimed on a net the wire pass will refuse, and the new `near()` rule merges every part bound to one anchor into a single cluster **regardless of wire span**, so the cluster overruns `MAX_WIRED_ENDPOINTS` and `:3939` skips it wholesale rather than routing a subset. Verified on a one-group repro (`U1` MCU8, `R1..R4`; `BIG = [U1.4, R1.1, R2.1, R3.1, R4.1]`, `GND = [U1.2, R1.2..R4.2]`): this PR produces **5 wires, all 2-unit stubs, 5 global `BIG` flags, and no wire joining the pin to any resistor**, while the four resistors occupy four lanes on U1's right shelf. The base commit on the same intent produces 11 `BIG` wires and 3 labels. With three resistors (4 endpoints) it wires cleanly, so the cliff is exactly at 5.
- **Note:** the `:3939` skip is inherited from #266, where I filed it as that PR's finding 2. This PR is what makes it reachable often enough to matter. Fixing it in #266 fixes it here.
- **Fix:** at `:3939` drop the upper bound and clamp the subset search instead: `if (cl.length < 2) continue;` with `for (let size = Math.min(cl.length, MAX_WIRED_ENDPOINTS); size >= 2 && !done; size--)`, keeping the enumeration at C(n,4) rather than 2^n. Regression test: the intent above, asserting `measureWiringStyle(...).twoPinAttached === 4`.

### 4. [medium] The multi-round re-draft paths that five ACs describe are never executed by the suite, and their tests pass by skipping

- **Where:** `test/draft-sheet-fit.test.ts:483` (squeeze), `:529` (AC-16.62), `:512` (AC-16.63).
- **Fails when:** a regression lands in any of them. Nothing in `test/` sets `COPPERHEAD_DRAFT_SQUEEZE`, so `engine.ts:847` always breaks at `squeezeRounds === 0` and `:851` writes `boxAreaBefore === boxAreaAfter` by construction (both from the same `best`), making `toBeLessThanOrEqual` unfailable and the `sq.rounds > 0` branch dead. The AC-16.62 test asserts its shape string only inside `if (notes.some(/laid in \d+ columns/))`, and its `sevenGroups()` fixture, added for this AC, actually produces a **rows** wrap, so the branch never runs: a single assertion that the note's k equals the number of distinct box `x1` values would have caught finding 2. The AC-16.63 test is wrapped in `if (!wrapped) return;`. I exercised the squeeze path manually with `COPPERHEAD_DRAFT_SQUEEZE=1`: it runs 2 rounds, 7945 to 7660 mm², and three consecutive `draftSchematicPlacement` calls emit byte-identical sheets, so the loop itself is sound, just untested. In total **15 of the 25 new ACs have no test at all** (16.45, 16.48 to 16.54, 16.56 to 16.61, 16.64).
- **Fix:** set `process.env.COPPERHEAD_DRAFT_SQUEEZE = '1'` in the squeeze test (restoring it in `finally`) and assert `sq.rounds >= 1 && sq.boxAreaAfter < sq.boxAreaBefore` plus the `cells measured:` note. Replace the two `if (...)` guards with fixtures that actually produce a column wrap, so the assertions are reached.

### 5. [medium] Three ACs assert more than the code does

- **Where:** `openspec/specs/SPEC.md` AC-16.45, AC-16.61, AC-16.50.
- **Fails when:** a reader or an acceptance test is written from the AC text. **AC-16.45** says "**every** engine clearance check between two texts ... pads the candidate by 0.8 mm", but `src/kicad/draft/engine.ts:4391` falls back to `TEXT_PAD / 4` (0.2mm) and then to `0`, its own comment calling touching boxes "the last rung", and seven further text-vs-text sites never pad at all (`:4542`, `:4543`, `:4544`, `:3548`, `:3626`, `:3632`, `:3700`, comparing power-symbol Values against net labels and member fields, and a power Value *is* a field, the category the AC enumerates). `docs/.../schematic-drafting.md:145` repeats the absolute claim. **AC-16.61** ("boxes in one row share the row's bottom edge") is false on half the corpus because the title-corner giveback at `:4866-4867` hands the shared bottom back: `npn-switch` row 2 has MCU y 51.181..127.762 against Driver y 51.181..96.901, and `usb-atmega-node` column 2 has MCU x 89.154..191.643 against Clock x 89.154..122.809. AC-16.64's "share their top / their left" holds unconditionally on every board I drafted, so **AC-16.64 is the correct one and AC-16.61 is the wrong one**. **AC-16.50**'s 110mm title corner is checked at `:3127-3134` on wrap-time rects only; the frame clamp (`:5062`) and the retry probe (`:771`) never re-check it, so on an `intoStrip` sheet text that grows into the corner raises an error-severity `out-of-frame` at `src/kicad/legibility.ts:610` with no engine-side repair. The reference corpus is gated so the fixtures are safe; unseen boards are not.
- **Fix:** reword AC-16.45 to "every clearance check that places a net label or a part's fields pads by 0.8 mm; the power-value sweep and the last-resort label rung relax it deliberately", or pad the five power-value comparisons. Add the title-corner exception to AC-16.61 and `schematic-drafting.md:157`, or delete AC-16.61 as redundant with AC-16.64. Re-check the corner after enclosure growth.

### 6. [medium] A stale second copy of the usb-atmega-node sheet was committed at the board root

- **Where:** `manual-tests/reference-boards/usb-atmega-node/usb-atmega-node.kicad_sch` (5390 lines, new), plus `usb-atmega-node.kicad_pro` and `sym-lib-table`.
- **Fails when:** anyone opens the project root. The other three reference boards contain only `docs/`, `reference/`, `schematic.intent.json` and `sym-lib-cache/` (I confirmed against `ldo-demo/`), and `manual-tests/reference-boards/README.md:19` states drafts materialise under the gitignored `manual-tests/runs/`. The committed root sheet is a **stale A3 draft** while the canonical `reference/usb-atmega-node.kicad_sch` is A4; the two differ by 932 lines, so the repo now carries two contradictory copies of the same board and a reader gets the wrong one. Nothing depends on them: `test/draft-reference-boards.test.ts:29` copies the directory and the draft overwrites them.
- **Fix:** `git rm manual-tests/reference-boards/usb-atmega-node/{usb-atmega-node.kicad_sch,usb-atmega-node.kicad_pro,sym-lib-table}`.

### 7. [medium] The label-nudge test's only regression sentinel was deleted, and the file now contradicts itself

- **Where:** `test/draft-layout.test.ts:718`, versus the docstring at `:677`.
- **Fails when:** the whole de-collision ladder regresses. `expect(nudgedOnce, 'no seeded intent produced a nudge').toBe(true)` became `void nudgedOnce;`, so `MAX_LABEL_NUDGE`, the ride, the flip and the one-unit retreat at `engine.ts:4464-4517` now have no test that ever executes them; `expect(seen).toBeGreaterThan(100)` only asserts the loop ran. The new comment at `:712-716` explains the removal, but the original docstring at `:677` still says "a placement change that leaves no nudge in a few hundred small boards would be a change to look at, **and fails here**", so the file states both that it does and does not catch this. (The companion loosening at `:707`, `STUB - 0.001` to `1 - 0.001`, is **not** a regression: the one-unit retreat rung exists unchanged at the base commit, so the old floor was simply wrong.)
- **Fix:** replace with a hand-built intent that provokes a nudge (two facing long-named pins in a tight channel, as in `draft-hardening.test.ts`'s facing-labels repro) and assert `len > STUB`. Delete whichever of the two comments is now false.

### 8. [low] `flip()` mutates `placed` before five guards that abandon the chain

- **Where:** `src/kicad/draft/engine.ts:2715-2724`, with `continue` paths at `:2725`, `:2727`, `:2728`, `:2740`, `:2745`, `:2752`.
- **Fails when:** a 5-part series chain whose only anchor sits below it and meets its top lead: flipped at `:2723`, abandoned at `:2725`. `flip()` writes a 180-degree-rotated `placeCell` for every chain member into `placed`, and none of the six `continue` paths undoes it, so the parts stay rotated in their column slots with no `finalizeMoves` clearance check on that path. `placeCell(key, pl.x, pl.y, (pl.rot + 180) % 360)` also silently drops `pl.mirror`. Impact is small because two-lead KiCad symbols are origin-centred, so the recomputed body barely moves.
- **Fix:** move both `flip()` calls below the `chain.length > 4`, `anchors.length` and both-ends-open guards, or compute the flip into a local map and apply it only on the success path.

### 9. [low] The bank row-break width estimate rotates the other caps by the current cap's rotation

- **Where:** `src/kicad/draft/engine.ts:2213-2214`, `bodyBoundsOf(rotatedSym(instByKey.get(capRefs[k])!.sym, rot))`.
- **Fails when:** a rail run mixes a lying-down part with upright ones, e.g. one TVS plus four caps. `rot` was computed at `:2191` for `ref` alone, so the loop measures the other caps' heights as widths, `railRunW` is wrong, and the bank wraps a row too early or too late.
- **Fix:** hoist the stand-up decision into a `rotForBank(ref)` helper and call it per `k` inside the loop.

### 10. [low] Two more test assertions were weakened, and a standing false-positive fixture was edited without mention

- **Where:** `test/draft-sheet-fit.test.ts:249-250`; `test/fixtures/open-key/hardware/open-key.kicad_sch:112,117`.
- **Fails when:** the connector and the MCU land at exactly the same x. `toBeLessThan` became `toBeLessThanOrEqual` on both pairs in the test named "the IC leads its group", so it now passes when the jack sits zero mm to the right. The stated reason (a column-major wrap may stack the two groups) is real, but the weaker form no longer distinguishes "same column" from "connector zero mm to the right". Separately, `open-key.kicad_sch` was edited to absorb a **genuine** new error finding: the new flag bounds at `src/kicad/legibility.ts:234-243` make a rot-90 flag overlap the "Keyer" caption. The repair is legitimate, but open-key is the standing false-positive net for AC-16.16 and the commit body does not mention it. Release-note-worthy consequence: `copperhead check` now measures global-label flags two heights tall and centred where it previously measured bare text, so **users' existing hand-drawn sheets can newly fail**.
- **Fix:** keep the strict `<` when the two groups are on different wrap lines and relax only when they share one. Note the open-key edit and the checker change in the description, and consider a release note.

### 11. [low] Small contract and doc drifts

- `COPPERHEAD_DRAFT_TRACE` uses a truthiness gate (`engine.ts:76`), so `=0` or `=false` **enables** it, unlike its sibling `COPPERHEAD_DRAFT_SQUEEZE`'s `!== '1'`. Off by default and harmless, but inconsistent with its own documented `=1` contract.
- AC-16.53's wording: "no refused hang" is coded as `refusals <= best.refusals` (no *more*), and "at least three percent" as `area < best.area * 0.97` (*more than* 3%), at `engine.ts:837`. `sheetFit.squeeze.rounds` counts *attempted* rounds, so a rejected round still reports `rounds: 1`.
- AC-16.47's bound is exactly 3 (`engine.ts:800`) and deterministic, but the counter is shared by three widening mechanisms and **resets per squeeze/look round** (`:824`), so total re-drafts are bounded by the outer 16, not 3. An overlap surviving the budget produces only a `trace()`, no draft note.
- `docs/.../schematic-drafting.md:147` says channels are six grid units; `engine.ts:45` has `CHANNEL = 4` (commit `dc5364d` moved the constant without the doc). Stale comment at `engine.ts:1927` says "twice its square side" against `2.5 *` at `:1930`.
- AC-16.47 is written after AC-16.64 in SPEC.md rather than in numeric order.

## Invariant checks performed

All six checked against the diff, not assumed. `openspec validate build-copperhead-phase-1` returns **"Change 'build-copperhead-phase-1' is valid"** (exit 0), and `npx tsc --noEmit` is clean.

- **Spec-gated in** - pass. `src/agent/` is untouched. `src/agent/tools.ts:820` still filters by `requiresUnlock`, with `edit_file` (`:230`) and `write_file` (`:298`) omitted from the list rather than prompt-discouraged.
- **Verification-gated out** - pass. No command or agent code in the diff; no repair or rollback path changed.
- **`check`/`verify` stays LLM-free and network-free** - pass, traced rather than assumed. `src/commands/check.ts`'s transitive closure is 12 files whose only external imports are node builtins plus `execa`. This PR touches two files in that closure (`src/kicad/legibility.ts`, `src/kicad/sexp.ts`); neither adds an import and both stay pure.
- **No sexp serialization** - pass. `src/kicad/sexp.ts` exports only parse and read helpers; this PR's change is read-side only (an extra `hide` location, and property rotation composed with symbol rotation). `src/kicad/emit.ts` imports only `node:crypto` and builds text from a typed `PlacementModel`; it never sees a parsed tree. This is the documented "engine emits a brand-new schematic" capability, not round-tripping. Escaping and precision also check out: `q()` escapes `\` before `"` (correct order) and `knum` rounds to 4 decimals and strips `-0`, so `107.95000000000002` emits as `107.95`.
- **Sync-obligations ledger** - pass, untouched. `src/agent/tools.ts:800-807` still refuses commit on un-passed ERC/DRC and open obligations.
- **Secrets** - pass. `src/util/redact.ts` unchanged with `sk-[A-Za-z0-9_-]+` intact; `.gitignore` still carries `.env` and `.copperhead/runs/`. **`COPPERHEAD_DRAFT_TRACE` cannot leak a secret**: every use is `console.error` and it emits only group, net and ref names plus coordinates from the IR; the engine has no access to credentials or config. I confirmed the emitted sheet is byte-identical with and without the flag, so it does not alter placement.

Four things I verified rather than read, because they were the highest-risk parts of a 2120-line src diff:

- **Mirror times rotation composition.** `transformSym` (`:471`) mirrors `x -> -x` and `angle -> 180-angle`, flips the body box, then rotates, matching KiCad's parser, which sets the transform from `(at ... angle)` and composes `SYM_MIRROR_Y` after it. It agrees term for term with the checker's `pinAbsolute` (`sexp.ts:190`), so pins, lead ends, body bounds and the emitted `(at ... rot)` plus `(mirror y)` all agree. No sin/cos sign errors, and no helper reads untransformed geometry.
- **Field-text angle round trip.** `emit.ts`'s `fieldRot = rot % 180 === 90 ? 90 : 0` and `sexp.ts`'s `rot = (stored + symRot) mod 360` reproduce KiCad's `SCH_FIELD::GetDrawRotation()` horizontal/vertical toggle for all four rotations and for `(mirror y)`; `legibility.ts`'s `textBounds` treats 180 as horizontal, so nothing is measured on end. Checked all eight combinations.
- **Chain-walk cycle safety.** `walk()` (`:2667`) uses the shared `chain` array as its visited set and checks it on both the advance and the anchor branches, so a net cycle, a self-loop and a part reachable on two chains all terminate. `claimHang`/`claimInline` are guarded by `chain.includes`, `hungKeys` and an `n < 3` cap.
- **Re-draft termination and state isolation.** `for (round = 0; round < 16)` with `retries < 3` and `squeezeRounds >= 3` is bounded. `best` is always assigned on the first fall-through, so `best!` cannot be null, and since a passing round always falls through, the last fall-through *is* the only candidate: the "returns the last, not the best" concern does not apply. `reserves`, `frameSlack` and `wrapGap` reset on every fall-through; `measured` and `reach` are rebuilt per `draftOnce`; nothing mutates `validated`, `intent.*` or the shared `ResolvedSymbol`s (`transformSym` always spreads); there are no module-level mutable caches. Confirmed empirically: three consecutive calls emit byte-identical sheets, with squeeze on and off.

Also clean: **empty and degenerate inputs** (every unseeded `Math.min/max` I traced is either guarded or its infinity is consumed only inside a `.map` over the same empty array; no NaN escapes into the emitted file); **attachment floors moved up** as claimed (`usb-atmega-node` 0.7 to 0.85, `npn-switch` 0.25 to 0.55, the other two unchanged) and `tierC-score.json` moved **up** 84.49 to 85.25; **AC numbering** is contiguous 16.1 to 16.64 with each defined exactly once, no gaps and no duplicates; and **new tests are deterministic** (no `Date.now`, no `Math.random`, the seeded PRNG is explicit, and every sort is total over deterministically-ordered input).

**One caveat about the validation result.** `openspec validate` passing carries no signal for these 25 ACs: they were written straight into `openspec/specs/SPEC.md`, and with the only drafting change archived, no active change carries an AC-16 delta, so validation structurally cannot detect AC-to-code drift. That is the established pattern for this line of work, not something this PR introduces, but it is why finding 5 went unnoticed.

**Untested surface** (manual mapping, since coverage did not complete): 15 of the 25 new ACs have no test (16.45, 16.48-16.54, 16.56-16.61, 16.64); the squeeze rounds, the column-wrap branch and the look round are guarded behind conditions the fixtures never satisfy (finding 4); the 5-or-more-endpoint cluster path reaches no test (finding 3); the six `flip()` abandon paths reach no test (finding 8); and the mixed-rotation bank row-break reaches no test (finding 9).

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 268, branch feat/attach-everything, based on
feat/net-colours-and-testpoints (this PR is stacked on PR 267, not on main).

Check it out first:
  gh pr checkout 268

Do not change these, they are verified correct: the mirror-times-rotation
composition in transformSym, the field-text angle round trip between emit.ts and
sexp.ts, the chain-walk cycle safety in walk(), and the bounded re-draft loop.

FINDING 1 [high] Drafting is 4x slower on the reference boards and up to 500x slower
on a mid-size board, and it already fails a test on its own 60s budget.
  Where: src/kicad/draft/engine.ts:3069 (wrapMasonry), called from :3157-3161
    (k=2..4 times 2 budgets = 6 calls) inside tryPaperBudgeted's 7x4 = 28-combination
    sweep at :3192, per candidate paper, inside a loop that drafts up to 16 times
  Fails when: a group count lands where k ** n slips under the
    "if (total <= 2_000_000)" guard, which is a wild over-estimate of the real deal
    count and so does not bound the work it is meant to bound. Measured
    draftSchematicPlacement on both branches: 4 groups 51ms -> 89ms, 8 groups 57ms
    -> 3029ms, 12 groups 103ms -> 53873ms, 16 groups 152ms -> 23839ms. The
    non-monotonic 12-versus-16 result confirms attribution: n=12 enumerates k=3
    (531k deals) while n=16 trips the guard and skips it. Independently reproduced
    on the committed reference boards: usb-atmega-node 93ms -> 341ms, buck-12v-5v
    41ms -> 178ms. Concrete failure today: test/draft-hardening.test.ts ("a drafted
    sheet with a derived symbol loads in KiCad") takes 52.2s on this PR against
    27.5s on base, inside its own hardcoded 60s budget. Beyond CI, draft_schematic
    is called repeatedly in the stage-4 repair loop, so a 15-to-20-group board burns
    minutes per turn and can trip the per-turn watchdog in src/config.ts.
  Fix: memoize wrapMasonry(k, budgetH) on the current groupRects signature and hoist
    it out of the frac/div sweep; bound it by deals actually judged rather than by
    k ** n. Disabling only the enumeration removes about 94% of the cost and picks
    the same sheet on every case tested. Add a test that drafts a 12-group intent
    within a few seconds.

FINDING 2 [high] The wrap note reports the group count instead of the column count,
and says so on a shipped reference board.
  Where: src/kicad/draft/engine.ts:3299 and :3302,
    new Set(wrap.deltas.map(d => d.dx)).size
  Fails when: any column wrap. Deltas are target minus the group's own x1, and the
    pre-wrap ribbon gives every group a distinct x1, so every delta is distinct and
    the count is always the number of groups. The rows branch three lines below
    (:3307) does it correctly from the absolute y and carries a comment explaining
    this exact hazard: the fix was applied to dy and not to dx. Reproduced on a
    committed board: drafting usb-atmega-node emits "groups wrapped onto 6 columns
    to fit the sheet" while the committed reference has exactly 3 distinct group-box
    x1 values (23.749, 89.154, 196.723) across 6 groups. Masonry only ever builds 2
    to 4 columns, so 6 is impossible by construction. This falsifies AC-16.49's "the
    draft notes say which was used" and AC-16.62's only mechanically checkable
    clause, and it is text the agent reads back during stage-4 repair.
  Fix: new Set(wrap.deltas.map((d, i) => Math.round((groupRects[i]!.x1 + d.dx) / U))).size
    in both places, matching the rows branch. Add an assertion that the note's k
    equals the number of distinct box x1 values.

FINDING 3 [medium] Dropping the endpoint-count gate leaves a 5-or-more-endpoint
local net with zero connecting wires.
  Where: src/kicad/draft/engine.ts:1534 (gate replaced by a trace), :3931
    (anchorKey(a.ep.ref) === anchorKey(b.ep.ref) in near()), :3939
    (if (cl.length < 2 || cl.length > MAX_WIRED_ENDPOINTS) continue;)
  Fails when: one pin's local net has 5 or more endpoints. The removed gate was
    "if (eps.length > MAX_WIRED_ENDPOINTS || eps.some(cross-group)) continue;". The
    cross-group half is harmless since hangable() still requires
    memberKeySet.has(key). The endpoint half is not: hangs are claimed on a net the
    wire pass will refuse, and the new near() rule merges every part bound to one
    anchor into a single cluster regardless of wire span, so the cluster overruns
    MAX_WIRED_ENDPOINTS and :3939 skips it wholesale rather than routing a subset.
    Repro, one group: U1 (MCU8), R1..R4 (Device:R); nets
    BIG = [U1.4, R1.1, R2.1, R3.1, R4.1], GND = [U1.2, R1.2..R4.2],
    VCC = [U1.1, U1.5]. This PR: 5 wires, all 2-unit stubs, 5 global BIG flags, and
    no wire joining the pin to any resistor, while the four resistors occupy four
    lanes on U1's right shelf. Base commit, same intent: 11 BIG wires and 3 labels.
    With three resistors (4 endpoints) it wires cleanly, so the cliff is exactly
    at 5. This also makes AC-16.37 and the "Local runs are wired" docs section
    ("clustered by group and wire span, and every cluster that routes cleanly is
    drawn as one wired run") no longer describe the code.
  Fix: at :3939 drop the upper bound and clamp the subset search instead:
    "if (cl.length < 2) continue;" with
    "for (let size = Math.min(cl.length, MAX_WIRED_ENDPOINTS); size >= 2 && !done; size--)",
    which keeps the enumeration at C(n,4) rather than 2^n. Regression test: the
    intent above, asserting measureWiringStyle(...).twoPinAttached === 4. Note the
    same skip is inherited from PR 266; fixing it there fixes it here.

FINDING 4 [medium] The multi-round re-draft paths that five ACs describe are never
executed by the suite, and their tests pass by skipping.
  Where: test/draft-sheet-fit.test.ts:483 (squeeze), :529 (AC-16.62), :512 (AC-16.63)
  Fails when: a regression lands in any of them. Nothing in test/ sets
    COPPERHEAD_DRAFT_SQUEEZE, so src/kicad/draft/engine.ts:847 always breaks at
    squeezeRounds === 0 and :851 writes boxAreaBefore === boxAreaAfter by
    construction (both from the same best), making toBeLessThanOrEqual unfailable
    and the sq.rounds > 0 branch dead. The AC-16.62 test asserts its shape string
    only inside "if (notes.some(/laid in \d+ columns/))", and its sevenGroups()
    fixture, added for this AC, actually produces a rows wrap, so the branch never
    runs; a single assertion that the note's k equals the number of distinct box x1
    values would have caught finding 2. The AC-16.63 test is wrapped in
    "if (!wrapped) return;". Exercised manually with COPPERHEAD_DRAFT_SQUEEZE=1 the
    squeeze runs 2 rounds, 7945 to 7660 mm-squared, and three consecutive
    draftSchematicPlacement calls emit byte-identical sheets, so the loop is sound,
    just untested. In total 15 of the 25 new ACs have no test (16.45, 16.48 to
    16.54, 16.56 to 16.61, 16.64).
  Fix: set process.env.COPPERHEAD_DRAFT_SQUEEZE = '1' in the squeeze test, restoring
    it in finally, and assert sq.rounds >= 1 and sq.boxAreaAfter < sq.boxAreaBefore
    plus the "cells measured:" note. Replace the two "if" guards with fixtures that
    actually produce a column wrap so the assertions are reached.

FINDING 5 [medium] Three ACs assert more than the code does.
  Where: openspec/specs/SPEC.md AC-16.45, AC-16.61, AC-16.50
  Fails when: a reader or an acceptance test is written from the AC text. AC-16.45
    says "every engine clearance check between two texts ... pads the candidate by
    0.8 mm", but src/kicad/draft/engine.ts:4391 falls back to TEXT_PAD / 4 (0.2mm)
    and then to 0, its own comment calling touching boxes "the last rung", and seven
    further text-vs-text sites never pad at all (:4542, :4543, :4544, :3548, :3626,
    :3632, :3700, comparing power-symbol Values against net labels and member
    fields; a power Value is a field, the category the AC enumerates).
    docs/src/content/docs/reference/schematic-drafting.md:145 repeats the absolute
    claim. AC-16.61 ("boxes in one row share the row's bottom edge") is false on
    half the corpus because the title-corner giveback at :4866-4867 hands the shared
    bottom back: npn-switch row 2 has MCU y 51.181..127.762 against Driver y
    51.181..96.901, and usb-atmega-node column 2 has MCU x 89.154..191.643 against
    Clock x 89.154..122.809. AC-16.64's "share their top / their left" holds
    unconditionally on every board drafted, so AC-16.64 is the correct one and
    AC-16.61 is the wrong one. AC-16.50's 110mm title corner is checked at
    :3127-3134 on wrap-time rects only; the frame clamp (:5062) and the retry probe
    (:771) never re-check it, so on an intoStrip sheet text that grows into the
    corner raises an error-severity out-of-frame at src/kicad/legibility.ts:610 with
    no engine-side repair.
  Fix: reword AC-16.45 to "every clearance check that places a net label or a part's
    fields pads by 0.8 mm; the power-value sweep and the last-resort label rung relax
    it deliberately", or pad the five power-value comparisons. Add the title-corner
    exception to AC-16.61 and schematic-drafting.md:157, or delete AC-16.61 as
    redundant with AC-16.64. Re-check the corner after enclosure growth.

FINDING 6 [medium] A stale second copy of the usb-atmega-node sheet was committed at
the board root.
  Where: manual-tests/reference-boards/usb-atmega-node/usb-atmega-node.kicad_sch
    (5390 lines, new), plus usb-atmega-node.kicad_pro and sym-lib-table
  Fails when: anyone opens the project root. The other three reference boards
    contain only docs/, reference/, schematic.intent.json and sym-lib-cache/, and
    manual-tests/reference-boards/README.md:19 states drafts materialise under the
    gitignored manual-tests/runs/. The committed root sheet is a stale A3 draft
    while the canonical reference/usb-atmega-node.kicad_sch is A4; the two differ by
    932 lines, so the repo carries two contradictory copies of the same board.
    Nothing depends on them: test/draft-reference-boards.test.ts:29 copies the
    directory and the draft overwrites them.
  Fix: git rm manual-tests/reference-boards/usb-atmega-node/usb-atmega-node.kicad_sch
    manual-tests/reference-boards/usb-atmega-node/usb-atmega-node.kicad_pro
    manual-tests/reference-boards/usb-atmega-node/sym-lib-table

FINDING 7 [medium] The label-nudge test's only regression sentinel was deleted, and
the file now contradicts itself.
  Where: test/draft-layout.test.ts:718, versus the docstring at :677
  Fails when: the whole de-collision ladder regresses.
    expect(nudgedOnce, 'no seeded intent produced a nudge').toBe(true) became
    "void nudgedOnce;", so MAX_LABEL_NUDGE, the ride, the flip and the one-unit
    retreat at src/kicad/draft/engine.ts:4464-4517 now have no test that ever
    executes them; expect(seen).toBeGreaterThan(100) only asserts the loop ran. The
    new comment at :712-716 explains the removal, but the original docstring at :677
    still says "a placement change that leaves no nudge in a few hundred small
    boards would be a change to look at, and fails here", so the file states both
    that it does and does not catch this. The companion loosening at :707
    (STUB - 0.001 to 1 - 0.001) is NOT a regression: the one-unit retreat rung
    exists unchanged at the base commit, so the old floor was simply wrong.
  Fix: replace with a hand-built intent that provokes a nudge (two facing
    long-named pins in a tight channel, as in draft-hardening.test.ts's
    facing-labels repro) and assert len > STUB. Delete whichever of the two comments
    is now false.

FINDING 8 [low] (optional) flip() mutates placed before five guards that abandon the
chain.
  Where: src/kicad/draft/engine.ts:2715-2724, with continue paths at :2725, :2727,
    :2728, :2740, :2745, :2752
  Fails when: a 5-part series chain whose only anchor sits below it and meets its
    top lead: flipped at :2723, abandoned at :2725. flip() writes a
    180-degree-rotated placeCell for every chain member into placed, and none of the
    six continue paths undoes it, so the parts stay rotated in their column slots
    with no finalizeMoves clearance check on that path.
    placeCell(key, pl.x, pl.y, (pl.rot + 180) % 360) also silently drops pl.mirror.
    Impact is small because two-lead KiCad symbols are origin-centred.
  Fix: move both flip() calls below the chain.length > 4, anchors.length and
    both-ends-open guards, or compute the flip into a local map and apply it only on
    the success path.

FINDING 9 [low] (optional) The bank row-break width estimate rotates the other caps
by the current cap's rotation.
  Where: src/kicad/draft/engine.ts:2213-2214,
    bodyBoundsOf(rotatedSym(instByKey.get(capRefs[k])!.sym, rot))
  Fails when: a rail run mixes a lying-down part with upright ones, e.g. one TVS
    plus four caps. rot was computed at :2191 for ref alone, so the loop measures
    the other caps' heights as widths, railRunW is wrong, and the bank wraps a row
    too early or too late.
  Fix: hoist the stand-up decision into a rotForBank(ref) helper and call it per k
    inside the loop.

FINDING 10 [low] (optional) Two more test assertions were weakened, and a standing
false-positive fixture was edited without mention.
  Where: test/draft-sheet-fit.test.ts:249-250;
    test/fixtures/open-key/hardware/open-key.kicad_sch:112,117
  Fails when: the connector and the MCU land at exactly the same x. toBeLessThan
    became toBeLessThanOrEqual on both pairs in the test named "the IC leads its
    group", so it passes when the jack sits zero mm to the right. The stated reason
    (a column-major wrap may stack the two groups) is real, but the weaker form no
    longer distinguishes "same column" from "connector zero mm to the right".
    Separately, open-key.kicad_sch was edited to absorb a genuine new error finding:
    the new flag bounds at src/kicad/legibility.ts:234-243 make a rot-90 flag
    overlap the "Keyer" caption. The repair is legitimate, but open-key is the
    standing false-positive net for AC-16.16 and the commit body does not mention
    it. Consequence worth a release note: copperhead check now measures global-label
    flags two heights tall and centred where it previously measured bare text, so
    users' existing hand-drawn sheets can newly fail.
  Fix: keep the strict < when the two groups are on different wrap lines and relax
    only when they share one. Note the open-key edit and the checker change in the
    description.

FINDING 11 [low] (optional) Small contract and doc drifts.
  Where: src/kicad/draft/engine.ts:76, :837, :800, :824, :1927, :45;
    docs/src/content/docs/reference/schematic-drafting.md:147;
    openspec/specs/SPEC.md
  Fails when: someone relies on the documented contract. COPPERHEAD_DRAFT_TRACE uses
    a truthiness gate at :76, so =0 or =false enables it, unlike its sibling
    COPPERHEAD_DRAFT_SQUEEZE's !== '1'. AC-16.53's "no refused hang" is coded as
    refusals <= best.refusals (no more) and "at least three percent" as
    area < best.area * 0.97 (more than 3%) at :837; sheetFit.squeeze.rounds counts
    attempted rounds, so a rejected round still reports rounds: 1. AC-16.47's bound
    is exactly 3 at :800 and deterministic, but the counter is shared by three
    widening mechanisms and resets per squeeze/look round at :824, so total
    re-drafts are bounded by the outer 16, not 3, and an overlap surviving the
    budget produces only a trace() with no draft note. schematic-drafting.md:147
    says channels are six grid units while engine.ts:45 has CHANNEL = 4. The comment
    at engine.ts:1927 says "twice its square side" against 2.5 * at :1930. AC-16.47
    is written after AC-16.64 in SPEC.md rather than in numeric order.
  Fix: make the trace gate === '1'; correct the three AC wordings; update the two
    docs and the stale comment; move AC-16.47 into numeric order.

Also do two things to the PR itself, which are not code changes:
  1. Update the description. It documents AC-16.40 to AC-16.47 and reports "838
     passed", but that is the first of nine commits. The PR now adds AC-16.40
     through AC-16.64, twenty-five acceptance criteria, and the measured figures in
     the body are all from commit a0ffb68. A reviewer cannot tell what they are
     approving.
  2. Fill in the "## Testing" section and the invariant checklist required by
     .github/pull_request_template.md; the template check is failing because they
     are missing.

These parts of the diff are reached by no test: 15 of the 25 new ACs (16.45,
16.48 to 16.54, 16.56 to 16.61, 16.64); the squeeze rounds, the column-wrap branch
and the look round, all guarded behind conditions the fixtures never satisfy
(finding 4); the 5-or-more-endpoint cluster path (finding 3); the six flip() abandon
paths (finding 8); and the mixed-rotation bank row-break (finding 9). Add tests for
the new branches and error paths they contain.

This PR changes spec-level behavior across 25 acceptance criteria, so
openspec/specs/SPEC.md and the docs pages have to describe what the code actually
does, and `openspec validate build-copperhead-phase-1` has to keep passing (it
passes today). Be aware that validation cannot detect AC-to-code drift here, because
these ACs live directly in SPEC.md and no active change carries an AC-16 delta, so
the tests are the only thing that will hold them honest.

    never by prompt text.
    snapshot.
    src/commands/check may touch a provider, an API key, or the network. Its closure
    is 12 files whose only external imports are node builtins plus execa, and this
    PR touches two of them (src/kicad/legibility.ts, src/kicad/sexp.ts); keep both
    pure and add no imports.
    edited only via anchored exact-match text replace, never round-tripped. The
    emitter builds text from a typed PlacementModel and never sees a parsed tree.
    COPPERHEAD_DRAFT_TRACE must stay console.error-only and must not alter placement.

Then verify:
  npm run typecheck
  npm run build
  npm test

not yours to fix. The draft-hardening timeout in finding 1 is different: that one is
caused by this PR and must go away once finding 1 is fixed. Do not silence any
failure by deleting an assertion, skipping a test, or loosening a threshold, and in
particular do not raise the 60s budget in draft-hardening.test.ts to hide finding 1.
If something fails and the fix is not obvious, stop and report it.

````

PR #271, opened 2026-09-05
docs(site): restyle the docs after GitBook's clean theme

animesh-chouhan:
## What

Restyles docs.copperhead.sh after GitBook's "clean" theme (the look of docs.quilter.ai): a neutral near-black palette with the copper accent, a floating rounded sidebar with per-page icons and a tinted active pill, a wide header search box with a "Get started" button, a section eyebrow and lead paragraph above each page, and a landing page rebuilt as a "Welcome" doc page with "Jump right in" rows and "Keep exploring" card groups. Still Astro Starlight, so search, edit links, last-updated stamps, and the llms.txt plugin are unchanged.

## Why

The stock Starlight look (blue accent, splash hero, boxed sidebar) reads as a template. The docs are the first thing most people open, and the GitBook clean layout is a proven docs pattern: denser navigation, a narrow readable column, quiet borders instead of shadows, and a landing page that routes readers instead of pitching them.

## Design

Everything lives in `docs/`; no runtime code changes.

- [custom.css](docs/src/styles/custom.css) holds the palette, type scale, and layout. It is deliberately unlayered so it overrides Starlight's own cascade layers without `!important`. Starlight names its gray scale from the dark palette, so `--sl-color-white` is the strongest text colour and `--sl-color-black` the page background in both modes; the file documents this.
- Component overrides in [docs/src/components/](docs/src/components/): [Header.astro](docs/src/components/Header.astro), [Sidebar.astro](docs/src/components/Sidebar.astro) plus [SidebarSublist.astro](docs/src/components/SidebarSublist.astro), [PageTitle.astro](docs/src/components/PageTitle.astro), and [ThemeSelect.astro](docs/src/components/ThemeSelect.astro). The sidebar keeps Starlight's persister and restore points so open/closed state and scroll position survive navigation. The theme toggle writes the same `starlight-theme` storage key Starlight's inline provider reads, so there is no flash on load.
- Per-page sidebar icons are set in [astro.config.mjs](docs/astro.config.mjs) via `attrs: { 'data-icon': '<starlight icon>' }` and rendered by the sublist override.
- [JumpLink.astro](docs/src/components/JumpLink.astro), [DocCard.astro](docs/src/components/DocCard.astro), and [DocCards.astro](docs/src/components/DocCards.astro) are the landing-page building blocks used by [index.mdx](docs/src/content/docs/index.mdx).
- Fonts: Inter Variable and IBM Plex Mono via `@fontsource-variable/inter` and `@fontsource/ibm-plex-mono` (self-hosted, no third-party requests). The lockfile diff also drops `libc` fields from sharp's optional-dependency entries; that is npm 11 normalising the lockfile, not a dependency change.
- Theme still follows the system preference by default. Quilter forces dark; easy to flip if preferred.

Gotcha for anyone iterating locally: after changing the `expressiveCode` config, Astro's content-layer cache keeps the old code-block stylesheet link on unchanged `.md` pages, which 404s in a local preview. `astro build --force` (or deleting `docs/.astro`) fixes it; CI builds from a clean checkout and is unaffected.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | `npm run typecheck` | n/a (docs-only, no `src/` changes) |
| Build | `npm run build` | n/a (docs-only) |
| Unit + offline | `npm test` | n/a (docs-only) |
| Docs build | `npm ci && npm run build` in `docs/` (fresh worktree from main) | pass, 17 pages |

- Sandbox variant used (`create` / `edit`): n/a, docs-only
- Commands run and outcome: built the site, served `docs/dist` locally, and screenshotted it with headless Chromium.

```text
$ npm --prefix docs run build
17 page(s) built, every page links the single emitted ec.*.css

Checked in the browser (1440, 1024, 820, and 390 px wide; dark and light):
  /                                   landing page, sidebar, right rail, pagination
  /getting-started/quickstart/        steps, terminal code frames, tip callout
  /reference/cli/ and /reference/configuration/   tables, inline code, highlighted JSON
  mobile drawer open, search dialog with results, theme toggle
```

## Docs

All changes are the docs site itself, under [docs/](docs/). No README, configuration reference, or CHANGELOG/DECISIONS entries touched.

---

## Invariant checklist

All N/A: this PR changes only the documentation site's theme and landing page, no runtime code.

- [ ] N/A **Spec-gated in**: `edit_file`/`write_file` stay structurally absent from the agent tool list until an OpenSpec proposal validates (gated by omission, not prompt text).
- [ ] N/A **Verification-gated out**: mutations still end in ERC (and DRC when the board changed) passing, with repair up to `maxRepairCycles` then rollback to the git snapshot.
- [ ] N/A **`check`/`verify` stays LLM-free and network-free**: nothing reachable from `src/commands/check` touches a provider, an API key, or the network.
- [ ] N/A **No sexp serialization**: the `src/kicad/` parser stays read-only; KiCad files are edited only via anchored exact-match text replace (no round-tripping).
- [ ] N/A **Sync-obligations ledger**: post-tool-call hooks keep feeding the ledger, and commit keeps refusing while obligations are open.
- [ ] N/A **Secrets**: transcripts and summaries redact `sk-[A-Za-z0-9_-]+` at write time; keys live only in env vars; `.gitignore` keeps `.env` and `.copperhead/runs/`.
- [ ] N/A **Spec coherence**: if spec-level behavior changed, SPEC.md and the change artifacts (`proposal.md`, `design.md`, delta specs, `tasks.md`) moved together and `openspec validate <change>` passes.

### DIFF
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -1,0 +1,37 @@
+name: Docs build
+
+# The deploy workflow only runs on push to main, so without this a docs change
+# reaches docs.copperhead.sh with its build unverified: a broken import goes
+# green on every PR check and fails afterwards, on main. This job builds the
+# site exactly as the deploy does, and publishes nothing.
+on:
+  pull_request:
+    paths:
+      - 'docs/**'
+      - '.github/workflows/docs-pr.yml'
+      - '.github/workflows/docs.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: docs-build-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # lastUpdated timestamps come from git history
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+          cache-dependency-path: docs/package-lock.json
+      - run: npm ci
+        working-directory: docs
+      - run: npm run build
+        working-directory: docs
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -28,6 +28,7 @@ export default defineConfig({
         Sidebar: './src/components/Sidebar.astro',
         PageTitle: './src/components/PageTitle.astro',
         ThemeSelect: './src/components/ThemeSelect.astro',
+        Footer: './src/components/Footer.astro',
       },
       customCss: [
         '@fontsource-variable/inter',
@@ -136,7 +137,7 @@ export default defineConfig({
             {
               label: 'Technical spec',
               link: `${REPO}/blob/main/openspec/specs/SPEC.md`,
-              attrs: { target: '_blank', 'data-icon': 'external' },
+              attrs: { target: '_blank', rel: 'noopener', 'data-icon': 'external' },
             },
           ],
         },
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -14,12 +14,41 @@ export default defineConfig({
       title: 'copperhead',
       description:
         'Cursor for circuit boards: an AI agent that designs, documents, and validates real PCBs on KiCad repositories',
+      logo: { src: './src/assets/mark.svg', alt: '' },
       social: [
         { icon: 'external', label: 'copperhead.sh', href: 'https://copperhead.sh' },
         { icon: 'github', label: 'GitHub', href: REPO },
       ],
       editLink: { baseUrl: `${REPO}/edit/main/docs/` },
       lastUpdated: true,
+      // Theme: see src/styles/custom.css for the palette and layout, and
+      // src/components/ for the overridden header, sidebar, title, and theme toggle.
+      components: {
+        Header: './src/components/Header.astro',
+        Sidebar: './src/components/Sidebar.astro',
+        PageTitle: './src/components/PageTitle.astro',
+        ThemeSelect: './src/components/ThemeSelect.astro',
+      },
+      customCss: [
+        '@fontsource-variable/inter',
+        '@fontsource/ibm-plex-mono/400.css',
+        '@fontsource/ibm-plex-mono/500.css',
+        './src/styles/custom.css',
+      ],
+      expressiveCode: {
+        styleOverrides: {
+          borderRadius: '0.75rem',
+          codeFontFamily: "'IBM Plex Mono', ui-monospace, SFMono-Regular, Menlo, Consolas, monospace",
+          codeFontSize: '0.875rem',
+          codeLineHeight: '1.65',
+          codePaddingBlock: '0.875rem',
+          codePaddingInline: '1rem',
+          frames: {
+            shadowColor: 'transparent',
+            frameBoxShadowCssValue: 'none',
+          },
+        },
+      },
       plugins: [
         starlightLlmsTxt({
           projectName: 'copperhead',
@@ -59,7 +88,6 @@ export default defineConfig({
           ],
         }),
       ],
-      customCss: ['./src/styles/custom.css'],
       head: [
         { tag: 'meta', attrs: { name: 'theme-color', content: '#b87333' } },
         { tag: 'link', attrs: { rel: 'icon', href: '/favicon.ico', sizes: '32x32' } },
@@ -69,46 +97,48 @@ export default defineConfig({
         { tag: 'meta', attrs: { name: 'twitter:card', content: 'summary_large_image' } },
         { tag: 'meta', attrs: { name: 'twitter:image', content: 'https://docs.copperhead.sh/og.png' } },
       ],
+      // `data-icon` picks the leading icon for each row; see SidebarSublist.astro.
       sidebar: [
         {
           label: 'Getting started',
           items: [
-            { label: 'Introduction', link: '/getting-started/introduction/' },
-            { label: 'Quickstart', link: '/getting-started/quickstart/' },
-            { label: 'Agent install', link: '/getting-started/agent-install/' },
-            { label: 'Simple demo', link: '/getting-started/demo/' },
+            { label: 'Welcome', link: '/', attrs: { 'data-icon': 'star' } },
+            { label: 'Introduction', link: '/getting-started/introduction/', attrs: { 'data-icon': 'information' } },
+            { label: 'Quickstart', link: '/getting-started/quickstart/', attrs: { 'data-icon': 'rocket' } },
+            { label: 'Agent install', link: '/getting-started/agent-install/', attrs: { 'data-icon': 'puzzle' } },
+            { label: 'Simple demo', link: '/getting-started/demo/', attrs: { 'data-icon': 'laptop' } },
           ],
         },
         {
           label: 'Concepts',
           items: [
-            { label: 'The agent loop', link: '/concepts/agent-loop/' },
-            { label: 'Guardrails', link: '/concepts/guardrails/' },
-            { label: 'Docs as memory', link: '/concepts/docs-as-memory/' },
+            { label: 'The agent loop', link: '/concepts/agent-loop/', attrs: { 'data-icon': 'random' } },
+            { label: 'Guardrails', link: '/concepts/guardrails/', attrs: { 'data-icon': 'padlock' } },
+            { label: 'Docs as memory', link: '/concepts/docs-as-memory/', attrs: { 'data-icon': 'open-book' } },
           ],
         },
         {
           label: 'Workflows',
           items: [
-            { label: 'Design from a brief', link: '/workflows/create-from-brief/' },
-            { label: 'Edit an existing board', link: '/workflows/edit-existing-board/' },
-            { label: 'Verify and sync', link: '/workflows/verify-and-sync/' },
+            { label: 'Design from a brief', link: '/workflows/create-from-brief/', attrs: { 'data-icon': 'add-document' } },
+            { label: 'Edit an existing board', link: '/workflows/edit-existing-board/', attrs: { 'data-icon': 'pencil' } },
+            { label: 'Verify and sync', link: '/workflows/verify-and-sync/', attrs: { 'data-icon': 'approve-check' } },
           ],
         },
         {
           label: 'Reference',
           items: [
-            { label: 'CLI', link: '/reference/cli/' },
-            { label: 'Configuration', link: '/reference/configuration/' },
-            { label: 'Schematic legibility rules', link: '/reference/schematic-legibility/' },
-            { label: 'How schematics are drafted', link: '/reference/schematic-drafting/' },
-            { label: 'The schematic intent file', link: '/reference/schematic-intent/' },
+            { label: 'CLI', link: '/reference/cli/', attrs: { 'data-icon': 'forward-slash' } },
+            { label: 'Configuration', link: '/reference/configuration/', attrs: { 'data-icon': 'setting' } },
+            { label: 'Schematic legibility rules', link: '/reference/schematic-legibility/', attrs: { 'data-icon': 'list-format' } },
+            { label: 'How schematics are drafted', link: '/reference/schematic-drafting/', attrs: { 'data-icon': 'pen' } },
+            { label: 'The schematic intent file', link: '/reference/schematic-intent/', attrs: { 'data-icon': 'document' } },
+            {
+              label: 'Technical spec',
+              link: `${REPO}/blob/main/openspec/specs/SPEC.md`,
+              attrs: { target: '_blank', 'data-icon': 'external' },
+            },
           ],
-        },
-        {
-          label: 'Technical spec',
-          link: `${REPO}/blob/main/openspec/specs/SPEC.md`,
-          attrs: { target: '_blank' },
         },
       ],
     }),
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -7,6 +7,8 @@
       "name": "copperhead-docs",
       "dependencies": {
         "@astrojs/starlight": "^0.41.3",
+        "@fontsource-variable/inter": "^5.3.0",
+        "@fontsource/ibm-plex-mono": "^5.3.0",
         "astro": "^7.1.1",
         "sharp": "^0.35.3",
         "starlight-llms-txt": "^0.11.0"
@@ -71,9 +73,6 @@
       "cpu": [
         "arm64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -89,9 +88,6 @@
       "integrity": "sha512-xqE8BVbDoBueK/B47w30PtkVofUWJKGkwoMVE+EOMLf11rnoANxIAdA9FPqY+rng4oNI5ndHGsri1yPj2k8vZQ==",
       "cpu": [
         "arm64"
-      ],
-      "libc": [
-        "musl"
       ],
       "license": "MIT",
       "optional": true,
@@ -109,9 +105,6 @@
       "cpu": [
         "x64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -127,9 +120,6 @@
       "integrity": "sha512-16q0fYf7kpbmdObZEeZJEup8hQv/whgNwVjrSvT8umrKwLDSnNIWiQpm09lQQu6bweZB0XyIvHwlPitvJhC+hg==",
       "cpu": [
         "x64"
-      ],
-      "libc": [
-        "musl"
       ],
       "license": "MIT",
       "optional": true,
@@ -451,9 +441,6 @@
       "cpu": [
         "arm64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -466,9 +453,6 @@
       "integrity": "sha512-v39HxiwGC5Rqm01HksP6+5Y+xKLPlsuVFgIgpEAo+SiQ22c+mJVhS3u7Z6ePAKdhL5NJoK1xq70kLz3L13AhpQ==",
       "cpu": [
         "arm64"
-      ],
-      "libc": [
-        "musl"
       ],
       "license": "MIT",
       "optional": true,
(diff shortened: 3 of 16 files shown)


animesh-chouhan:
_Automated `pr-review` pass (Claude Code). Not a human review._

## Verdict: approve with nits

The theme is well-built: it mirrors Starlight's own component structure rather than forking it, keeps the sidebar persister and restore points, reuses the `starlight-theme` storage key so there is no theme flash, and every claim in the description checked out against the built output. One real regression is worth fixing before merge (printing renders the dark palette), the rest are nits.

I built the site from a clean `npm ci` in an isolated worktree: **17 pages, exit 0**. That matters here because CI never builds the docs on a PR (see L2).

## Metrics

```
lines: +1423 / -180 (net 1243) across 14 files  ·  src +0/-0, test +0/-0, docs +1403/-78, spec +0/-0, other +0/-0
generated (excluded from areas): 1 file(s), +20/-102
new vs net: 9 new files (0 in src), 5 modified; 0 new src lines (the net-new surface)
tests: +0 test lines; suite not measured -> 821/21/3 (pass/skip/fail, base -> head)
docs build: PASS — npm ci + astro build in the review worktree, 17 pages, exit 0
```

**Method.** Measured in the review worktree `/tmp/pr-review-worktrees/copperhead-eea95c6f/pr-271` at head `199d690`; the user's checkout was never entered and stayed on `feat/vertex-ai-provider`. Merge base `060f92f` (`git merge-base origin/main HEAD`). Areas from `git diff --numstat`, generated files excluded. Suite via `npx vitest run --reporter=json`. Base suite not run (base CI is green). Diff coverage skipped — zero changed `src/` lines, so there is nothing to cover. Docs build measured by me, not by the script: `npm ci && npm run build` in `docs/`.

**On the 3 suite failures:** pre-existing and environmental, not caused by this PR — it changes zero `src/` and zero `test/` lines, and the `test` job on GitHub is green at head. All three shell out to `kicad-cli` ERC and fail on the reviewer's local KiCad config (they need an isolated `KICAD_CONFIG_HOME` to match CI):

```
gating-sync.test.ts      :: finish blocks on open obligations and unverified ERC
init-check.test.ts       :: clean fixture: everything green (AC-2.1) and stable JSON keys (AC-2.4)
stage-completion.test.ts :: error-severity findings keep the stage active; a clean sheet completes
```

**Untested surface:** 9 new `.astro` components and 432 new CSS lines with no automated assertions of any kind. This is normal for a docs theme, but combined with L2 it means the entire diff has no CI gate.

## Findings

### M1 · medium · Printing any docs page renders the dark theme; Starlight's print stylesheet is defeated

`docs/src/styles/custom.css:14` (and `:69`)

The palette is deliberately unlayered so it beats Starlight's `@layer starlight.*` rules. Starlight's print stylesheet is **also unlayered** (`@media print { :root { … } }`, "force light mode when printing"), so the two `:root` blocks have identical specificity `(0,1,0)` and `@media print` adds none — document order decides, and `custom.css` loses nothing to `print.css` because it comes second.

Verified against the emitted CSS rather than inferred:

```
index.html:  <link rel="stylesheet" href="/_astro/print.ehPL0gv-.css" media="print">   <-- first
             <link rel="stylesheet" href="/_astro/common.GWb5bhwJ.css">                <-- second
print.ehPL0gv-.css:  starts `@media print{:root{--sl-color-black:#fff; …}`  (unlayered)
common.GWb5bhwJ.css: copperhead `:root` palette at byte 5693; first `@layer` at 14329  (unlayered, later)
```

**Failure scenario:** a reader on a dark-mode OS (no stored preference → Starlight's inline provider sets `data-theme="dark"`) opens `/reference/cli/` and hits Print or Save as PDF. `--sl-color-black` resolves to `#1b1b1c` instead of `#fff` and `--sl-color-white` to `#f7f8f8` instead of near-black, so the page prints as light text on a near-black background — an ink-flooded or (with backgrounds off) invisible page. Light mode is unaffected, because `:root[data-theme='light']` is `(0,2,0)` and outranks `print.css` anyway. A CLI and configuration reference is exactly the kind of page people print.

**Fix:** append a print block at the end of `custom.css` that outranks the dark `:root` in both modes:

```css
/* Starlight's print.css is unlayered too, so this file would otherwise win
   and print the dark palette. Restore light values for print. */
@media print {
	:root,
	:root[data-theme='dark'] {
		--sl-color-white: #1a1b1e;
		--sl-color-gray-1: #2b2d32;
		--sl-color-gray-2: #545962;
		--sl-color-gray-3: #6d747f;
		--sl-color-gray-4: #c2c8d2;
		--sl-color-gray-5: #e3e7ed;
		--sl-color-gray-6: #eff2f6;
		--sl-color-black: #ffffff;
		--sl-color-accent: #a35d1f;
		--sl-color-text-accent: #9a5619;
		--ch-hover-bg: transparent;
		--ch-pill-bg: transparent;
	}
}
```

### L2 · low · CI never builds the docs on a pull request

`.github/workflows/docs.yml:4-8`

`Deploy docs` triggers only on `push` to `main`. No workflow builds `docs/` for a PR, so a 1,423-line docs change merges with its build unverified and the first place a break would surface is the production deploy to `docs.copperhead.sh`. The gap is pre-existing, not introduced here — but this PR is the first one large enough for it to matter, and it is why I built the site by hand for this review.

**Failure scenario:** a future docs PR adds a component with a bad Starlight import. Every check goes green, it merges, `Deploy docs` fails on `main`, and the site is stale until someone notices.

**Fix:** add a build-only job gated on `pull_request` with `paths: ['docs/**']`, reusing the existing build steps without the deploy job.

### L3 · low · Footer visual order no longer matches keyboard focus order

`docs/src/styles/custom.css:379-393`

Starlight's `Footer.astro` emits `.meta` (Edit page, Last updated) **before** `<Pagination />`. The new rules flip them visually with `footer .meta { order: 2 }` / `.pagination-links { order: 1 }`, but `order` does not move DOM or tab order. Confirmed in the built markup for `/reference/cli/`: `<div class="meta">` with the "Edit page" link, then `<div class="pagination-links">`.

**Failure scenario:** a keyboard user tabbing past the last content link lands on "Edit page" — visually the bottom-most element on the page — then tabs *backwards up the page* into the Previous/Next cards. That is a WCAG 2.4.3 (Focus Order) mismatch.

**Fix:** keep the DOM order and restyle in place (drop both `order` declarations and let the meta row sit above the pagination cards), or override `Footer.astro` too so the DOM order matches the intended visual order.

### L4 · low · The mobile "On this page" dropdown still lists the Overview entry the desktop rail hides

`docs/src/styles/custom.css:260`

```css
starlight-toc li:has(> a[href='#_top']) { display: none; }
```

The mobile table of contents is a separate custom element, `<mobile-starlight-toc>`, and it does **not** nest a `<starlight-toc>`, so the selector never matches it. Confirmed in the built markup for `/reference/cli/`: the mobile dropdown still contains `<li><a href="#_top"><span>Overview</span></a></li>` while the desktop rail hides it.

**Failure scenario:** on a phone, the "On this page" dropdown opens with a leading "Overview" row that no desktop reader ever sees — the two navigations disagree about the page's headings.

**Fix:** widen the selector to cover both, e.g. `:is(starlight-toc, mobile-starlight-toc) li:has(> a[href='#_top']) { display: none; }`.

### Nits

- `docs/astro.config.mjs:141` — the "Technical spec" sidebar entry sets `target: '_blank'` with no `rel`. Confirmed in the built HTML: `<a … class="item" target="_blank">`. Browsers imply `noopener` these days so this is not a security hole, but it is inconsistent with `DocCard.astro:24`, which sets `rel="noopener"` explicitly for exactly this case.
- `docs/src/styles/custom.css:298` — `:is(p, li, td, dd, blockquote) > a` is a direct-child match, so a link wrapped in emphasis (`**[text](url)**` → `<p><strong><a>`) misses the custom underline treatment and silently falls back to Starlight's plain link style. No current page triggers it (I checked all 16), but docs get edited; dropping the `>` fixes it and `:not(:where(.not-content *))` still guards the components.
- `docs/src/components/PageTitle.astro:31` — hardcodes `id="_top"` where Starlight's own `PageTitle.astro` imports `PAGE_TITLE_ID` from `@astrojs/starlight/constants`. It is `'_top'` today, so this is correct now; importing the constant would keep the TOC "Overview" anchor and the skip target from breaking silently on an upstream change.
- `docs/src/styles/custom.css:52` — the comment says "h1 36px" but `--sl-text-h1: 2.125rem` is 34px.

## Checks performed

Repo invariants — all verified, none applicable: the diff touches `docs/` only (`git diff --name-only` confirms zero files outside it), so `edit_file`/`write_file` spec-gating, ERC/DRC verification-gating, the LLM-free/network-free `check` path, the read-only sexp parser, and the sync-obligations ledger are all untouched. Secrets: zero `sk-[A-Za-z0-9_-]+` matches in added lines; `.gitignore` unmodified. Spec coherence: no spec-level behavior changed, and `openspec validate build-copperhead-phase-1` reports valid in the worktree.

Also checked and clean, so a reader can tell these were examined rather than skipped:

- **Lockfile is in sync.** Both font packages are present in `docs/package-lock.json` with integrity hashes, and `npm ci` succeeded — `Deploy docs` will not fail on a drifted lockfile. The `libc` removals on sharp's optional deps are npm normalisation, as the description says.
- **All 21 `data-icon` / `Icon` names resolve** against Starlight 0.41.3's icon set. An unknown name would render a silent blank box (the `as StarlightIcon` cast at `SidebarSublist.astro:35` defeats typechecking, and `astro build` runs no `astro check`), but none are wrong: the built `index.html` has 17 `item-icon` SVGs and zero empty ones.
- **The theme toggle renders twice** (header + mobile drawer, confirmed: 2 `<ch-theme-toggle>` in the built HTML) and handles it correctly — `paintButtons` queries document-wide, so both stay in sync.
- **`''` for auto matches Starlight's convention.** Starlight's inline `ThemeProvider` reads `storedTheme || system`, so the empty string written for `auto` correctly falls through to the system preference. No flash, as claimed.
- **`--sl-font: 'Inter Variable'` needs no fallback stack** — Starlight composes `--__sl-font: var(--sl-font, var(--sl-font-system)), var(--sl-font-system)`.
- **Font payload is fine.** 17 woff2 files / 310 KB total on disk, but `unicode-range` means an English reader downloads only the latin subsets (~77 KB).
- **`--sl-main-pad` and `.sidebar-pane` overrides are valid.** Both are real Starlight props; `.sidebar-pane` is `position: fixed`, so the inset floating panel works, and the desktop overrides are inside `@media (min-width: 50rem)` so the mobile drawer is untouched.
- **No lead-paragraph duplication.** All 16 content pages render a `PageTitle` lead that differs from their first body paragraph.
- **`--sl-color-gray-7` undefined in dark mode is harmless** — Starlight only references it inside its light and print blocks, matching upstream.
- **Contrast passes AA.** `--sl-color-gray-3` is 5.3:1 on dark and 4.7:1 on light; `--sl-color-text-accent` is 8.0:1 and 5.7:1.
- **Version badge resolves.** `Sidebar.astro` builds `releases/tag/v0.10.0` from the root `package.json`; that release exists.

Review ran in an isolated `git worktree` at head `199d690`; the main checkout was never entered and stays on `feat/vertex-ai-provider`.

## Fix prompt

````text
(copperheadhq/copperhead): a TypeScript CLI agent that designs and edits real KiCad
projects. Node >= 20. The PR is #271, branch docs/gitbook-style-theme, based on main.
It is a docs-only change: it restyles the Astro Starlight site under docs/ after
GitBook's clean theme. It touches no files outside docs/.

Check it out first:
  gh pr checkout 271

FINDING 1 [medium] Printing any docs page renders the dark theme, because the
unlayered palette defeats Starlight's print stylesheet.
  Where: docs/src/styles/custom.css:14 (the :root palette) and :69
  Fails when: A reader whose OS is in dark mode (no stored preference, so Starlight's
    inline provider sets data-theme="dark") opens a page such as /reference/cli/ and
    hits Print or Save as PDF. Starlight's print.css is unlayered too
    (@media print { :root { --sl-color-black: #fff; ... } }, commented "force light
    mode when printing"), so it has the same (0,1,0) specificity as this file's :root
    block and @media print adds none. Document order decides, and custom.css is loaded
    second: index.html links /_astro/print.*.css with media="print" first, then
    /_astro/common.*.css, which carries the copperhead :root palette unlayered (at
    byte 5693, before the first @layer at 14329). So --sl-color-black resolves to
    #1b1b1c instead of #fff and --sl-color-white to #f7f8f8 instead of near-black, and
    the page prints as light text on a near-black background: ink-flooded, or invisible
    when the browser drops backgrounds. Light mode is unaffected, because
    :root[data-theme='light'] is (0,2,0) and outranks print.css already.
  Fix: Append a print block at the end of docs/src/styles/custom.css that outranks the
    dark :root in both modes, restoring the light values for print:

      /* Starlight's print.css is unlayered too, so this file would otherwise win
         and print the dark palette. Restore light values for print. */
      @media print {
        :root,
        :root[data-theme='dark'] {
          --sl-color-white: #1a1b1e;
          --sl-color-gray-1: #2b2d32;
          --sl-color-gray-2: #545962;
          --sl-color-gray-3: #6d747f;
          --sl-color-gray-4: #c2c8d2;
          --sl-color-gray-5: #e3e7ed;
          --sl-color-gray-6: #eff2f6;
          --sl-color-black: #ffffff;
          --sl-color-accent: #a35d1f;
          --sl-color-text-accent: #9a5619;
          --ch-hover-bg: transparent;
          --ch-pill-bg: transparent;
        }
      }

FINDING 2 [low] (optional) CI never builds the docs on a pull request.
  Where: .github/workflows/docs.yml:4-8
  Fails when: The "Deploy docs" workflow triggers only on push to main, and no other
    workflow builds docs/ for a pull request. A future docs PR that adds a component
    with a bad Starlight import goes green on every check, merges, then fails "Deploy
    docs" on main, leaving docs.copperhead.sh stale until someone notices. The gap
    predates this PR, but this PR is the first one large enough (1,423 added lines,
    all in docs/) for it to matter.
  Fix: Add a build-only job triggered on pull_request with paths: ['docs/**'], reusing
    the existing checkout / setup-node / npm ci / npm run build steps from the build
    job, and without the deploy job.

FINDING 3 [low] (optional) Footer visual order no longer matches keyboard focus order.
  Where: docs/src/styles/custom.css:379-393
  Fails when: Starlight's Footer.astro emits <div class="meta"> (Edit page, Last
    updated) before <Pagination />. These rules flip them visually with
    footer .meta { order: 2 } and .pagination-links { order: 1 }, but CSS order does
    not move DOM or tab order. Confirmed in the built markup for /reference/cli/. A
    keyboard user tabbing past the last content link lands on "Edit page", which is
    visually the bottom-most element on the page, then tabs backwards up the page into
    the Previous/Next cards. That is a WCAG 2.4.3 (Focus Order) mismatch.
  Fix: Drop both order declarations and restyle in place, letting the meta row sit
    above the pagination cards, or override Footer.astro as well so the DOM order
    matches the intended visual order.

FINDING 4 [low] (optional) The mobile "On this page" dropdown still lists the Overview
entry that the desktop rail hides.
  Where: docs/src/styles/custom.css:260
  Fails when: The rule is
    starlight-toc li:has(> a[href='#_top']) { display: none; }
    but the mobile table of contents is a separate custom element,
    <mobile-starlight-toc>, and it does not nest a <starlight-toc>, so the selector
    never matches it. Confirmed in the built markup for /reference/cli/: the mobile
    dropdown still contains <li><a href="#_top"><span>Overview</span></a></li> while
    the desktop rail hides it. On a phone the dropdown opens with a leading "Overview"
    row that no desktop reader ever sees, so the two navigations disagree about the
    page's headings.
  Fix: Widen the selector to cover both elements, for example
    :is(starlight-toc, mobile-starlight-toc) li:has(> a[href='#_top']) { display: none; }

FINDING 5 [low] (optional) The external sidebar link opens a new tab with no rel.
  Where: docs/astro.config.mjs:141
  Fails when: The "Technical spec" sidebar entry sets target: '_blank' with no rel.
    Confirmed in the built HTML: <a ... class="item" target="_blank">. Browsers imply
    noopener for target=_blank these days, so this is not a security hole, but it is
    inconsistent with docs/src/components/DocCard.astro:24, which sets rel="noopener"
    explicitly for exactly this case.
  Fix: Add 'rel': 'noopener' alongside target in that entry's attrs.

FINDING 6 [low] (optional) Links wrapped in emphasis miss the custom link styling.
  Where: docs/src/styles/custom.css:298
  Fails when: The selector is
    .sl-markdown-content :is(p, li, td, dd, blockquote) > a:not(:where(.not-content *))
    and the > is a direct-child match, so a link wrapped in emphasis, for example
    **[text](url)** rendering as <p><strong><a>, does not match and silently falls back
    to Starlight's plain link style. No page triggers it today (all 16 were checked),
    but docs get edited.
  Fix: Drop the > from both that rule and its :hover companion. The
    :not(:where(.not-content *)) guard still keeps the components unstyled.

FINDING 7 [low] (optional) The page title hardcodes an id that Starlight exports as a
constant.
  Where: docs/src/components/PageTitle.astro:31
  Fails when: The override writes <h1 id="_top"> where Starlight's own PageTitle.astro
    imports PAGE_TITLE_ID from @astrojs/starlight/constants. It is '_top' today, so the
    markup is correct now, but an upstream change to that constant would silently break
    the table-of-contents "Overview" anchor and the skip target with no build error.
  Fix: Import PAGE_TITLE_ID from @astrojs/starlight/constants and use it for the id.

FINDING 8 [low] (optional) A code comment states the wrong h1 size.
  Where: docs/src/styles/custom.css:52
  Fails when: The comment reads "Type scale (GitBook-ish: h1 36px, h2 24px, h3 20px)"
    but --sl-text-h1 is 2.125rem, which is 34px.
  Fix: Correct the comment to 34px, or change the value to 2.25rem if 36px was intended.

This PR adds no runtime code and no tests, and it changes no spec-level behavior, so no
regression test and no SPEC.md or OpenSpec artifact update is expected. Do not add one.

Do not break these repo invariants (they are hard requirements from SPEC.md). None of
them are in scope for this change, so the practical rule is: stay inside docs/ and
.github/workflows/docs.yml, and touch nothing under src/.
    by prompt text.

Then verify:
  npm --prefix docs ci
  npm --prefix docs run build
  npm run lint:md

The docs build must complete and emit 17 pages. Then confirm the print fix by hand:
open docs/dist in a browser, force dark mode, and check the print preview of
/reference/cli/ renders black text on white.

The root suite (npm run typecheck, npm run build, npm test) is unaffected by a
docs-only change, but run it if you touch anything outside docs/. Three tests
(gating-sync, init-check, stage-completion) fail on a machine whose local KiCad config
is not isolated via KICAD_CONFIG_HOME; that is pre-existing and unrelated.

````

Issue #258, opened 2026-08-30
Run protocol stream: version the CLI/consumer boundary with a handshake, an event stream, and a result document


animesh-chouhan:
## Summary

This stream owns the **wire between the CLI and any non-interactive consumer**: how a consumer discovers what the installed CLI can do, how a run reports what it is doing while it runs, and how it states what it did when it finished.

It owns the **transport, not the semantics**. Nothing here changes what a stage does, what a tool returns, when a run is retried, or where an artifact is stored. It fixes the form those facts take on the way out.

The other four streams each assume this wire exists. #251 needs an exit status that carries a retry class and progress events a lease can trust; #250 needs artifact records carrying the producing tool call; #249 wants its cost numbers emitted "as structured transcript events so hosted runs can aggregate them without scraping the CLI table"; #246 needs its tool-result envelope to survive the trip. Each assumes it, none owns it, and today a consumer reconstructs all of it by tailing a human audit trail and matching regexes against `--plain` output.

## Boundary rule

- **In scope:** capability discovery and protocol versioning, the machine event stream and its envelope, the result document, the form (not the policy) of exit classification, typed findings, and conformance fixtures.
- **Out of scope:** retry policy, leases, cancellation policy (#251); artifact storage, workspace lifecycle, payload limits (#250); caching and model routing (#249); the tool registry and the contents of its envelope (#246); anything that changes circuit-design behaviour.
- Structural invariants are untouched: edit tools stay absent from the tool list until an OpenSpec proposal validates, and `check` / `verify` stays LLM-free and network-free (AC-2.1).
- **No clause of this protocol may name a hosted platform.** Every field has to be meaningful to someone running `copperhead do --event-stream run.ndjson` in their own CI with no account. That rule is what keeps this an open-core capability rather than cloud plumbing living in the CLI, and it is testable: the conformance fixtures in Phase 4 are generated on a bench.

## Current state (baseline on `main`)

**1. There is no way to ask the CLI what it supports.** `doctor --json` reports host health (`DoctorReport { ok, checks }`, [doctor.ts:38](src/commands/doctor.ts#L38)) and is the right shape for a human, not for a negotiation. A consumer pins a version in a Dockerfile and hopes. The copperhead-cloud runner image pins `copperhead@0.9.0` while `0.10.0` is published, and nothing on either side detects the gap.

[...]

PR #273, opened 2026-09-06
feat(mcp): serve the gated pipeline to MCP hosts over stdio

### LINKED ISSUE #40: Add MCP server: expose the gated copperhead pipeline to MCP hosts

animesh-chouhan:
## What

Adds `copperhead mcp`, a stdio MCP server exposing the gated pipeline to any MCP host as five opaque, outcome-level tools: `copperhead_check`, `copperhead_do`, `copperhead_sync`, `copperhead_init`, and `copperhead_doctor`. Closes #40.

## Why

Hosts already point coding agents at KiCad repos through generic file tools, so a host agent can rewrite a `.kicad_sch` with no spec gate, no verification, and no rollback. copperhead's invariants only protected people whose entry point is the CLI — a shrinking share of how people work. This makes the gated pipeline the thing the host calls.

## Design

**Opacity is the security boundary.** The server exposes outcomes, not operations: no file-edit tool, no raw-KiCad tool, no partial-loop tool. Each tool dispatches into the existing entry points ([runCheck](src/commands/check.ts), [runAgentLoop](src/agent/loop.ts), [syncVerify / syncResolve](src/commands/sync.ts), [runInit](src/memory/scaffold.ts), [runDoctor](src/commands/doctor.ts)), so dirty-tree refusal, path sandbox, budgets and rollback are inherited by construction. `--allow-dirty` is deliberately not exposed: a tool input that switches a safety rail off is a rail a host can bypass.

**Experimental on purpose (D7).** A `0.` protocol major, announced on startup, no registry listing until the stabilization criteria in the proposal are met. A stable tool schema is a contract with hosts we cannot migrate, on a CLI that shipped four minor versions in six weeks. A test asserts the `0.`, so stabilizing is a deliberate edit rather than a drift. This is also why the change does **not** block on the run-protocol handshake (#258): an unstable surface makes no promise to reconcile against. Adopting `capabilities --json` is written down as the stabilization trigger instead.

**Shared envelope, separate catalog (D9).** Results are built with `seal()` and errors carry the shared `ToolErrorKind` values, so redaction is inherited rather than reimplemented. But the five tools are deliberately **absent** from the agent capability catalog: a `CatalogEntry` is gated on `RunContext` and therefore callable by the loop, so registering `copperhead_do` there would make the gated pipeline reachable from inside the pipeline — a recursion hazard and a hole straight through the spec gate. A test asserts they resolve to nothing in the registry.

**`doctor` as the fifth tool (D10).** LLM-free, network-free, mutates nothing, so it widens the surface without widening what the surface can *do*. It earns its place because "is this host configured correctly" is the largest predictable support cost of shipping an MCP server. It deliberately skips the `kicad-cli` preflight every other tool performs — gating a diagnostic on the thing it diagnoses would fail exactly when it is needed.

Behaviour worth calling out:

- A rollback is a successful tool call whose result says `rolled_back`, not a protocol error (D6).
- `copperhead_do` runs non-interactively with no confirm callback and no budget-extension prompt: there is no human on this transport, so the loop must fail rather than block.
- Progress streams as MCP progress notifications. All logs go to stderr; stdout carries only JSON-RPC.
- `copperhead_do` and `copperhead_sync --resolve` refuse with a typed `unavailable` naming the missing variable when no model resolves.
- Mutating tools are serialized per repo with a typed busy error; `check` and `doctor` run concurrently.

## Spec / OpenSpec

OpenSpec change `add-mcp-server`, re-specced before implementation: `openspec validate add-mcp-server` passes, 23/25 tasks checked off. The 2 open tasks are the explicitly deferred stabilization items (adopt the handshake; submit registry listings). Two incoherences in the prior spec were fixed rather than implemented around — it promised `fab`/`strict` inputs on `copperhead_check` that `check` cannot accept, and it assumed no tool registry existed when #246 Phase 0 has since landed.

## Testing

| Check | Command | Status |
| --- | --- | --- |
| Markdown lint | `npm run lint:md` | pass (186 files) |
| Full suite | `npm test` | **918 passed, 0 failed, 21 skipped** |
| Spec | `openspec validate add-mcp-server` | valid |

24 new tests across three files: tool surface and opacity, a surface audit that fails if a tool starts accepting a filesystem path, catalog absence, redaction of a key reaching a result, `check`/`doctor` parity with the direct entry points, keyless degradation, per-repo serialization, run-status mapping, a **real-stdio handshake** proving stdout carries only JSON-RPC, and a **broken-`kicad-cli` case** proving `doctor` answers where `check` correctly refuses with `unavailable`.

[test/entry-point-parity.test.ts](test/entry-point-parity.test.ts) pins `check`/`sync`/`init` entry-point results byte-identical to their `--json` CLI output, so the server and the CLI cannot silently drift.

> Note for reviewers: three suites (`init-check`, `stage-completion`, `gating-sync`) fail on a dev box whose user-level KiCad `sym-lib-table` exists, which CI does not have. Runs above used an isolated `KICAD_CONFIG_HOME`.

## Docs

[README.md](README.md) gets an MCP section; [integrations/README.md](integrations/README.md) covers host configuration (Claude Code, generic stdio), the Codex gap, and the stabilization criteria. The companion skill is [integrations/claude-code/copperhead/SKILL.md](integrations/claude-code/copperhead/SKILL.md) — note it is prompt-level guidance and explicitly *not* a substitute for the server's structural opacity.

## Dependency

`@modelcontextprotocol/sdk@^1.30.0`. It was already vendored as a peer dependency of `@anthropic-ai/claude-agent-sdk`, so the lockfile change is one line.

## Invariant checklist

- [x] **Spec-gated in**: the MCP surface exposes no file-edit tool, so the agent tool list and its spec gate are untouched and cannot be reached around; a surface-audit test enforces it.
- [x] **Verification-gated out**: `copperhead_do` runs the unchanged loop; a failure returns `rolled_back` after repair and restores the git snapshot.
- [x] **`check` / `verify` stays LLM-free and network-free**: `copperhead_check` dispatches `runCheck` with no provider and no key; a test runs it with no credential.
- [ ] **No sexp serialization**: n/a (the KiCad parser is untouched).
- [ ] **Sync-obligations ledger**: n/a (the loop and commit hooks are untouched).
- [x] **Secrets**: keys are read only from env vars, are never a tool input, and every result goes through `seal()`; tests assert a key in a result and in `doctor`'s report comes back redacted.
- [x] **Spec coherence**: the delta specs, design decisions and `tasks.md` moved with the code; `openspec validate` passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ copperhead doctor                    # env preflight: node, kicad-cli, git, open
 copperhead sync [--dry-run]          # verify the whole design state, resolve drift
 copperhead create --brief brief.md   # brief → full output package
 copperhead export bom --supplier jlcpcb   # supplier-ready ordering file from docs/BOM.md
+copperhead mcp                       # EXPERIMENTAL: serve the gated pipeline to MCP hosts over stdio
 ```
 
 Global flags: `--repo <path>` (default: cwd) and `--json` for machine-readable output. `--model` is available on `do`, `sync`, `create`, `skill run`, and `doctor`; `--interactive` only on `do` and `create`; `do` also takes `--dry-run`, `--max-turns`, and `--allow-dirty`.
@@ -173,6 +174,39 @@ copperhead export bom --supplier mouser --spares 15     # Mouser cart CSV, 15% s
 
 Nothing is a black box: decisions land in an append-only `docs/DECISIONS.md`, every run writes a human-readable summary next to its transcript, and a per-run `docs/CHANGELOG.md` narrates the design history.
 
+## MCP server (experimental)
+
+Coding agents are already pointed at KiCad repos through generic file tools, which
+means a host agent can rewrite a `.kicad_sch` with no spec gate, no verification,
+and no rollback. `copperhead mcp` closes that: it serves the gated pipeline to any
+MCP host as five opaque, outcome-level tools — `copperhead_check`, `copperhead_do`,
+`copperhead_sync`, `copperhead_init`, `copperhead_doctor` — and nothing finer. There is no file-edit
+tool, no raw KiCad tool, and no partial-loop tool to reach around, so a host agent
+cannot skip spec-gating or verification by any sequence of calls.
+
+```jsonc
+// .mcp.json
+{
+  "mcpServers": {
+    "copperhead": {
+      "command": "copperhead",
+      "args": ["mcp", "--repo", "/absolute/path/to/your/kicad/project"]
+    }
+  }
+}
+```
+
+> **The surface is experimental and unstable.** It declares a `0.` protocol major:
+> tool names, inputs, and result shapes may change in any release, and there is no
+> registry listing until the stabilization criteria are met. Host configuration for
+> Claude Code and generic stdio hosts, the companion skill, the Codex gap, and those
+> criteria are all in [`integrations/`](integrations/README.md).
+
+`copperhead_check`, `copperhead_init` and `copperhead_doctor` need no credential; `copperhead_do` and
+`copperhead_sync` with `resolve: true` refuse with a typed error naming the missing
+environment variable. The server opens no network transport of its own — stdio only,
+and LLM calls happen exactly where the CLI already makes them.
+
 ## What it is not
 
 - **Not an autorouter.** Routing stays human or delegated; copperhead produces the DRC-clean draft that layout tools optimize from.
@@ -218,6 +252,7 @@ Contributions are welcome; see [CONTRIBUTING.md](CONTRIBUTING.md) for setup and 
 
 - [`AGENTS.md`](AGENTS.md): repository instructions loaded automatically by Codex and compatible coding agents; [`CLAUDE.md`](CLAUDE.md) provides the corresponding Claude Code guidance
 - [`src/`](src/): CLI ([`cli.ts`](src/cli.ts), [`commands/`](src/commands/)), the provider-agnostic agent loop ([`agent/`](src/agent/)), the `kicad-cli` wrapper and s-expression reader ([`kicad/`](src/kicad/)), and doc/constraint memory ([`memory/`](src/memory/))
+- [`integrations/`](integrations/README.md): MCP host configuration and the companion Claude Code skill
 - [`test/`](test/): offline suite plus [`fixtures/`](test/fixtures/), a tiny known-good KiCad project
 - [`openspec/specs/SPEC.md`](openspec/specs/SPEC.md): the full technical specification, including binary acceptance criteria
 - [`openspec/changes/build-copperhead-phase-1/`](openspec/changes/build-copperhead-phase-1/): the implementation plan with proposal, design decisions, capability specs, and task checklist
--- a/integrations/README.md
+++ b/integrations/README.md
@@ -15,7 +15,7 @@ Exactly five tools, and nothing finer:
 | Tool | LLM | Mutates | Purpose |
 | --- | --- | --- | --- |
 | `copperhead_check` | no | no | ERC, DRC, doc drift, spec validation |
-| `copperhead_init` | no | yes | scaffold docs memory from an existing schematic |
+| `copperhead_init` | no | yes | scaffold docs memory, and install a `pre-commit` hook running `copperhead check` |
 | `copperhead_do` | yes | yes | the full gated change pipeline |
 | `copperhead_sync` | only with `resolve: true` | only with `resolve: true` | design-state consistency, and optionally fix drift |
 | `copperhead_doctor` | no | no | probe node, kicad-cli, git, openspec and the model credential |
--- a/integrations/README.md
+++ b/integrations/README.md
@@ -1,0 +1,88 @@
+# Integrations
+
+Host-side configuration for the `copperhead mcp` server, and the companion skill
+that tells a host agent to use it.
+
+> **Experimental.** The MCP surface declares a `0.` protocol major. Tool names,
+> input schemas, and result shapes may change in any release. Pin a copperhead
+> version if you depend on the shapes, and read the stabilization criteria below
+> before building anything durable on it.
+
+## What the server exposes
+
+Exactly five tools, and nothing finer:
+
+| Tool | LLM | Mutates | Purpose |
+| --- | --- | --- | --- |
+| `copperhead_check` | no | no | ERC, DRC, doc drift, spec validation |
+| `copperhead_init` | no | yes | scaffold docs memory from an existing schematic |
+| `copperhead_do` | yes | yes | the full gated change pipeline |
+| `copperhead_sync` | only with `resolve: true` | only with `resolve: true` | design-state consistency, and optionally fix drift |
+| `copperhead_doctor` | no | no | probe node, kicad-cli, git, openspec and the model credential |
+
+There is deliberately no file-edit tool, no raw KiCad tool, and no way to drive a
+single step of the loop. That is the point: a host agent integrating copperhead
+cannot bypass spec-gating or verification-gating by any sequence of calls.
+
+`copperhead_check`, `copperhead_init` and `copperhead_doctor` need no credential. `copperhead_doctor`
+is also the one tool that does not require `kicad-cli` to be present — it reports a missing one as a
+failed check, so a host can diagnose its own setup rather than meeting an opaque error elsewhere. `copperhead_do` and
+`copperhead_sync --resolve` return a typed `unavailable` error naming the missing
+variable when no model can be resolved.
+
+## Claude Code
+
+```jsonc
+// .mcp.json in the project, or the equivalent under `claude mcp add`
+{
+  "mcpServers": {
+    "copperhead": {
+      "command": "copperhead",
+      "args": ["mcp", "--repo", "/absolute/path/to/your/kicad/project"]
+    }
+  }
+}
+```
+
+Then install the companion skill so the agent prefers these tools over its own
+file tools:
+
+```bash
+cp -r integrations/claude-code/copperhead ~/.claude/skills/
+```
+
+## Any stdio MCP host
+
+The server speaks MCP over stdio and takes no network transport:
+
+```bash
+copperhead mcp --repo /absolute/path/to/your/kicad/project
+```
+
+stdout carries JSON-RPC alone; every human-readable line goes to stderr. A host
+that merges the two streams will corrupt the protocol.
+
+## Codex
+
+Codex reads [`AGENTS.md`](../AGENTS.md) rather than Claude Code's skill format, so
+there is no drop-in equivalent of the skill above. Until that gap closes, paste the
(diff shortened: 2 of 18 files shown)


animesh-chouhan:
🤖 **Automated pr-review pass** (Claude Code `/pr-review`) — a comment, not a human review and not a merge gate.

**Disclosure: the reviewer authored every commit under review.** Steps 3 and 4 were therefore run as independent adversarial subagents rather than inline, and the findings below were each reproduced before being reported. The review ran in an isolated worktree at `b6af08e`; the author's checkout was never entered.

## Verdict: request changes → all blocking findings fixed in `68ca8af` / `91a7fa6`

## Metrics

- **lines**: +1362 / -44 across 16 files · src +562/-3, test +474/-0, docs +192/-0, spec +132/-41
- **new vs net**: 6 new files (1 in src); 538 net-new src lines
- **tests**: 918 pass / 21 skip / 0 fail at review time → **924 pass / 0 fail** after fixes
- **diff coverage**: `src/mcp/server.ts` 71.6% lines, 55.2% branch, 85.7% funcs (v8, installed into the review worktree)
- **deps**: +`@modelcontextprotocol/sdk`, +`zod` (added during review)
- **ci**: 5 pass / 0 fail / 0 pending; no required checks reported
- **method**: measured in the review worktree against merge base `26d9b51`; suite via `vitest run`; coverage intersected with changed src lines. Suite runs used an isolated `KICAD_CONFIG_HOME`, because a user-level KiCad `sym-lib-table` fails three suites that CI never sees.

## Findings

### 1. HIGH — `copperhead_init`'s `path` input escaped the repo sandbox (AC-4.2) — FIXED

`src/memory/scaffold.ts:175` resolved the host-supplied search path with bare `path.resolve` instead of `resolveInRepo` (`src/util/paths.ts:15`), whose docstring reads "All file tools must go through this."

Reproduced: `copperhead_init {path: '../foreign'}` returned `ok: true`, scaffolded this repo's `docs/` from a schematic outside it, and durably wrote `config.schematic = "../foreign/hardware/open-key.kicad_sch"`. Every later `check` / `do` then operates on that out-of-repo file. The blast radius is larger than one read: `src/kicad/draft/draft.ts:109` *writes* to that path, and rollback (`git reset --hard` + `clean -fd`, `src/util/git.ts:270`) only covers `repoRoot` — so a clobbered out-of-repo schematic had **no rollback**.

`scaffold.ts:175` predates this PR, but until now the input was reachable only from a human typing `copperhead init --path`. Exposing it over MCP handed it to a host agent. The PR's own surface-audit test whitelisted this exact input on a false premise ("consumed by the scaffolder, which sandboxes to the repo root"), so the one test guarding "no filesystem path into the pipeline" had a hole justified by an incorrect comment.

Fixed in `runInit`, which hardens the CLI path too; `SandboxError` maps to a typed `validation` error; regression tests for `../..` and absolute paths; new delta-spec requirement; whitelist comment corrected.

### 2. MEDIUM — `zod` imported but undeclared — FIXED

`src/mcp/server.ts` imports `zod` directly while `package.json` did not declare it; it resolved only transitively through the MCP SDK's own dependency.

*A reported HIGH failure scenario here was refuted:* `npm ci --omit=optional` does **not** prune it — npm re-derives reachability from `package.json`. Tested in a scratch tree; both packages install. Downgraded accordingly. Fixed by declaring `zod` and regenerating the lockfile, which also cleared stale `optional`/`peer` flags left by an earlier hand-splice.

### 3. MEDIUM — MCP SDK loaded eagerly on every command — FIXED

`src/cli.ts` imported `./mcp/server.js` at module top level, adding ~150ms of SDK + zod load to *every* invocation including the pre-commit-hook `check`, and making any resolution failure fatal to the whole CLI rather than to `copperhead mcp` alone. Now lazily imported, matching the existing `skill` pattern.

### 4. MEDIUM — the concurrency test asserted nothing — FIXED

`test/mcp-server.test.ts` had its only `expect` inside `if (!env.ok && …)`, so removing the lock entirely left it green. Now asserts exactly one busy refusal and one success, unconditionally.

### 5. MEDIUM — `mcpRenderer` had zero coverage — FIXED

The whole progress-notification adapter (the thing that keeps long runs from timing out in a host) was executed by no test, despite needing neither a model nor a network. Now unit-tested for the monotonic counter, message content, and the no-emit paths.

### 6. MEDIUM — task boxes checked for tests that do not exist — FIXED (by unchecking)

`tasks.md` claimed golden request/response fixtures, a stdio progress-notification test, and a rollback-as-result test. None existed. Now unchecked as `3.1a` / `3.3a` with the reason: driving `copperhead_do` in a test needs a provider seam on `createMcpServer`, which this PR does not add.

### 7. MEDIUM — cancellation is ignored — DEFERRED

Neither mutating handler reads `extra.signal`. The SDK aborts in-flight handlers on transport close and on `notifications/cancelled`, so a user who cancels a long `copperhead_do` or closes their editor gets the run finishing and **committing** anyway. Deferred rather than fixed: threading abort into `runAgentLoop` is a change to the loop's contract, not to this adapter.

### 8. LOW — fixed in passing

- `copperhead_init` installs a git `pre-commit` hook with no host opt-out; the tool description said only "scaffold docs". Now disclosed in the description, `integrations/README.md`, and `SKILL.md`.
- The module docstring claimed the lock stops two hosts interleaving; it is in-memory and per **server process**. Corrected.
- `sync --resolve` computed its report before taking the lock, so it could ask the loop to fix drift another run had already resolved. Now re-verifies once the lock is held.
- The unused `onProgress` "test seam" was dead code that would have silenced real progress notifications the moment anyone set it. Removed.
- `test/mcp-stdio.test.ts` leaked a temp dir per run.
- **Spec coherence**: the PR added a CLI command and a new entry point without touching `openspec/specs/SPEC.md`. §3 now lists `copperhead mcp`; §7 states that every safety rail applies identically to the MCP entry point.

## Invariant checks — all six performed

- **Spec-gated in — PASS.** `runAgentLoop` hardcodes `editsUnlocked: false`; `RunOptions` exposes no field that can preset it, so the MCP path cannot. Gating remains by omission in `registry.list(ctx)`, not by prompt text.
- **Verification-gated out — PASS.** `copperhead_do` passes only `repoRoot, request, model, allowDirty:false, dryRun, interactive:false, renderer, meta` — no `maxTurns`, `confirm`, `onBudgetExhausted`, `provider`, or `stagePrompt`. The `finish` gate still requires ERC-since-schematic-edit and DRC-since-board-edit; `repairCycles > maxRepairCycles` still fails to rollback.
- **`check`/`verify` LLM-free and network-free — PASS.** The transitive import closure of `src/commands/check.ts` is 12 files; package imports are `execa` and `node:*` only. ERC/DRC reports are written to a temp dir, never into the repo.
- **No sexp serialization — PASS.** The diff touches nothing under `src/kicad/`; `sexp.ts` still exports no serializer.
- **Sync-obligations ledger — PASS.** Unchanged; the MCP path routes through the same loop and inherits it.
- **Secrets — PASS.** All 23 return statements in `src/mcp/server.ts` were enumerated: every one passes through `seal()`, `failure()`, or `verdict()`. `.gitignore` still carries `.env` and `.copperhead/runs/`. A test asserts a key in a result — and in `doctor`'s report — comes back redacted.

Also refuted and not reported: stdout corruption on the `do`/`sync` paths (the only two `console.log` sites are short-circuited when a `renderer` is supplied), progress notifications leaking unsealed text, auto-approval of the spec gate, a lock leak between `tryAcquire` and `try`, and any new network surface.

## Fix prompt

All findings except #7 are already applied in `68ca8af` and `91a7fa6`. The one open item, for a follow-up:

> Thread MCP cancellation into the agent loop. `src/mcp/server.ts`'s `copperhead_do` and `copperhead_sync` handlers ignore `extra.signal`, which the MCP SDK aborts on transport close and on `notifications/cancelled`. Today a user who cancels a long run, or whose editor disconnects, gets the run finishing and committing to their repo anyway, and the per-repo lock stays held for the duration. Add an abort path to `runAgentLoop` (`src/agent/loop.ts`) that triggers its existing rollback, and have both handlers forward `extra.signal`; return a `rolled_back` result when aborted. Repo invariants: mutation tools stay structurally absent until an OpenSpec proposal validates; `check`/`verify` stays LLM-free and network-free; no sexp round-tripping; secrets only in env vars and redacted at write time. Verify with `npm run typecheck && npm test && npm run build`, and run the suite with an isolated `KICAD_CONFIG_HOME` containing only `10.0/fp-lib-table`.

PR #279, opened 2026-09-07
feat(kicad): add static circuit pattern library (voltage-regulator, u…

DevSars24:
## What
Adds a static, read-only pattern library of three known-good circuit blocks (AMS1117 linear voltage regulator, USB-C power input sink, crystal oscillator) as JSON files under `src/kicad/patterns/`, along with a JSON schema, a standalone validation script, and `docs/patterns.md`.

## Why
During the schematic step, the AI currently has to re-derive every common sub-circuit (regulator, USB-C input, crystal osc) from scratch, pin by pin. This leads to avoidable ERC failures and retries (wrong pull-up rail, missing decoupling cap, miswired USB signal pins), burning both time and compute. There is currently no reusable, named source of known-good multi-part circuit blocks that the AI or a human reviewer could reference. Closes #274.

## Design
This is intentionally a small, purely additive first step, not the Phase 4 "template corpus (tscircuit-style idiom matching)" mentioned in `ROADMAP.md`.

- Pattern files use the same shape already accepted by `SchematicIntent` (`IntentPart` / `IntentNet` from [ir.ts](src/kicad/draft/ir.ts)), so no new concepts are introduced, just static data expressed in an existing shape.
- `src/kicad/patterns/pattern.schema.json` formally defines that shape (`name`, `description`, `parts[]`, `nets[]`) so pattern files can be validated independently of the drafting pipeline.
- `scripts/validate-patterns.ts` loads every pattern file, validates it against the schema, and checks referential integrity (every `REF.pin` in `nets[].pins` must resolve to a `ref` declared in that same file's `parts[]`). It is wired in via `npm run validate:patterns` and does not touch any existing script.
- Pattern values (part refs, decoupling caps, pull-down resistor values, load cap values) were based on the existing multi-part sub-circuits already visible in `manual-tests/real-designs/examples` (the regulator stage in `cm5_minima`, the crystal block in `ecc83`), not invented from scratch.
- Nothing in [ir.ts](src/kicad/draft/ir.ts), [engine.ts](src/kicad/draft/engine.ts), part-selection, or any prompt/validator is modified. The patterns are not consumed anywhere yet; consuming them from part-selection or the schematic drafting step is explicitly left as a separate, future proposal, per the discussion in #274.

### Invariants
- **Spec-gated edits**: no effect. Only new static data files and docs are added; no new write tools, commands, or agent capabilities are introduced.
[...]

Issue #278, opened 2026-09-07
Make the layout engine pattern-aware: keep pattern-instance parts visually grouped


DevSars24:
### The problem

Once a schematic.intent.json can carry pattern provenance via patternRefs (see 
#<issue-number-of-intent-expansion-issue>), the drafting engine still has no idea 
that a given set of parts belongs to one coherent circuit block. Today it only 
groups parts by the free-text `group` field on IntentPart (SUBSYSTEMS.md heading), 
via the existing group-box logic in engine.ts (L616-L636, L1193-L1200).

This means a pattern instance — e.g. a voltage-regulator-ams1117 pattern's U1/C1/C2 
— gets placed exactly like any other loose collection of same-subsystem parts. It 
can end up:
- Split across the width of a large subsystem group instead of drawn as the tight, 
  textbook-standard footprint a regulator stage normally has (IC with input cap on 
  one side, output cap on the other).
- Ordered by whatever the AI happened to list the parts in, rather than in the 
  pattern's own canonical internal layout.
- Visually indistinguishable from a group of unrelated parts that merely share a 
  subsystem label, even though the pattern file already encodes a known-good, 
  human-reviewed spatial relationship between its parts.

The four existing geometric idioms (decoupling rows, crystal flanking, drop chains, 
subsystem group boxing) are all narrow and part-type-driven — none of them currently 
key off "these N parts are one pattern instance."

### Proposed behavior

Extend the existing subsystem-group-boxing pass in engine.ts to additionally 
recognize patternRefs entries as a finer-grained sub-grouping *inside* their 
subsystem box, and apply each pattern's own internal layout hint rather than 
generic group placement.

Concretely:

1. Each pattern file (src/kicad/patterns/*.json) gains an optional `layoutHint` 
   block describing the pattern's canonical internal arrangement, e.g.:

   "layoutHint": {
     "anchor": "U1",
     "relative": [
       { "ref": "C1", "side": "left",  "offset": 1 },
       { "ref": "C2", "side": "right", "offset": 1 }
     ]
   }

2. In engine.ts, when a schematic.intent.json includes patternRefs, the layout pass 
   (right before the existing group-box logic runs) reads each pattern's layoutHint 
   and places that instance's parts relative to its anchor FIRST, using the same 
   grid-alignment primitives the engine already uses for decoupling rows / drop 
   chains — no new placement math, just reusing existing helpers with a different 
   input source.

[...]

Issue #277, opened 2026-09-07
Add optional pattern-expansion support to SchematicIntent (ir.ts + engine.ts)


DevSars24:
### The problem

Even after part-selection can reference a named pattern (see #<issue-number-of-
part-selection-issue>), by the time BOM rows reach Stage 4, they get flattened back 
into individual, unrelated IntentPart/IntentNet entries in schematic.intent.json. 
The drafting engine (src/kicad/draft/engine.ts) never learns that a given AMS1117 + 
2 caps, or a given USB-C connector + CC pull-downs, actually came from one known-good 
pattern.

This means:
- The AI still has to manually re-derive every REF.PIN net endpoint for a pattern's 
  internal wiring at the schematic stage, even though that wiring is already fully 
  known from src/kicad/patterns/*.json — reintroducing exactly the wiring-mistake 
  risk that patterns are meant to remove.
- validateIntent() in ir.ts (L157) has no way to cross-check a set of parts+nets 
  against the pattern they supposedly came from, so a pattern's internal wiring 
  invariants (e.g. "output cap must sit between VOUT and GND") aren't reusable at 
  validation time.
- Downstream, the layout engine (engine.ts L845-L1250) has no signal that a group 
  of parts is "one pattern instance" versus an arbitrary handful of unrelated parts 
  that merely share a subsystem group.

This is explicitly the highest-risk part of the overall pattern-library effort, 
because SchematicIntent is a validated, load-bearing schema consumed by the 
drafting engine, sexp emission, and existing tests (test/draft-idioms.test.ts, 
test/draft-engine.test.ts, etc.).

### Proposed behavior

Add pattern expansion as a strictly additive, backward-compatible pre-processing 
step — not a change to the core schema's required shape.

1. ir.ts: add ONE new optional field to SchematicIntent, not to IntentPart/IntentNet:

   SchematicIntent {
     version: number,
     parts: IntentPart[],
     nets: IntentNet[],
     noConnect?: string[],
     hints?: {...},
     patternRefs?: { patternName: string, instanceId: string, refPrefix: string }[]
   }

   patternRefs is purely informational metadata: "these refs came from this pattern, 
   with this instance id." It does NOT replace parts/nets — every part and net must 
   still be fully present and valid exactly as today. validateIntent() continues to 
   require fully-resolved REF.PIN endpoints; it additionally (optionally) cross-checks 
   that if patternRefs is present, the referenced parts/nets structurally match the 
[...]

Issue #276, opened 2026-09-07
Allow part-selection to reference a named pattern instead of enumerating every part


DevSars24:
### The problem

Today, Stage 3 (part-selection) forces the AI to enumerate every single component 
as an individual BOM.md row (| Refdes | Value | Footprint | MPN | Rationale |), even 
for extremely common, well-understood sub-circuits like a linear voltage regulator 
stage or a USB-C power input. This means:

- The AI re-derives the same handful of parts (regulator + 2 caps, or a USB-C 
  connector + CC pull-downs + bulk cap) from first principles on every single design, 
  with no memory of "the way this is always done."
- Small BOM mistakes (missing a decoupling cap, wrong cap value, wrong connector 
  pinout) surface only later, at the schematic/ERC stage, causing avoidable retries.
- There's no way for a human reviewing docs/BOM.md to tell "this is a standard 
  regulator block" at a glance — every design reads as a flat, undifferentiated 
  parts list.

This builds directly on the static pattern library proposed in #<issue-number-of-
pattern-library-issue> (src/kicad/patterns/*.json). Without named patterns to 
reference, this proposal has nothing to point at — that issue should land first.

### Proposed behavior

Extend Stage 3 (src/commands/create.ts L167-L189) so the AI-facing prompt can 
reference a pattern by name from src/kicad/patterns/, and have that expand into 
its constituent BOM rows automatically, tagged so the origin is traceable.

Example BOM.md entry, before:
| U1 | AMS1117-3.3 | SOT-223 | AMS1117-3.3   | Linear regulator |
| C1 | 10uF        | 0603    | GRM188...     | Input decoupling |
| C2 | 10uF        | 0603    | GRM188...     | Output decoupling |

Example BOM.md entry, after (pattern-referenced):
| Pattern: voltage-regulator-ams1117 | source: src/kicad/patterns/voltage-regulator-ams1117.json |
| U1 | AMS1117-3.3 | SOT-223 | AMS1117-3.3 | from pattern |
| C1 | 10uF        | 0603    | GRM188...   | from pattern |
| C2 | 10uF        | 0603    | GRM188...   | from pattern |

CLI-wise:
$ copperhead create --stage part-selection
...
Resolved pattern "voltage-regulator-ams1117" -> 3 parts added to BOM.md
Resolved pattern "usb-c-power-input" -> 5 parts added to BOM.md

Concretely this needs:
1. A small lookup/expansion helper that reads src/kicad/patterns/*.json and turns 
[...]

PR #280, opened 2026-09-07
Feature/pattern expansion ir

DevSars24:
> **Draft note:** Issue #277 recommends this land only after #276 and #278 are in. Filing as a draft to keep the work visible and unblock review discussion, but not requesting merge until sequencing is confirmed with maintainers. This branch is currently based on top of `feature/circuit-pattern-library` (#279), which is still pending merge into main. The diff below will include #279's commits until it merges; that is expected and will clean up automatically once #279 lands and this branch is rebased onto main.

## What
Adds [patternExpand.ts](src/kicad/draft/patternExpand.ts), a pure helper function that loads a named circuit pattern from `src/kicad/patterns/*.json` and expands it into a `{ parts: IntentPart[], nets: IntentNet[] }` block with renumbered reference designators and rewritten net endpoint pins, for manual merging into `schematic.intent.json`.

## Why
Even after part-selection can reference a named pattern (#276), the drafting engine never learns that a set of parts came from a known-good pattern, so the AI still has to manually re-derive every `REF.PIN` net endpoint for a pattern's internal wiring at the schematic stage, even though that wiring is already fully known from `src/kicad/patterns/*.json`. This reintroduces exactly the wiring-mistake risk that patterns are meant to remove. Refs #277.

## Design
This implements only the "expansion-only" variant discussed in #277, which the issue itself flags as the safer MVP compared to the full metadata-tracking version.

- **Pure function, no file I/O beyond loading the pattern**: `expandPatternRef(patternName, options)` reads the requested pattern definition and returns `{ parts, nets }`. It does not read or write `schematic.intent.json`, and the caller is responsible for merging the result into their own intent.
- **No schema changes**: does not add a `patternRefs?` field to `SchematicIntent` in [ir.ts](src/kicad/draft/ir.ts), and does not change `validateIntent()`'s required shape or logic in any way.
- **No engine changes**: [engine.ts](src/kicad/draft/engine.ts) is untouched. Layout-level pattern-awareness (grouping expanded instances visually) is explicitly deferred to #278.
- **No pipeline wiring**: [create.ts](src/commands/create.ts) is untouched. Automatically calling this helper during Stage 4 is out of scope for this PR.
- **Ref and pin renumbering**: remaps each component reference (e.g. `U1` -> `REG1_U1`) using the given ref prefix and instance id, and rewrites every corresponding net pin connection (e.g. `U1.3` -> `REG1_U1.3`) so the expanded output is structurally indistinguishable from hand-written parts and nets.
- **Out of scope, deferred to future issues**: the `patternRefs` metadata field on `SchematicIntent`, structural cross-checking of pattern instances in `validateIntent()`, and layout-engine grouping hints. These are intentionally left out per the descope decision in #277, and should only be attempted after this expansion-only helper has been used in practice.

## Spec / OpenSpec
This adds a new OpenSpec change: `add-pattern-expansion-helper` (see [proposal.md](openspec/changes/add-pattern-expansion-helper/proposal.md), [design.md](openspec/changes/add-pattern-expansion-helper/design.md), [tasks.md](openspec/changes/add-pattern-expansion-helper/tasks.md), and the delta spec at `specs/pattern-expansion/spec.md`). The proposal documents the problem, the descoped expansion-only solution, and explicitly lists what is out of scope (the `patternRefs` field, `validateIntent()` cross-checks, and engine.ts layout hints) so a future issue can pick those up separately. `openspec validate add-pattern-expansion-helper` passes.

### Automated test status
| Check | Command | Status |
| --- | --- | --- |
| Build | `npm run build` | ✅ Pass |
| Unit + offline | `npm test` | ✅ Pass (no existing tests affected) |
| Pattern expansion (new) | `npx vitest run test/pattern-expand.test.ts` | ✅ Pass (7/7 tests) |
| Pattern validation | `npm run validate:patterns` | ✅ Pass (3/3 patterns) |
| OpenSpec validation (new) | `openspec validate add-pattern-expansion-helper` | ✅ Pass |
| Live-LLM (opt-in) | n/a | n/a — no LLM-facing code changed |

New tests: [pattern-expand.test.ts](test/pattern-expand.test.ts) covers expansion of all three existing patterns (`voltage-regulator-ams1117`, `usb-c-power-input`, `crystal-oscillator`), ref-prefix and instance-id renumbering, verification that expanded output passes the existing `validateIntent()` unchanged when merged into a minimal `SchematicIntent`, and error handling for unknown pattern names.

Known pre-existing failures unrelated to this PR: none observed.

- Sandbox variant used: n/a
- Reason: this PR does not change CLI behavior, the agent loop, providers, or runtime KiCad generation. `expandPatternRef` is a standalone pure function not yet wired into `create`/`edit`/`check`, so there is no CLI-level flow to exercise manually. Verified instead via the automated suite above:
```text
$ npx vitest run test/pattern-expand.test.ts
✓ expands voltage-regulator-ams1117 with given ref prefix
✓ expands usb-c-power-input with given ref prefix
✓ expands crystal-oscillator with given ref prefix
✓ renumbers refs and rewrites net pins consistently across multiple instances
✓ expanded output merges into SchematicIntent and passes validateIntent() unchanged
✓ throws a descriptive error for an unknown pattern name
✓ throws on ref-prefix / part-count collisions
7 passed (7)

$ openspec validate add-pattern-expansion-helper
Change 'add-pattern-expansion-helper' is valid

$ git diff main...HEAD -- src/kicad/draft/ir.ts src/kicad/draft/engine.ts src/commands/create.ts
(no output — 0 changed lines in all three guardrail files)
```

## Docs
No user-facing docs changes in this PR. The OpenSpec proposal itself documents the helper's interface and scope; a `docs/` update for consuming this helper is deferred until it's wired into the actual CLI pipeline.

---

## Invariant checklist
- [x] **Spec-gated in**: N/A for this PR directly (no new write tool exposed to the agent yet), but the OpenSpec proposal (`add-pattern-expansion-helper`) is included ahead of any future write-tool exposure, per the spec-gated-edits rule.
- [x] **Verification-gated out**: N/A — this PR introduces no new mutation path. Any future caller that merges expanded output into `schematic.intent.json` still goes through the existing, unmodified `validateIntent()` and ERC/DRC gates exactly as hand-written parts/nets do today.
- [x] **`check`/`verify` stays LLM-free and network-free**: N/A — no changes under `src/commands/check`.
- [x] **No sexp serialization**: N/A — no KiCad file parsing/writing touched; `src/kicad/` parser untouched.
- [x] **Sync-obligations ledger**: N/A — no post-tool-call hooks or ledger-relevant behavior touched.
- [x] **Secrets**: N/A — no secrets, env vars, or transcript/log code touched.
- [x] **Spec coherence**: `openspec/changes/add-pattern-expansion-helper/` (proposal.md, design.md, tasks.md, delta spec) added together in this PR, and `openspec validate add-pattern-expansion-helper` passes. No changes to `SPEC.md` itself since this is a purely additive helper, not a change to existing spec-level behavior.

### DIFF
--- a/docs/patterns.md
+++ b/docs/patterns.md
@@ -1,0 +1,53 @@
+# Circuit Pattern Library
+
+This directory contains static, read-only reference definitions of standard, proven circuit blocks under `src/kicad/patterns/`.
+
+> [!IMPORTANT]
+> **Read-Only Reference Corpus**
+>
+> These are read-only reference data. Consuming these patterns from part-selection or the schematic drafting step is a separate, future proposal (see [issue #274](https://github.com/copperheadhq/copperhead/issues/274)) and is not implemented here.
+
+## Overview
+
+Each pattern file defines a self-contained circuit block using the declarative component (`parts`) and netlist (`nets`) structure compatible with `SchematicIntent`.
+
+All pattern files are validated against `src/kicad/patterns/pattern.schema.json` via:
+
+```bash
+npm run validate:patterns
+```
+
+## Available Patterns
+
+| Pattern File | Pattern Name | Description | Part Count |
+|:---|:---|:---|:---:|
+| `voltage-regulator-ams1117.json` | `voltage-regulator-ams1117` | AMS1117 3.3V linear voltage regulator subcircuit with input and output ceramic decoupling capacitors. | 3 |
+| `usb-c-power-input.json` | `usb-c-power-input` | USB Type-C 16-pin USB 2.0 power input sink block with 5.1k CC1/CC2 pull-down resistors and VBUS bulk capacitor. | 4 |
+| `crystal-oscillator.json` | `crystal-oscillator` | Standard crystal oscillator resonant circuit with dual load capacitors, matching canonical crystal flanking topology. | 3 |
+
+## Schema and Structure
+
+Pattern files follow the top-level schema defined in `src/kicad/patterns/pattern.schema.json`:
+
+```json
+{
+  "name": "pattern-identifier",
+  "description": "Human-readable description of the subcircuit.",
+  "parts": [
+    {
+      "ref": "U1",
+      "libId": "Regulator_Linear:AMS1117-3.3",
+      "value": "AMS1117-3.3",
+      "footprint": "Package_TO_SOT_SMD:SOT-223-3_TabPin2",
+      "group": "Power"
+    }
+  ],
+  "nets": [
+    {
+      "name": "VIN",
+      "pins": ["U1.3", "C1.1"],
+      "kind": "power"
+    }
+  ]
+}
+```
--- a/openspec/changes/add-pattern-expansion-helper/.openspec.yaml
+++ b/openspec/changes/add-pattern-expansion-helper/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-09-07
--- a/openspec/changes/add-pattern-expansion-helper/design.md
+++ b/openspec/changes/add-pattern-expansion-helper/design.md
@@ -1,0 +1,71 @@
+# add-pattern-expansion-helper: Design
+
+## Architecture
+
+The pattern expansion helper is designed as a pure, side-effect-free function residing in `src/kicad/draft/patternExpand.ts`.
+
+```text
+src/kicad/patterns/<name>.json  ────────┐
+                                        │
+refPrefix / instanceId options ─────────┼──► expandPatternRef() ──► { parts, nets }
+                                        │                             (ready to merge into
+                                        │                              schematic.intent.json)
+```
+
+## Interface Specification
+
+```typescript
+export interface ExpandPatternOptions {
+  /**
+   * Prefix or naming modifier applied to all component reference designators in the pattern.
+   * Example: "PWR_" turns "U1" into "PWR_U1" and "C1" into "PWR_C1".
+   */
+  refPrefix?: string;
+
+  /**
+   * Optional instance identifier or suffix.
+   */
+  instanceId?: string;
+
+  /**
+   * Optional custom directory to load pattern JSON files from (defaults to src/kicad/patterns).
+   */
+  patternsDir?: string;
+
+  /**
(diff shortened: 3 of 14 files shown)


adhavan18 (commented):
reviewed the diff and openspec artifacts. the core `expandPatternRef` logic is solid: refPrefix/instanceId renumbering, net endpoint rewriting, and the error paths are all covered by real assertions in test/pattern-expand.test.ts, not just smoke tests.

two things worth raising before this lands:

1. **branch base / overlap with #279**: this PR includes the exact same docs/patterns.md, scripts/validate-patterns.ts, package.json entry, and all three pattern JSONs that #279 adds (diffed it against #279's PR — byte-identical on those files). it's branched off main directly rather than off #279's branch, so whichever of the two merges second is going to conflict on every one of those files. worth rebasing this on top of #279 (or #279 merging first) so this PR's diff is just patternExpand.ts + its test.

2. **net names aren't renamed on expansion, only part refs are.** `expandPatternRef` renumbers `U1` -> `PWR_U1` etc, and rewrites `nets[].pins` endpoints to match, but `nets[].name` is passed through unchanged (patternExpand.ts, the net-rewriting loop near the end). so calling this twice for two regulator instances (`REG1_`, `REG2_`) produces two separate nets both still named `VIN`, `+3V3`, `GND`. GND colliding is probably intentional (single ground plane), but VIN/+3V3 colliding is not: merge both expansions into one SchematicIntent and the two regulators' VIN/+3V3 nets get silently shorted together under one net name instead of staying separate per the docs' framing of prefix/instanceId as *isolating* each instance. the design.md and delta spec don't mention this at all, and there's no test that expands the same pattern twice and checks the resulting net set. given the whole point of instanceId/refPrefix is "use this pattern more than once in one board," this is the one scenario that actually needs a test and doesn't have one.

design/proposal/tasks docs are otherwise well-scoped (explicitly not touching ir.ts, engine.ts, or create.ts, matches the "expansion-only" descope from #277). tasks.md checkboxes are all left unchecked even though the work described is implemented and tested — worth ticking those before merge per AGENTS.md's review expectations.

PR #282, opened 2026-09-07
fix(mcp): make per-repo mutating-call lock acquisition atomic

DevSars24:
## What
Fixes a race condition in the MCP server where two near-simultaneous mutating tool calls (`copperhead_init`, `copperhead_do`, `copperhead_sync` with `resolve=true`) against the same repo could both proceed instead of one being correctly refused as busy.

## Why
The per-repo mutating-call lock (`locks.tryAcquire(repoRoot)`) was called after one or more `await` points inside each handler (`await requireKicad()`, `await resolveModelOrFail()`). This left an async gap: both concurrent calls could suspend during preflight and reach the lock check before either had acquired it, or the first call could finish and release the lock before the second call even checked it. Either way, both calls got admitted, violating the documented per-repo serialization spec and defeating the purpose of the lock (protecting a repo from concurrent mutating runs stepping on each other).

## Design
- **Root cause**: non-atomic check-then-set. The lock check (`locks.tryAcquire`) happened after async work, not at the synchronous entry of the handler, so two calls racing on the same event-loop tick (or close enough) could both slip past.
- **Fix**: moved `locks.tryAcquire(repoRoot)` to the very first synchronous line of each mutating handler in [server.ts](src/mcp/server.ts), before any `await`. The existing busy-error response (`unavailable`, "another copperhead run is in progress for this repo; retry shortly") is unchanged. All prior handler logic (`requireKicad()`, `resolveModelOrFail()`, the actual run) now executes inside the existing `try` block, with `locks.release(repoRoot)` still in `finally`.
- **Scope**: touches only lock-acquisition ordering in `copperhead_init`, `copperhead_do`, and `copperhead_sync`. No change to the lock implementation itself (`locks.tryAcquire`/`release` stay as-is, already a synchronous Map/Set-based check-and-set), no change to read-only tools (`copperhead_check`, `copperhead_doctor`), no change to any other package (`ir`, `engine`, `commands`, other tools). The diff is 48 insertions / 49 deletions in a single file, mostly a 2-space indentation shift from wrapping existing await calls one level deeper inside the `try` block.
- **Independent of #276 / PR #281**: confirmed via `git diff main feature/pattern-bom-expansion -- src/mcp/ test/mcp-server.test.ts` showing 0 changed lines before this fix; this bug is pre-existing on `main` and unrelated to the pattern-BOM-expansion work.

### Automated test status
| Check | Command | Status |
| --- | --- | --- |
| Build | `npm run build` | ✅ Pass |
| Unit + offline | `npm test` | ✅ Pass (902/947 on this Windows dev machine; all 24 failures verified pre-existing/environment-specific, see below — no failures attributable to this change) |
| Targeted race test | `npx vitest run test/mcp-server.test.ts -t "rejects a concurrent mutating call with a typed busy error"` | ✅ Pass 20/20 consecutive runs |
| Live-LLM (opt-in) | n/a | n/a — no provider-facing code touched |

**Pre-existing/environment failures (verified, not introduced by this PR):** on this Windows machine, both `main` and this branch show the same class of failures: 60s timeouts on `kicad-cli` subprocess spawns under parallel Vitest workers (`init-check`, `create-preflight`, `draft-*`, `tool-registry`, etc.), Windows-specific assertion mismatches (CRLF vs `\n`, `chmod 000` not restricting access on NTFS, Windows home-path formatting, Windows NTSTATUS exit codes instead of POSIX), and one test (`draft-hardening`) that fails only under full-suite CPU saturation but passes 32/32 in isolation. None of these touch `src/mcp/server.ts` locking logic. This PR's CI run (Linux) is the authoritative signal for merge, since none of the above Windows-specific causes apply there.

The one test this PR directly fixes — `rejects a concurrent mutating call with a typed busy error` — fails on `main` and passes reliably (20/20) on this branch.

- Sandbox variant used: n/a
- Reason: this is an internal MCP server concurrency fix with no change to CLI-facing behavior, the agent loop, or KiCad file handling. Exercised instead via the automated concurrency test above, run 20 times consecutively to rule out a lucky pass.
```text
$ for i in $(seq 1 20); do npx vitest run test/mcp-server.test.ts -t "rejects a concurrent mutating call with a typed busy error"; done
20 passed (20)

$ git diff main --stat
 src/mcp/server.ts | 97 +++++++++++++++++++++++++++----------------------------
 1 file changed, 48 insertions(+), 49 deletions(-)
```

## Docs
No user-facing docs changes. This is an internal correctness fix to MCP tool-call serialization; the documented spec behavior ("mutating tools are serialized per repo") is unchanged, this PR just makes the implementation actually satisfy it.

---

## Invariant checklist
- [ ] **Spec-gated in**: N/A, no new tool exposed.
- [x] **Verification-gated out**: N/A directly, but this strengthens an existing safety property (mutating tools are serialized per repo) that protects against concurrent mutation of the same repo stepping on ERC/DRC gating and rollback logic.
- [x] **`check`/`verify` stays LLM-free and network-free**: N/A — no changes under `src/commands/check`; `copperhead_check`/`copperhead_doctor` (read-only tools) are untouched by this fix.
- [x] **No sexp serialization**: N/A — no KiCad file parsing/writing touched.
- [x] **Sync-obligations ledger**: N/A — no post-tool-call hooks or ledger code touched.
- [x] **Secrets**: N/A — no secrets, env vars, or transcript/log code touched.
- [x] **Spec coherence**: N/A — no `SPEC.md`-level behavior changed; the documented serialization spec is unchanged, only the implementation's fidelity to it.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -308,12 +308,12 @@ export function createMcpServer(opts: ServerOptions): McpServer {
       _meta: { schemaVersion: TOOL_SCHEMA_VERSIONS.copperhead_init },
     },
     async ({ path: searchPath }) => {
-      const kicad = await requireKicad();
-      if ('error' in kicad) return toMcp(kicad.error);
       if (!locks.tryAcquire(repoRoot)) {
         return toMcp(failure('unavailable', 'another copperhead run is in progress for this repo; retry shortly'));
       }
       try {
+        const kicad = await requireKicad();
+        if ('error' in kicad) return toMcp(kicad.error);
         const res = await runInit({
           repoRoot,
           ...(searchPath ? { searchPath } : {}),
@@ -361,26 +361,26 @@ export function createMcpServer(opts: ServerOptions): McpServer {
       _meta: { schemaVersion: TOOL_SCHEMA_VERSIONS.copperhead_do },
     },
     async ({ request, dry_run: dryRun }, extra) => {
-      const kicad = await requireKicad();
-      if ('error' in kicad) return toMcp(kicad.error);
-      const resolved = await resolveModelOrFail(repoRoot);
-      if ('error' in resolved) return toMcp(resolved.error);
       if (!locks.tryAcquire(repoRoot)) {
         return toMcp(failure('unavailable', 'another copperhead run is in progress for this repo; retry shortly'));
       }
-      const progressToken = extra?._meta?.progressToken;
-      const sink: ProgressSink = (update): void => {
-        if (progressToken === undefined) return;
-        void extra
-          ?.sendNotification({
-            method: 'notifications/progress',
-            params: { progressToken, progress: update.progress, message: update.message },
-          })
-          .catch(() => {
-            // A host that stopped listening must not fail the run.
-          });
-      };
       try {
+        const kicad = await requireKicad();
+        if ('error' in kicad) return toMcp(kicad.error);
+        const resolved = await resolveModelOrFail(repoRoot);
+        if ('error' in resolved) return toMcp(resolved.error);
+        const progressToken = extra?._meta?.progressToken;
+        const sink: ProgressSink = (update): void => {
+          if (progressToken === undefined) return;
+          void extra
+            ?.sendNotification({
+              method: 'notifications/progress',
+              params: { progressToken, progress: update.progress, message: update.message },
+            })
+            .catch(() => {
+              // A host that stopped listening must not fail the run.
+            });
+        };
         const res = await runAgentLoop({
           repoRoot,
           request,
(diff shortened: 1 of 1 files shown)


adhavan18 (commented):
reviewed the diff against src/mcp/server.ts on main. the fix is correct and minimal: in all three handlers (copperhead_init, copperhead_do, sync --resolve), `locks.tryAcquire` now runs before any `await` (requireKicad / resolveModelOrFail / syncVerify), instead of after. on main today the async probe runs first, so two concurrent calls can both pass the probe before either takes the lock, which is exactly the MCP-lock race #281 hit.

checked the release side too: every early-return path that follows a successful tryAcquire either has an explicit `locks.release(repoRoot)` before returning or is covered by the trailing `finally`, so i don't see a double-release or a leaked lock on any of the three handlers.

this is a real fix for a real race, but it ships with zero test coverage — the diff only touches src/mcp/server.ts, no changes to test/mcp-server.test.ts. CONTRIBUTING.md asks for tests on anything you change, and this repo already has a concurrency test for exactly this lock (`rejects a concurrent mutating call with a typed busy error`) that would be the natural place to extend: right now that test's own comment says "each call awaits a kicad-cli probe before it tries the lock," which describes the pre-fix ordering this PR removes. worth adding a case (or extending that test) that makes requireKicad/resolveModelOrFail slow via a mock and asserts the lock is still taken before either resolves, so a regression back to the old ordering gets caught by CI instead of relying on manual review.

everything else checks out: no behavior change to the actual init/do/resolve logic, just call-order, and CI is green.

PR #286, opened 2026-09-08
docs(site): make the GitBook theme hold up on a phone

animesh-chouhan:
## What

Makes the docs site readable on a narrow column. Tables stop being content-sized scrolling blocks and become labelled cards below 34rem of column width, inline code stops breaking mid-token, long lines in code blocks wrap instead of being cut off, and a code block nested two lists deep gets its width back. Follow-up to #271.

## Why

#271 restyled the site after GitBook's clean theme, but only the wide case was checked. On a phone, and on a tablet with the sidebar open, five things broke:

- Starlight renders a table as a content-sized scrolling block, which left a ragged right edge next to full-width code blocks. Worse, the binding constraint is the content column, not the viewport: with the sidebar open the column is about 500px, and a four-column table was crushed into one-word ribbons.
- Inline code broke mid-token, so `--json` split at the hyphen and `kicad-cli` at its own.
- Long lines in code blocks were cut off rather than wrapped. Sample briefs and prompts are prose-shaped, so the tail simply vanished.
- A code block nested two lists deep was squeezed to about two thirds of the column.
- Two cosmetic leftovers: the first [JumpLink.astro](docs/src/components/JumpLink.astro) sat flush against its heading, because Starlight's content-spacing rule skips `<a>` siblings, and the Steps guide line ran on past the last step, connecting to nothing.

## Design

[rehype-table-labels.mjs](docs/src/plugins/rehype-table-labels.mjs) is a build-time pass over the content HTML, wired in through `markdown.rehypePlugins` in [astro.config.mjs](docs/astro.config.mjs). It does two things:

- wraps each `<table>` in `.table-wrap`, so the table itself can be a real full-width table while the wrapper owns horizontal overflow, and stamps every body cell with `data-label="<its column header>"`;
- marks a short single-token inline code span `nowrap`, and splits a multi-word span into per-token `<span class="nowrap">`, so a break can only land on a space. Tokens over 30 characters are left free to wrap. An `inPre` guard keeps the pass out of code blocks entirely.

In [custom.css](docs/src/styles/custom.css), `.sl-markdown-content` becomes a size container, so the table rules key off the width the table actually has rather than the viewport. Below 34rem of column width each row becomes a labelled card; the header row is clipped out of the flow but stays in the DOM for assistive tech.

Headings that were a bare path (`.copperhead/config.json`) now name the file and give the path in the prose, so a heading is not the widest thing on the page. That changes their generated slugs, so each renamed heading keeps an explicit `<a id>` carrying its pre-rename slug (`#copperheadconfigjson` and friends) and links published against the old headings still land on their section.

## Spec / OpenSpec

n/a, docs-only and presentational. No spec-level behavior changes, so SPEC.md and the change artifacts correctly stay put.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Build | `npm run build` (in `docs/`) | pass, clean at 17 pages |
| Unit + offline | `npm test` | 923 pass / 21 skip / 15 fail locally. The 12 new tests all pass; the failures are environment-specific to this machine (create, gating, init, preflight, mcp-server, and the set fluctuates between runs), they are in files this PR does not touch, and CI runs the same suite green. |

New tests: [docs-rehype-table-labels.test.ts](test/docs-rehype-table-labels.test.ts), 12 cases. `docs/` has no harness of its own, but the plugin only ever touches plain objects, so the root vitest suite can exercise it with no new dependency. It pins the table wrapping and labelling (including a row wider than its header, and a table with no header row), the nowrap split, the over-long-token case, and that nothing inside a `<pre>` is touched.

- Sandbox variant used (`create` / `edit`): n/a, this is a docs-site style change with no CLI surface.
- Commands run and outcome: built the site and asserted the plugin's output in `dist/`, including that the `inPre` guard holds across all 17 pages, that both slugs (old and new) are present on the renamed headings, and that the nested-list rule ships as a container query. Also rebuilt the stashed baseline to confirm the two build warnings (the `gitignore` language and the `/404.html` entry) are pre-existing and not introduced here.

```text
$ cd docs && npm run build
[build] 17 page(s) built in 2.79s
[build] Complete!

$ # tables wrapped, on the CLI reference
15

$ # data-label stamped from the column headers
data-label="Option"  data-label="Meaning"  data-label="Exit code"  ...

$ # single tokens marked nowrap, multi-word spans split at the space
<code class="nowrap" dir="auto">config.json</code>
<code dir="auto"><span class="nowrap">copperhead</span> <span class="nowrap">init</span></code>

$ # no nowrap markup leaked into any code block, across all pages
code blocks containing nowrap: 0

$ # the pre-rename slugs still resolve, alongside the new ones
legacy anchors present: ['copperheadconfigjson', 'copperheadconstraintsjson']
new slugs still present: True True

$ # the nested-list rule ships as a container query (minified range syntax)
@container (width<=39.99rem){.sl-markdown-content :is(ul,ol) :is(ul,ol)...{padding-inline-start:1rem}

$ npx vitest run test/docs-rehype-table-labels.test.ts
Test Files  1 passed (1)
     Tests  12 passed (12)
```

## Docs

The two content edits are in this PR: [configuration.md](docs/src/content/docs/reference/configuration.md) and [docs-as-memory.md](docs/src/content/docs/concepts/docs-as-memory.md), where a heading that was a bare path now names the file and gives the path in the prose.

---

## Invariant checklist

This diff changes zero lines under `src/` (the only non-docs file is a new test), so every gated path is unreachable from it. Each box is marked N/A on that basis, and the reviewer confirmed it independently.

- [x] **`check`/`verify` stays LLM-free and network-free**: N/A, no change to `src/commands/check` or anything it imports.
- [x] **No sexp serialization**: N/A, no change under `src/kicad/`.
- [x] **Sync-obligations ledger**: N/A, no change to the hooks or the commit gate.
- [x] **Secrets**: N/A, no new logging or transcript path.
- [x] **Spec coherence**: N/A, no spec-level behavior changed.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -1,6 +1,7 @@
 import { defineConfig } from 'astro/config';
 import starlight from '@astrojs/starlight';
 import starlightLlmsTxt from 'starlight-llms-txt';
+import rehypeTableLabels from './src/plugins/rehype-table-labels.mjs';
 
 const REPO = 'https://github.com/copperheadhq/copperhead';
 
@@ -9,6 +10,7 @@ const REPO = 'https://github.com/copperheadhq/copperhead';
 // so Pages cannot own a path under it.
 export default defineConfig({
   site: 'https://docs.copperhead.sh',
+  markdown: { rehypePlugins: [rehypeTableLabels] },
   integrations: [
     starlight({
       title: 'copperhead',
@@ -37,6 +39,9 @@ export default defineConfig({
         './src/styles/custom.css',
       ],
       expressiveCode: {
+        // Long lines wrap instead of scrolling: sample briefs, prompts, and
+        // commands are prose-like and were being cut off on phones.
+        defaultProps: { wrap: true, preserveIndent: true },
         styleOverrides: {
           borderRadius: '0.75rem',
           codeFontFamily: "'IBM Plex Mono', ui-monospace, SFMono-Regular, Menlo, Consolas, monospace",
--- a/docs/src/components/JumpLink.astro
+++ b/docs/src/components/JumpLink.astro
@@ -40,6 +40,11 @@ const { href, title, description, icon } = Astro.props;
 			border-color 0.15s,
 			transform 0.15s;
 	}
+	/* Starlight's content-spacing rule skips <a> siblings, so the first link
+	   would otherwise sit flush against the heading above it. */
+	.jump-link {
+		margin-top: 1rem;
+	}
 	.jump-link + .jump-link {
 		margin-top: 0.75rem;
 	}
--- a/docs/src/content/docs/concepts/docs-as-memory.md
+++ b/docs/src/content/docs/concepts/docs-as-memory.md
@@ -41,6 +41,8 @@ Docs that disagree with the schematic are the failure mode this whole tool exist
 
 Hand edits that desync the docs, the constraint registry, or the schematic fail at commit time. See [Verify and sync](/workflows/verify-and-sync/).
 
+<a id="copperheadreadmemd"></a>
+
 ## `README.md` in `.copperhead/`
 
 The config directory describes itself. `init` writes a README into `.copperhead/` documenting every key in `config.json` and what `constraints.json` is for, so someone who clones the repo and has never used copperhead can still tell what these files are.
--- a/docs/src/content/docs/concepts/docs-as-memory.md
+++ b/docs/src/content/docs/concepts/docs-as-memory.md
@@ -41,6 +41,6 @@ Docs that disagree with the schematic are the failure mode this whole tool exist
 
 Hand edits that desync the docs, the constraint registry, or the schematic fail at commit time. See [Verify and sync](/workflows/verify-and-sync/).
 
-## `.copperhead/README.md`
+## `README.md` in `.copperhead/`
 
 The config directory describes itself. `init` writes a README into `.copperhead/` documenting every key in `config.json` and what `constraints.json` is for, so someone who clones the repo and has never used copperhead can still tell what these files are.
--- a/docs/src/content/docs/reference/configuration.md
+++ b/docs/src/content/docs/reference/configuration.md
@@ -5,9 +5,9 @@ sidebar:
   order: 2
 ---
 
-## `.copperhead/config.json`
+## `config.json`
 
-Written by `copperhead init`. Every key is optional; the defaults below apply when a key is absent or the file does not exist.
+`.copperhead/config.json`, written by `copperhead init`. Every key is optional; the defaults below apply when a key is absent or the file does not exist.
 
 ```json
 {
@@ -52,9 +52,9 @@ Budgets are hard constraints, not hints. A change that would exceed one is refus
 
 The names are yours. copperhead passes them through verbatim and expects the units to be in the key, as in `sleep_current_uA`.
 
-## `.copperhead/constraints.json`
+## `constraints.json`
 
-The constraint registry: machine-readable counterparts to the constraints stated in your design docs. Constraints are dual-written, to the doc and to the registry, and the sync-obligations ledger refuses to let a run commit if one was updated without the other.
+`.copperhead/constraints.json` is the constraint registry: machine-readable counterparts to the constraints stated in your design docs. Constraints are dual-written, to the doc and to the registry, and the sync-obligations ledger refuses to let a run commit if one was updated without the other.
 
 ## Environment variables
 
--- a/docs/src/content/docs/reference/configuration.md
+++ b/docs/src/content/docs/reference/configuration.md
@@ -5,6 +5,8 @@ sidebar:
   order: 2
 ---
 
+<a id="copperheadconfigjson"></a>
+
 ## `config.json`
 
 `.copperhead/config.json`, written by `copperhead init`. Every key is optional; the defaults below apply when a key is absent or the file does not exist.
@@ -51,6 +53,8 @@ Budgets are hard constraints, not hints. A change that would exceed one is refus
 ```
 
 The names are yours. copperhead passes them through verbatim and expects the units to be in the key, as in `sleep_current_uA`.
+
+<a id="copperheadconstraintsjson"></a>
 
 ## `constraints.json`
 
--- a/docs/src/plugins/rehype-table-labels.mjs
+++ b/docs/src/plugins/rehype-table-labels.mjs
@@ -1,0 +1,93 @@
+/**
+ * Build-time HTML tweaks for the markdown content:
+ *
+ * - Every <table> is wrapped in <div class="table-wrap"> so the table itself
+ *   can be a real full-width table while the wrapper takes care of horizontal
+ *   scrolling if it is ever too wide.
+ * - Every body cell gets `data-label="<its column header>"`, so the
+ *   stylesheet can stack rows into cards on narrow screens and still show
+ *   which column each value came from.
+ * - Inline code never breaks inside a token: a short single-token span
+ *   (`--json`, `kicad-cli`) gets class "nowrap", and a multi-word span
+ *   (`--model claude-code`) has each token wrapped in <span class="nowrap">
+ *   so it can only wrap at the spaces. Tokens longer than NOWRAP_MAX_CHARS
+ *   are left alone so nothing can overflow a phone column.
+ *
+ * See the "Tables" and "Inline code" blocks in src/styles/custom.css.
+ */
+const NOWRAP_MAX_CHARS = 30;
+
+function textOf(node) {
+  if (node.type === 'text') return node.value;
+  return (node.children ?? []).map(textOf).join('');
+}
+
+function children(node, tag) {
+  return (node.children ?? []).filter((c) => c.type === 'element' && c.tagName === tag);
+}
+
+function addClass(node, name) {
+  const props = (node.properties ??= {});
+  const existing = Array.isArray(props.className)
+    ? props.className
+    : typeof props.className === 'string'
+      ? props.className.split(/\s+/).filter(Boolean)
+      : [];
+  if (!existing.includes(name)) props.className = [...existing, name];
+}
+
+function visit(node, fn, parent = null, inPre = false) {
+  const replacement = fn(node, parent, inPre);
+  const next = inPre || (node.type === 'element' && node.tagName === 'pre');
+  for (const child of [...(node.children ?? [])]) visit(child, fn, node, next);
+  return replacement;
+}
+
+export default function rehypeTableLabels() {
+  return (tree) => {
+    visit(tree, (node, parent, inPre) => {
+      if (node.type !== 'element') return;
+
+      if (node.tagName === 'code' && !inPre) {
+        const text = textOf(node);
+        if (!/\s/.test(text)) {
+          if (text.length <= NOWRAP_MAX_CHARS) addClass(node, 'nowrap');
+          return;
+        }
+        // Only touch the plain case: one text node, nothing nested.
+        if (node.children.length === 1 && node.children[0].type === 'text') {
+          node.children = text.split(/(\s+)/).map((part) =>
+            /^\s+$/.test(part) || part.length > NOWRAP_MAX_CHARS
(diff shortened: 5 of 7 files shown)


animesh-chouhan:
_Automated pr-review pass (the `pr-review` skill), not a human review._

**Disclosure: this is a self-review.** The single commit under review (`f0adcc4`) was authored by this agent, so treat the findings below as self-scrutiny, which warrants more skepticism rather than less.

## Verdict

**Approve with nits, once the red check is fixed.** The change does what it says: verified in the built HTML, not just read. One CI check is red and it is the description, not the code. The rest are low.

```
== metrics block ==
lines: +220 / -6 (net 214) across 6 files  ·  src +0/-0, test +0/-0, docs +220/-6, spec +0/-0, other +0/-0
new vs net: 1 new file (0 in src), 5 modified; 0 new src lines (the net-new surface)
tests: +0 test lines; CI `test` passes at head. Local suite in the review worktree: 911 pass / 21 skip / 14 fail
ci: 5 pass / 1 fail (`template`); required checks: none reported (gh pr checks)

== method ==
measured in: /tmp/pr-review-worktrees/.../pr-286 (isolated worktree; the user's checkout was never entered and stayed on its own branch)
head: f0adcc4  ·  merge base: c1be00d (git merge-base origin/main HEAD)
areas: git diff --numstat against the merge base
suite: npx vitest run --reporter=json in the worktree
docs build: npm run build in docs/, clean at 17 pages, and CI "Docs build" is green
note: the setup script warned that this PR changes package-lock.json. That is a false positive: it compared against a stale local `main`. gh reports the six docs/ files only.
note: the 14 local failures (create-resilience, create-stage-turns, gating-sync, init-check, preflight, stage-completion) are environment-specific to this machine. CI runs the same suite green at this head, and the PR changes zero src and zero test lines, so nothing in the suite's input differs from the base.
```

## Findings

### 1. medium: the PR description fails the repo's own `pr-lint` check

`.github/workflows/pr-lint.yml` requires `## What`, `## Why` and `## Testing` sections with content, plus a surviving invariant checklist (`- [ ]` items). The current body has none of them, so `template` is the one red check on this PR. The body also uses an em dash on the "Inline code broke mid-token" line, which `.github/pull_request_template.md` explicitly forbids.

Fix: rewrite the body to the template. The content is already there; it needs to be under the required headings.

### 2. low: the plugin's "nothing can overflow" guarantee is not actually enforced

[rehype-table-labels.mjs:14-16](docs/src/plugins/rehype-table-labels.mjs#L14-L16) says tokens longer than `NOWRAP_MAX_CHARS` "are left alone so nothing can overflow a phone column". Leaving them alone only omits `nowrap`; no rule gives inline code an `overflow-wrap`, so a long token wraps only where the browser finds a break opportunity. `overflow-wrap: anywhere` is set on `td` in card mode only, not on inline code in prose.

Today's content is safe: all five inline tokens over 30 chars contain a `/` or a `-`, so they break. A future token of unbroken word characters (a hex digest, a long `snake_case` identifier) would overflow the column instead.

Fix: add `overflow-wrap: anywhere` to the inline-code rule, or soften the comment to say the guarantee depends on the token containing a break opportunity.

### 3. low: the nested-list fix uses a viewport query where the table fix correctly uses a container query

[custom.css:346](docs/src/styles/custom.css#L346) gates the nested-list padding on `@media (max-width: 39.99rem)`, while the table rules below it key off `@container`, with a comment explaining exactly why the viewport is the wrong axis: with the sidebar open on a tablet the content column is only about 500px. The nested-list squeeze has the same cause, so it will not be fixed in the case the PR itself identifies as the interesting one.

Fix: move the rule into `@container (max-width: 39.99rem)`.

### 4. low: the renamed headings change public anchor slugs

`## .copperhead/config.json` becomes `## config.json`, so `#copperheadconfigjson` becomes `#configjson`; likewise `constraints.json` and, in [docs-as-memory.md](docs/src/content/docs/concepts/docs-as-memory.md), `.copperhead/README.md`. I grepped the repo and nothing links to the old anchors, so no in-repo link breaks. External bookmarks and any inbound links to those anchors will land at page top instead of the section. The PR description does call this out, so it is a known and accepted cost rather than a surprise.

Fix, if the cost is not wanted: keep an empty `<a id="...">` at each renamed heading.

### 5. low: the new plugin has no test, and docs/ has no harness to hold one

93 new lines of build-time HTML rewriting with no automated check. `docs/` has no test setup, so this is a gap in the area rather than a missed test file, and the verification that was done is real: the build is clean, and the built HTML was asserted against (15 tables wrapped on the CLI reference, `data-label` correct, single tokens carrying `class="nowrap"`, `copperhead init` split into two nowrap spans, and zero `nowrap` leakage into any `<pre>` across all 17 pages, which is the failure mode the `inPre` guard exists to prevent). Worth noting so the next change to this file knows nothing will catch a regression.

### 6. low, nit: dead code in `visit()`

[rehype-table-labels.mjs:39-44](docs/src/plugins/rehype-table-labels.mjs#L39-L44): `const replacement = fn(node, parent, inPre)` is returned from `visit`, but `fn` never returns a meaningful value and no caller reads the result. Drop the variable and the return.

### 7. low, nit: duplicated selector

[custom.css:335-341](docs/src/styles/custom.css#L335-L341) opens a second `.sl-markdown-content code:not(:where(.not-content *))` block for `box-decoration-break`, about 25 lines below the existing block that sets the same selector's padding. Folding it into the first keeps the inline-code styling in one place.

## Invariant checks

All six were checked and all pass trivially: the diff changes zero lines under `src/`, so none of the gated paths is reachable from it.

- Spec-gated in (`edit_file`/`write_file` absent until a proposal validates): unaffected, no change to the agent tool list.
- Verification-gated out (ERC, and DRC when the board changed, then repair up to `maxRepairCycles`, then rollback): unaffected.
- `check`/`verify` stays LLM-free and network-free: unaffected, no change to `src/commands/check` or anything it imports.
- No sexp serialization: unaffected, no change under `src/kicad/`.
- Sync-obligations ledger: unaffected.
- Secrets (`sk-` redaction at write time, `.env` and `.copperhead/runs/` ignored): unaffected, no new logging or transcript path.

Spec coherence: docs-only and presentational. No spec-level behavior changes, so SPEC.md and the OpenSpec change artifacts correctly stay put.

## Fix prompt

```
Resolve these pr-review findings on PR #286 (branch docs/mobile-layout, head f0adcc4).
The PR restyles the docs site for narrow columns: a rehype plugin
(docs/src/plugins/rehype-table-labels.mjs) wraps tables and stamps data-label
attributes and nowrap classes, and docs/src/styles/custom.css turns table rows into
cards below 34rem of *column* width using container queries.

1. medium: the PR description fails .github/workflows/pr-lint.yml. Rewrite the PR body
   to follow .github/pull_request_template.md: sections "## What", "## Why" and
   "## Testing" must be present and non-empty, and the invariant checklist ("- [ ]"
   items) must survive. Keep the prose em-dash-free: the current body has an em dash on
   the "Inline code broke mid-token" line. The existing content is fine, it just needs
   to sit under the required headings.

2. low: docs/src/plugins/rehype-table-labels.mjs lines 14-16 claim tokens longer than
   NOWRAP_MAX_CHARS "are left alone so nothing can overflow a phone column", but no CSS
   gives inline code an overflow-wrap, so such a token only wraps where the browser finds
   a break opportunity. Either add `overflow-wrap: anywhere` to the inline-code rule in
   docs/src/styles/custom.css, or reword the comment to say the guarantee depends on the
   token containing a break opportunity.

3. low: docs/src/styles/custom.css line 346 gates the nested-list padding fix on
   `@media (max-width: 39.99rem)`, while the table rules just below correctly use
   `@container`, with a comment explaining that the content column, not the viewport, is
   the binding constraint (sidebar open on a tablet leaves about 500px). Move the
   nested-list rule into `@container (max-width: 39.99rem)` so it fires in that case too.

4. low: the renamed headings change public anchor slugs (#copperheadconfigjson ->
   #configjson, and the same for constraints.json and .copperhead/README.md). Nothing in
   the repo links to the old anchors. If inbound links matter, add an empty
   <a id="copperheadconfigjson"></a> style target at each renamed heading; otherwise
   leave as is and treat this as accepted.

5. low: docs/src/plugins/rehype-table-labels.mjs has no test and docs/ has no test
   harness. If you add one, the behaviours worth pinning are: a table gets .table-wrap
   and data-label per body cell; a single-token inline code gets class "nowrap"; a
   multi-word one is split into per-token nowrap spans; a token over 30 chars is left
   alone; and nothing inside <pre> is touched.

6. low, nit: docs/src/plugins/rehype-table-labels.mjs lines 39-44, `visit()` assigns
   `const replacement = fn(...)` and returns it, but fn returns nothing a caller uses.
   Drop the variable and the return.

7. low, nit: docs/src/styles/custom.css lines 335-341 open a second
   `.sl-markdown-content code:not(:where(.not-content *))` block about 25 lines below the
   existing one. Fold the box-decoration-break declaration into the first block.

Repo invariants that must keep holding (all currently unaffected: this diff changes zero
src lines, and it must stay that way):
- Spec-gated in: edit_file/write_file stay structurally absent from the agent tool list
  until an OpenSpec proposal validates. Gate by omission, never by prompt text.
- Verification-gated out: no mutation is done until ERC passes (and DRC if the board
  changed), with repair up to maxRepairCycles then rollback to the git snapshot.
- check/verify stays LLM-free and network-free: src/commands/check and everything it
  imports must not touch a provider, an API key, or the network.
- No sexp serialization: the parser in src/kicad/ is read-only; KiCad files are edited
  only by anchored exact-match text replace.
- The sync-obligations ledger keeps being fed by post-tool-call hooks, and commit keeps
  refusing while obligations are open.
- Secrets: transcripts and summaries redact sk-[A-Za-z0-9_-]+ at write time, keys live
  only in env vars, and .gitignore keeps .env and .copperhead/runs/.

Verify with:
  cd docs && npm run build          # must stay clean at 17 pages
  npm run typecheck && npm test     # from the repo root, for any src change
And re-check the built HTML for the plugin's output, for example that no <pre> block in
dist/ contains the string "nowrap".
```

animesh-chouhan:
_Automated pr-review follow-up: fixes applied in c62f540._

All seven findings from the [review above](https://github.com/copperheadhq/copperhead/pull/286#issuecomment-5587139052) are addressed.

| # | Finding | Resolution |
| --- | --- | --- |
| 1 | medium: PR body failed `pr-lint` | Rewritten to the template; `template` check is green. |
| 2 | Long-token overflow guarantee unenforced | `overflow-wrap: break-word` added to the inline-code rule, and the plugin comment now describes what the CSS actually does. |
| 3 | Nested-list fix used a viewport query | Moved into `@container (max-width: 39.99rem)`. Ships as `@container (width<=39.99rem)` after minification. |
| 4 | Renamed headings changed public slugs | Each renamed heading keeps an explicit `<a id>` with its pre-rename slug, so old links still land on the section. Both slugs verified in the built HTML. |
| 5 | Plugin had no test | [docs-rehype-table-labels.test.ts](test/docs-rehype-table-labels.test.ts), 12 cases in the root suite. The plugin only touches plain objects, so no new dependency was needed. |
| 6 | Dead `replacement` variable in `visit()` | Dropped, and replaced with comments saying why the child list is copied and why `<pre>` is skipped. |
| 7 | Duplicated inline-code selector | Folded into the single block. |

Verification: `npm run typecheck` passes, `npm run build` in `docs/` stays clean at 17 pages, the 12 new tests pass, and the built HTML still shows 15 wrapped tables on the CLI reference, correct `data-label` stamping, and zero `nowrap` leakage into any `<pre>` across all 17 pages.

The one finding worth a second opinion is 3: the threshold stayed at 39.99rem when it moved from viewport to column, so it now fires slightly earlier than before (the column is narrower than the viewport by the page padding). That is the intended direction, but it is a behavior change rather than a pure relocation.

animesh-chouhan:
_Note on the red CI run in this PR's history:_ the `test` job failed once on c62f540 at `test/mcp-server.test.ts > rejects a concurrent mutating call with a typed busy error`, which is the per-repo mutating-call lock race that #282 is open to fix. The 12 new tests passed in that same run (937 passed overall), and this PR touches neither `src/mcp/` nor that test file. A re-run of the identical commit is green.

PR #287, opened 2026-09-08
docs(site): send docs traffic to the same GA property as the apex

animesh-chouhan:
## What

Adds Google Analytics to the docs site, using the same measurement ID the apex already uses, so docs.copperhead.sh and copperhead.sh report into one property.

## Why

The docs site had no analytics at all. copperhead.sh has been measured since it launched ([Base.astro](https://github.com/copperheadhq/copperhead-site) in the `copperhead-site` repo), so every visit to the docs was invisible: there was no way to tell which pages people actually read, or how much traffic the docs get relative to the landing page.

## Design

Wired through Starlight's `head` config in [astro.config.mjs](docs/astro.config.mjs) rather than a component override, so it reaches every page without touching the theme components. The snippet is byte-for-byte the one the apex serves: the async `gtag/js` loader, then the `dataLayer` bootstrap and `gtag('config', ...)` call.

The measurement ID sits in a named constant next to `REPO`, with a comment pointing at the apex as the other place it lives. It is a public identifier that ships in the HTML of every page, so it is a literal here exactly as it is there, not a secret.

## Spec / OpenSpec

n/a, docs-site configuration only. No spec-level behavior changes.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | `npm run typecheck` | n/a (zero src lines changed) |
| Build | `npm run build` (in `docs/`) | pass, clean at 17 pages |
| Unit + offline | `npm test` | n/a (no src or test lines changed) |

- Sandbox variant used (`create` / `edit`): n/a, docs-site configuration with no CLI surface.
- Commands run and outcome: built the site and confirmed both tags reach every page, and that the inline snippet matches the apex's.

```text
$ cd docs && npm run build
[build] 17 page(s) built in 2.03s
[build] Complete!

$ # both tags present on every generated page
pages total: 17
with loader script: 17
with config call: 17

$ # as emitted, on the home page
<script async src="https://www.googletagmanager.com/gtag/js?id=G-KGEWCR918W"></script>
<script>window.dataLayer = window.dataLayer || [];
function gtag(){dataLayer.push(arguments);}
gtag('js', new Date());
gtag('config', 'G-KGEWCR918W');</script>
```

## Docs

No content changes; this is site configuration only.

---

## Invariant checklist

This diff changes one file, `docs/astro.config.mjs`, and zero lines under `src/`, so every gated path is unreachable from it.

- [x] **`check`/`verify` stays LLM-free and network-free**: N/A, no change to `src/commands/check` or anything it imports. The added network call is in the published site's HTML, not in the CLI.
- [x] **No sexp serialization**: N/A, no change under `src/kicad/`.
- [x] **Sync-obligations ledger**: N/A.
- [x] **Secrets**: N/A. A GA measurement ID is a public identifier that ships in every page's HTML, not a credential.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -4,6 +4,8 @@ import starlightLlmsTxt from 'starlight-llms-txt';
 import rehypeTableLabels from './src/plugins/rehype-table-labels.mjs';
 
 const REPO = 'https://github.com/copperheadhq/copperhead';
+// The same GA property as the apex, copperhead.sh (copperhead-site).
+const GA_MEASUREMENT_ID = 'G-KGEWCR918W';
 
 // Served at the root of its own subdomain, docs.copperhead.sh. The apex,
 // copperhead.sh, is a separate Cloudflare Worker (the copperhead-site repo),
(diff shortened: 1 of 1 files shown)


animesh-chouhan:
_CI note: the red `test` check on this PR is unrelated to the change._

This PR's entire diff is `docs/astro.config.mjs`, which vitest never loads. The failure is the same one every time:

```
FAIL test/mcp-server.test.ts > mutating tools are serialized per repo (spec: serialization)
     > rejects a concurrent mutating call with a typed busy error
AssertionError: one of two concurrent mutating calls must be refused as busy:
  expected [] to have a length of 1 but got +0
```

That is the per-repo mutating-call lock race that #282 is open to fix. Some evidence on what it takes to trigger:

- It failed twice in a row here, on two separate CI runs of the same commit.
- It also failed once on #286 (a docs-only PR) and passed on a re-run of the identical commit.
- Run in isolation on this branch it passes 5 out of 5, so it needs the full suite's contention to show. That fits a lock acquisition that is not atomic rather than anything in the test.

So this is a real defect in `src/mcp/`, surfacing at a rate that is now blocking unrelated PRs. Landing #282 first is the clean fix.

PR #295, opened 2026-09-10
Implement Phase 1/2 part research with credential-free JLCSearch

Devesh36:
## What

Add credential-free part research and a live, model-free part-availability audit from issue #246.

`copperhead audit <file>` reads a Markdown table of MPNs and reports three clear groups:

- **Available now**
- **Not available**
- **Needs review**

It uses the public JLCSearch API by default, requires no model or supplier credential, and does not modify `BOM.md` or `.copperhead/constraints.json`.

## Why

Part research should work for users without paid provider credentials. Engineers also need a fast answer to “which of these parts can I source now?” without starting an LLM agent run or mutating their design.

`copperhead audit demo.md` is deliberately separate from `copperhead check`: audit is explicit live network I/O, while `check` stays deterministic, LLM-free, network-free, and CI-safe.

## Design

- `JlcSearchProvider` calls the public JLCSearch endpoint and normalizes LCSC/JLC results.
- Audit input is a repository-relative Markdown table with required `MPN` and optional `Refdes` / `Required qty` columns.
- The audit requires an exact returned MPN. Missing exact matches, zero/insufficient stock, and EOL lifecycle fail the command; absent lifecycle or datasheet metadata appears under **Needs review**.
- All network traffic uses the existing allowlist, redirect validation, response limits, timeout, transcript logging, and secret redaction boundary.
- `--output <path>` writes an optional Markdown report; otherwise audit writes only its ignored run transcript.
- Existing JLCSearch, optional Nexar/Brave providers, datasource cache, citation validation, and offline sourceability validation remain intact.

## Example

```bash
copperhead audit demo.md --output demo.audit.md
```

```md
| Refdes | MPN | Required qty |
|---|---|---:|
| R1 | 0603WAF1001T5E | 10 |
```

## Testing

| Check | Status |
| --- | --- |
| `npm run typecheck` | pass |
| `npm run build` | pass |
| `npm test -- test/research.test.ts` | pass, 18/18 |
| `openspec validate add-part-research-tools` | pass |
| Docs build | pass |
| Live `copperhead audit parts.md` | pass; public JLCSearch HTTP 200 and exact MPN returned |
| Full `npm test` | pre-existing environment-specific Windows KiCad-path failures remain |
| Live LLM tests | not run; audit itself is model-free |

## Invariants

- [x] `edit_file` / `write_file` remain structurally OpenSpec-gated.
- [x] KiCad mutations still require ERC, and DRC for board changes.
- [x] `check` / `verify` remains network-free; audit is an explicit separate command.
- [x] Audit makes no BOM or constraint writes, so it creates no sourcing dual-write obligation.
- [x] Specs, OpenSpec artifacts, README, tests, and implementation were updated together.


Devesh36:
Screenshot 2026-09-11 at 10 53 00 PM

Devesh36:
@animesh-chouhan check this out

animesh-chouhan:
checking

animesh-chouhan:
🤖 **Automated `pr-review` pass** (Claude Code). Not a human review, and not a formal approve/request-changes — merge gating is unaffected. Review ran in an isolated throwaway `git worktree`; no branch, index, or tracked file in anyone's checkout was moved.

**Measured commit:** `814038b` (PR head) vs merge base `b24de54` (= current `main`, so the branch is up to date).

## Verdict: request changes

Genuinely good work — the egress boundary re-validates **every** redirect hop, the allowlist held up against `user@host`, case, trailing-dot, IDN and subdomain-suffix probes, `--output` is sandboxed through `resolveInRepo`, the tests are deterministic (they stub `globalThis.fetch` and restore it), and the new env-var value sweep in `redact.ts` genuinely closes the Brave/Nexar key gap. The `check` network-free invariant — the highest-risk item in this PR — is **clean** under a full hand-traced transitive import closure.

But two high findings are violations of this PR's **own delta spec**, which states at `openspec/changes/add-part-research-tools/specs/part-research/spec.md:95`:

> Research tools SHALL be read-only with respect to repo files except the datasheet cache, and fetched content SHALL NOT bypass spec gating, verification gates, or the obligations ledger.

`search_parts` and `fetch_datasheet` violate both halves. Two further high findings are plain correctness bugs on the documented path.

## Metrics

```text
lines: +1747 / -59 (net +1688) across 32 files  ·  src +1198/-9, test +321/-0, spec +169/-50, docs +59/-0
new vs net: 11 new files, 21 modified; 1198 new src lines (the net-new surface, ~69% of the diff)
tests: +321 test lines, 18 tests in 3 describe blocks; suite NOT MEASURED locally (no node runtime
       or node_modules on this machine); CI `test` job passed at head (2m35s, run 35067807916)
diff coverage: NOT MEASURED (no node/vitest available — could not run coverage instrumentation).
       Manual symbol/branch mapping used instead; see "Untested surface" findings below.
deps: none changed — package.json and package-lock.json untouched (verified by git diff)
ci: 4 pass / 0 fail / 0 pending (cla, label, template, test); required checks: 0 failing, 0 pending
mergeable: MERGEABLE · not a draft · no prior automated pr-review pass on this PR
```

Method notes: change size and new-vs-net from `git diff --numstat`/`--diff-filter=A` against the merge base. Suite and coverage could not be run at all locally, so they are reported unmeasured rather than estimated; CI is the only suite evidence and it is green.

## Invariant checks performed

| Invariant | Result |
|---|---|
| **`check`/`verify` LLM-free and network-free** | **PASS** — 13-module transitive closure of `src/commands/check.ts` hand-traced; the new `src/memory/sourceability.ts` imports only node builtins, a *type-only* `ResearchConfig`, `constraints.ts` and `bom-table.ts`. No path reaches `src/research/`. No dynamic `import()` escapes it. |
| **Verification-gated out** (ERC/DRC, `maxRepairCycles`, rollback) | **PASS** — `src/agent/loop.ts` changes are additive run-summary fields only; no path skips verification or ends a run early. |
| **No sexp serialization** | **PASS (by absence)** — no file under `src/kicad/` is in the diff. |
| **Secrets** | **PASS** — `sk-` pattern intact; Brave key travels as an `x-subscription-token` header, Nexar secret as a POST body, bearer as a header; `net.ts` logs no headers or bodies; `.gitignore` still has `.env` and `.copperhead/runs/`. |
| **Spec-gated in** | **FAIL** — finding 1. |
| **Sync-obligations ledger** | **FAIL** — finding 2. |
| **Spec coherence** | **PARTIAL** — findings 12 and 13. `openspec validate` NOT MEASURED (no node). |

---

## Findings

**[high] 1 · Research tools mutate `BOM.md` and `constraints.json` without the spec gate** — `src/capabilities/handlers.ts:73`, `src/capabilities/handlers.ts:98`

`search_parts` and `fetch_datasheet` are registered `requiresUnlock: false`, yet both write tracked repo files: `recordPartSelection` calls `saveConstraint(...)` and rewrites `docs/BOM.md` (`src/research/selection.ts:19,46`), and `recordDatasheetEvidence` does the same (`:68,84`). Every pre-existing capability that writes the same registry is gated — `record_constraint` is `requiresUnlock: true` (`src/capabilities/handlers.ts:671`), as are `record_decision` and `resolve_affected`. This is a structural gate bypass, and it contradicts the PR's own `proposal.md` and delta spec `:95`, and `design.md` D10.

*Failure scenario*: a `do` run on a repo with `research.enabled: true` never validates an OpenSpec proposal, so `ctx.editsUnlocked === false`. The model calls `search_parts {query:"1k 0603", mpn:"…", refdes:"R1"}`. `.copperhead/constraints.json` gains `sourcing.R1` and the `R1` row of `docs/BOM.md` is rewritten — then both land in the run's commit, with no proposal and no unlock ever having occurred.

*Fix*: set `requiresUnlock: true` on `search_parts` and `fetch_datasheet`, or gate only the writing branch (the `refdes`-present path and `recordDatasheetEvidence`) on `ctx.editsUnlocked` and keep the read-only search path ungated.

---

**[high] 2 · Research file writes bypass the obligations ledger** — `src/research/selection.ts:51`, `src/research/selection.ts:85`

`recordPartSelection` and `recordDatasheetEvidence` call `ctx.filesTouched.add(...)` directly instead of `markTouched` (`src/capabilities/helpers.ts:28-37`), so the `rel.endsWith('.md')` → `ctx.ledger.onDocEdit(rel)` branch never fires and **no drift obligation ever opens** for the `BOM.md` write. `recordPartSelection` then *clears* `constraint-dual-write` and `affects-revisit` (`:54-55`) for a key no hook opened, and writes the constraint via `saveConstraint` directly, bypassing the `ctx.ledger.onConstraintChange` hook that `record_constraint` uses (`handlers.ts:697`). `recordDatasheetEvidence` touches the ledger not at all. The commit gate only refuses while an obligation is *open*, so an obligation that never opens is a silently weakened gate.

*Failure scenario*: a run whose only doc mutation is `search_parts {refdes:"U1", …}` rewrites `docs/BOM.md`, then calls `finish`. `handlers.ts:836` finds `ctx.ledger.openObligations` empty, the run commits, and `check_drift` is never required against the changed BOM — exactly the silent desync the ledger exists to prevent. `tasks.md` 4.3 is checked `[x]` claiming this hook exists.

*Fix*: replace both `ctx.filesTouched.add(...)` pairs with `markTouched(ctx, rel)`, and route the constraint write so `onConstraintChange` fires; remove the unearned `ledger.clear(...)` calls.

---

**[high] 3 · HTTP error responses are treated as successful payloads** — `src/research/net.ts:88`

`request()` returns the response for every status except 429; neither it nor `requestJson` checks `response.ok`. Each provider then does `data.<field> ?? []`, so a JSON-bodied 4xx/5xx silently becomes "zero results" rather than an error.

*Failure scenario*: JLCSearch returns `503 {"error":"service unavailable"}` → `jlcsearch.ts:79` yields `[]` → `auditRow` returns `status:'failure'` with `issues:['exact MPN was not returned by the selected provider']` → `copperhead audit parts.md` lists **every** part under "Not available" and exits 1. A provider outage is indistinguishable from "none of my parts are sourceable", and the user may re-design around a phantom shortage.

*Fix*: in `request()`, throw an `EgressError` carrying `.status` for any non-2xx, as is already done for 429.

---

**[high] 4 · A non-matching MPN is silently recorded as the selected part** — `src/capabilities/handlers.ts:81`

`const selected = typeof args.mpn === 'string' ? results.find((p) => p.mpn === args.mpn) ?? results[0]! : results[0]!;` — when the caller names an MPN and the provider does not return that exact string, the code falls back to the **first search hit** and dual-writes it into `BOM.md` and `constraints.json`. The comparison is also case-sensitive and untrimmed, unlike `exactMpn` in `src/commands/audit.ts:90`, which trims and upper-cases. The audit command is strict about exact MPNs; the agent tool writing the actual design memory is not.

*Failure scenario*: `search_parts {query:"STM32F103", mpn:"stm32f103c8t6", refdes:"U1"}` against a provider returning `"STM32F103C8T6"` first and `"STM32F103RBT6"` second → `find` misses on case, `results[0]` is taken. With different ordering, `U1`'s BOM row and `sourcing.U1` are written with a part the caller never asked for, and the tool result reports `selected … for U1` as though it matched.

*Fix*: when `mpn` is supplied, normalise both sides with `.trim().toUpperCase()` and return an error instead of falling back to `results[0]`.

---

**[medium] 5 · Redirects re-send credentials, and HTTPS→HTTP downgrade is allowed** — `src/research/net.ts:62`

The hop loop passes the original `init` unchanged — `fetch(current, { ...init, signal: controller.signal, redirect: 'manual' })` — so the `authorization: Bearer …` header and the POST body follow a redirect to a *different* host. `checkUrl` (`:30`) also accepts `http:`. Standard `fetch` drops `Authorization` on a cross-origin redirect; this manual loop does not. The default allowlist contains wildcards (`*.mouser.com`, `*.digikey.com`, `*.espressif.com`), so "another allowlisted host" is a broad set.

*Failure scenario*: `api.nexar.com` answers the GraphQL POST with `302 Location: http://wmsc.lcsc.com/x`. The Nexar bearer token is re-sent, in cleartext, to an unrelated operator's host. On the token endpoint the same applies to the `client_id`/`client_secret` form body.

*Fix*: strip `authorization` and `cookie` (and drop the body on a 303) whenever the redirect target's origin differs from the previous hop's, and require `https:` unless the host is explicitly configured otherwise.

---

**[medium] 6 · The response size cap is applied only after the whole body is buffered** — `src/research/net.ts:76`

`const body = new Uint8Array(await response.arrayBuffer()); if (body.byteLength > maxBytes)` — the full body is read into memory *before* the `maxPdfMB` check, and there is no `content-length` pre-check. The declared 10 MB cap therefore does not bound memory; only the 30 s abort does.

*Failure scenario*: an allowlisted host serves a multi-GB body to `fetch_datasheet` → the process buffers all of it and OOMs, and the "exceeds configured size cap" error is never reached.

*Fix*: stream via `response.body`'s reader and abort once accumulated bytes exceed `maxBytes`; reject early on an over-cap `content-length`.

---

**[medium] 7 · `recordPartSelection` writes `BOM.md` by fixed column index while the rest of the repo resolves by header name** — `src/research/selection.ts:40`

It hardcodes `raw[3] = part.mpn` and `raw[4] = rationale`; `recordDatasheetEvidence` does the same at `:78-79`. By contrast `parseBomTable` (`src/memory/bom-table.ts:331-338`) deliberately resolves Refdes/Value/Footprint/MPN by header *name*, with a comment explaining that this is so a doc that reorders or drops an optional column is still read correctly "instead of silently shifting every cell". The writer breaks the invariant the reader was written to uphold.

*Failure scenario*: `docs/BOM.md` is hand-maintained as `| Refdes | MPN | Value | Footprint | Rationale |`. `search_parts {refdes:"U1", …}` writes the selected MPN into column index 3 — the *Footprint* cell — while the real MPN column keeps `UNVERIFIED`. The BOM is silently corrupted and `check`'s sourceability pass still sees `UNVERIFIED` for U1.

*Fix*: resolve the MPN and Rationale column indices from the header row, as `parseBomTable` does, and fail loudly if they are absent.

---

**[medium] 8 · `search_parts` strips the `UNVERIFIED` flag on mere supplier confirmation** — `src/research/selection.ts:40`

`raw[3] = part.mpn` overwrites the MPN cell, so a row that read `UNVERIFIED` becomes a concrete MPN with **no datasheet evidence** — only supplier stock/lifecycle metadata. `parseBomTable` then yields `flags: []` (neither `UNVERIFIED` nor `VERIFIED(datasheet)`), `sourceability.ts:70` stops filtering it out, and `create.ts:184`'s `!mpn.startsWith('UNVERIFIED')` treats it as a human-confirmed part. `openspec/changes/build-copperhead-phase-1/specs/safety-rails/spec.md:43` states: *"The agent SHALL never present an MPN as verified: any newly introduced part MUST carry a datasheet-verifiable justification in BOM.md and be flagged `UNVERIFIED` for human review."* `design.md` describes promotion as a two-step process (select → attach datasheet evidence); this collapses it to one.

*Failure scenario*: `search_parts {query:"1k 0603", refdes:"R1"}` flips R1 from `UNVERIFIED` to `0603WAF1001T5E` purely because LCSC has stock. No datasheet was ever fetched, yet every downstream consumer now treats R1 as a settled part.

*Fix*: keep the `UNVERIFIED` marker in the MPN cell (e.g. `UNVERIFIED: <mpn>`) until `recordDatasheetEvidence` promotes the row to `VERIFIED(datasheet)`.

---

**[medium] 9 · Wrong-typed provider JSON throws a raw `TypeError` instead of a handled error** — `src/research/jlcsearch.ts:79`

`(data.components ?? []).flatMap(...)` assumes `data` is a non-null object and `components` is an array; `requestJson` only guarantees the bytes parsed as JSON. `src/research/nexar.ts:19-22` has the same problem with `for (const seller of part.sellers ?? [])`. The per-field normalisers in this file are otherwise commendably defensive — it is only the top-level shape that is unchecked.

*Failure scenario*: body `null` → `TypeError: Cannot read properties of null (reading 'components')`; body `{"components":{"0":{…}}}` → `TypeError: data.components.flatMap is not a function`. In `audit` this aborts the whole run and surfaces as a bare `TypeError` with exit 1; in the agent loop it throws out of the tool handler rather than returning a tool error the model can recover from.

*Fix*: `const components = Array.isArray((data as {components?: unknown})?.components) ? … : []`, and guard `Array.isArray(part.sellers)` in `nexar.ts`.

---

**[medium] 10 · A corrupt `datasheets/index.json` breaks both `fetch_datasheet` and `check`** — `src/research/cache.ts:35`

`JSON.parse(await readFile(p,'utf8'))` is unguarded, as is the identical read in `src/memory/sourceability.ts:29`. `saveIndex` (`cache.ts:41`) writes in place with no tmp+rename, so an interrupted write leaves a truncated file; two concurrent `fetchDatasheet` calls also load-modify-save independently, so the last writer silently drops the other's entry.

*Failure scenario*: a run is killed mid-`saveIndex` → `index.json` is truncated → every later `fetch_datasheet` throws `SyntaxError: Unexpected end of JSON input`, and `copperhead check` — which now calls `checkSourceability` whenever any `sourcing.*` constraint exists (`src/commands/check.ts:127`) — fails in CI with the same unhandled `SyntaxError`. That turns a corrupt cache file into a broken CI gate.

*Fix*: wrap both parses in try/catch returning an empty index, and write via temp file + `rename`.

---

**[medium] 11 · The oversized-datasheet `not-cached` path is unreachable beyond one byte of overage** — `src/research/cache.ts:73`

`fetchDatasheet` requests with `Math.floor(maxBytes) + 1`, and the comment at `:71-72` says this exists so the cache can "write a traceable `not-cached` entry instead of losing the URL to the generic egress cap". But `net.ts:77` throws as soon as the body exceeds the cap it was passed, so the `not-cached` branch at `:78-83` is reachable only when the body is *exactly* `floor(maxBytes)+1` bytes. The test passes only because it uses a 2-byte body against a 1-byte cap (`test/research.test.ts:121-127`).

*Failure scenario*: `research.maxPdfMB: 10` and a 12 MB PDF → `request` throws, the handler returns an error, and `index.json` gains no entry. The URL is lost, contradicting delta spec scenario "Oversized file refused but traceable" (`spec.md:41-43`) and task 3.3, which is checked `[x]`.

*Fix*: have `requestBytes` report over-cap as a typed, catchable outcome (or pass `Infinity` and let `fetchDatasheet` enforce the cap itself) so the index entry is always written.

---

**[medium] 12 · The delta spec's gating requirement was not updated alongside `design.md` D2** — `openspec/changes/add-part-research-tools/specs/part-research/spec.md:6`

The PR rewrote D2 to provider-specific gating (credential-free JLCSearch default; `web_search` off unless Brave is selected) but left the delta spec's normative requirement and both its scenarios at the old "AND the selected providers' API keys are present in environment variables" wording. The artifacts did not move together, and the delta spec's WHEN/THEN no longer describes shipped behavior.

*Failure scenario*: with `{"research":{"enabled":true}}` and an empty environment, scenario "Tools absent without keys" (`:8-10`) asserts no research tool appears; `researchPartToolGate` (`src/research/config.ts:42`) returns `true` for the default `jlcsearch` provider and `researchDatasheetToolGate` (`:51`) returns `true` on `enabled` alone, so `search_parts` and `fetch_datasheet` **do** appear. Symmetrically, "Tools present when configured" (`:12-14`) asserts all three appear, but with the shipped default `searchProvider: 'none'`, `web_search` never does.

*Fix*: rewrite the requirement and both scenarios to match provider-specific gating.

---

**[medium] 13 · SPEC.md does not document `check --strict-sourcing` or the new `check` failure mode** — `openspec/specs/SPEC.md:293`

The PR adds `--strict-sourcing` (`src/cli.ts:180`) and makes `check` fail on sourceability errors (`src/commands/check.ts:127-139`, where `ok` now ANDs `sourceability?.ok`), and README documents the flag — but SPEC.md's `copperhead check` CLI block and AC-2 are unchanged: no flag, and no `sourceability` key in the `--json` contract that AC-2.4 promises has "stable keys". SPEC.md is the repo's declared source of truth for the CLI surface, and per CLAUDE.md the delta spec's scenarios should map 1:1 onto SPEC.md's acceptance criteria. The delta spec has a full requirement plus two scenarios for this (`spec.md:60-69`) with no SPEC.md counterpart.

*Failure scenario*: an existing repo has a `sourcing.*` constraint whose lifecycle is `eol` and no `research` block in config. `check.ts:127`'s `Object.keys(registry).some(k => k.startsWith('sourcing.'))` is true, `checkSourceability` emits a severity-`error` finding, and a previously-green CI step exits 1 — a failure mode no line of SPEC.md describes.

*Fix*: add `--strict-sourcing` to SPEC.md's `check` block, document the `sourceability` JSON key, and add the matching acceptance criterion.

---

**[low] 14 · `MISSING_MPN` is suppressed by an unintended `else` rebinding, and the new regex has no left boundary** — `src/memory/bom-table.ts:348`

The new `if (row.cells.some((c) => /VERIFIED\(datasheet\)/i.test(c)))` was inserted *between* `if (mpn === 'UNVERIFIED')` and its `else if (!mpn) flags.push('MISSING_MPN')`, so the `else` now chains to the new condition. Separately, `/VERIFIED\(datasheet\)/i` matches inside the string `UNVERIFIED(datasheet)`, so a cell meaning the opposite would be flagged as verified — and per `openspec/specs/SPEC.md:474`, `VERIFIED(datasheet)` is a load-bearing evidence claim. The doc comment at `:306` still describes the old two-flag contract. Impact is limited today: `MISSING_MPN` is only consumed for display in `repl inspect` and by tests.

*Failure scenario*: a BOM row with an empty MPN cell whose Rationale carries a stale `VERIFIED(datasheet) …` note parses as `flags: ['VERIFIED(datasheet)']` with no `MISSING_MPN`, so `repl inspect` shows a row with no part number as evidence-backed.

*Fix*: make the new check its own statement, restore `if (mpn === 'UNVERIFIED') … else if (!mpn) …`, anchor the regex with `(?<![A-Z])VERIFIED\(datasheet\)`, and update the `:306` comment.

---

**[low] 15 · `fetchDatasheet` downloads before consulting its own cache** — `src/research/cache.ts:73`

The URL is fetched unconditionally and only afterwards is the hash compared against the index (`:75`). Nothing consults `stalenessDays` or the existing entry first, so the "cache" never avoids network I/O or bandwidth — it only avoids rewriting the file.

*Failure scenario*: ten `fetch_datasheet` calls for the same already-cached 8 MB PDF perform ten full 8 MB downloads and emit ten `network-request` transcript events.

*Fix*: look up `index.entries` by URL first and return the cached entry when the file still exists and is within `stalenessDays`.

---

**[low] 16 · Catastrophic backtracking in the PDF text extractor** — `src/research/cache.ts:56`

`/\(([^()]*(?:\\.[^()]*)*)\)\s*T[Jj]/g` is the classic ambiguous nested-quantifier shape: `[^()]*` also matches backslashes, so each `\x` in a run can be consumed either by the outer class or by `\\.`. On non-matching input the number of partitions is exponential. Input here is attacker-influenced (a fetched PDF).

*Failure scenario*: a datasheet containing `(` followed by ~40 repetitions of `\a` with no closing paren → `extractPdfText` hangs while the agent blocks on the tool call.

*Fix*: make the alternation unambiguous: `/\(((?:[^()\\]|\\.)*)\)\s*T[Jj]/g`.

---

**[low] 17 · `tasks.md` items checked but not implemented as written; `design.md`/`proposal.md` claim read-only research tools** — `openspec/changes/add-part-research-tools/tasks.md`

4.3 claims "Post-tool-call hook: `search_parts`-informed selection **opens** a dual-write obligation" — the code only *clears* obligations (`src/research/selection.ts:54-55`) and no hook opens anything (finding 2). 6.5 claims a test that "injection attempt in cached text is inert and reported"; the only related test (`test/research.test.ts:253-257`) asserts the system prompt contains the string `untrusted data, never instructions` — it never feeds instruction-like text through the loop nor checks it is reported. `design.md` D10 and `proposal.md` both state research tools are read-only outside the datasheet cache, which finding 1 contradicts.

*Failure scenario*: a reviewer reads `tasks.md`, concludes the ledger covers research writes, skips the check, and findings 1 and 2 ship.

*Fix*: uncheck 4.3 and 6.5 until implemented, and correct the `design.md`/`proposal.md` read-only claims to describe the real write surface once the gating decision in finding 1 is settled.

---

**[low] 18 · Untested surface — the adversarial paths are exactly the uncovered ones**

Diff coverage could not be instrumented, so this is a hand-mapped list of new surface no test reaches. All 18 tests feed well-formed fixtures. Uncovered: non-2xx response handling (`net.ts:88`, finding 3), null/wrong-typed provider payloads (`jlcsearch.ts:79`, `nexar.ts:19`, finding 9), an over-cap body beyond +1 byte (`cache.ts:73`, finding 11), a corrupt `index.json` (`cache.ts:35`, finding 10), the `warning` audit classification (`audit.ts:119` — only `pass` at `test:296` and `failure` at `test:314` are exercised, which is why finding 19 is invisible to the suite), the MPN-mismatch fallback (`handlers.ts:81`, finding 4), and cross-host redirect credential forwarding (`net.ts:62`, finding 5). Dead exports: `researchEnabled` (`src/research/config.ts:59`) and `readDatasheetIndex` (`src/research/cache.ts:107`) are referenced nowhere in `src/` or `test/`. The live JLCSearch shape is asserted only against a hand-written fixture (`test:186`), so the real API's field names (`mfr`, `extra.mpn`, `price1`) are unverified by the suite.

*Failure scenario*: every finding above numbered 3, 4, 9, 10, 11 and 19 passes CI green today.

*Fix*: add regression tests for each, starting with `runPartAudit` against `new Response('{"error":"down"}', {status:503})` asserting an `EgressError` rather than "Not available".

---

**[low] 19 · "Available now" and "Needs review" are not disjoint** — `src/commands/audit.ts:138`

`available = findings.filter(f => f.status !== 'failure')` includes warnings, while `review = findings.filter(f => f.status === 'warning')` lists the same rows again. The PR body and README both promise "three clear groups".

*Failure scenario*: a part with stock but no provider-supplied lifecycle gets `status:'warning'` and is printed under "Available now — 25 in stock" *and* under "Needs review — lifecycle was not supplied by the provider".

*Fix*: `const available = result.findings.filter((f) => f.status === 'pass');`

---

**[low] 20 · `researchSummary()` is used at one of four identical call sites** — `src/agent/loop.ts:86`

The helper is defined and used at `:424`, but the failure, refusal, dry-run and commit paths at `:675`, `:726` and `:809` inline the same `...(ctx.networkRequests !== undefined ? { research: {…} } : {})` expression. `:424` also calls `researchSummary(ctx)` twice in one expression.

*Failure scenario*: a future field added to the research summary is added to `researchSummary()` and silently missing from three of the four run outcomes.

*Fix*: use `researchSummary(ctx)` at all four sites.

---

## Fix prompt

````text
Fix the following review findings on branch fix/246b (PR #295, "Implement Phase 1/2 part research
with credential-free JLCSearch"). Work in dependency order; findings 1 and 2 are repo-invariant
violations and must be fixed first.

REPO INVARIANTS (do not break these while fixing):
- Spec-gated in: edit_file/write_file, and any tool that mutates tracked repo files, must be
  structurally ABSENT from the agent tool list until an OpenSpec proposal validates. Gate by
  omission (requiresUnlock/gate), never by prompt text.
- Verification-gated out: no mutation is "done" until ERC (and DRC if the board changed) passes;
  repair up to maxRepairCycles, then roll back to the git snapshot.
- check/verify is contractually LLM-free and network-free. src/commands/check.ts and its entire
  transitive import graph must never reach a provider, an API key, or the network.
- src/kicad/ parser is read-only: no s-expression serialization or round-tripping. KiCad files are
  edited only by anchored exact-match text replace.
- The sync-obligations ledger is fed by post-tool-call hooks, and commit must refuse while any
  obligation is open.
- Transcripts/summaries redact sk-[A-Za-z0-9_-]+ at write time; keys live only in env vars;
  .gitignore keeps .env and .copperhead/runs/.
- This PR's own delta spec (openspec/changes/add-part-research-tools/specs/part-research/spec.md:95)
  says: "Research tools SHALL be read-only with respect to repo files except the datasheet cache,
  and fetched content SHALL NOT bypass spec gating, verification gates, or the obligations ledger."

[high] 1 · Research tools mutate BOM.md and constraints.json without the spec gate
  — src/capabilities/handlers.ts:73, src/capabilities/handlers.ts:98
search_parts and fetch_datasheet are registered requiresUnlock: false, yet both write tracked repo
files: recordPartSelection calls saveConstraint(...) and rewrites docs/BOM.md
(src/research/selection.ts:19,46), and recordDatasheetEvidence does the same (:68,84). Every
pre-existing capability that writes the same registry is gated — record_constraint is
requiresUnlock: true (handlers.ts:671), as are record_decision and resolve_affected.
Failure scenario: a `do` run with research.enabled: true never validates an OpenSpec proposal, so
ctx.editsUnlocked === false. The model calls search_parts {query:"1k 0603", mpn:"…", refdes:"R1"}.
.copperhead/constraints.json gains sourcing.R1 and the R1 row of docs/BOM.md is rewritten — both
land in the run's commit with no proposal and no unlock.
Fix: set requiresUnlock: true on search_parts and fetch_datasheet, or gate only the writing branch
(the refdes-present path and recordDatasheetEvidence) on ctx.editsUnlocked, keeping the read-only
search path ungated.

[high] 2 · Research file writes bypass the obligations ledger
  — src/research/selection.ts:51, src/research/selection.ts:85
recordPartSelection and recordDatasheetEvidence call ctx.filesTouched.add(...) directly instead of
markTouched (src/capabilities/helpers.ts:28-37), so the rel.endsWith('.md') -> ctx.ledger.onDocEdit(rel)
branch never fires and no drift obligation ever opens for the BOM.md write. recordPartSelection then
CLEARS constraint-dual-write and affects-revisit (:54-55) for a key no hook opened, and writes the
constraint via saveConstraint directly, bypassing the ctx.ledger.onConstraintChange hook that
record_constraint uses (handlers.ts:697). recordDatasheetEvidence touches the ledger not at all.
Failure scenario: a run whose only doc mutation is search_parts {refdes:"U1"} rewrites docs/BOM.md,
then calls finish. handlers.ts:836 finds ctx.ledger.openObligations empty, the run commits, and
check_drift is never required against the changed BOM.
Fix: replace both ctx.filesTouched.add(...) pairs with markTouched(ctx, rel), route the constraint
write so onConstraintChange fires, and remove the unearned ledger.clear(...) calls.

[high] 3 · HTTP error responses are treated as successful payloads — src/research/net.ts:88
request() returns the response for every status except 429; neither it nor requestJson checks
response.ok. Each provider then does data.<field> ?? [], so a JSON-bodied 4xx/5xx silently becomes
"zero results".
Failure scenario: JLCSearch returns 503 {"error":"service unavailable"} -> jlcsearch.ts:79 yields []
-> auditRow returns status:'failure' with issues:['exact MPN was not returned by the selected
provider'] -> `copperhead audit parts.md` lists EVERY part under "Not available" and exits 1. A
provider outage is indistinguishable from a real sourcing shortage.
Fix: in request(), throw an EgressError carrying .status for any non-2xx, as already done for 429.
Add a regression test: runPartAudit against new Response('{"error":"down"}', {status:503}) must
surface an EgressError, not report "Not available".

[high] 4 · A non-matching MPN is silently recorded as the selected part
  — src/capabilities/handlers.ts:81
`const selected = typeof args.mpn === 'string' ? results.find((p) => p.mpn === args.mpn) ?? results[0]!
: results[0]!;` — when the caller names an MPN and the provider does not return that exact string,
the code falls back to the first search hit and dual-writes it into BOM.md and constraints.json. The
comparison is case-sensitive and untrimmed, unlike exactMpn in src/commands/audit.ts:90.
Failure scenario: search_parts {query:"STM32F103", mpn:"stm32f103c8t6", refdes:"U1"} against a
provider returning "STM32F103C8T6" first and "STM32F103RBT6" second -> find misses on case,
results[0] is taken. With different ordering, U1's BOM row and sourcing.U1 are written with a part
the caller never asked for, while the tool result says `selected … for U1`.
Fix: when mpn is supplied, normalise both sides with .trim().toUpperCase() and return an error
instead of falling back to results[0].

[medium] 5 · Redirects re-send credentials, and HTTPS->HTTP downgrade is allowed
  — src/research/net.ts:62
The hop loop passes the original init unchanged — fetch(current, { ...init, signal: controller.signal,
redirect: 'manual' }) — so the `authorization: Bearer …` header and the POST body follow a redirect to
a DIFFERENT host. checkUrl (:30) also accepts http:. Standard fetch drops Authorization on a
cross-origin redirect; this manual loop does not. The default allowlist contains wildcards
(*.mouser.com, *.digikey.com, *.espressif.com).
Failure scenario: api.nexar.com answers the GraphQL POST with 302 Location: http://wmsc.lcsc.com/x.
The Nexar bearer token is re-sent, in cleartext, to an unrelated operator's host. On the token
endpoint the same applies to the client_id/client_secret form body.
Fix: strip authorization and cookie (and drop the body on a 303) whenever the redirect target's
origin differs from the previous hop's, and require https: unless the host is explicitly configured
otherwise.

[medium] 6 · The response size cap is applied only after the whole body is buffered
  — src/research/net.ts:76
`const body = new Uint8Array(await response.arrayBuffer()); if (body.byteLength > maxBytes)` — the
full body is read into memory BEFORE the maxPdfMB check, with no content-length pre-check. The
declared 10 MB cap does not bound memory; only the 30 s abort does.
Failure scenario: an allowlisted host serves a multi-GB body to fetch_datasheet -> the process
buffers all of it and OOMs, and the "exceeds configured size cap" error is never reached.
Fix: stream via response.body's reader and abort once accumulated bytes exceed maxBytes; reject
early on an over-cap content-length.

[medium] 7 · recordPartSelection writes BOM.md by fixed column index while the rest of the repo
  resolves by header name — src/research/selection.ts:40
It hardcodes raw[3] = part.mpn and raw[4] = rationale; recordDatasheetEvidence does the same at
:78-79. parseBomTable (src/memory/bom-table.ts:331-338) deliberately resolves Refdes/Value/Footprint/
MPN by header NAME, with a comment explaining this is so a doc that reorders or drops an optional
column is still read correctly "instead of silently shifting every cell".
Failure scenario: docs/BOM.md is hand-maintained as | Refdes | MPN | Value | Footprint | Rationale |.
search_parts {refdes:"U1"} writes the selected MPN into column index 3 — the Footprint cell — while
the real MPN column keeps UNVERIFIED. The BOM is silently corrupted and check's sourceability pass
still sees UNVERIFIED for U1.
Fix: resolve the MPN and Rationale column indices from the header row as parseBomTable does, and
fail loudly if they are absent.

[medium] 8 · search_parts strips the UNVERIFIED flag on mere supplier confirmation
  — src/research/selection.ts:40
raw[3] = part.mpn overwrites the MPN cell, so a row that read UNVERIFIED becomes a concrete MPN with
NO datasheet evidence — only supplier stock/lifecycle metadata. parseBomTable then yields flags: []
(neither UNVERIFIED nor VERIFIED(datasheet)), sourceability.ts:70 stops filtering it out, and
create.ts:184's !mpn.startsWith('UNVERIFIED') treats it as a human-confirmed part.
openspec/changes/build-copperhead-phase-1/specs/safety-rails/spec.md:43 states: "The agent SHALL
never present an MPN as verified: any newly introduced part MUST carry a datasheet-verifiable
justification in BOM.md and be flagged UNVERIFIED for human review." design.md describes promotion
as two steps (select -> attach datasheet evidence); this collapses it to one.
Failure scenario: search_parts {query:"1k 0603", refdes:"R1"} flips R1 from UNVERIFIED to
0603WAF1001T5E purely because LCSC has stock. No datasheet was ever fetched, yet every downstream
consumer now treats R1 as settled.
Fix: keep the UNVERIFIED marker in the MPN cell (e.g. `UNVERIFIED: <mpn>`) until
recordDatasheetEvidence promotes the row to VERIFIED(datasheet).

[medium] 9 · Wrong-typed provider JSON throws a raw TypeError instead of a handled error
  — src/research/jlcsearch.ts:79
(data.components ?? []).flatMap(...) assumes data is a non-null object and components is an array;
requestJson only guarantees the bytes parsed as JSON. src/research/nexar.ts:19-22 has the same
problem with `for (const seller of part.sellers ?? [])`.
Failure scenario: body null -> TypeError: Cannot read properties of null (reading 'components');
body {"components":{"0":{…}}} -> TypeError: data.components.flatMap is not a function. In audit this
aborts the run as a bare TypeError with exit 1; in the agent loop it throws out of the tool handler
rather than returning a recoverable tool error.
Fix: const components = Array.isArray((data as {components?: unknown})?.components) ? … : [], and
guard Array.isArray(part.sellers) in nexar.ts.

[medium] 10 · A corrupt datasheets/index.json breaks both fetch_datasheet and check
  — src/research/cache.ts:35
JSON.parse(await readFile(p,'utf8')) is unguarded, as is the identical read in
src/memory/sourceability.ts:29. saveIndex (cache.ts:41) writes in place with no tmp+rename, so an
interrupted write leaves a truncated file; two concurrent fetchDatasheet calls also load-modify-save
independently, so the last writer silently drops the other's entry.
Failure scenario: a run is killed mid-saveIndex -> index.json is truncated -> every later
fetch_datasheet throws SyntaxError: Unexpected end of JSON input, and `copperhead check` — which now
calls checkSourceability whenever any sourcing.* constraint exists (src/commands/check.ts:127) —
fails in CI with the same unhandled SyntaxError.
Fix: wrap both parses in try/catch returning an empty index, and write via temp file + rename.

[medium] 11 · The oversized-datasheet not-cached path is unreachable beyond one byte of overage
  — src/research/cache.ts:73
fetchDatasheet requests with Math.floor(maxBytes) + 1, and the comment at :71-72 says this exists so
the cache can "write a traceable not-cached entry instead of losing the URL to the generic egress
cap". But net.ts:77 throws as soon as the body exceeds the cap it was passed, so the not-cached
branch at :78-83 is reachable only when the body is EXACTLY floor(maxBytes)+1 bytes. The test passes
only because it uses a 2-byte body against a 1-byte cap (test/research.test.ts:121-127).
Failure scenario: research.maxPdfMB: 10 and a 12 MB PDF -> request throws, the handler returns an
error, and index.json gains no entry. The URL is lost, contradicting delta spec scenario "Oversized
file refused but traceable" (spec.md:41-43) and task 3.3, which is checked [x].
Fix: have requestBytes report over-cap as a typed, catchable outcome (or pass Infinity and let
fetchDatasheet enforce the cap itself) so the index entry is always written.

[medium] 12 · The delta spec's gating requirement was not updated alongside design.md D2
  — openspec/changes/add-part-research-tools/specs/part-research/spec.md:6
The PR rewrote D2 to provider-specific gating (credential-free JLCSearch default; web_search off
unless Brave is selected) but left the delta spec's normative requirement and both its scenarios at
the old "AND the selected providers' API keys are present in environment variables" wording.
Failure scenario: with {"research":{"enabled":true}} and an empty environment, scenario "Tools absent
without keys" (:8-10) asserts no research tool appears; researchPartToolGate (src/research/config.ts:42)
returns true for the default jlcsearch provider and researchDatasheetToolGate (:51) returns true on
enabled alone, so search_parts and fetch_datasheet DO appear. Symmetrically, "Tools present when
configured" (:12-14) asserts all three appear, but with the shipped default searchProvider: 'none',
web_search never does.
Fix: rewrite the requirement and both scenarios to match provider-specific gating.
Then run: openspec validate add-part-research-tools

[medium] 13 · SPEC.md does not document check --strict-sourcing or the new check failure mode
  — openspec/specs/SPEC.md:293
The PR adds --strict-sourcing (src/cli.ts:180) and makes check fail on sourceability errors
(src/commands/check.ts:127-139, where ok now ANDs sourceability?.ok), and README documents the flag —
but SPEC.md's `copperhead check` CLI block and AC-2 are unchanged: no flag, and no sourceability key
in the --json contract that AC-2.4 promises has "stable keys". Per CLAUDE.md the delta spec's
scenarios should map 1:1 onto SPEC.md's acceptance criteria; the delta spec has a full requirement
plus two scenarios for this (spec.md:60-69) with no SPEC.md counterpart.
Failure scenario: an existing repo has a sourcing.* constraint whose lifecycle is eol and no research
block in config. check.ts:127's Object.keys(registry).some(k => k.startsWith('sourcing.')) is true,
checkSourceability emits a severity-error finding, and a previously-green CI step exits 1 — a failure
mode no line of SPEC.md describes.
Fix: add --strict-sourcing to SPEC.md's check block, document the sourceability JSON key, and add the
matching acceptance criterion.

[low] 14 · MISSING_MPN is suppressed by an unintended else rebinding, and the new regex has no left
  boundary — src/memory/bom-table.ts:348
The new `if (row.cells.some((c) => /VERIFIED\(datasheet\)/i.test(c)))` was inserted BETWEEN
`if (mpn === 'UNVERIFIED')` and its `else if (!mpn) flags.push('MISSING_MPN')`, so the else now chains
to the new condition. Separately, /VERIFIED\(datasheet\)/i matches inside the string
"UNVERIFIED(datasheet)", so a cell meaning the opposite would be flagged as verified — and per
openspec/specs/SPEC.md:474, VERIFIED(datasheet) is a load-bearing evidence claim. The doc comment at
:306 still describes the old two-flag contract.
Failure scenario: a BOM row with an empty MPN cell whose Rationale carries a stale
"VERIFIED(datasheet) …" note parses as flags: ['VERIFIED(datasheet)'] with no MISSING_MPN, so
`repl inspect` shows a row with no part number as evidence-backed.
Fix: make the new check its own statement, restore `if (mpn === 'UNVERIFIED') … else if (!mpn) …`,
anchor the regex with (?<![A-Z])VERIFIED\(datasheet\), and update the :306 comment.

[low] 15 · fetchDatasheet downloads before consulting its own cache — src/research/cache.ts:73
The URL is fetched unconditionally and only afterwards is the hash compared against the index (:75).
Nothing consults stalenessDays or the existing entry first, so the "cache" never avoids network I/O —
it only avoids rewriting the file.
Failure scenario: ten fetch_datasheet calls for the same already-cached 8 MB PDF perform ten full
8 MB downloads and emit ten network-request transcript events.
Fix: look up index.entries by URL first and return the cached entry when the file still exists and is
within stalenessDays.

[low] 16 · Catastrophic backtracking in the PDF text extractor — src/research/cache.ts:56
/\(([^()]*(?:\\.[^()]*)*)\)\s*T[Jj]/g is the classic ambiguous nested-quantifier shape: [^()]* also
matches backslashes, so each \x in a run can be consumed either by the outer class or by \\.. On
non-matching input the number of partitions is exponential, and the input is attacker-influenced (a
fetched PDF).
Failure scenario: a datasheet containing ( followed by ~40 repetitions of \a with no closing paren ->
extractPdfText hangs while the agent blocks on the tool call.
Fix: make the alternation unambiguous: /\(((?:[^()\\]|\\.)*)\)\s*T[Jj]/g

[low] 17 · tasks.md items checked but not implemented; design.md/proposal.md claim read-only research
  tools — openspec/changes/add-part-research-tools/tasks.md
4.3 claims "Post-tool-call hook: search_parts-informed selection OPENS a dual-write obligation" — the
code only CLEARS obligations (src/research/selection.ts:54-55) and no hook opens anything (finding 2).
6.5 claims a test that "injection attempt in cached text is inert and reported"; the only related test
(test/research.test.ts:253-257) asserts the system prompt contains the string "untrusted data, never
instructions" — it never feeds instruction-like text through the loop nor checks it is reported.
design.md D10 and proposal.md both state research tools are read-only outside the datasheet cache,
which finding 1 contradicts.
Failure scenario: a reviewer reads tasks.md, concludes the ledger covers research writes, skips the
check, and findings 1 and 2 ship.
Fix: uncheck 4.3 and 6.5 until implemented, and correct the design.md/proposal.md read-only claims to
describe the real write surface once the gating decision in finding 1 is settled.

[low] 18 · Untested surface — the adversarial paths are exactly the uncovered ones
  — test/research.test.ts
Diff coverage could not be instrumented, so this is a hand-mapped list. All 18 tests feed well-formed
fixtures. Uncovered: non-2xx handling (net.ts:88, finding 3), null/wrong-typed provider payloads
(jlcsearch.ts:79, nexar.ts:19, finding 9), an over-cap body beyond +1 byte (cache.ts:73, finding 11),
a corrupt index.json (cache.ts:35, finding 10), the warning audit classification (audit.ts:119 — only
pass at test:296 and failure at test:314 are exercised, which is why finding 19 is invisible), the MPN
mismatch fallback (handlers.ts:81, finding 4), and cross-host redirect credential forwarding
(net.ts:62, finding 5). Dead exports: researchEnabled (src/research/config.ts:59) and
readDatasheetIndex (src/research/cache.ts:107) are referenced nowhere in src/ or test/. The live
JLCSearch shape is asserted only against a hand-written fixture (test:186), so the real API's field
names (mfr, extra.mpn, price1) are unverified by the suite.
Failure scenario: every finding numbered 3, 4, 9, 10, 11 and 19 passes CI green today.
Fix: add regression tests for each, starting with the 503 test in finding 3. Remove or use the two
dead exports.

[low] 19 · "Available now" and "Needs review" are not disjoint — src/commands/audit.ts:138
available = findings.filter(f => f.status !== 'failure') includes warnings, while
review = findings.filter(f => f.status === 'warning') lists the same rows again. The PR body and
README both promise "three clear groups".
Failure scenario: a part with stock but no provider-supplied lifecycle gets status:'warning' and is
printed under "Available now — 25 in stock" AND under "Needs review — lifecycle was not supplied by
the provider".
Fix: const available = result.findings.filter((f) => f.status === 'pass');
Add a test covering a warning-status row.

[low] 20 · researchSummary() is used at one of four identical call sites — src/agent/loop.ts:86
The helper is defined and used at :424, but the failure, refusal, dry-run and commit paths at :675,
:726 and :809 inline the same ...(ctx.networkRequests !== undefined ? { research: {…} } : {})
expression. :424 also calls researchSummary(ctx) twice in one expression.
Failure scenario: a future field added to the research summary is added to researchSummary() and
silently missing from three of the four run outcomes.
Fix: use researchSummary(ctx) at all four sites.

VERIFICATION (run all of these after the fixes):
  npm run typecheck
  npm run build
  npm test
  openspec validate add-part-research-tools
Confirm specifically that:
  - `copperhead check` and everything it imports still reach no provider, API key, or network
    (src/memory/sourceability.ts must keep importing only node builtins, a type-only ResearchConfig,
    constraints.ts and bom-table.ts).
  - No tool that writes a tracked repo file is reachable while ctx.editsUnlocked is false.
  - A BOM.md write from research opens a drift obligation, and commit refuses while it is open.
````

animesh-chouhan:
apart from this, how are you planning to integrate with the full workflow

Devesh36:
@animesh-chouhan thanks for the detailed review — I addressed the findings in `ee075d4`.

Key changes:

- `search_parts` remains usable as a read-only query before edit unlock, but selecting with `refdes` is rejected until an OpenSpec change validates. `fetch_datasheet` can populate only the evidence cache before unlock; attaching evidence to BOM/constraints is gated the same way.
- Research writes now go through the obligations hooks: BOM edits open drift, constraint changes open dual-write/affected-artifact obligations, and the dual-write obligation clears only after both BOM.md and constraints.json are updated. Affected-artifact obligations remain open for explicit reconciliation.
- Exact MPN selection is trimmed/case-normalized and never falls back to the first result. Supplier confirmation leaves the BOM as `UNVERIFIED: <mpn>` until cached datasheet evidence is attached.
- Egress is HTTPS-only, rejects non-2xx responses with status-bearing errors, strips credentials on cross-origin redirects, refuses unsafe cross-origin bodies, and streams through the response cap instead of buffering unbounded bodies.
- Datasheet index reads tolerate corruption, writes use temp+rename, concurrent in-process updates are serialized, oversized files always get traceable `not-cached` entries, and fresh cached URLs avoid repeat downloads.
- BOM writes resolve Refdes/MPN/Rationale by header name; provider payloads are shape-checked; audit warning/pass groups are disjoint; the PDF extractor regex is linear; and the run-summary helper is used on every exit path.
- The delta spec, normative SPEC.md, tasks, and README now match provider-specific gating, `check --strict-sourcing`, and the actual write surface. The unimplemented live prompt-injection scenario is no longer marked complete.

For the full workflow integration: `do` / `create` use supplier search during planning, then after the proposal validates an exact selection dual-writes an `UNVERIFIED` BOM row and sourcing snapshot. Datasheet fetch creates reproducible cached evidence, attaching a cited section promotes the row to `VERIFIED(datasheet)`, and the normal drift/affected-artifact plus ERC/DRC gates still control completion. Later `copperhead check` validates the saved sourcing and citations fully offline.

Verification completed: typecheck, 32 focused research tests, package build, docs build, `git diff --check`, and `openspec validate add-part-research-tools`. Live LLM acceptance tests were not run. The full local suite is still affected by the fixture under KiCad 10.0.5, which reports two existing `lib_symbol_mismatch` ERC warnings for R1/R2.

PR #298, opened 2026-09-13
fix(agent): time out turns on inactivity, not total length

animesh-chouhan:
## What

Provider turns are no longer killed just for running past 10 minutes. `turnTimeoutMs` (still 10 minutes) now measures how long a turn goes without any progress, and a new `turnMaxMs` (default 60 minutes) caps a turn that keeps producing output. The claude-code provider now reports progress while it streams, so a long turn that keeps producing output completes. A second commit fixes the Codex provider's teardown, which left a retried turn without its conversation history. The branch also carries the two `test/mcp-server.test.ts` commits from #299 (already merged to `main`), cherry-picked here so CI stays green; they merge cleanly because the file is identical on both sides.

## Why

A turn had a fixed 10-minute deadline, and a timed-out turn was resent unchanged. A turn that legitimately needs longer (a large claude-code capture turn) was killed at 10, 20, 30 and 40 minutes, then the run failed as `provider-error` and rolled back. In `create`, the stage diagnosis saw only "provider-error" and its prompt called timeouts transient, so it retried the stage, which changed every response-cache key and hit the same wall again (worst case over two hours per stage). The heartbeat could not tell slow from hung: claude-code emitted nothing until its reply was complete.

## Design

- **Watchdog** ([recovery.ts](src/agent/recovery.ts)): `withWatchdog(fn, { idleMs, maxMs, onTimeout })` hands `fn` an `activity()` callback that restarts the idle deadline. `maxMs` is a hard cap that activity does not extend, with two rules: it only trips once `fn` has reported progress (progress that first arrives after the cap came due trips it at once), and it is never shorter than `idleMs`. `TurnTimeoutError` carries `kind: 'idle' | 'max'`. `withTimeout` is a thin wrapper over it, so its other callers behave as before. The timer rejects before running `onTimeout`, so a provider whose `close()` rejects its call synchronously still surfaces `TurnTimeoutError`.
- **Compatibility**: a turn that reports no progress (openai, anthropic, vertex, cursor, codex) never restarts the deadline and is never capped, so it gets exactly the old whole-turn `turnTimeoutMs` deadline, including "no deadline" when `turnTimeoutMs <= 0` (`turnMaxMs` then defaults to disabled). A streaming turn's cap is at least `turnTimeoutMs`. No turn that finished in time before times out now, and a hung call is still caught within `turnTimeoutMs`.
[...]

PR #299, opened 2026-09-13
test(mcp): make the concurrent-init busy test deterministic

animesh-chouhan:
## What

Makes `rejects a concurrent mutating call with a typed busy error` in [test/mcp-server.test.ts](test/mcp-server.test.ts) deterministic. Test-only: no production code changes.

## Why

The test fails intermittently on `main` and has turned CI red on unrelated PRs: #298 ([run](https://github.com/copperheadhq/copperhead/actions/runs/34763632865)) and `fix/296-locked-write-file-silent-drop` ([run](https://github.com/copperheadhq/copperhead/actions/runs/34583101130), green on rerun). Locally it failed in 5 of 16 runs on unmodified `main`.

## Design

**Root cause.** The test fires two `copperhead_init` calls together and expects exactly one to be refused as busy, but nothing guarantees the calls overlap. The handler awaits `requireKicad()`, a `kicad-cli` subprocess probe, before it calls `locks.tryAcquire` ([server.ts](src/mcp/server.ts)), and `init` itself is quick. When one call's probe is slow, the other call acquires the lock, writes the docs and releases it before the slow call arrives. Envelopes logged from failing runs show exactly that: `[{ ok: true, "init wrote 0 file(s)" }, { ok: true, "init wrote 8 file(s)" }]`, so the calls ran one after the other.

**The lock is sound.** `RepoLocks.tryAcquire` is a synchronous check-and-set, so two calls can never both hold it. Nothing in the server changes.

**Fix.** The test spies on `RepoLocks.prototype.release` and holds every release until both calls have answered, then applies them. Whichever call acquires first keeps the lock for as long as the other can still reach `tryAcquire`, so exactly one call is refused whatever the probe timing. The test still drives two real concurrent calls through the real server and handler. It also asserts that only the holder releases (the refused call returns before it takes the lock) and that a follow-up `copperhead_init` succeeds once the held release is applied, so a handler that never releases is caught as well. The spy is restored in a `finally`.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | **938 passed, 0 failed, 21 skipped** (66 files passed, 2 skipped) |
| Live-LLM (opt-in) | keyed on `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `COPPERHEAD_TEST_CODEX` | n/a (no provider code touched) |

Runs used an isolated `KICAD_CONFIG_HOME` holding only `fp-lib-table`, matching CI.

### Test details

- **Before**: the old test failed in 5 of 16 local runs on unmodified `main` (2 of 10 in one batch with envelope logging, 3 of 6 in another).
- **After**: the new test passed 25 of 25 consecutive runs, with the full suite running at the same time for extra load.
- **It still catches regressions.** Each mutation below was applied to the `copperhead_init` handler on `main` and run 3 times against the new test; every run failed:

| Mutation | Failure |
| --- | --- |
| Lock check removed (`if (!locks.tryAcquire(...))` never refuses) | `one of two concurrent mutating calls must be refused as busy: expected [] to have a length of 1 but got +0` |
| Release removed (`locks.release(...)` never called) | `expected "release" to be called 1 times, but got 0 times` |

- Commands run and outcome: n/a, test-only change. No CLI, agent loop, provider or KiCad code changed, so there is nothing to exercise at runtime.

## Docs

None.

---

## Invariant checklist

- [ ] **Spec-gated in**: n/a (test-only change).
- [ ] **Verification-gated out**: n/a (test-only change).
- [ ] **`check`/`verify` stays LLM-free and network-free**: n/a (test-only change).
- [ ] **No sexp serialization**: n/a (test-only change).
- [ ] **Sync-obligations ledger**: n/a (test-only change).
- [ ] **Secrets**: n/a (test-only change).
- [ ] **Spec coherence**: n/a (no spec-level behavior changed).


animesh-chouhan:
**Automated pr-review pass** (the repo's `pr-review` skill run by Claude Code; not a human review).

Disclosure: the reviewer authored the commit under review (c1e8421), so this is a self-review and was run with extra skepticism, including mutation checks against the handler.

## Verdict: approve with nits

The test is now deterministic and still fails when the lock check or the release is removed. One low-severity coverage gap: holding every release hides a release that comes too early.

## Metrics

```text
lines: +34 / -10 (net 24) across 1 files  ·  src +0/-0, test +34/-10, docs +0/-0, spec +0/-0, other +0/-0
new vs net: 0 new files (0 in src), 1 modified; 0 new src lines (the net-new surface)
tests: +34 test lines; suite not measured -> 938/21/0 (pass/skip/fail, base -> head)
```

Method: `scripts/metrics.mjs` in an isolated review worktree after `npm ci`, merge base ef5a129 (`git merge-base origin/main HEAD`); areas from `git diff --numstat`; suite from `npx vitest run --reporter=json`; diff coverage skipped because no src lines changed; base suite not run (base CI is green on `main`); CI from `gh pr checks`. `mergeable=MERGEABLE`, not a draft.

## Findings

### 1. [low] Holding every release hides a lock released before the run finishes

- **Claim**: the spy holds every `RepoLocks.release` until both calls answer, so the test no longer detects a `copperhead_init` handler that releases the lock before `runInit` completes.
- **Where**: `test/mcp-server.test.ts:257` (the `mockImplementation` on `RepoLocks.prototype.release`), guarding `src/mcp/server.ts:344`.
- **Fails when**: the handler's `locks.release(repoRoot)` moves out of the `finally` to just before `await runInit(...)`. The lock is then free while init writes docs, so concurrent mutating calls can interleave, but the spy still holds that early release, the loser is still refused, `release` is still called once, and the test stays green. Verified: with that mutation the test passed 3 of 3 runs. (The previous version of the test would usually have failed, because both calls would proceed.)
- **Fix**: in the spy, record whether the run has written its docs when it releases, and require it:

```ts
const releasedAfterDocs: boolean[] = [];
// in mockImplementation, before held.push(...):
releasedAfterDocs.push(existsSync(path.join(root, 'docs', 'SPEC.md')));
// after expect(release).toHaveBeenCalledTimes(1):
expect(releasedAfterDocs, 'the lock must stay held until the run has written its docs').toEqual([true]);
```

Verified: this assertion fails 3 of 3 runs against the early-release mutation and passes 5 of 5 against the real handler (the whole `test/mcp-server.test.ts` file, 29 tests each run).

No other finding survived verification. Uncovered-lines list: none (no src lines changed), so there is no untested-surface finding.

## What was checked

- **Determinism (checklist)**: no network, no `Date.now()` / `Math.random()`, no order dependence; every test asserts. The prototype spy is restored in a `finally`, and held releases are applied there, so a failing assertion cannot leak a held lock or a patched prototype into later tests (tests within a file run sequentially here; other files run in separate workers).
- **Still catches the original regression classes** (mutations on `main`, 3 runs each): lock check removed fails with `expected [] to have a length of 1`; release removed fails with `expected "release" to be called 1 times, but got 0 times`.
- **Stability**: 25 of 25 consecutive runs passed on the PR head with the full suite running concurrently; the old test failed 5 of 16 runs on `main`.
- **Scope and description**: single test file, no production change; the description's root cause (the `requireKicad()` probe awaited before `tryAcquire`, `src/mcp/server.ts:311-313`) and numbers match what was reproduced.
- **Repo invariants** (all performed, all pass; this PR changes no production code): Spec-gated in, Verification-gated out, `check`/`verify` LLM-free and network-free, No sexp serialization, Sync-obligations ledger, Secrets.
- **Spec coherence**: no spec-level behavior changed; `openspec validate` not required.

Review ran in an isolated git worktree, measuring commit c1e8421 (PR head).

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 299, branch fix/mcp-busy-test-race, based on main.

Check it out first:
  gh pr checkout 299

FINDING 1 [low] (optional) Holding every release hides a lock released before the run finishes: the spy holds every RepoLocks.release until both calls answer, so the test no longer detects a copperhead_init handler that releases the lock before runInit completes.
  Where: test/mcp-server.test.ts:257 (the mockImplementation on RepoLocks.prototype.release), guarding src/mcp/server.ts:344
  Fails when: the handler's locks.release(repoRoot) moves out of the finally to just before await runInit(...). The lock is then free while init writes docs, so concurrent mutating calls can interleave, but the spy still holds that early release, the loser is still refused, release is still called once, and the test stays green. Verified: with that mutation the test passed 3 of 3 runs.
  Fix: in the spy, record whether the run has written its docs when it releases, and require it. Add `const releasedAfterDocs: boolean[] = [];` next to `held`; in the mockImplementation, before held.push(...), add `releasedAfterDocs.push(existsSync(path.join(root, 'docs', 'SPEC.md')));`; after `expect(release).toHaveBeenCalledTimes(1);` add `expect(releasedAfterDocs, 'the lock must stay held until the run has written its docs').toEqual([true]);`. Verified: fails 3 of 3 runs against the early-release mutation, passes 5 of 5 against the real handler.

    by prompt text.

Then verify:
  npm run typecheck
  npm run build
  npm test

and report it.

````

animesh-chouhan:
**Automated pr-review pass (follow-up)**, not a human review. Reviewer disclosure as before: self-review.

Changes since the previous pass (c1e8421 to fe758f0):

- **Finding 1 [low], held releases hide a lock released before the run finishes: resolved** by fe758f0. The spy now records whether `docs/SPEC.md` exists when the lock is released, and the test asserts `[true]`. Verified before pushing: against a `copperhead_init` handler mutated to release before `runInit`, the test fails 3 of 3 runs; against the real handler the whole `test/mcp-server.test.ts` file passes 5 of 5 runs (29 tests each). Typecheck passes and the full suite passes (938 passed, 0 failed, 21 skipped).
- CI on fe758f0: all checks pass (test job 2m8s).

No findings remain open, so there is no fix prompt.

Verdict: approve.

PR #303, opened 2026-09-14
fix: pass compat settings through to skill run providers

### LINKED ISSUE #301: Skill commands drop compatible-provider settings in providerForSkillRun
### What happened

When running a skill command configured to use a compatible/custom provider
(OpenAI-compatible endpoint), the command fails with "requires an endpoint",
even though the same model works when invoked directly (not via a skill).

Expected: skill commands should resolve and pass the same compatibility
settings (endpoint, etc.) that direct provider calls receive.

Got instead: providerForSkillRun calls makeProvider(model) without passing
the resolved compatibility settings, so the endpoint config never reaches
the provider factory when invoked through a skill command.

Root cause: src/commands/skill.ts, lines 46-50 — makeProvider(model) is
called without resolveCompatSettings(config, {}), unlike other call sites
that pass compat settings explicitly.

### Steps to reproduce

1. Configure a compat/OpenAI-compatible provider in .copperhead/config.json
   with an endpoint (e.g. a local model backend).
2. Run a skill command that uses this provider, e.g.:
     copperhead skill run local-audit
   (or the equivalent skill invocation in your setup)
3. Compare against calling the same provider/model directly (not via a
   skill command) — that path works.

Isolated repro (no KiCad project needed):
   node node_modules/tsx/dist/cli.mjs repro-new-findings.ts skill-compat

This calls providerForSkillRun and the direct provider factory side by
side with the same config, and prints both results.

### Output

```shell
SKILL direct factory: openai-compat
command factory error: compat:local-audit requires an endpoint
```

### Which command

other or not command-specific

### Model backend

Anthropic API key

### copperhead version

5ed4c1f (main branch, source checkout)

### Node, KiCad, and OS

Node v24.16.0 kicad-cli 10.0.5 Windows 11

### Before submitting

- [x] I searched existing issues and this is not a duplicate.
- [x] I removed API keys and other secrets from everything I pasted.

adhavan18:
## What

`copperhead skill run <name>` now resolves the repo's compat/OpenAI-compatible settings (`baseURL`, `apiKeyEnv`) the same way every other entry point does, so a compat provider configured in `.copperhead/config.json` actually works when invoked through a skill command.

## Why

Closes #301. `providerForSkillRun` (src/commands/skill.ts) called `makeProvider(model)` with no compat settings, while `create.ts`, `loop.ts`, and `doctor.ts` all pass `resolveCompatSettings(config)` explicitly. So a `compat:<model>` configured with an endpoint worked fine via `create`/`do`, but failed the instant it ran through `skill run` with `compat:<model> requires an endpoint`, exactly as reported, because the config's `baseURL` never reached the provider factory.

## Design

One-line fix: resolve compat settings from the loaded config and pass them into `makeProvider`, matching the existing pattern at every other call site. No change to `resolveCompatSettings` itself or to the compat provider's behavior, just closing the one call site that skipped it.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline (targeted) | `npx vitest run test/tool-registry.test.ts test/openai-compat.test.ts test/doctor.test.ts` | pass, 99/99 |
| Unit + offline (full) | `npm test` | 951 passed / 36 failed / 21 skipped — the 36 failures are pre-existing and unrelated to this change (kicad-cli not on PATH in this environment, and a couple of POSIX-permission tests that don't hold on Windows); none touch src/commands/skill.ts or test/tool-registry.test.ts |
| Live-LLM (opt-in) | n/a | not run, no API key in this environment |

New test: `test/tool-registry.test.ts` — "resolves compat baseURL/apiKeyEnv from repo config, same as a direct provider call (#301)". It configures `.copperhead/config.json` with a `baseURL`/`apiKeyEnv` and calls `providerForSkillRun` directly. Verified it fails with the exact reported error (`compat:llama3-70b-8192 requires an endpoint`) against the unfixed code, and passes once the fix is applied.

### Manual test log

- Sandbox variant used: n/a, this is a provider-wiring fix in a helper function, exercised directly by the new unit test against a real (non-mocked) `providerForSkillRun` + `makeProvider` + `OpenAIProvider` chain, which is the same code path `skill run` calls.

## Docs

None needed, no user-facing config or CLI surface changed.

---

## Invariant checklist

- [ ] Spec-gated in: n/a, doesn't touch the tool catalog or edit-tool gating
- [ ] Verification-gated out: n/a, no KiCad mutation involved
- [ ] `check`/`verify` stays LLM-free and network-free: n/a, not touched
- [ ] No sexp serialization: n/a, not touched
- [ ] Sync-obligations ledger: n/a, not touched
- [ ] Secrets: n/a, no new secret handling; still resolves the key only through the named env var, never a literal
- [ ] Spec coherence: n/a, this is a bugfix to existing behavior, not a spec-level change


animesh-chouhan:
ack, checking

animesh-chouhan:
🤖 **Automated `pr-review` pass** (not a human review). Ran in an isolated git worktree against commit `22e99be`; the reviewer's checkout was never modified.

## Verdict: approve with nits

The fix is correct, minimal, matches the established pattern at the other call sites, and ships a regression test I independently verified is genuine. Two low-severity findings, both about an error path the fix newly makes reachable.

## Metrics

```
lines: +38 / -2 (net 36) across 2 files  ·  src +2/-2, test +36/-0, docs +0/-0, spec +0/-0, other +0/-0
tests: +36 test lines; suite at head 983 passed / 21 skipped / 4 failed
diff coverage: 1/1 changed src line covered (hand-derived, see method)
ci: 5/5 checks pass (cla, greeting, label, template, test)
mergeable: MERGEABLE · draft: false
```

**Method** — measured in the review worktree; merge base `5ed4c1f`. Areas via `git diff --numstat`. Suite via `npx vitest run --reporter=json` in the worktree. The 4 failures are all `test/draft-reference-boards.test.ts` (`kicad-cli erc ... Failed to load schematic`) — environmental, the review machine's KiCad cannot load the reference schematics; none touch the changed code, and CI's `test` job passes. **Base suite: not run** (base CI is green). **Diff coverage: the metrics script could not measure it** — it failed to install `@vitest/coverage-v8` offline. Derived by hand instead: the sole changed src line is `src/commands/skill.ts:50`, and I confirmed the new test executes it by reverting that line in the worktree and watching the test fail, then restoring it. So there are no uncovered changed lines, and the untested-surface findings below concern *unchanged* code that the fix newly makes reachable.

**Regression test verified.** Reverting `src/commands/skill.ts:50` to `makeProvider(model)` makes the new test fail with exactly the reported error — `compat:llama3-70b-8192 requires an endpoint; set baseURL ...` — and it passes with the fix. The test is also tighter than it looks: deleting `OPENAI_API_KEY` while setting `GROQ_API_KEY` means it would fail if `apiKeyEnv` failed to resolve from config, not just `baseURL`.

## Findings

### 1. low — `providerForSkillRun` appends credential advice naming the wrong variables for compat models

`src/commands/skill.ts:54`

The catch block appends "skill run needs `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, or `--model` for a saved-login provider" whenever the message matches `msg.includes('API_KEY')`. The fix newly routes compat failures through that filter: with `baseURL` configured but the named key missing, `makeProvider` now throws `GROQ_API_KEY is not set`, which matches, so the user is told to set two variables irrelevant to their configured endpoint.

Confirmed empirically against the PR head — a repo configured with `baseURL: https://api.groq.com/openai/v1`, `apiKeyEnv: GROQ_API_KEY`, and no `GROQ_API_KEY` in env produces:

```
GROQ_API_KEY is not set — skill run needs OPENAI_API_KEY, ANTHROPIC_API_KEY, or --model for a saved-login provider (codex, claude-code, cursor).
```

Before this PR that case threw `requires an endpoint`, which did not match the filter and passed through clean — so this is a new user-visible regression in the advice, though not in the outcome. It does not violate AC-3.14, which only requires the error to name the missing variable (it does); the appended suffix is what misleads. Fix by skipping the suffix for compat models: hoist the resolved model into a `let` declared outside the `try` (it is currently scoped inside it) and gate the ternary on `!isCompatModel(resolved)`.

### 2. low — untested surface: the compat credential-failure path through `skill run`

`test/tool-registry.test.ts:706`

The new test covers only the success path. The sibling branch the fix opens up — `baseURL` resolved from config but the configured key absent — has no test at this entry point, which is why finding 1 shipped unnoticed. The existing error-path test at `test/tool-registry.test.ts:673` asserts `/API_KEY|login|no model/`, loose enough to pass on the misleading message. Add a case asserting the error names the configured variable and does *not* mention `OPENAI_API_KEY`.

## Invariant checks performed

All six were checked, and all pass. The diff touches only `src/commands/skill.ts` and one test, so most are untouched by construction, but each was verified rather than assumed:

- **Spec-gated in** — no change to the tool catalog or `edit_file`/`write_file` gating.
- **Verification-gated out** — no KiCad mutation path touched.
- **`check`/`verify` LLM-free and network-free** — confirmed `src/commands/check.ts` does not import `commands/skill.ts` or any provider; its import list is unchanged.
- **No sexp serialization** — `src/kicad/` untouched.
- **Sync-obligations ledger** — untouched.
- **Secrets** — the key is still resolved only through the named env var; no literal ever enters config, transcript, or error text beyond the variable's *name*.

**Spec coherence: no artifact change needed.** AC-3.13 already requires compat credentials to resolve from `.copperhead/config.json`, so this is a conformance bugfix against an existing criterion, not a spec-level change. The gap is now fully closed: `makeProvider` has exactly three call sites in `src/` (`skill.ts:50`, `create.ts:481`, `loop.ts:277`), all three now pass compat settings, and `src/cli.ts:357` is the only entry into `providerForSkillRun`.

One note on the PR description: it says `doctor.ts` passes `resolveCompatSettings(config)` to `makeProvider`. `doctor.ts` never calls `makeProvider` — it passes the settings to `checkCredential` instead. The conclusion is unaffected; noting it only because the description reads as a call-site inventory.

## Fix prompt

```
In the copperhead repo, on branch fix/skill-compat-settings (PR #303), resolve the two
findings below. Both are low severity; neither blocks the merge, but both are worth
fixing before it lands.

FINDING 1 (low) — src/commands/skill.ts:54 — providerForSkillRun appends credential
advice naming the wrong variables for compat models.

The catch block in providerForSkillRun appends the suffix "— skill run needs
OPENAI_API_KEY, ANTHROPIC_API_KEY, or --model for a saved-login provider (codex,
claude-code, cursor)." whenever msg.includes('API_KEY') (or 'no model' or
'configured'). Now that this PR passes resolveCompatSettings(config) into
makeProvider, a compat run whose baseURL resolves but whose configured key is
missing throws "GROQ_API_KEY is not set", which matches that filter. The user is
then told to set OPENAI_API_KEY/ANTHROPIC_API_KEY, neither of which is the right
remedy for their configured endpoint. Verified against the PR head: a repo with
baseURL https://api.groq.com/openai/v1 and apiKeyEnv GROQ_API_KEY, with
GROQ_API_KEY unset, produces:

  GROQ_API_KEY is not set — skill run needs OPENAI_API_KEY, ANTHROPIC_API_KEY, or
  --model for a saved-login provider (codex, claude-code, cursor).

Before this PR that case threw "requires an endpoint", which did not match the
filter and passed through clean, so this is a new user-visible regression in the
advice (not in the outcome). It does not violate AC-3.14, which only requires the
error to name the missing variable — it does; the appended suffix is what misleads.

Fix: skip the suffix for compat models. The resolved model is currently scoped
inside the try block, so it is not visible in the catch. Hoist it — declare
something like `let resolved: string | undefined` outside the try, assign it from
resolveModel inside, and gate the ternary on `!(resolved && isCompatModel(resolved))`.
Import isCompatModel from ../config.js. Keep resolveModel's own failures ("no model
configured") getting the suffix, since those genuinely are about the default
providers.

FINDING 2 (low) — test/tool-registry.test.ts:706 — untested surface: the compat
credential-failure path through skill run.

The new test covers only the success path. The sibling branch this PR opens up —
baseURL resolved from config but the configured key absent — has no test at this
entry point, which is why finding 1 shipped unnoticed. The existing error-path test
at test/tool-registry.test.ts:673 asserts /API_KEY|login|no model/, loose enough to
pass on the misleading message.

Fix: add a test alongside the new one that configures .copperhead/config.json with
baseURL and apiKeyEnv: 'GROQ_API_KEY', deletes GROQ_API_KEY (and OPENAI_API_KEY,
COPPERHEAD_BASE_URL, COPPERHEAD_API_KEY_ENV) from process.env, calls
providerForSkillRun(repo, 'compat:llama3-70b-8192'), and asserts the rejection
message names GROQ_API_KEY and does NOT contain 'OPENAI_API_KEY'. Follow the
save/restore env pattern the new test already uses, and restore every variable in a
finally block.

REPO INVARIANTS — these are hard requirements from openspec/specs/SPEC.md; do not
break any of them while fixing the above:
- Spec-gated in: edit_file/write_file must remain structurally ABSENT from the agent
  tool list until an OpenSpec proposal validates — gated by omission, never by
  prompt text.
- Verification-gated out: mutations must end in ERC (and DRC when the board changed)
  passing, repairing up to maxRepairCycles, then rolling back to the git snapshot.
- check/verify stays LLM-free and network-free: src/commands/check.ts and everything
  it imports must never touch a provider, an API key, or the network.
- No sexp serialization: the parser in src/kicad/ is read-only; KiCad files are
  edited only via anchored exact-match text replace.
- Secrets: transcripts and summaries redact sk-[A-Za-z0-9_-]+ at write time; keys
  live only in env vars; .gitignore keeps .env and .copperhead/runs/.

Neither fix is a spec-level change: AC-3.13 already requires compat credentials to
resolve from .copperhead/config.json, so this PR and these fixes are conformance
work against an existing criterion. Do not edit SPEC.md or the change artifacts.

VERIFY:
  npm run typecheck
  npm run build
  npx vitest run test/tool-registry.test.ts test/openai-compat.test.ts test/doctor.test.ts
  npm test
Expect test/tool-registry.test.ts and test/openai-compat.test.ts fully green. In an
environment without a working kicad-cli, test/draft-reference-boards.test.ts fails
with "Failed to load schematic" — that is environmental and pre-existing, not caused
by these changes. Confirm CI stays green.
```

adhavan18:
fixed both. pushed c1d6c2a.

finding 1: hoisted resolved out of the try block, gated the default-provider suffix on !isCompatModel(resolved). a missing compat key now just says "GROQ_API_KEY is not set" without the irrelevant OPENAI_API_KEY/ANTHROPIC_API_KEY suggestion.

finding 2: added the missing test at the entry point you flagged. reverted skill.ts locally first to confirm it reproduces the exact misleading message from the review (GROQ_API_KEY is not set — skill run needs OPENAI_API_KEY...), then restored the fix and confirmed it passes.

ran the full verify block: typecheck clean, build clean, npx vitest run test/tool-registry.test.ts test/openai-compat.test.ts test/doctor.test.ts -> 100 passed. didn't run the full npm test since draft-reference-boards.test.ts needs a working kicad-cli i don't have here, matches what you flagged as environmental.

adhavan18:

PR #309, opened 2026-09-16
Finish linear-regulator DC operating-point checks in `check` (closes #308)

### LINKED ISSUE #308: Finish SPICE verification: DC operating-point checks in `check`
### The problem

Copperhead currently verifies designs through ERC, DRC, and document drift checks — all of which confirm the schematic is internally consistent. None of them confirm the circuit actually behaves correctly. A regulator can pass ERC/DRC with the wrong feedback resistor values and still output the wrong voltage, and nothing in check will catch it.

The repo already has a SPICE netlist parser and assertion-related helpers, but there is no execution path: nothing runs a simulation, and nothing is wired into the check command or the agent's verification gate. So today, behavioral correctness is verified manually, if at all — after the design has already passed automated review.

### Proposed behavior

Add SPICE-based DC operating-point verification to check, scoped to circuits with declared output constraints (starting with linear regulator outputs).

copperhead check --spice

Expected output:

SPICE verification

U3 (regulator)
  Circuit: linear regulator, VIN=5V, VOUT target=3.3V ±5%
  Simulation: converged
  Result: VOUT = 3.28V — within tolerance

U7 (regulator)
  Circuit: linear regulator, VIN=12V, VOUT target=3.3V ±5%
  Simulation: converged
  Result: VOUT = 4.1V — OUT OF TOLERANCE

2 circuits checked, 1 passed, 1 failed

And with --json for CI/MCP consumers, matching the existing output convention used by other check subchecks.

The command should:

Translate the relevant schematic subsection into a SPICE netlist using the existing parser output.
Run the simulation via a local ngspice invocation (subprocess, no network).
Extract the resulting operating point and compare it against the constraint already declared for that net/component.
Report pass/fail per circuit, with the simulated value and the declared tolerance — and report clearly (not as a pass) when a circuit has no usable model or the simulation fails to converge.

Scope for v1: DC operating-point analysis only, for circuits copperhead can already recognize as regulators from existing constraint metadata. Transient analysis, arbitrary topologies, and full behavioral verification are explicitly out of scope for this proposal.

### Alternatives you considered

Manual review only (status quo) — works today but doesn't scale, and defeats the purpose of automated verification gates that the rest of the tool already relies on.
A full general-purpose SPICE deck builder for arbitrary circuits — considered and set aside for v1; without a bounded circuit vocabulary, model coverage and assertion design become open-ended and risk shipping something that silently mis-simulates unsupported topologies.
Symbolic/analytical checks instead of simulation (e.g. resistor-divider math for a regulator's feedback network) — simpler to implement, but doesn't generalize past the simplest cases and duplicates what a simulator already does correctly.

### Effect on the invariants

No effect on either invariant.

Spec-gated in: unaffected — this adds a new read-only check, not a write tool. No new mutation capability is introduced.
Verification-gated out: this proposal strengthens this invariant rather than touching it — it adds SPICE as an additional pass/fail gate alongside existing ERC/DRC, so finish/mutation flows have one more offline check they can be required to pass.

The command runs locally via subprocess (ngspice), makes no network calls, and does not access or modify any files outside the existing check output convention.

### Before submitting

- [x] I searched existing issues and the roadmap, and this is not already tracked.

DevSars24:
## What

Adds `copperhead check --spice` for declared three-terminal linear regulators. The check runs a bounded local ngspice DC operating-point simulation and reports the measured output voltage against the declared target and tolerance in text and `--json`.

## Why

Copperhead could already verify ERC, DRC, drift, and constraints, but it could not verify whether a declared regulator output actually reached its required voltage. A regulator could pass every existing check while still producing the wrong voltage on the board. This adds a read-only, offline electrical operating-point check for the first supported circuit type, closing that gap.

## Design

- Reuses the existing schematic connectivity reader, Simulation block parser, and constraint registry — no duplicate parsing was introduced.
- Supports linear-regulator DC operating-point checks only for this first slice.
- Reports explicit failures for missing ngspice, missing or unusable models, non-convergence, timeouts, and missing measurements — none of these silently pass.
- Reports `not_checked` when a regulator has no declared voltage constraint, rather than a false pass.
- Runs ngspice only as a bounded local subprocess.
- Does not access the network, invoke a provider, modify project files, modify the Git index, or add any mutation tool.
- Transient analysis, other circuit types, arbitrary topologies, and generalized deck generation are intentionally deferred. The scope boundary is documented in `openspec/changes/add-spice-verification-gate/` and the CLI documentation.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Docs build | `npm run docs:build` (Astro telemetry disabled) | Pass, 17 pages |
| OpenSpec validation | `openspec validate add-spice-verification-gate` | Pass |
| Focused SPICE + adjacent `check` suites | — | 23 passed, 1 skipped |
| Focused SPICE suite with real ngspice 47 | `npm test -- test/regulator-op.test.ts` | 7 passed, 1 guarded skip (missing-binary case, skipped because ngspice was available) |
| Full unit + offline suite | `npm test` | 981 passed, 27 failed, 22 skipped — see note below |
| Live-LLM (opt-in) | n/a | Not applicable to this change |

**Note on full-suite failures:** The clean parent commit (`f2944e4`) recorded 980 passed, 22 failed, 21 skipped on this machine before this branch's changes. This branch recorded 981 passed, 27 failed, 22 skipped. The failures are existing Windows newline/path assertions, pre-existing safety-behavior differences, and parallel-load timeouts across unrelated draft, gating, MCP, REPL, check, and registry tests — not new deterministic failures introduced by this change. Five full-suite reruns showed the same load-sensitive failures varying between 21 and 25 failed tests; the new `regulator-op` CLI test passed in all five reruns.

- Sandbox variant used (`create` / `edit`): `edit`
- ngspice version: 47, verified on Windows
- Fixtures: generated known-good and deliberately wrong-output regulator fixtures

```text
$ npm test -- test/regulator-op.test.ts --maxWorkers=1 --minWorkers=1 --reporter=verbose
✓ ngspice integration > runs the local ngspice binary on the modeled regulator fixture
Test Files  1 passed (1)
Tests  7 passed | 1 skipped (8)

$ node --import tsx src/cli.ts --repo <known-good-fixture> check --spice
SPICE U1 (1V5): PASS — 1.5 V; target 1.5 V, tolerance 1.4–1.6 V

$ node --import tsx src/cli.ts --repo <wrong-output-fixture> check --spice
SPICE U1 (1V5): FAIL — 1.8 V; target 1.5 V, tolerance 1.4–1.6 V
```

Both fixtures also reported 16 unrelated ERC violations, so each run's overall `check` exit code was 1; the SPICE subcheck result above was distinct and correct regardless of those unrelated findings.

## Docs

- Updated `README.md` and `docs/src/content/docs/reference/cli.md` with the new `--spice` flag and expected output.
- Updated `openspec/specs/SPEC.md` and `openspec/changes/add-spice-verification-gate/` with the command contract and the explicitly deferred scope (transient analysis, other circuit types, arbitrary topologies).

## Invariant checklist

- [x] **Spec-gated in:** Not affected. This adds a new read-only check, not a write tool. No new mutation capability is introduced.
- [x] **Verification-gated out:** Not weakened. This adds SPICE as an additional read-only check alongside existing ERC/DRC/drift checks; no existing mutation, rollback, or obligations protection is changed.
- [x] **`check`/`verify` stays LLM-free and network-free:** Confirmed. ngspice runs only as a bounded local subprocess; no network call, provider invocation, or API key is used.
- [x] **No sexp serialization:** Confirmed. The existing S-expression parser is used only to read schematic data; no schematic file is serialized or modified.
- [x] **Sync-obligations ledger:** Not applicable. The command performs no mutations and does not enter the agent loop.
- [x] **Secrets:** Not applicable. No credentials are read and no transcript is created.
- [x] **Spec coherence:** `SPEC.md` and the OpenSpec change under `add-spice-verification-gate/` were updated together and pass `openspec validate`.

## Known limitations

- ngspice 47 was verified on Windows only. Cross-platform simulator behavior and vendor-specific regulator models were not exhaustively tested.
- The manual fixtures use a behavioral surrogate model and intentionally retain unrelated ERC issues so the SPICE pass/fail result could be verified in isolation from other checks.


animesh-chouhan:
great work but do lookup a bit about: https://fang.copperhead.sh

PR #315, opened 2026-09-21
fix(create): e2e harness + P0/P1 fixes from real pipeline run

stffinfcti:
## What

Adds a deterministic end-to-end replay harness for `copperhead create` ([create-e2e-replay.test.ts](test/create-e2e-replay.test.ts) plus [fixtures](test/fixtures/create-e2e)), and lands the P0/P1 fixes a real `create` run surfaced: the openspec spec-delta deadlock ([cli.ts](src/openspec/cli.ts), [create.ts](src/commands/create.ts)) and the BOM dossier passive-row drop ([dossier.ts](src/kicad/dossier.ts)). Full defect list in [FINDINGS.md](FINDINGS.md).

## Why

Closes #66, which asked for e2e coverage of the create pipeline. The live run also proved the pipeline could not finish a real brief end to end:

- P0: openspec 1.13.x rejects `init --no-interactive`, so initialized workspaces never got spec deltas; `validate_change` then demanded deltas that `propose_change` could not author, deadlocking the run.
- P1: `bomSymbolDossier` silently dropped R/C/L passives and never listed installed symbol libraries, so the stage-4 model pattern-copied Footprint values into `lib_ids` and refused with a false "libraries are missing".

## Design

Fixes stay inside the existing stage machinery rather than adding wrappers: `init` falls back to the interactive-compatible path when `--no-interactive` is rejected, `propose_change` accepts a `spec_deltas` argument so `validate_change` has a real artifact to check, and the dossier now names skipped refdes and discloses Device:R/C/L instead of dropping them. Test steering in [recovery.ts](src/agent/recovery.ts) and [filetools.ts](src/agent/filetools.ts) routes minified-JSON anchor misses and engine-drafted edits back to `draft_schematic`, which is where several observed stalls came from.

## Spec / OpenSpec

No spec-level behavior change: these are bug fixes to existing stages, so SPEC.md and the OpenSpec change set do not move.

## Testing

Real results on this branch (Ubuntu, Node 20). Every failing test reports `kicad-cli not found on PATH`; there are no assertion failures.

| Check | Command | Status |
| --- | --- | --- |
| Unit + offline | `npm test` | 977 pass, 29 fail (all: `kicad-cli not found on PATH`), 25 skipped |
| Live-LLM (opt-in) | keyed on provider env var | n/a: no provider key in this environment |

### Invariants

- [ ] Agent loop: tool surface unchanged; new steering lives inside existing recovery and draft-tool paths.
- [ ] Safety gates: no new filesystem or network writes outside the workspace; the replay harness is fully offline (fixture transcripts, no provider calls).
- [ ] KiCad layer: dossier output is additive only (skipped refdes and installed libraries disclosed); no change for complete symbol sets.
- [ ] Providers: no changes to request/response shape beyond what tests already cover.


stffinfcti:

PR #317, opened 2026-09-22
fix(legibility): accurately measure group caption advance and check caption overflow (fixes #307)

### LINKED ISSUE #307: legibility: TEXT_ADVANCE 0.6 under-measures stroke text by ~22%, so group captions overflow their boxes undetected
## What happened

`TEXT_ADVANCE` in `src/kicad/legibility.ts:79` is `0.6`, but KiCad's stroke font advances at roughly `0.76` of the glyph height. Every text box the legibility checker computes is therefore about 22% narrower than what KiCad actually draws.

The comment on `textBounds` at line 185 calls this a "conservative text box". It is optimistic, not conservative, and it under-measures in the direction that produces false negatives.

## Measured on a real sheet

From a `copperhead create` run (stage 4, ERC clean, legibility reported score 89.91 with **0 error-severity findings**). Estimated width is `len * TEXT_ADVANCE * height` with height 3.5mm. Rendered width is measured from the `<g class="stroked-text">` path extents in the `kicad-cli sch export svg` output:

| caption | chars | estimated | rendered | under by | implied advance |
|---|---|---|---|---|---|
| `B. Input bypass` | 15 | 31.50 | **40.00** | 21% | 0.762 |
| `C. PTC` | 6 | 12.60 | **16.17** | 22% | 0.770 |
| `E. Outputs` | 10 | 21.00 | **26.00** | 19% | 0.743 |
| `F. Mechanical` | 13 | 27.30 | **35.17** | 22% | 0.773 |

## Consequence: group captions escape their own boxes

Group box geometry from the same sheet, with captions placed at `box_start + 2.0mm`:

| group | box x-range | caption rendered x-range | verdict |
|---|---|---|---|
| B | 100.08 to **135.89** | 103.37 to **143.37** | 7.48mm outside its box, drawn inside group C's box |
| F | 248.16 to **264.16** | 251.45 to **286.62** | 22.46mm outside its box, close to the frame edge |

Group B's caption is physically rendered inside group C's rectangle, having crossed the full 5.08mm gutter. On the render, `B. Input bypass` and `C. PTC` read as one run of text. They do not strictly overlap (0.73mm apart), which is exactly why the `text-collision` check at line 526 does not fire: with the 0.6 advance the checker believes B's caption ends at 133.58, comfortably clear.

So the checker reports a clean sheet while the captions are visibly wrong.

## Steps to reproduce

1. Draft any sheet whose group captions are long relative to their group box.
2. `copperhead score` / the legibility gate reports 0 error-severity findings.
3. `kicad-cli sch export svg` and measure the `stroked-text` path extents against the group `rectangle` bounds.

## Suggested fix

Two parts, either of which helps on its own:

1. Raise `TEXT_ADVANCE` to about `0.78`. That makes the box genuinely conservative and lets the existing `text-collision` logic do its job. Worth re-running the golden corpus afterwards, since some currently-passing sheets may start reporting real findings.
2. Add a containment check: a group caption should be flagged when its text box is not inside its own group rectangle. `group-overlap` and `unlabeled-group` already exist as error-severity kinds, so a `caption-overflow` sibling would fit the existing model. The engine could also shrink the caption or widen the box when the two do not fit.

## Environment

- copperhead 0.11.0 at `b24de54`
- Windows 11, Node v22.17.0, kicad-cli 10.0.4

jackspiece:
## What

Fixes group captions overflowing group rectangles undetected by using an accurate stroke-font advance for group captions (`CAPTION_ADVANCE = 0.78`) and adding a containment check for captions against their group rectangles.

## Why

Closes #307. KiCad's stroke font advances at ~0.76 to 0.78 of glyph height for uppercase text, while the general label `TEXT_ADVANCE` was set to 0.60 to avoid false positive collisions near bodies (design C3). This caused long subsystem group captions to spill outside their group rectangles into neighboring boxes without triggering findings.

## Design

- Introduced `CAPTION_ADVANCE = 0.78` specifically for group caption text items, preserving `TEXT_ADVANCE = 0.6` for wire labels and design C3 constraints.
- Extended `unlabeled-group` checks to flag when a caption's text box extends past its group rectangle right boundary by more than `TOL`.
- Added unit test covering caption overflow detection.

### Automated test status

| Check | Command | Status |
| --- | --- | --- |
| Typecheck | `pnpm run typecheck` | pass |
| Build | `pnpm run build` | pass |
| Unit + offline | `pnpm test test/legibility.test.ts` | pass (24/24) |

Exercised `checkLegibility` against clean, ugly, and overflowing group caption schematic fixtures. All 24 tests in `test/legibility.test.ts` passed cleanly.

## Docs

N/A

---

## Invariant checklist

- [x] **Spec-gated in**: N/A
- [x] **Verification-gated out**: N/A
- [x] **`check`/`verify` stays LLM-free and network-free**: Legibility checks remain 100% geometric and deterministic.
- [x] **No sexp serialization**: S-expression parser/checker is read-only; no round-trip serialization.
- [x] **Sync-obligations ledger**: Preserved.
- [x] **Secrets**: No secrets touched.
- [x] **Spec coherence**: N/A


jackspiece:

jackspiece:
recheck

jackspiece:
recheck

animesh-chouhan:
ack, checking

animesh-chouhan:
@jackspiece can you add before and after render of a sample board in comments

animesh-chouhan:
> 🤖 **Automated pr-review pass** (Claude Code `/pr-review` skill). This is not a human review and does not gate the merge.

# Review: #317 fix(legibility): accurately measure group caption advance and check caption overflow (fixes #307)

**Verdict: request changes.** The idea is right: measure captions at KiCad's real stroke-font advance and flag captions that run out of their group box. But the draft engine still sizes group boxes without looking at caption width, so the new error-severity finding can now block `create` on the exact sheet from #307, and the model has no way to fix it. The new test also passes with the old 0.6 advance, so it doesn't pin down the fix.

## Metrics

```
lines: +54 / -18 (net 36) across 2 files  ·  src +35/-18, test +19/-0, docs +0/-0, spec +0/-0, other +0/-0
tests: +19 test lines; head suite 984 passed / 21 skipped / 4 failed (the 4 failures also fail on base; environmental, see below)
diff coverage: not measured (instrumented); traced by hand, every new branch is reached by some test
ci: 4/4 checks pass (cla, label, template, test); mergeable=MERGEABLE, draft=false
dependencies: none changed
```

**Method.** Line counts come from `git diff --numstat` against merge base `c9045af`, with generated files excluded. The metrics script's own suite run could not resolve vitest because the worktree was missing its `node_modules` link. I linked it by hand and ran `npx vitest run` in the review worktree. The 4 failures are all in `test/draft-reference-boards.test.ts` ("kicad-cli erc produced no report"). The same 4 fail on base `main` in this environment. CI installs KiCad 10, and those tests pass there, so the pinned reference boards stay legibility-clean with this PR. `@vitest/coverage-v8` is not installed, so diff coverage was not instrumented. CI state comes from `gh pr checks`. Typecheck (`tsc --noEmit`) is clean.

Ran in an isolated worktree at head `5cbf6ef7fb91a60777d9016797ea886aa4ed979e`. The maintainer's checkout was not touched.

## Findings

**1. [medium] The draft engine sizes group boxes from their cells only, so an engine draft with a long caption on a narrow group now fails the legibility gate, and the model cannot fix it.**
- Where: `src/kicad/draft/engine.ts:2884` (group `maxX` is taken from cell bodies and field reach only). Related: `src/kicad/legibility.ts:565-570`.
- Fails when: an IR names a group with a caption wider than its cells, like #307's `F. Mechanical` around a single connector. The engine draws the rect at cell width, and the new check reports `unlabeled-group` at error severity. Under AC-16.22, `create` reports the schematic stage as unmet and does not advance. The finding says "widen the group rectangle", but the engine owns every coordinate (SPEC §16). The model can only rename the group, so re-runs hit the same finding. Today the engine checks caption width only against the frame edge (`engine.ts:5142-5145`, at `LABEL_ADVANCE * CAPTION_SIZE`), never against the group's own box.
- Fix: in the engine, widen each group rect to `max(cell extent, x1 + 2 + len(name) * LABEL_ADVANCE * CAPTION_SIZE + margin)`. `LABEL_ADVANCE` is 0.8, at or above the checker's 0.78, so the engine stays conservative. Add a draft test: a one-part group with a long caption, drafted and checked, has zero `overflows its group rectangle` findings.

**2. [low] The new test passes with the old advance, so it does not pin down the fix.**
- Where: `test/legibility.test.ts:343-358`
- Fails when: `CAPTION_ADVANCE` is set back to `0.6`. The test still passes (checked by mutation in the review worktree). The caption `B. Input bypass and filtering` is 29 characters, not the 30 the comment says. The caption is anchored at x=22 in a rect that ends at x=50, so 28 mm is available. At 0.6 the caption measures 29 × 0.6 × 2 = 34.8 mm, which overflows too. Line 354 also has stray indentation.
- Fix: use a caption of 18 to 23 characters at height 2. For example, `B. Input bypass caps` (20 characters) is 24.0 mm at 0.6, which fits, and 31.2 mm at 0.78, which overflows. Correct the comment's arithmetic. Add a negative case: a caption that fits at 0.78 produces no overflow finding.

**3. [low] Caption overflow is reported under the `unlabeled-group` kind, and the docs don't mention it.**
- Where: `src/kicad/legibility.ts:568`; `docs/src/content/docs/reference/schematic-legibility.md:41`
- Fails when: a user reads the kinds table. It says `unlabeled-group` means "no caption, or a caption naming nothing in SUBSYSTEMS.md or BOM.md", so a finding for a correctly named caption that is too long is surprising. Setting `legibility.severity.unlabeled-group` to `off` to silence caption-name checks also silences overflow. The two can't be tuned separately.
- Fix: add a `caption-overflow` kind with an `error` default severity, as #307 suggested, plus a row in the docs table. The minimum fix is to update the `unlabeled-group` row to include overflow.

**4. [low] Only the right edge is checked. A caption that is centered, right-justified, or vertical can spill out of the left edge or the bottom and not be flagged.**
- Where: `src/kicad/legibility.ts:567`
- Fails when: a caption has no `(justify left)`, so `textBounds` centers it on its anchor. Its anchor sits in the top band near the rect's left edge, and its text extends past `g.bounds.minX`. The finding does not fire. Hand-authored sheets can hit this. Engine output can't, because engine captions are `left top`.
- Fix: flag when `!boundsContain(g.bounds, cBox)` (with `TOL`) instead of comparing `maxX` alone, and report the overflow on the offending edge.

**5. [low] A stale doc comment on `TEXT_ADVANCE` contradicts the constant.**
- Where: `src/kicad/legibility.ts:78-79`
- Fails when: someone reads the code. Two JSDoc blocks are stacked on the constant, and the new one says "KiCad stroke font advances at ~0.76-0.78 (closes #307)" above `TEXT_ADVANCE = 0.6`, directly under the retained design C3 comment saying it is tuned below the true average.
- Fix: delete the added line at 79 and keep the C3 comment.

**Untested surface.** No test exercises a caption that fits at 0.78, and no test checks a non-left-justified caption against the box (findings 2 and 4). Every other new branch is reached by some test.

**Also outstanding:** @animesh-chouhan asked for before/after renders of a sample board. None have been posted yet.

## Invariant checks

- **Spec-gated in:** passed. No agent or tool-list code is touched.
- **Verification-gated out:** passed. No run or commit flow is touched. Finding 1 is about the legibility gate blocking progress, not skipping verification.
- **`check`/`verify` LLM-free and network-free:** passed. `legibility.ts` stays pure geometry, with no new imports.
- **No sexp serialization:** passed. The checker stays read-only.
- **Sync-obligations ledger:** passed. It is not touched.
- **Secrets:** passed. Nothing is touched.
- **Spec coherence:** no SPEC.md or change artifacts changed. The legibility kinds are not listed in SPEC.md, so no divergence. AC-16.3 and AC-16.22 are affected in behavior by finding 1.

## Fix prompt

````text
KiCad projects. Node >= 20. The PR is 317, branch fix/legibility-text-advance, based on main.

Check it out first:
  gh pr checkout 317

FINDING 1 [medium] The draft engine sizes group boxes from their cells only, so an engine draft with a long caption on a narrow group now fails the legibility gate, and the model cannot fix it.
  Where: src/kicad/draft/engine.ts:2884 (group maxX is taken from cell bodies and field reach only); related src/kicad/legibility.ts:565-570
  Fails when: an IR names a group with a caption wider than its cells (for example "F. Mechanical" around a single connector). The engine draws the rect at cell width, and the new check reports unlabeled-group at error severity. Under AC-16.22, create reports the schematic stage as unmet and does not advance. The finding says "widen the group rectangle", but the engine owns every coordinate, so the model can only rename the group and re-runs hit the same finding. Today the engine checks caption width only against the frame edge (engine.ts:5142-5145, at LABEL_ADVANCE * CAPTION_SIZE), never against the group's own box.
  Fix: in the engine, widen each group rect to max(cell extent, x1 + 2 + len(name) * LABEL_ADVANCE * CAPTION_SIZE + margin). LABEL_ADVANCE is 0.8, at or above the checker's 0.78, so the engine stays conservative. Add a draft test: a one-part group with a long caption, drafted and checked, has zero "overflows its group rectangle" findings.

FINDING 2 [low] (optional) The new test passes with the old advance, so it does not pin down the fix.
  Where: test/legibility.test.ts:343-358
  Fails when: CAPTION_ADVANCE is set back to 0.6. The test still passes. The caption "B. Input bypass and filtering" is 29 characters, not 30. It is anchored at x=22 in a rect that ends at x=50, so 28 mm is available, and at 0.6 it measures 29 * 0.6 * 2 = 34.8 mm, which overflows too. Line 354 also has stray indentation.
  Fix: use a caption of 18 to 23 characters at height 2. For example, "B. Input bypass caps" (20 characters) is 24.0 mm at 0.6, which fits, and 31.2 mm at 0.78, which overflows. Correct the comment's arithmetic. Add a negative case: a caption that fits at 0.78 produces no overflow finding.

FINDING 3 [low] (optional) Caption overflow is reported under the unlabeled-group kind, and the docs don't mention it.
  Where: src/kicad/legibility.ts:568; docs/src/content/docs/reference/schematic-legibility.md:41
  Fails when: a user reads the kinds table, which says unlabeled-group means "no caption, or a caption naming nothing in SUBSYSTEMS.md or BOM.md", so a finding for a correctly named caption that is too long is surprising. Setting legibility.severity.unlabeled-group to off to silence caption-name checks also silences overflow.
  Fix: add a caption-overflow kind with an error default severity, plus a row in the docs table. The minimum fix is to update the unlabeled-group row to include overflow.

FINDING 4 [low] (optional) Only the right edge is checked. A caption that is centered, right-justified, or vertical can spill out of the left edge or the bottom and not be flagged.
  Where: src/kicad/legibility.ts:567
  Fails when: a caption has no (justify left), so textBounds centers it on its anchor. Its anchor sits in the top band near the rect's left edge and its text extends past g.bounds.minX. The finding does not fire.
  Fix: flag when !boundsContain(g.bounds, cBox) (with TOL) instead of comparing maxX alone, and report the overflow on the offending edge.

FINDING 5 [low] (optional) A stale doc comment on TEXT_ADVANCE contradicts the constant.
  Where: src/kicad/legibility.ts:78-79
  Fails when: someone reads the code. Two JSDoc blocks are stacked on the constant, and the new one says "KiCad stroke font advances at ~0.76-0.78 (closes #307)" above TEXT_ADVANCE = 0.6, directly under the retained design C3 comment saying it is tuned below the true average.
  Fix: delete the added line at 79 and keep the C3 comment.

No test exercises a caption that fits at 0.78, and no test checks a
non-left-justified caption against the box. Add tests for both.

    by prompt text.

Then verify:
  npm run typecheck
  npm run build
  npm test

All three must pass. The reference-board tests (test/draft-reference-boards.test.ts)
need kicad-cli 10 and must stay byte-identical and legibility-clean; if the engine
change in finding 1 changes a pinned output, regenerate it deliberately and say so.

````

animesh-chouhan:
image
image

animesh-chouhan:
**Before/after renders for caption overflow (#307)**

I rendered each sheet with `kicad-cli sch export svg` (9.0.8). The dashed box is the caption box each checker version measures; the blue fill is where KiCad actually draws the caption.

### 1. Hand-drawn sheet with the #307 layout

3.5 mm bold captions, left/top justified, 2 mm in from the group corner.

**`main` @ c9045af:** the sheet is reported clean (0 errors), even though two captions leave their groups.



**This PR @ 5cbf6ef:** the overflow is caught, but the flat 0.78 box still ends short of the drawn text.



**Proposed fix (local commit `ef294ca`, not pushed yet):** the box is measured glyph by glyph, to the edge of the ink.



| caption | group box x (mm) | KiCad ink x (mm) | box right edge @0.6 / @0.78 / proposed (mm) |
|---|---|---|---|
| B. Input bypass | 76.20–111.76 | 79.14–**119.84** | 109.70 / 119.15 / 120.26 |
| F. Mechanical | 157.48–175.26 | 160.42–**196.29** | 186.78 / 194.97 / 196.71 |

### 2. Engine draft: one resistor in a group with a long caption

The engine sizes a group box from the parts inside it, so the caption runs out of its box. This already fails `main`'s gate (as a text collision), and the PR's new overflow check fails it again. The model can't fix either one, because the engine sets every coordinate.



With the proposed fix, the engine widens the box to fit its caption, using the same measurement the checker uses. All three checker versions report 0 errors.

animesh-chouhan:
> 🤖 **Automated pr-review pass: follow-up** (Claude Code `/pr-review` skill). This is not a human review and does not gate the merge.

**Follow-up to the earlier review. New finding (medium): the caption box must contain the drawn caption, and a flat per-character advance can't guarantee that.**

KiCad's stroke font is proportional. W and M advance 1.14 of the text height, m 1.33, and i and l less than 0.53. Bold ink also starts 0.8–1.3 mm past the anchor. So any single average under-reads some captions. I measured 176 captions against KiCad renders:

| box model | captions with ink outside the box |
|---|---|
| 0.6 × chars | 157 / 176 |
| 0.78 × chars (this PR) | 140 / 176 |
| 0.8 × chars + 0.5 h | 52 / 176 |

Two cases the PR gets wrong:
- `WWWWMMMM` passes a 0.78 check but draws about 9 mm past its box.
- `illi.illi` fits its box but gets flagged at 0.78.

**Proposed fix (ready locally as commit `ef294ca` on top of `5cbf6ef`, not pushed yet).** It also resolves every finding from the first pass. The plan is to push it to this branch, since maintainer edits are enabled. @jackspiece, say if you'd rather apply it yourself.
- **`src/kicad/strokefont.ts` (new):** per-glyph metrics for KiCad's stroke font (advance and ink extent), measured from kicad-cli SVG stroke paths. `strokeTextExtent()` returns a box that contains the ink for left, center or right alignment, bold or regular. Across 400 random KiCad renders, no ink fell outside the box, and the box was 0.09–0.33 h wider than the ink per side.
- **Checker:** captions are measured with it, and every edge is checked, which fixes finding 4. The general 0.6 estimate (design C3) is unchanged, and the stale comment is gone (finding 5).
- **Engine:** group boxes widen to fit their caption using the same measurement, so drafts contain their captions by construction (finding 1). The pinned reference boards and the Tier C golden are byte-identical.
- **Tests:** they cover a wide-glyph overflow, a narrow-glyph fit, a left-edge spill, and an engine draft of a one-part group under a long caption. Each fails with its half of the fix reverted, which covers finding 2.
- **Docs:** the `unlabeled-group` entry now covers caption overflow (finding 3, minimum fix).
- **Checks:** typecheck and build are clean, and the full suite has 992 passed, 0 failed.

**Possible follow-up: wrap long captions instead of only widening the box.** Widening is the smallest change that guarantees a fit, and it keeps the pinned outputs byte-identical. But it can leave a very wide box around a small group: 117 mm around one resistor in the engine example, which can push a draft onto larger paper. A cleaner layout would wrap the caption at word boundaries to fit the width the parts already need, up to 2 lines. The box would widen only when that fails: a single long word, or text that still doesn't fit in two lines. In the engine example that gives two lines of about 52 mm, roughly half the width. Wrapping needs three more changes, and the first one would otherwise make wrapped captions fail the gate:

1. **Name matching:** `captionMatches` (`src/kicad/legibility.ts`) compares the caption text as-is, so `"Mechanical connector\nand mounting holes"` wouldn't match the SUBSYSTEMS.md name and would be reported as naming nothing. It should collapse whitespace before comparing.
2. **Engine caption strip:** `CAPTION_BAND` reserves 8.9 mm, enough for one line. A two-line 3.5 mm bold caption draws 10.3 mm tall (the second line starts 5.6 mm below the first), so the band has to grow by about 5.6 mm per extra line, or the second line sits on the parts.
3. **Checker caption height:** the caption box is one line tall, so the second line is ignored by the collision and fit checks. The box's height should grow with the line count. The width side already handles multiple lines (`strokeTextExtent` measures the widest line).

Captions that fit on one line would draw as before, so the pinned outputs should stay byte-identical. That needs confirming with a test run. This could go in the same change or a separate PR. It isn't part of the proposed commit.

**Correction to the first pass:** the 4 `draft-reference-boards` failures I reported were a local glitch, not a KiCad-version issue. They pass locally now, on this PR and on `main`.

Renders are in the comment above.

